PR #8, opened 2019-11-27
Rolling Buffer changes to handle retransmission

zhiyua-git:
*Issue #, if available:*
N/A
*Description
Rolling buffer that holds data and let data out of range fall off and freed.
Rtp rolling buffer to keep packets and return packet for resend if required.


src/source/Rtcp/RollingBuffer.c
```diff
@@ -0,0 +1,155 @@
...
+#include "../Include_i.h"
+
+STATUS createRollingBuffer(UINT32 capacity, FreeDataFunc freeDataFunc, PRollingBuffer* ppRollingBuffer)
+{
+    ENTERS();
+    STATUS retStatus = STATUS_SUCCESS;
+    PRollingBuffer pRollingBuffer = NULL;
+    CHK(capacity != 0, STATUS_INVALID_ARG);
+
+    pRollingBuffer = (PRollingBuffer) MEMALLOC(SIZEOF(RollingBuffer) + SIZEOF(UINT64) * capacity);
```
Sean-Der: Mind doing `MEMCALLOC` instead of setting 0 explicitly, just in case new members get added.

Sean-Der:
We really need to get ASAN going, looks like we had memory corruption on this build https://travis-ci.com/awslabs/amazon-kinesis-video-streams-webrtc-sdk-c/jobs/261136514

Possibly unrelated!

src/source/Rtcp/RollingBuffer.c
zhiyua-git: The performance is better on using MEMALLOC, given the rolling buffer could be very large, I would rather do MEMALLOC.

src/source/Rtcp/RollingBuffer.c
```diff
@@ -0,0 +1,155 @@
...
+    pRollingBuffer->dataBuffer = (PUINT64) (pRollingBuffer + SIZEOF(RollingBuffer));
+    MEMSET(pRollingBuffer->dataBuffer, 0, SIZEOF(UINT64) * pRollingBuffer->capacity);
+
+CleanUp:
+    if (STATUS_FAILED(retStatus) && pRollingBuffer != NULL) {
+        freeRollingBuffer(&pRollingBuffer);
+        pRollingBuffer = NULL;
+    }
+
+    if (pRollingBuffer != NULL) {
```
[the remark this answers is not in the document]
zhiyua-git: I should actually check for ppRollingBuffer != NULL instead. Updated.

PR #16, opened 2019-11-28
Remove private beta URLs for STUN and ControlPlane

Sean-Der:


samples/Samples.h
```diff
@@ -19,8 +19,7 @@ extern "C" {
 #define NUMBER_OF_OPUS_FRAME_FILES                                              618
 #define DEFAULT_FPS_VALUE                                                       25
 #define DEFAULT_TERMINATION_POLLING_INTERVAL                                    5 * HUNDREDS_OF_NANOS_IN_A_SECOND
-#define KINESIS_VIDEO_BETA_CONTROL_PLANE_URL                                    "https://beta.kinesisvideo.us-west-2.amazonaws.com"
-#define KINESIS_VIDEO_BETA_STUN_URL                                             "stun:stun.beta.kinesisvideo.us-west-2.amazonaws.com:443"
+#define KINESIS_VIDEO_STUN_URL                                                  "stun:stun.kinesisvideo.us-west-2.amazonaws.com:443"
```
zhiyua-git: Stun endpoint need to be region specific. Or it will get high latency.

Sean-Der: how do we determine what region we should use?

MushMal: Make dynamic based on the region

Sean-Der: done!

samples/Samples.h
```diff
@@ -19,8 +19,7 @@ extern "C" {
 #define NUMBER_OF_OPUS_FRAME_FILES                                              618
 #define DEFAULT_FPS_VALUE                                                       25
 #define DEFAULT_TERMINATION_POLLING_INTERVAL                                    5 * HUNDREDS_OF_NANOS_IN_A_SECOND
-#define KINESIS_VIDEO_BETA_CONTROL_PLANE_URL                                    "https://beta.kinesisvideo.us-west-2.amazonaws.com"
-#define KINESIS_VIDEO_BETA_STUN_URL                                             "stun:stun.beta.kinesisvideo.us-west-2.amazonaws.com:443"
+#define KINESIS_VIDEO_STUN_URL                                                  "stun:stun.kinesisvideo.%s.amazonaws.com:443"
```
MushMal: I think we should move this to the public header.?

PR #22, opened 2019-11-30
If CMAKE_OSX_SYSROOT is set include it in libsrtp CFLAGS

Sean-Der:


cmake-scripts/libsrtp-CMakeLists.txt
```diff
@@ -4,14 +4,21 @@ project(libsrtp-download NONE)
 
 find_program(MAKE_EXE NAMES make)
 
+if (DEFINED CMAKE_OSX_SYSROOT)
```
MushMal: You could merge these in a single if

Sean-Der: done!

PR #23, opened 2019-12-02
Fix off-by-one in RtcIceCandidateInit parsing

Sean-Der:
```
==18070==ERROR: AddressSanitizer: stack-buffer-overflow on address 0x7ffee67b67b4 at pc 0x00010983f1cf bp 0x7ffee67b6710 sp 0x7ffee67b6708
READ of size 4 at 0x7ffee67b67b4 thread T0
    #0 0x10983f1ce in deserializeRtcIceCandidateInit SessionDescription.c:648
    #1 0x1094946e4 in com::amazonaws::kinesis::video::webrtcclient::PeerConnectionApiTest_deserializeRtcIceCandidateInit_Test::TestBody() PeerConnectionApiTest.cpp:24
```

### DIFF
--- a/src/source/PeerConnection/SessionDescription.c
+++ b/src/source/PeerConnection/SessionDescription.c
@@ -642,9 +642,8 @@ STATUS deserializeRtcIceCandidateInit(PCHAR pJson, UINT32 jsonLen, PRtcIceCandid
     tokenCount = jsmn_parse(&parser, pJson, jsonLen, tokens, ICE_CANDIDATE_INIT_TOKENS);
     CHK(tokenCount > 1, STATUS_INVALID_API_CALL_RETURN_JSON);
     CHK(tokens[0].type == JSMN_OBJECT, STATUS_ICE_CANDIDATE_INIT_MALFORMED);
-    CHK(tokenCount == ICE_CANDIDATE_INIT_TOKENS, STATUS_ICE_CANDIDATE_INIT_MALFORMED);
 
-    for (i = 1; i < tokenCount; i++) {
+    for (i = 1; i < (tokenCount - 1); i++) {
         if (STRNCMP(CANDIDATE_KEY, pJson + tokens[i].start, ARRAY_SIZE(CANDIDATE_KEY) - 1) == 0) {
             STRNCPY(pRtcIceCandidateInit->candidate, pJson + tokens[i + 1].start, (tokens[i + 1].end - tokens[i + 1].start));
         }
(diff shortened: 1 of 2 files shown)


src/source/PeerConnection/SessionDescription.c
```diff
@@ -642,9 +642,8 @@ STATUS deserializeRtcIceCandidateInit(PCHAR pJson, UINT32 jsonLen, PRtcIceCandid
     tokenCount = jsmn_parse(&parser, pJson, jsonLen, tokens, ICE_CANDIDATE_INIT_TOKENS);
     CHK(tokenCount > 1, STATUS_INVALID_API_CALL_RETURN_JSON);
     CHK(tokens[0].type == JSMN_OBJECT, STATUS_ICE_CANDIDATE_INIT_MALFORMED);
-    CHK(tokenCount == ICE_CANDIDATE_INIT_TOKENS, STATUS_ICE_CANDIDATE_INIT_MALFORMED);
```
zhiyua-git: Why don't we need to track ICE_CANDIDATE_INIT_TOKENS here?

chehefen:
did we check if end-to-end is affected?

src/source/PeerConnection/SessionDescription.c
Sean-Der: The other values are actually [optional](https://developer.mozilla.org/en-US/docs/Web/API/RTCIceCandidateInit)! When I wrote it the first time I just asserted it was the length that Chromium sent. 

It should have no effect, but I just wanted to remove the overly assertive check.

Sean-Der:
> did we check if end-to-end is affected?

All the samples still work. We don't have E2E tests in CI yet, will work on getting that done ASAP

src/source/PeerConnection/SessionDescription.c
```diff
@@ -642,9 +642,8 @@ STATUS deserializeRtcIceCandidateInit(PCHAR pJson, UINT32 jsonLen, PRtcIceCandid
...
 
-    for (i = 1; i < tokenCount; i++) {
+    for (i = 1; i < (tokenCount - 1); i++) {
```
MushMal: Can we get more info on i < tokenCount

Sean-Der: When looping for key/value pairs it checks for `candidate`.

 It would incorrectly accept a key (without a value) at the end and then do an invalid read.

MushMal: I see, we normally bump up the value if i when "consuming" the value too.

PR #27, opened 2019-12-02
when process dtls packet, only try send data back if handshake is not…

chehefen:
… done. In dtls timer callback, do handle dtls timeout.

*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -11,3 +11,4 @@ open-source/libusrsctp
 open-source/libwebsockets
 open-source/local
 outputs
+open-source/amazon-kinesis-video-streams-producer-sdk-cpp
--- a/src/source/Dtls/Dtls.c
+++ b/src/source/Dtls/Dtls.c
@@ -34,16 +34,33 @@ STATUS dtlsTransmissionTimerCallback(UINT32 timerID, UINT64 currentTime, UINT64 
 {
     STATUS retStatus = STATUS_SUCCESS;
     PDtlsSession pDtlsSession = (PDtlsSession) customData;
+    BOOL locked = FALSE;
+    struct timeval timeout = {0};
+    UINT64 timeoutValDefaultTimeUnit = 0;
 
     CHK(pDtlsSession != NULL, STATUS_NULL_ARG);
 
-    CHK(!SSL_is_init_finished(pDtlsSession->pSsl), STATUS_TIMER_QUEUE_STOP_SCHEDULING);
+    MUTEX_LOCK(pDtlsSession->sslLock);
+    locked = TRUE;
 
-    // do SSL_do_handshake to drive ssl state machine
-    SSL_do_handshake(pDtlsSession->pSsl);
-    CHK_STATUS(dtlsCheckOutgoingDataBuffer(pDtlsSession));
+    CHK(!SSL_is_init_finished(pDtlsSession->pSsl), STATUS_TIMER_QUEUE_STOP_SCHEDULING);
+    // if not timeout yet. try again on next iteration
+    CHK(DTLSv1_get_timeout(pDtlsSession->pSsl, &timeout) == 1, retStatus);
+
+    timeoutValDefaultTimeUnit = (UINT64) timeout.tv_sec * HUNDREDS_OF_NANOS_IN_A_SECOND +
+            (UINT64) timeout.tv_usec * HUNDREDS_OF_NANOS_IN_A_MICROSECOND;
+
+    if (timeoutValDefaultTimeUnit == 0) {
+        // Retransmit the packet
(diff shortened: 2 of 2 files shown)


src/source/Dtls/Dtls.c
```diff
@@ -34,17 +34,37 @@ STATUS dtlsTransmissionTimerCallback(UINT32 timerID, UINT64 currentTime, UINT64
...
 
+    MUTEX_LOCK(pDtlsSession->sslLock);
+    locked = TRUE;
+
     CHK(!SSL_is_init_finished(pDtlsSession->pSsl), STATUS_TIMER_QUEUE_STOP_SCHEDULING);
 
-    // do SSL_do_handshake to drive ssl state machine
-    SSL_do_handshake(pDtlsSession->pSsl);
-    CHK_STATUS(dtlsCheckOutgoingDataBuffer(pDtlsSession));
+    if (DTLSv1_get_timeout(pDtlsSession->pSsl, &timeout) == 0) {
```
Sean-Der: Possible to put `DTLSv1_get_timeout` directly inside `CHK` call?

src/source/Dtls/Dtls.c
```diff
@@ -342,13 +362,21 @@ STATUS dtlsSessionProcessPacket(PDtlsSession pDtlsSession, PBYTE pData, PINT32 p
...
+    } else {
+        // if dtls handshake is done, and SSL_read did not fail, then sslRet and number of sctp bytes read
+        dataLen = sslRet < 0 ? 0 : sslRet;
+    }
 
 CleanUp:
 
     CHK_LOG_ERR(retStatus);
 
+    if (pDataLen != NULL) {
```
Sean-Der: I would drop `pDataLen`, we have a `CHK` call when we enter the function

src/source/Dtls/Dtls.c
```diff
@@ -342,13 +362,21 @@ STATUS dtlsSessionProcessPacket(PDtlsSession pDtlsSession, PBYTE pData, PINT32 p
         LOG_OPENSSL_ERROR("SSL_read");
     }
 
-    *pDataLen = sslRet;
-    CHK_STATUS(dtlsCheckOutgoingDataBuffer(pDtlsSession));
+    if (!SSL_is_init_finished(pDtlsSession->pSsl)) {
+        CHK_STATUS(dtlsCheckOutgoingDataBuffer(pDtlsSession));
+    } else {
+        // if dtls handshake is done, and SSL_read did not fail, then sslRet and number of sctp bytes read
+        dataLen = sslRet < 0 ? 0 : sslRet;
```
Sean-Der: I would just set `*pDataLen` directly, I don't believe the indirection is needed.

src/source/Dtls/Dtls.c
chehefen: fixed

src/source/Dtls/Dtls.c
chehefen: fix

.gitignore
```diff
@@ -11,3 +11,4 @@ open-source/libusrsctp
 open-source/libwebsockets
 open-source/local
 outputs
+open-source/amazon-kinesis-video-streams-producer-sdk-cpp
```
MushMal: Are we sure we need to add this to ignore? I mean, there is a way to modify/send CR for the linked project but perhaps we don't want this now

src/source/Dtls/Dtls.c
```diff
@@ -44,11 +44,8 @@ STATUS dtlsTransmissionTimerCallback(UINT32 timerID, UINT64 currentTime, UINT64
     locked = TRUE;
 
     CHK(!SSL_is_init_finished(pDtlsSession->pSsl), STATUS_TIMER_QUEUE_STOP_SCHEDULING);
-
-    if (DTLSv1_get_timeout(pDtlsSession->pSsl, &timeout) == 0) {
-        // failed to get timeout. try again on next iteration
-        CHK(FALSE, retStatus);
-    }
+    // if not timeout yet. try again on next iteration
+    CHK(DTLSv1_get_timeout(pDtlsSession->pSsl, &timeout) == 1, retStatus);
```
MushMal: Not sure about checking part. The documentation doesn't say anything about the successful return being 1

https://boringssl.googlesource.com/boringssl/+/63fa118f3a3d065adb6ca17227bb9e407ffadccb/include/openssl/ssl.h

Perhaps, this should be CHK(DTLSv1_get_timeout(pDtlsSession->pSsl, &timeout) != 0, retStatus);

.gitignore
Sean-Der: I don't believe we want this, @chehefen was something happening that you needed this? Maybe there is something we need to fix to make this not needed.

PR #37, opened 2019-12-05
Handle case when Rtx payload type doesn't exist in offer/answer

zhiyua-git:
*Issue #, if available:* N/A

*Description of changes:* Update code logic to accept case when rtx payload type doesn't exist, it can still continue to check other codecs instead of exiting.

### DIFF
--- a/src/source/PeerConnection/SessionDescription.c
+++ b/src/source/PeerConnection/SessionDescription.c
@@ -135,7 +135,7 @@ CleanUp:
 /*
  * Populate map with PayloadTypes for codecs a KvsPeerConnection has enabled.
  */
-STATUS setPayloadTypesFromOffer(PHashTable codecTable, PSessionDescription pSessionDescription)
+STATUS setPayloadTypesFromOffer(PHashTable codecTable, PHashTable rtxTable, PSessionDescription pSessionDescription)
 {
     ENTERS();
     STATUS retStatus = STATUS_SUCCESS;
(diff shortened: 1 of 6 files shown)


src/source/PeerConnection/SessionDescription.c
```diff
@@ -230,7 +230,7 @@ STATUS setPayloadTypesFromOffer(PHashTable codecTable, PSessionDescription pSess
     return retStatus;
 }
 
-STATUS setTransceiverPayloadTypes(PHashTable codecTable, PDoubleList pTransceievers) {
+STATUS setTransceiverPayloadTypes(PHashTable codecTable, PDoubleList pTransceiver) {
```
Sean-Der: `pTransceievers` I think is the right name since it is a list of them

Sean-Der:
@zhiyua-git would you mind also getting rid of `RTC_RTX_CODEC_H264_PROFILE_42E01F_LEVEL_ASYMMETRY_ALLOWED_PACKETIZATION_MODE`? We should do a table just for RTX PayloadTypes, I just don't want users to be confused when they see it in the public API.

src/source/PeerConnection/SessionDescription.c
MushMal: +1

zhiyua-git: hmm, one typo to another typo, will fix it.

zhiyua-git:
> @zhiyua-git would you mind also getting rid of `RTC_RTX_CODEC_H264_PROFILE_42E01F_LEVEL_ASYMMETRY_ALLOWED_PACKETIZATION_MODE`? We should do a table just for RTX PayloadTypes, I just don't want users to be confused when they see it in the public API.

Split into RTX type and RTX table.

PR #55, opened 2019-12-07
Decouple rtp rolling buffer logic from rolling buffer

zhiyua-git:
*Issue #, if available:* N/A

*Description of changes:* Previous impl of RTP rolling buffer rely on the header index and tail index impl of rolling buffer. Refactor to decouple two implementations. 
1. Add last index to track which index to look at for the rtp rolling buffer.
2. Bug fix of  sequence number % MAX_UINT16 which results in sequence number lands in a range of [0..65534]. 
3. Add unit test for rtp rolling buffer and fix other tests due to refactoring.

### DIFF
--- a/src/source/Rtp/RtpPacket.h
+++ b/src/source/Rtp/RtpPacket.h
@@ -101,6 +101,8 @@ STATUS createBytesFromRtpPacket(PRtpPacket, PBYTE*, PUINT32);
 STATUS setBytesFromRtpPacket(PRtpPacket, PBYTE, UINT32);
 STATUS constructRtpPackets(PPayloadArray, UINT8, UINT16, UINT32, UINT32, PRtpPacket, UINT32);
 
+#define GET_UINT16_SEQ_NUM(seqIndex) ((UINT16) ((seqIndex) % (MAX_UINT16 + 1)))
+
 #pragma pack(pop, include_i)
 
 #ifdef  __cplusplus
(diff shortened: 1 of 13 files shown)


src/source/Rtp/RtpPacket.h
```diff
@@ -101,6 +101,8 @@ STATUS createBytesFromRtpPacket(PRtpPacket, PBYTE*, PUINT32);
 STATUS setBytesFromRtpPacket(PRtpPacket, PBYTE, UINT32);
 STATUS constructRtpPackets(PPayloadArray, UINT8, UINT16, UINT32, UINT32, PRtpPacket, UINT32);
 
+#define GET_UINT16_SEQ_NUM(seqIndex) ((UINT16) ((seqIndex) % (MAX_UINT16 + 1)))
```
MushMal: careful with MAX_UINT6 + 1. Need to check the standard but this might overflow if the compiler doesn't auto upgrade the type if seqIndex just happed to be UINT16. Better to use explicit cast like (UINT32)MAX_UINT16 + 1

zhiyua-git: MAX_UINT16 is defined as 65535 not as (UINT16) 65535, so if it is running on 32/64 bit device, that MAX_UINT16+1 should be good enough.

MushMal: Actually, MAX_INT + 1 should be promoted to INT32 by the compiler, this should be OK

Issue #56, opened 2019-12-07
Sdk crashes when new PeerConnection establishes new connection


chehefen:
Steps to reproduce:
- start kvsWebrtcClientMasterSample with any channel name. 
- start js viewer in any connection mode. Ensure data channel is enabled. Connection should be established successfully and starts streaming.
- stop the js viewer.
- wait until connectionStateChange callback is called and trigger peerConnection teardown, the following log should be seen
```
2019-12-07 01:53:57 DEBUG   main(): Completed freeing peerConnection and Reinitializing peerConnection
2019-12-07 01:53:57 VERBOSE signalingClientGetIceConfigInfoCout(): Signaling Client Get ICE Config Info Count
2019-12-07 01:53:57 VERBOSE signalingClientGetIceConfigInfo(): Signaling Client Get ICE Config Info
2019-12-07 01:53:57 DEBUG   stepIceAgentStateMachine(): Ice agent state changed from ICE_AGENT_STATE_NONE to ICE_AGENT_STATE_NEW
```
- start the js viewer again, then the program will crash.
- stacktrace:
```
* thread #8, stop reason = EXC_BAD_ACCESS (code=1, address=0x10)
  * frame #0: 0x0000000100101c8d libkvsWebrtcClient.dylib`onSctpOutboundPacket(addr=0x0000000000000000, data=0x0000000100731940, length=16, tos='\0', set_df='\0') at Sctp.c:150:5
    frame #1: 0x0000000100387dbc libusrsctp.1.dylib`sctp_send_resp_msg(src=0x000070000b992ba0, dst=0x000070000b992b90, sh=0x000000010073cf60, vtag=4152928290, type='\x06', cause=0x0000000000000000, vrf_id=0, port=0) at sctp_output.c:12021:10
    frame #2: 0x000000010037ab15 libusrsctp.1.dylib`sctp_send_abort(m=0x000000010073cf10, iphlen=0, src=0x000070000b992ba0, dst=0x000070000b992b90, sh=0x000000010073cf60, vtag=4152928290, cause=0x0000000000000000, vrf_id=0, port=0) at sctp_output.c:13091:2
[...]

PR #57, opened 2019-12-07
fix sctp lifecycle handling

chehefen:
*Issue #, if available:*
#56 
*Description of changes:*
Properly manage sctp life cycle. Now multiple peerConnection with sctp should work. And sctp is shutdown when its parent peerConnection is freed.

### DIFF
--- a/src/include/com/amazonaws/kinesis/video/webrtcclient/Include.h
+++ b/src/include/com/amazonaws/kinesis/video/webrtcclient/Include.h
@@ -1175,6 +1175,13 @@ PUBLIC_API STATUS freeTransceiver(PRtcRtpTransceiver*);
 PUBLIC_API STATUS initKvsWebRtc(VOID);
 
 /*
+ * deinitKvsWebRtc deinitializes global state needed for all RtcPeerConnection's it must only be called once
+ *
+ * @return - STATUS code of the execution
+ */
+PUBLIC_API STATUS deinitKvsWebRtc(VOID);
+
+/*
  * addSupportedCodec adds to the list of codecs we support receiving. The remote MUST only send codecs we declare
  *
  * @param - RTC_CODEC - IN - Codec that we support receiving.
--- a/src/source/PeerConnection/PeerConnection.c
+++ b/src/source/PeerConnection/PeerConnection.c
@@ -866,10 +866,20 @@ STATUS initKvsWebRtc(VOID)
     // SSL_library_init() always returns "1", so it is safe to discard the return value.
     UNUSED_PARAM(SSL_library_init());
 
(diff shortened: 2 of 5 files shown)


src/include/com/amazonaws/kinesis/video/webrtcclient/Include.h
```diff
@@ -1174,6 +1174,13 @@ PUBLIC_API STATUS freeTransceiver(PRtcRtpTransceiver*);
  */
 PUBLIC_API STATUS initKvsWebRtc(VOID);
 
+/*
+ * deinitKvsWebRtc deinitializes global state needed for all RtcPeerConnection's it must only be called once
+ *
+ * @return - STATUS code of the execution
+ */
+PUBLIC_API STATUS deinitKvsWebRtc(VOID);
```
MushMal: Let's come up with a better name for this API. So far we have pairs of  createXXX freeXXX which create and free the entire object. The initialization has been named like initializeXXX and releaseXXX. Though, I don't see a major difference - just suggest to have full names - like initialize instead of init

src/source/PeerConnection/PeerConnection.c
```diff
@@ -866,10 +866,20 @@ STATUS initKvsWebRtc(VOID)
     // SSL_library_init() always returns "1", so it is safe to discard the return value.
     UNUSED_PARAM(SSL_library_init());
 
-    usrsctp_init(0, &onSctpOutboundPacket, NULL);
+    CHK_STATUS(initSctpSession());
```
MushMal: same for initSctp ->initializeSctp

src/source/PeerConnection/PeerConnection.c
```diff
@@ -866,10 +866,20 @@ STATUS initKvsWebRtc(VOID)
...
+CleanUp:
+
+    LEAVES();
+    return retStatus;
+
+}
+
+STATUS deinitKvsWebRtc(VOID)
```
MushMal: deinitializeKvs... think also about using uninitializeXXX and releaseXXX

src/source/Sctp/Sctp.c
```diff
@@ -55,6 +55,26 @@ STATUS configureSctpSocket(struct socket *socket)
...
+    // Disable Explicit Congestion Notification
+    usrsctp_sysctl_set_sctp_ecn_enable(0);
+
+    return retStatus;
+}
+
+VOID deinitSctpSession()
+{
+    // need to block until usrsctp_finish or sctp thread could be calling free objects and cause segfault
+    while (usrsctp_finish() != 0) {
```
MushMal: Break out after some larger timeout. You can't loop indefinitely

PR #58, opened 2019-12-07
Enable compiler warnings when building tests

Sean-Der:
Before we only built our samples/library code with warnings, this builds
the tests with them enabled as well


tst/SignalingApiFunctionalityTest.cpp
```diff
@@ -43,17 +43,15 @@ STATUS masterMessageReceived(UINT64 customData, PReceivedSignalingMessage pRecei
 
 STATUS signalingClientStateChanged(UINT64 customData, SIGNALING_CLIENT_STATE state)
 {
-    SignalingApiFunctionalityTest *pTest = (SignalingApiFunctionalityTest*) customData;
-
+    UNUSED_PARAM(customData);
```
unicornss: nice.. wondering if we should add a comment on the intent of this param on local contexts...

PR #59, opened 2019-12-09
Don't treat warnings as errors when building usrsctp

Sean-Der:
Remove our patch to remove the warnings we were hitting on builders


unicornss:
can we confirm if the build succeeds in Ubuntu?

Sean-Der:
@unicornss builds on Mac+Ubuntu on travis-ci!

On Debian Sid builds locally as well. We can wait until tomorrow to merge though (so we can confirm on jenkins as well) no rush from me. There is just a new build error for me using GCC 9

PR #60, opened 2019-12-09
Setting correlation id to empty in the samples and the tests

MushMal:
*Issue #, if available:*

*Description of changes:*


cmake-scripts/libwebsockets-CMakeLists.txt
```diff
@@ -25,7 +25,7 @@ ExternalProject_Add(project_libwebsockets
         -DLWS_WITH_SHARED=1
         -DLWS_STATIC_PIC=1
         -DLWS_WITH_ZLIB=0
-# enable for debug output        -DCMAKE_BUILD_TYPE=DEBUG
+        -DCMAKE_BUILD_TYPE=DEBUG
```
Sean-Der: Should we remove? Worth making an argument?

src/source/PeerConnection/PeerConnection.h
```diff
@@ -99,6 +99,8 @@ VOID onSctpSessionDataChannelOpen(UINT64, UINT32, PBYTE, UINT32);
 
 STATUS sendPacketToRtpReceiver(PKvsPeerConnection, PBYTE, UINT32);
 
+STATUS generateJSONSafeString(PCHAR, UINT32);
```
Sean-Der: Is it a good pattern to expose functions that aren't used outside of a compilation file?

MushMal: Not the best idea really. Even for the tests.. Let me fix this asap

PR #68, opened 2019-12-12
Aligned memory access

MushMal:
*Issue #, if available:*

*Description of changes:*


MushMal:
This is still a WIP as need to be tested on kernel trap

Sean-Der:
@MushMal anyway you know of that we can test this in travis (to make sure we don't regress?)

src/include/com/amazonaws/kinesis/video/webrtcclient/Include.h
```diff
@@ -432,8 +432,8 @@ typedef VOID (*RtcOnMessage)(UINT64, BOOL, PBYTE, UINT32);
  *
  * https://www.w3.org/TR/webrtc/#dom-rtcpeerconnection-ondatachannel
  */
-typedef struct __RtcDataChannel RtcDataChannel;
-typedef VOID (*RtcOnDataChannel)(UINT64, RtcDataChannel*);
+struct __RtcDataChannel;
```
MushMal: Why these changes? If we don't typedef then we need to use struct keyword in every place of usage

src/source/Ice/IceAgent.h
```diff
@@ -151,7 +151,7 @@ typedef struct {
 
     UINT32 foundationCounter;
 
-    PTurnConnection pTurnConnection;
+    struct __TurnConnection* pTurnConnection;
```
MushMal: same

src/include/com/amazonaws/kinesis/video/webrtcclient/Include.h
```diff
@@ -432,8 +432,8 @@ typedef VOID (*RtcOnMessage)(UINT64, BOOL, PBYTE, UINT32);
...
+typedef VOID (*RtcOnDataChannel)(UINT64, struct __RtcDataChannel*);
```
MushMal: same

src/source/Ice/IceUtils.h
```diff
@@ -34,7 +34,7 @@ STATUS iceUtilsGenerateTransactionId(PBYTE, UINT32);
 
 // Stun packaging and sending functions
 STATUS iceUtilsPackageStunPacket(PStunPacket, PBYTE, UINT32, PBYTE, PUINT32);
-STATUS iceUtilsSendStunPacket(PStunPacket, PBYTE, UINT32, PKvsIpAddress, PSocketConnection, PTurnConnection, BOOL);
+STATUS iceUtilsSendStunPacket(PStunPacket, PBYTE, UINT32, PKvsIpAddress, PSocketConnection, struct __TurnConnection*, BOOL);
```
MushMal: same for here and all the instances where we need to use struct keyword

src/include/com/amazonaws/kinesis/video/webrtcclient/Include.h
MushMal: Could you please describe which compilers don't do forward ref? Every one of the compilers should handle it

zhiyua-git: https://stackoverflow.com/questions/8367646/redefinition-of-typedef
This doesn’t work for pure C.

MushMal: I believe it's about type re-declaration and not forward decl

zhiyua-git: It is the forwarding typedef. We had same issue on C producer on the old gcc compiler before.

PR #73, opened 2019-12-16
Add gcc 4.4 build to travis

zhiyua-git:
*Issue #, if available:* N/A

*Description of changes:* 
1. Check if we can add old version of gcc to travis build.
2. Add fix for the build failure.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -22,7 +22,7 @@ install:
 script:
   - make
   - ulimit -c unlimited -S
-  - timeout --signal=SIGABRT 10m ./tst/webrtc_client_test
+  - if [ "$CC" != "gcc-4.4" ]; then timeout --signal=SIGABRT 10m ./tst/webrtc_client_test; fi
 
 after_failure:
   - for i in $(find ./ -maxdepth 1 -name 'core*' -print); do gdb $(pwd)/tst/webrtc_client_test core* -ex "thread apply all bt" -ex "set pagination 0" -batch; done;
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,6 +9,11 @@ cache:
 - directories:
   - $HOME/.cache
 
+addons:
+  apt:
+    packages:
+      - gcc-4.4
+      - g++-4.4
 install:
   - sudo apt-get install -y gdb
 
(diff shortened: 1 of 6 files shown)


Sean-Der:
@zhiyua-git maybe you can do an old version of Linux just for that build? You can specify https://docs.travis-ci.com/user/reference/precise/ and that should get you something legacy!

.travis.yml
```diff
@@ -9,13 +9,20 @@ cache:
...
+      - gcc-4.4
+      - g++-4.4
 install:
   - sudo apt-get install -y gdb
 
 script:
   - make
   - ulimit -c unlimited -S
-  - timeout --signal=SIGABRT 10m ./tst/webrtc_client_test
+  - if [ "$CC" != "gcc-4.4" ]; then timeout --signal=SIGABRT 10m ./tst/webrtc_client_test; fi
```
Sean-Der: @zhiyua-git You can remove this change, the `timeout` command isn't compiler specific. It just kills the test executable after 10 minutes.

It is useful for detecting deadlocks.

.travis.yml
```diff
@@ -85,3 +92,11 @@ matrix:
       os: linux
       compiler: clang
       before_script: mkdir build && cd build && cmake .. -DBUILD_TEST=TRUE -DUNDEFINED_BEHAVIOR_SANITIZER=TRUE
+
+    # Old Version GCC 4.4
+    - name: "Linux GCC 4.4 Build"
+      os: linux
+      compiler: gcc
+      before_script:
+        - export CC=gcc-4.4 && export CXX=g++-4.4 && mkdir build && cd build && cmake .. -DBUILD_TEST=FALSE
```
Sean-Der: You don't have to set `BUILD_TEST=FALSE` it should be off by default!

If not mind turning it off by default? I don't want users to be force to build C++ code if they don't have to!

.travis.yml
zhiyua-git: We are using c++11 standard to write our unit/integ tests and that is not available in the old gcc compiler we use. Maybe I can try to only override CC instead of doing both. Will check if that is possible in another PR given this one is merged.

.travis.yml
zhiyua-git: Will remove in a separate PR.

PR #76, opened 2019-12-16
use MAX_JSON_TOKEN_COUNT in parsing sdp json

chehefen:
*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -20,6 +20,7 @@ install:
   - sudo apt-get install -y gdb
 
 script:
+  - export AWS_KVS_LOG_LEVEL=2
   - make
   - ulimit -c unlimited -S
   - if [ "$CC" != "gcc-4.4" ]; then timeout --signal=SIGABRT 10m ./tst/webrtc_client_test; fi
--- a/src/source/PeerConnection/SessionDescription.h
+++ b/src/source/PeerConnection/SessionDescription.h
@@ -10,9 +10,6 @@ SessionDescription internal include file
 extern "C" {
 #endif
 
-#define SESSION_DESCRIPTION_TOKENS 5
-#define ICE_CANDIDATE_INIT_TOKENS 7
-
 #define SESSION_DESCRIPTION_INIT_LINE_ENDING "\\r\\n"
 
 #define SESSION_DESCRIPTION_INIT_TEMPLATE_HEAD "{\"type\": \"%s\", \"sdp\": \""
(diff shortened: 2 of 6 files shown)


.travis.yml
```diff
@@ -20,6 +20,7 @@ install:
   - sudo apt-get install -y gdb
 
 script:
+  - export AWS_KVS_LOG_LEVEL=2
```
Sean-Der: Do we need this?

MushMal: This is the debug level. We can get some info from the runs.

src/source/PeerConnection/SessionDescription.h
```diff
@@ -10,8 +10,8 @@ SessionDescription internal include file
 extern "C" {
 #endif
 
-#define SESSION_DESCRIPTION_TOKENS 5
-#define ICE_CANDIDATE_INIT_TOKENS 7
+// For tight packing
+#pragma pack(push, include_i, 1) // for byte alignment
```
MushMal: Remove these - we don't tight pack any longer or merge from mainline

tst/PeerConnectionFunctionalityTest.cpp
```diff
@@ -19,17 +34,28 @@ TEST_F(PeerConnectionFunctionalityTest, connectTwoPeers)
...
 
     auto onICECandidateHdlr = [](UINT64 customData, PCHAR candidateStr) -> void {
-        RtcIceCandidateInit iceCandidate;
+        PAddIceCandidateRoutineArgs pAddIceCandidateRoutineArgs = NULL;
+        TID tid;
 
         if (candidateStr != NULL) {
-            EXPECT_EQ(deserializeRtcIceCandidateInit(candidateStr, STRLEN(candidateStr), &iceCandidate), STATUS_SUCCESS);
-            EXPECT_EQ(addIceCandidate((PRtcPeerConnection) customData, iceCandidate.candidate), STATUS_SUCCESS);
+            pAddIceCandidateRoutineArgs = (PAddIceCandidateRoutineArgs) MEMALLOC(SIZEOF(AddIceCandidateRoutineArgs));
```
MushMal: Doesn't need to be allocated/freed - can be on the stack variable

chehefen:
> Thank you so much for digging into this! How did you find/diagnose the deadlock? Would love to know what helped/what can be improved debugging wise.

I couldnt pass the travis checks so I run the test locally myself and found the deadlock.

PR #78, opened 2019-12-18
Refreshing ICE should not require re-connection

MushMal:
*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/src/source/Signaling/Signaling.c
+++ b/src/source/Signaling/Signaling.c
@@ -14,6 +14,7 @@ STATUS createSignalingSync(PSignalingClientInfo pClientInfo, PChannelInfo pChann
     PCHAR userLogLevelStr = NULL;
     UINT32 userLogLevel;
     struct lws_context_creation_info creationInfo;
+    PStateMachineState pStateMachineState;
 
     CHK(pClientInfo != NULL &&
          pChannelInfo != NULL &&
@@ -137,6 +138,14 @@ STATUS createSignalingSync(PSignalingClientInfo pClientInfo, PChannelInfo pChann
 
     // Set the time out before execution
     pSignalingClient->stepUntil = GETTIME() + SIGNALING_CREATE_TIMEOUT;
+
+    // Notify of the state change initially as the state machinery is already in the NEW state
+    if (pSignalingClient->signalingClientCallbacks.stateChangeFn != NULL) {
+        CHK_STATUS(getStateMachineCurrentState(pSignalingClient->pStateMachine, &pStateMachineState));
+        CHK_STATUS(pSignalingClient->signalingClientCallbacks.stateChangeFn(
+                pSignalingClient->signalingClientCallbacks.customData,
+                getSignalingStateFromStateMachineState(pStateMachineState->state)));
+    }
 
     // Prime the state machine
     CHK_STATUS(stepSignalingStateMachine(pSignalingClient, STATUS_SUCCESS));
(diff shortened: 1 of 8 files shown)


src/source/Signaling/Signaling.c
```diff
@@ -434,12 +461,14 @@ STATUS refreshIceConfigurationCallback(UINT32 timerId, UINT64 scheduledTime, UIN
...
 
-    // Check if we are in a connected state and if not bail
+    // Check if we are in a connected state or ready state and if not bail.
+    // The ICE state will be called in any other states
     CHK_STATUS(getStateMachineCurrentState(pSignalingClient->pStateMachine, &pStateMachineState));
-    CHK(pStateMachineState->state == SIGNALING_STATE_CONNECTED, retStatus);
+    CHK(pStateMachineState->state == SIGNALING_STATE_CONNECTED ||
+        pStateMachineState->state == SIGNALING_STATE_READY, retStatus);
 
     // Force terminate the connection for now as LWS doesn't allow parallel processing
```
chehefen: is this comment still valid?

src/source/Signaling/Signaling.c
```diff
@@ -434,12 +461,14 @@ STATUS refreshIceConfigurationCallback(UINT32 timerId, UINT64 scheduledTime, UIN
...
-    terminateConnectionWithStatus(pSignalingClient, SERVICE_CALL_RESULT_SIGNALING_RECONNECT_ICE);
+    ATOMIC_STORE(&pSignalingClient->result, (SIZE_T) SERVICE_CALL_RESULT_SIGNALING_RECONNECT_ICE);
```
chehefen: It looks like there are lots of signals passed around using the pSignalingClient ->result variable. Any chance it can get overridden? If so are we ok with that?

src/source/Signaling/Signaling.c
```diff
@@ -373,11 +389,19 @@ STATUS validateSignalingClientInfo(PSignalingClient pSignalingClient, PSignaling
...
-    CHK(pClientInfo->version <= SIGNALING_CLIENT_INFO_CURRENT_VERSION, STATUS_SIGNALING_INVALID_CLIENT_INFO_VERSION);
+    CHK((pClientInfo->version & (~SIGNALING_CLIENT_INFO_INTERNAL_SENTINEL_BIT)) <= SIGNALING_CLIENT_INFO_CURRENT_VERSION, STATUS_SIGNALING_INVALID_CLIENT_INFO_VERSION);
     CHK(STRNLEN(pClientInfo->clientId, MAX_SIGNALING_CLIENT_ID_LEN + 1) <= MAX_SIGNALING_CLIENT_ID_LEN, STATUS_SIGNALING_INVALID_CLIENT_INFO_CLIENT_LENGTH);
 
-    // Store and validate
-    pSignalingClient->clientInfo = *pClientInfo;
+    // Store and validate.
+    // NOTE: We will check whether an internal structure has been passed in for testing purposes by
+    // checking the sentinel bit is set.
+    if ((pClientInfo->version & SIGNALING_CLIENT_INFO_INTERNAL_SENTINEL_BIT) != 0) {
```
chehefen: It's not very obvious what SIGNALING_CLIENT_INFO_INTERNAL_SENTINEL_BIT is supposed to, and it feels hacky to overload version with this extra functionality. Cant we just add a boolean?

src/source/Signaling/Signaling.c
MushMal: D'oh - will change the comment. Thanks for pointing out

src/source/Signaling/Signaling.c
MushMal: The state machine that's checking these have a serialized access

src/source/Signaling/Signaling.c
MushMal: The issue is that we are trying to not expose this to the customers code. Yes it's kind of hacky but this is supposed to be used internally and for test injection purposes only

chehefen: what about adding a boolean and a compile time flag to check against the boolean? The issue with your approach is that it is gonna set a style here and once it spreads we will have two versions of version which just adds more unnecessary complexity.

MushMal: The main idea here is to NOT expose any of this in the public headers.

src/source/Signaling/StateMachine.c
```diff
@@ -649,6 +708,13 @@ STATUS executeDisconnectedSignalingState(UINT64 customData, UINT64 time)
 
     CHK(pSignalingClient != NULL, STATUS_NULL_ARG);
 
+    // Notify of the state change
+    if (pSignalingClient->signalingClientCallbacks.stateChangeFn != NULL) {
```
Sean-Der: Can we make `stateChangeFn` mandatory?

I am afraid that in the future a NULL check may be missed and could cause an issue.

MushMal: Not sure if we need to make this a mandatory. What we do in KVS Producer SDK is we provide a default dummy implementation if the customer doesn't specify the functionality. We can add it later - not in this PR

PR #97, opened 2019-12-24
add peerConnection detect disconnection test

chehefen:
*Issue #, if available:*

*Description of changes:*

add peerConnection detect disconnection test


tst/PeerConnectionFunctionalityTest.cpp
```diff
@@ -106,6 +109,77 @@ TEST_F(PeerConnectionFunctionalityTest, connectTwoPeersForcedTURN)
...
+            THREAD_SLEEP(HUNDREDS_OF_NANOS_IN_A_SECOND);
+        }
+    }
+
+    EXPECT_TRUE(connected == TRUE);
+
+    // free offerPc so it wont send anymore keep alives and answerPc will detect disconnection
+    freePeerConnection(&offerPc);
+
+    THREAD_SLEEP((UINT64) (KVS_ICE_ENTER_STATE_DISCONNECTION_GRACE_PERIOD * 1.2));
```
Sean-Der: Is it possible this `1.2` can be racey? I would poll in a loop and do a higher time, in the good case we bail quickly but still avoid a possibly flakey test

src/source/PeerConnection/PeerConnection.c
```diff
@@ -84,12 +85,20 @@ VOID onInboundPacket(UINT64 customData, PBYTE buff, UINT32 buffLen)
...
+            locked = TRUE;
+            if (pKvsPeerConnection->pSrtpSession == NULL) {
+                CHK_STATUS(allocateSrtp(pKvsPeerConnection));
+            }
+
+            if (pKvsPeerConnection->pSctpSession == NULL) {
                 CHK_STATUS(allocateSctp(pKvsPeerConnection));
             }
+
+            MUTEX_UNLOCK(pKvsPeerConnection->peerConnectionObjLock);
```
Sean-Der: Do we need to drop the lock? Looking at the code nothing else happens.

Do you anticipate lock contention?

src/source/PeerConnection/PeerConnection.c
```diff
@@ -882,6 +899,10 @@ STATUS deinitKvsWebRtc(VOID)
 
     deinitSctpSession();
 
+    srtp_shutdown();
```
Sean-Der: Nice! How did you find this? Did tooling alert you of a leak?

src/source/PeerConnection/PeerConnection.c
```diff
@@ -84,12 +85,20 @@ VOID onInboundPacket(UINT64 customData, PBYTE buff, UINT32 buffLen)
...
 
-        if (pKvsPeerConnection->pSrtpSession == NULL) {
-            dtlsSessionIsInitFinished(pKvsPeerConnection->pDtlsSession, &isDtlsConnected);
-            if (isDtlsConnected) {
-                allocateSrtp(pKvsPeerConnection);
+        CHK_STATUS(dtlsSessionIsInitFinished(pKvsPeerConnection->pDtlsSession, &isDtlsConnected));
+        if (isDtlsConnected) {
+            MUTEX_LOCK(pKvsPeerConnection->peerConnectionObjLock);
+            locked = TRUE;
+            if (pKvsPeerConnection->pSrtpSession == NULL) {
```
Sean-Der: These `NULL` checks seem fragile.

Maybe we should have the struct exist, but then pass the keys into SRTP (and start SCTP) when we are actually ready?

tst/PeerConnectionFunctionalityTest.cpp
```diff
@@ -106,6 +109,77 @@ TEST_F(PeerConnectionFunctionalityTest, connectTwoPeersForcedTURN)
...
+    if (!mAccessKeyIdSet) {
+        return;
+    }
+
+    RtcConfiguration configuration;
+    PRtcPeerConnection offerPc = NULL, answerPc = NULL;
+    UINT32 i;
+    RtcSessionDescriptionInit sdp;
+    BOOL connected = FALSE;
+    SIZE_T stateChangeCount[7];
```
Sean-Der: Can we determine the `7` at runtime? Just so it is a little more self documenting what this contains.

src/source/PeerConnection/PeerConnection.c
chehefen: no, the new test I added fails when running the entire test suite. Turns out that its initKvsWebRtc fails at srtp_init.

src/source/PeerConnection/PeerConnection.c
chehefen: The NULL check is inside `isDtlsConnected` though, so we are doing sctp and srtp allocate when we are actually ready. Because both SRTP and SCTP are initialized by the allocate call (key passing, start sctp etc.), the only way to check whether SRTP and SCTP has started seems to be the null check. To avoid duplicate allocate calls, I put them inside the lock.

src/source/PeerConnection/PeerConnection.c
chehefen: I'm using the lock to avoid duplicate allocateSrtp and allocateSctp calls, so I dont need to hold it anymore after allocate calls are done.

tst/PeerConnectionFunctionalityTest.cpp
chehefen: fixed

tst/PeerConnectionFunctionalityTest.cpp
chehefen: added a new enum for counting the total number of states.

src/source/PeerConnection/PeerConnection.c
Sean-Der: Does the ICE Agent dispatch packets concurrently? A lot of the PeerConnection code is written to assume it will be processing one packet a time

chehefen: actually it doesnt dispatch packets concurrently. so we dont actually need the lock here.

Issue #45, opened 2019-12-06
Support non-rsize RTCP


Sean-Der:
Currently when we parse RTCP we only look at the first in the compound packet. Because of this NACKs only work with clients that support (breaking FireFox)

PR #104, opened 2019-12-30
refactor some ice code. Refactor peerConnectionFunctionality test to …

chehefen:
…store rtcStateTransition count in test base class.

*Issue #, if available:*

*Description of changes:*

refactor some ice code. Refactor peerConnectionFunctionality test to store rtcStateTransition count in test base class.


src/source/Ice/IceAgentStateMachine.c
```diff
@@ -383,10 +383,14 @@ STATUS executeGatheringIceAgentState(UINT64 customData, UINT64 time)
             }
             CHK_STATUS(turnConnectionStart(pIceAgent->pTurnConnection));
 
+            turnServerCount++;
+
             break;
         }
     }
 
+    CHK(pIceAgent->iceTransportPolicy != ICE_TRANSPORT_POLICY_RELAY || turnServerCount > 0, STATUS_ICE_MISSING_TURN_SERVER_IN_TURN_ONLY_MODE);
```
Sean-Der: I don't think we need this check `new RTCPeerConnection({iceTransportPolicy: 'relay'})` doesn't error out in the browser.

I need to confirm, but the following *might* be true.
* The browser might still be gathering+sending connectivity checks (but just not emitting host/prflx candidates)
* There might be a case where relayonly (but no TURN server is given is useful) I can't think of any, but just better not to deviate IMO.

src/source/Ice/IceAgentStateMachine.c
```diff
@@ -405,6 +405,15 @@ STATUS executeGatheringIceAgentState(UINT64 customData, UINT64 time)
 
     CHK_LOG_ERR_NV(retStatus);
 
+    if (STATUS_FAILED(retStatus)) {
+        pIceAgent->iceAgentStatus = retStatus;
+        // step into failed state
+        stepStateMachine(pIceAgent->pStateMachine);
+
+        // fix up retStatus so we can successfully transition to failed state.
+        retStatus = STATUS_SUCCESS;
```
Sean-Der: So only `STATUS_SUCCESS` is possible? Is it worth having these functions return a status if they don't mean anything?

chehefen: the state machine is a pic construct. The way it works is that if any of the state callbacks return fail status then it bails immediately. But I want it to move to failed state instead, so I have to store the error status in pIceAgent->iceAgentStatus and  fix up the retStatus.

Sean-Der: Aye, ok will approve!

src/source/Rtcp/RtcpPacket.c
```diff
@@ -14,7 +15,7 @@ STATUS setRtcpPacketFromBytes(PBYTE pRawPacket, UINT32 pRawPacketsLen, PRtcpPack
     // We don't assert exact length since this may be a compound packet, it
     // is the callers responsibility to parse subsequent entries
     packetLen = getInt16(*(PUINT16) (pRawPacket + RTCP_PACKET_LEN_OFFSET));
-    CHK((packetLen + 1) * RTCP_PACKET_LEN_WORD_SIZE >= pRawPacketsLen, STATUS_RTCP_INPUT_PACKET_LEN_MISMATCH);
+    CHK((packetLen + 1) * RTCP_PACKET_LEN_WORD_SIZE >= pRawPacketsLen, STATUS_SUCCESS);
```
yuma-m: I think this line checks the packet length defined in [RFC 4585 Section 6](https://tools.ietf.org/html/rfc4585#section-6).
Why is this line changed from error to success?
Do you mean C SDK will ignore compound RTCP packet?

Sean-Der: Yes it does ignore compound packets, we do want to fix this though in the short term.

https://github.com/awslabs/amazon-kinesis-video-streams-webrtc-sdk-c/issues/45 is the ticket.

PR #118, opened 2020-01-04
in iceAgent, pre-allocate stun packets needed to establish connection…

chehefen:
…. Fix deadlock when using turn.

*Issue #, if available:*

*Description of changes:*

in iceAgent, pre-allocate stun packets needed to establish connection. Fix deadlock when using turn.

### DIFF
--- a/src/source/Ice/ConnectionListener.c
+++ b/src/source/Ice/ConnectionListener.c
@@ -153,6 +153,41 @@ CleanUp:
     return retStatus;
 }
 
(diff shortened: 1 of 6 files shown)


Sean-Der:
@chehefen sorry for the pushback, but do you have any data on how this improves performance?

chehefen:
> @chehefen sorry for the pushback, but do you have any data on how this improves performance? Is this also a code cleanliness change?

previously it was allocating and freeing new stunPacket in everytime ice timerCallback fires. now it only allocate once at each state and reuse it in subsequence timer calls.

Issue #128, opened 2020-01-07
(BUG) CHK_WARN CHK_ERR should not be inside MUTEX_LOCK


ek9852:
There are many case  where CHK_WARN or CHK_ERR or CKH_STATUS are inside a MUTEX_LOCK block, which would skip the mutex unlock when error occurs.

Such as :  https://github.com/awslabs/amazon-kinesis-video-streams-webrtc-sdk-c/blob/master/samples/Common.c#L109 AND  https://github.com/awslabs/amazon-kinesis-video-streams-webrtc-sdk-c/blob/master/samples/Common.c#L114

The CHK_WARN, CHK_ERR just use a goto cleanup when error occurs, and will not unlock the mutex. 

Pls make some test cases and polish the code.

chertov:
but mutex will unlock here https://github.com/awslabs/amazon-kinesis-video-streams-webrtc-sdk-c/blob/master/samples/Common.c#L70

ek9852:
Not in https://github.com/awslabs/amazon-kinesis-video-streams-webrtc-sdk-c/blob/master/samples/Common.c#L109 AND https://github.com/awslabs/amazon-kinesis-video-streams-webrtc-sdk-c/blob/master/samples/Common.c#L114

Sean-Der:
Fixed with d633a918a2dd125bb1e0f4d0196056a0e15dac0d

PR #131, opened 2020-01-08
Fix missing mutex_unlock on error

ek9852:
Issue: https://github.com/awslabs/amazon-kinesis-video-streams-webrtc-sdk-c/issues/128

The CHK_WARN, CHK_ERR, CHK_STATUS will do a goto
and might skip the MUTEX_UNLOCK, there
are some code path does not happened that properly.

Fixing it.

### DIFF
--- a/samples/Common.c
+++ b/samples/Common.c
@@ -92,10 +92,12 @@ STATUS masterMessageReceived(UINT64 customData, PReceivedSignalingMessage pRecei
     PSampleConfiguration pSampleConfiguration = (PSampleConfiguration) customData;
     PSampleStreamingSession pSampleStreamingSession = NULL;
     UINT32 i;
+    BOOL locked = FALSE;
 
     CHK(pSampleConfiguration != NULL, STATUS_NULL_ARG);
 
     MUTEX_LOCK(pSampleConfiguration->sampleConfigurationObjLock);
+    locked = TRUE;
 
     // ice candidate message and offer message can come at any order. Therefore, if we see a new peerId, then create
     // a new SampleStreamingSession, which in turn creates a new peerConnection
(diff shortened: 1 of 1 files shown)


ek9852:
@chehefen, @Sean-Der

Sean-Der:
Fantastic, thank you so much @ek9852 

I wish we had a better way to this, have a scoped locks would be really nice (but not portable unfortunately). Maybe we can come up with some tooling that requires `unlock` can only be after `CleanUp:` ?

ek9852:
This might increase the locked time for the code if only unlock in cleanup and depends on how long those are. Unfortunately this is a c library, otherwise we can use c++ object lock to lock a scope of code and release it after it is out of scope.

Sean-Der:
> This might increase the locked time for the code if only unlock in cleanup and depends on how long those are. Unfortunately this is a c library, otherwise we can use c++ object lock to lock a scope of code and release it after it is out of scope.

That's a good point. Maybe if you are doing a lock you need to push the code that takes the lock into a smaller function.

I just really hate walking away from an issue with fixing the 'root cause'

ek9852:
me too, but there are quite a number of code inside the sdk using the locked variable to do that.
I don't have a brilliant idea on that just using pure portable C. Possible that the  reason why there is another implement using Go for webrtc. :)

Sean-Der:
@ek9852 merged with d633a918a2dd125bb1e0f4d0196056a0e15dac0d

Whoever did Go in WebRTC sounds like they really to make life hard on themselves, I would avoid them :p 

I really appreciate you bringing this issue up, and anything you bring up on this repo I would love to improve! I am all over the place, but I really love getting others involved/getting feedback :)

PR #149, opened 2020-01-16
Unit test for TURN connection send data lock

zhiyua-git:
*Issue #, if available:*
KinesisVideo - 5340

*Description of changes:*
Adding unit test for the lock added for https://github.com/awslabs/amazon-kinesis-video-streams-webrtc-sdk-c/commit/75e47eb09d2ef491f962a8d04dafd5a0882cd748.
Thread sanitizer would fail if the above commit is reverted.

### DIFF
--- a/tst/TurnConnectionFunctionalityTest.cpp
+++ b/tst/TurnConnectionFunctionalityTest.cpp
@@ -84,6 +84,7 @@ namespace com { namespace amazonaws { namespace kinesis { namespace video { name
 
         // wait until channel is created
         while(!turnReady) {
+            THREAD_SLEEP(100 * HUNDREDS_OF_NANOS_IN_A_MILLISECOND);
             MUTEX_LOCK(pTurnConnection->lock);
             if (pTurnConnection->state == TURN_STATE_READY) {
                 turnReady = TRUE;
@@ -111,6 +112,7 @@ namespace com { namespace amazonaws { namespace kinesis { namespace video { name
         turnReady = FALSE;
 
         while(!turnReady) {
+            THREAD_SLEEP(100 * HUNDREDS_OF_NANOS_IN_A_MILLISECOND);
             MUTEX_LOCK(pTurnConnection->lock);
             if (pTurnConnection->state == TURN_STATE_READY) {
                 turnReady = TRUE;
@@ -158,6 +160,7 @@ namespace com { namespace amazonaws { namespace kinesis { namespace video { name
 
         // wait until channel is created
         while(!turnReady) {
+            THREAD_SLEEP(100 * HUNDREDS_OF_NANOS_IN_A_MILLISECOND);
             MUTEX_LOCK(pTurnConnection->lock);
             if (pTurnConnection->state == TURN_STATE_READY) {
                 turnReady = TRUE;
@@ -246,6 +249,7 @@ namespace com { namespace amazonaws { namespace kinesis { namespace video { name
 
         // wait until channel is created
         while(!turnReady) {
+            THREAD_SLEEP(100 * HUNDREDS_OF_NANOS_IN_A_MILLISECOND);
             MUTEX_LOCK(pTurnConnection->lock);
             if (pTurnConnection->state == TURN_STATE_READY) {
                 turnReady = TRUE;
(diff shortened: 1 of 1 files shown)


tst/TurnConnectionFunctionalityTest.cpp
```diff
@@ -263,6 +263,92 @@ namespace com { namespace amazonaws { namespace kinesis { namespace video { name
...
     }
 
+    struct TurnCustomData {
+        PTurnConnection pTurnConnection;
+        PBYTE pBuf;
+        UINT32 bufLen;
+        PKvsIpAddress pKvsIpAddress;
+    };
+
+    PVOID testTurnSendDataRoutine(PVOID arg) {
```
Sean-Der: Would you mind making this a lambda/keeping it internal to the test if possible?

Just keeps the code a little cleaner for the reader/don't have to worry about others re-using incorrectly.

tst/TurnConnectionFunctionalityTest.cpp
```diff
@@ -263,6 +263,92 @@ namespace com { namespace amazonaws { namespace kinesis { namespace video { name
...
+            if (pTurnConnection->state == TURN_STATE_READY) {
+                turnReady = TRUE;
+            }
+            MUTEX_UNLOCK(pTurnConnection->lock);
+        }
+
+        EXPECT_TRUE(turnReady == TRUE);
+
+        for (i = 0; i < reqCount; i++) {
+            buf[i] = (PBYTE) MEMALLOC(bufLen);
```
Sean-Der: If possible I would use a `unique_ptr` here, better then having to free manually.

tst/TurnConnectionFunctionalityTest.cpp
```diff
@@ -263,6 +263,92 @@ namespace com { namespace amazonaws { namespace kinesis { namespace video { name
...
+            for (j = 0; j < bufLen; j++) {
+                buf[i][j] = i;
+            }
+            args[i].pBuf = buf[i];
+            args[i].bufLen = bufLen;
+            args[i].pKvsIpAddress = &turnPeerAddr;
+            args[i].pTurnConnection = pTurnConnection;
+            EXPECT_EQ(STATUS_SUCCESS, THREAD_CREATE(&threads[i], testTurnSendDataRoutine, (PVOID) &args[i]));
```
Sean-Der: I would use `std::thread`, you can capture multiple things and don't have to worry about putting in a struct. The PeerConnectionFunctionality tests do that.

tst/TurnConnectionFunctionalityTest.cpp
zhiyua-git: It is only defined in this cpp and not in any .h file, I suppose that would not be reused and just internal to this test file.

tst/TurnConnectionFunctionalityTest.cpp
```diff
@@ -263,6 +263,73 @@ namespace com { namespace amazonaws { namespace kinesis { namespace video { name
...
+        BOOL turnReady = FALSE;
+        KvsIpAddress turnPeerAddr;
+        UINT32 bufLen = 5;
+        UINT32 reqCount = 10;
+        PBYTE *buf;
+        std::thread *threads;
+        UINT32 i, j;
+
+        buf = (PBYTE*) MEMALLOC(SIZEOF(PBYTE) * reqCount);
+        threads = (std::thread*) MEMALLOC(SIZEOF(std::thread) * reqCount);
```
MushMal: Use stack based arrays - these are tiny

tst/TurnConnectionFunctionalityTest.cpp
```diff
@@ -263,6 +263,73 @@ namespace com { namespace amazonaws { namespace kinesis { namespace video { name
...
+        EXPECT_EQ(STATUS_SUCCESS, turnConnectionAddPeer(pTurnConnection, &turnPeerAddr));
+        EXPECT_EQ(STATUS_SUCCESS, turnConnectionStart(pTurnConnection));
+
+        // wait until channel is created
+        while(!turnReady) {
+            MUTEX_LOCK(pTurnConnection->lock);
+            if (pTurnConnection->state == TURN_STATE_READY) {
+                turnReady = TRUE;
+            }
+            MUTEX_UNLOCK(pTurnConnection->lock);
```
MushMal: This is a tight loop. At least add some thread sleeps

PR #150, opened 2020-01-16
Update CMake to make cross-compiles easier

Sean-Der:
* CC/CXX is now respected across all dependencies
* OpenSSL and LibSRTP can now be configured via arguments to CMake
* Add travis build that ensures cross compiles do not regress

### DIFF
--- a/README.md
+++ b/README.md
@@ -37,6 +37,7 @@
 * Portable
   - Tested on Linux/MacOS
   - Tested on x64, ARMv5
+  - Build system designed for pleasant cross-compilation
 * Small Install Size
   - Sub 200k library size
   - OpenSSL, libsrtp, libjsmn, libusrsctp and libwebsockets dependencies.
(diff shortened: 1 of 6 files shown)


README.md
```diff
@@ -65,6 +68,9 @@ You can pass the following options to `cmake ..`.
 * `-DADD_MUCLIBC` -- Add -muclibc c flag
 * `-DBUILD_DEPENDENCIES` -- Whether or not to build depending libraries from source
 * `-DBUILD_OPENSSL` -- If building dependencies, whether or not building openssl from source
+* `-DBUILD_OPENSSL_PLATFORM` -- If buildng OpenSSL what is the target platform
+* `-DBUILD_LIBSRTP_HOST_PLATFORM` -- If buildng LibSRTP what is the current platform
+* `-DBUILD_LIBSRTP_DESTINATION_PLATFOR` -- If buildng LibSRTP what is the destination platform
```
MushMal: Not quite sure yet why we need host and destination platforms only for libsrtp? Shouldn't they be used for all dependencies?

Sean-Der: These are options that are just custom to the `openssl` and `libsrtp` configure scripts, the other libraries don't ask or have anyway for us to pass these values in.

Sean-Der:
Awesome! Glad it will be helpful @staowyze :)

Is there anything else we can do better, would make using the lib better for you?

README.md
```diff
@@ -60,11 +61,21 @@ GStreamer is installed on your system.
 By default we download all the libraries from GitHub and build them locally, so should require nothing to be installed ahead of time.
 If you do wish to link to existing libraries you can use the following flags to customize your build.
 
+
+#### Cross-Compilation
+
+If you wish to cross-compile `CC` and `CXX` are respected when building the library and all its dependencies. You will also need to set `BUILD_OPENSSL_PLATFORM`, `BUILD_LIBSRTP_HOST_PLATFORM` and `BUILD_LIBSRTP_DESTINATION_PLATFORM`. See our [.travis.yml](.travis.yml) for an example of this. Every commit is cross compiled to ensure that it continues to work.
```
unicornss: are we saying that the following environment variables are to be defined properly to match the toolchain settings.
CC, CXX ... BUILD_LIBSRTP_DESTINATION_PLATFORM

Sean-Der: `CC` and `CXX` need to be set.

`BUILD_LIBSRTP_DESTINATION_PLATFORM` are CMake arguments. Should I reword the documentation or maybe update the README in other ways?

thanks

Issue #165, opened 2020-01-29
[bug] deadlock when TURN is on


staowyze:
See the bt of these two threads
```
Thread 6 (Thread 0x7f3ad8b43700 (LWP 20818)):
#0  0x00007f3aecf7410d in __lll_lock_wait () at ../sysdeps/unix/sysv/linux/x86_64/lowlevellock.S:135
#1  0x00007f3aecf6d098 in __GI___pthread_mutex_lock (mutex=0x7f3ab8006270) at ../nptl/pthread_mutex_lock.c:113
#2  0x00007f3af01757ba in defaultLockMutex (mutex=139890171863664) at amazon-kinesis-video-streams-producer-sdk-cpp/kinesis-video-pic/src/utils/src/Mutex.c:202
#3  0x00007f3af015ef9c in turnConnectionSendData (pTurnConnection=0x7f3ab8000b20, pBuf=0x7f3ad8b42260 "", bufLen=28, pDestIp=0x7f3ac8018014) at amazon-kinesis-video-streams-webrtc-sdk-c/src/source/Ice/TurnConnection.c:551
#4  0x00007f3af015b214 in iceUtilsSendStunPacket (pStunPacket=0x55c886db3fd0, password=0x0, passwordLen=0, pDest=0x7f3ac8018014, pSocketConnection=0x7f3ab80062f0, pTurnConnection=0x7f3ab8000b20, useTurn=1)
    at amazon-kinesis-video-streams-webrtc-sdk-c/src/source/Ice/IceUtils.c:157
#5  0x00007f3af0156a07 in iceAgentSendKeepAliveTimerCallback (timerId=3, currentTime=15803165393161839, customData=94319744328448) at amazon-kinesis-video-streams-webrtc-sdk-c/src/source/Ice/IceAgent.c:1079
#6  0x00007f3af01785e2 in timerQueueExecutor (args=0x55c886d88240) at amazon-kinesis-video-streams-webrtc-sdk-c/open-source/amazon-kinesis-video-streams-producer-sdk-cpp/kinesis-video-pic/src/utils/src/TimerQueue.c:561
#7  0x00007f3aecf6a6db in start_thread (arg=0x7f3ad8b43700) at pthread_create.c:463
#8  0x00007f3aecc9388f in clone () at ../sysdeps/unix/sysv/linux/x86_64/clone.S:95
```
```
[...]

PR #166, opened 2020-01-30
Deadlock fix in Turn Connection.

MushMal:
Issue: https://github.com/awslabs/amazon-kinesis-video-streams-webrtc-sdk-c/issues/165

*Issue #165 , if available:*

Circular lock dependency leading to a dead lock. Traced most of the codepaths which seem to be OK with the exception of the callback. The cross-dependency of ICE and Turn connection are hard to break as both are low-level layers.

### DIFF
--- a/src/source/Ice/TurnConnection.c
+++ b/src/source/Ice/TurnConnection.c
@@ -764,6 +764,8 @@ STATUS turnConnectionDeliverChannelData(PTurnConnection pTurnConnection, PBYTE p
     UINT64 data;
     PTurnPeer pTurnPeer = NULL;
     UINT32 paddedDataLen;
+    PConnectionDataAvailableWrapper pWrapper = NULL;
+    TID callbackTid = INVALID_TID_VALUE;
 
     CHK(pTurnConnection != NULL, STATUS_NULL_ARG);
     CHK(pChannelMsg != NULL && channelMsgLen > 0, STATUS_INVALID_ARG);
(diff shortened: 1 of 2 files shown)


src/source/Ice/TurnConnection.c
```diff
@@ -781,34 +783,83 @@ STATUS turnConnectionDeliverChannelData(PTurnConnection pTurnConnection, PBYTE p
...
+            // Stop the loop iteration
+            pCurNode = NULL;
         }
     }
 
 CleanUp:
 
     CHK_LOG_ERR_NV(retStatus);
 
+    if (STATUS_FAILED(retStatus) && pWrapper != NULL) {
```
staowyze: would it be memory leak if `retStatus` looks okay? as we won't `MEMFREE(pWrapper)`.

src/source/Ice/TurnConnection.c
```diff
@@ -781,34 +783,83 @@ STATUS turnConnectionDeliverChannelData(PTurnConnection pTurnConnection, PBYTE p
...
+            pWrapper->customData = pTurnConnection->turnConnectionCallbacks.customData;
+            pWrapper->dataSize = channelDataSize;
+            pWrapper->channelNumber = channelNumber;
+            pWrapper->pTurnConnection = pTurnConnection;
+
+            // Set the pointer and store the data within the wrapper
+            pWrapper->pData = (PBYTE)(pWrapper + 1);
+            MEMCPY(pWrapper->pData, pChannelMsg + TURN_DATA_CHANNEL_SEND_OVERHEAD, channelDataSize);
+
+            CHK_STATUS(THREAD_CREATE(&callbackTid, turnDataAvailableWrapper, (PVOID) pWrapper));
```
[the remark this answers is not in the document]
MushMal: No they are not. This bugs me too. I assume some applications might use data channel extensively though. Let me rethink this path.

src/source/Ice/TurnConnection.c
MushMal: No, in that case the ownership of the allocation is in the thread routine which would free it

PR #168, opened 2020-01-30
Removing the previous fix of kicking off a separate thread as it's to…

MushMal:
…o expensive for some of the scenarios where incoming data can happen at high rate. Instead, fixing ICE agent interlocking.

*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/src/source/Ice/IceAgent.c
+++ b/src/source/Ice/IceAgent.c
@@ -1288,7 +1288,7 @@ STATUS incomingDataHandler(UINT64 customData, PSocketConnection pSocketConnectio
 {
     STATUS retStatus = STATUS_SUCCESS;
     PIceAgent pIceAgent = (PIceAgent) customData;
-    PStunPacket pStunResponse;
+    PStunPacket pStunResponse = NULL;
     BOOL locked = FALSE, useTurn;
 
     CHK(pIceAgent != NULL && pSocketConnection != NULL, STATUS_NULL_ARG);
@@ -1329,6 +1329,10 @@ CleanUp:
 
     if (locked) {
         MUTEX_UNLOCK(pIceAgent->lock);
+    }
+
+    if (pStunResponse != NULL) {
+        freeStunPacket(&pStunResponse);
     }
 
     return retStatus;
--- a/src/source/Ice/IceAgent.c
+++ b/src/source/Ice/IceAgent.c
@@ -1288,7 +1288,8 @@ STATUS incomingDataHandler(UINT64 customData, PSocketConnection pSocketConnectio
 {
     STATUS retStatus = STATUS_SUCCESS;
     PIceAgent pIceAgent = (PIceAgent) customData;
-    BOOL locked = FALSE;
+    PStunPacket pStunResponse;
+    BOOL locked = FALSE, useTurn;
 
     CHK(pIceAgent != NULL && pSocketConnection != NULL, STATUS_NULL_ARG);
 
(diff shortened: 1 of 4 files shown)


src/source/Ice/TurnConnection.c
```diff
@@ -399,8 +403,8 @@ STATUS turnConnectionHandleChannelDataTcpMode(PTurnConnection pTurnConnection, P
 
     pCurPos = pBuffer;
     remainingBufLen = bufferLen;
-    while(remainingBufLen > 0) {
-        if (pTurnConnection->currRecvDataLen > 0) {
+    while(remainingBufLen != 0) {
```
staowyze: probably not a big difference, may I know why changing from `>0` to `!=0`?

src/source/Ice/TurnConnection.c
```diff
@@ -160,6 +160,10 @@ STATUS turnConnectionIncomingDataHandler(UINT64 customData, PSocketConnection pS
         CHK_STATUS(turnConnectionHandleChannelDataTcpMode(pTurnConnection, pBuffer, bufferLen));
     }
 
+    // Unlock before checking if need to deliver to data channel callback
+    MUTEX_UNLOCK(pTurnConnection->lock);
```
staowyze: I guess this is typo, mutex is released in cleanup, a couple of lines below.

src/source/Ice/IceAgent.c
```diff
@@ -1304,10 +1305,24 @@ STATUS incomingDataHandler(UINT64 customData, PSocketConnection pSocketConnectio
         locked = FALSE;
 
         pIceAgent->iceAgentCallbacks.inboundPacketFn(pIceAgent->customData, pBuffer, bufferLen);
-        CHK(FALSE, retStatus);
-    }
+    } else {
+        CHK_STATUS(handleStunPacket(pIceAgent, pBuffer, bufferLen, pSocketConnection, pSrc, pDest,
+                &pStunResponse, &useTurn));
+
+        if (pStunResponse != NULL) {
```
staowyze: this way is definitely better than creating threads.

src/source/Ice/IceAgent.c
```diff
@@ -1304,10 +1305,24 @@ STATUS incomingDataHandler(UINT64 customData, PSocketConnection pSocketConnectio
...
+            locked = FALSE;
 
-    CHK_STATUS(handleStunPacket(pIceAgent, pBuffer, bufferLen, pSocketConnection, pSrc, pDest));
+            CHK_STATUS(iceUtilsSendStunPacket(pStunResponse,
+                    (PBYTE) pIceAgent->localPassword,
+                    (UINT32) STRLEN(pIceAgent->localPassword) * SIZEOF(CHAR),
+                    pSrc,
+                    pSocketConnection,
+                    pIceAgent->pTurnConnection,
+                    useTurn));
```
staowyze: dummy question, when do we need to send stun packet without using turn connection?
appears `iceUtilsSendStunPacket` is taking too many parameters.

src/source/Ice/TurnConnection.c
MushMal: Both are unsigned so won't be less than zero. It's better to be 'precise' in these cases

src/source/Ice/TurnConnection.c
MushMal: Yes, it does look weird to do it again. Let me change this. Thanks for spotting it.

src/source/Ice/IceAgent.c
MushMal: Thanks. I do think that we will need a separate service thread that services all of the outgoing packets with some buffering.

src/source/Ice/IceAgent.c
MushMal: This is a util function that can send stun packets via turn or direct P2P.

PR #169, opened 2020-01-31
use non-block socket. add configuration in KvsRtcConfig for send buff…

chehefen:
…er size setting

*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/src/source/Ice/ConnectionListener.c
+++ b/src/source/Ice/ConnectionListener.c
@@ -277,53 +277,62 @@ PVOID connectionListenerReceiveDataRoutine(PVOID arg)
                 pSocketConnection = (PSocketConnection) data;
 
                 if (FD_ISSET(pSocketConnection->localSocket, &rfds)) {
-                    readLen = recvfrom(pSocketConnection->localSocket, pConnectionListener->pBuffer, pConnectionListener->bufferLen, 0,
-                                       (struct sockaddr *) &srcAddrBuff, &srcAddrBuffLen);
-                    if (readLen == -1 ) {
-                        DLOGE("recvfrom() failed with errno %s", strerror(errno));
-                    } else if (readLen == 0) {
-                        CHK_STATUS(doubleListRemoveNode(pConnectionListener->connectionList, pCurNode));
-                    }
 
-                    if (readLen >= 0 &&
-                        pSocketConnection->dataAvailableCallbackFn != NULL &&
-                        // data could be encrypted so they need to be decrypted through socketConnectionReadData
-                        // and get the decrypted data length.
-                        STATUS_SUCCEEDED(socketConnectionReadData(pSocketConnection,
-                                                                  pConnectionListener->pBuffer,
-                                                                  pConnectionListener->bufferLen,
-                                                                  (PUINT32) &readLen))) {
-
-
-
-                        if (pSocketConnection->protocol == KVS_SOCKET_PROTOCOL_UDP) {
-                            if (srcAddrBuff.ss_family == AF_INET) {
-                                srcAddr.family = KVS_IP_FAMILY_TYPE_IPV4;
-                                pIpv4Addr = (struct sockaddr_in *) &srcAddrBuff;
-                                MEMCPY(srcAddr.address, (PBYTE) &pIpv4Addr->sin_addr, IPV4_ADDRESS_LENGTH);
-                                srcAddr.port = pIpv4Addr->sin_port;
-                            } else if (srcAddrBuff.ss_family == AF_INET6) {
-                                srcAddr.family = KVS_IP_FAMILY_TYPE_IPV6;
-                                pIpv6Addr = (struct sockaddr_in6 *) &srcAddrBuff;
-                                MEMCPY(srcAddr.address, (PBYTE) &pIpv6Addr->sin6_addr, IPV6_ADDRESS_LENGTH);
-                                srcAddr.port = pIpv6Addr->sin6_port;
+                    for(;;) {
+                        readLen = recvfrom(pSocketConnection->localSocket, pConnectionListener->pBuffer, pConnectionListener->bufferLen, 0,
+                                           (struct sockaddr *) &srcAddrBuff, &srcAddrBuffLen);
+                        if (readLen < 0 ) {
+                            if (errno != EWOULDBLOCK) {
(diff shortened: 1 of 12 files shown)


src/source/Ice/ConnectionListener.c
```diff
@@ -277,53 +277,62 @@ PVOID connectionListenerReceiveDataRoutine(PVOID arg)
...
-                                srcAddr.family = KVS_IP_FAMILY_TYPE_IPV4;
-                                pIpv4Addr = (struct sockaddr_in *) &srcAddrBuff;
-                                MEMCPY(srcAddr.address, (PBYTE) &pIpv4Addr->sin_addr, IPV4_ADDRESS_LENGTH);
-                                srcAddr.port = pIpv4Addr->sin_port;
-                            } else if (srcAddrBuff.ss_family == AF_INET6) {
-                                srcAddr.family = KVS_IP_FAMILY_TYPE_IPV6;
-                                pIpv6Addr = (struct sockaddr_in6 *) &srcAddrBuff;
-                                MEMCPY(srcAddr.address, (PBYTE) &pIpv6Addr->sin6_addr, IPV6_ADDRESS_LENGTH);
-                                srcAddr.port = pIpv6Addr->sin6_port;
+                    for(;;) {
```
MushMal: minor: use while (iterate) {} or while (TRUE) {} with break; in the body (less desired)

src/source/Ice/ConnectionListener.c
```diff
@@ -277,53 +277,62 @@ PVOID connectionListenerReceiveDataRoutine(PVOID arg)
...
+                        readLen = recvfrom(pSocketConnection->localSocket, pConnectionListener->pBuffer, pConnectionListener->bufferLen, 0,
+                                           (struct sockaddr *) &srcAddrBuff, &srcAddrBuffLen);
+                        if (readLen < 0 ) {
+                            if (errno != EWOULDBLOCK) {
+                                DLOGE("recvfrom() failed with errno %s", strerror(errno));
```
MushMal: Are we continuing to poll in case of the error? Doesn't seem right

src/source/Ice/ConnectionListener.c
```diff
@@ -277,53 +277,62 @@ PVOID connectionListenerReceiveDataRoutine(PVOID arg)
...
                         }
 
-                        pSocketConnection->dataAvailableCallbackFn(pSocketConnection->dataAvailableCallbackCustomData,
-                                                                   pSocketConnection,
-                                                                   pConnectionListener->pBuffer,
-                                                                   (UINT32) readLen,
-                                                                   pSrcAddr,
-                                                                   NULL); // no dest information available right now.
-                    }
+                        if (readLen >= 0 &&
```
MushMal: These if/else statements seem sub-optimal. Is there a need to do any of these in case the callback is not specified? Can we do a set of simple ifs up-front for error case, etc and save data to be called later outside of the if statements after we check whether the callback is specified.

src/source/Ice/Network.c
```diff
@@ -78,6 +78,12 @@ STATUS createSocket(PKvsIpAddress pHostIpAddress, PKvsIpAddress pPeerAddress, KV
     sockType = protocol == KVS_SOCKET_PROTOCOL_UDP ? SOCK_DGRAM : SOCK_STREAM;
 
     sockfd = socket(pHostIpAddress->family == KVS_IP_FAMILY_TYPE_IPV4 ? AF_INET : AF_INET6, sockType, 0);
+
+    if (sendBufLen > 0 && setsockopt(sockfd, SOL_SOCKET, SO_SNDBUF, &sendBufLen, SIZEOF(sendBufLen)) < 0) {
+        DLOGW("setsockopt() failed with errno %s", strerror(errno));
+        CHK(FALSE, STATUS_INTERNAL_ERROR);
```
MushMal: Why are we bailing out with an internal error? Doesn't seem right. This is a setsockopt failure which might still be OK. Also, I wonder if this can fail due to permissions?

MushMal: One more thing. Do we need to move the setsockopt after the check for failed sockfd below?

src/source/Ice/Network.c
```diff
@@ -133,12 +139,15 @@ STATUS createSocket(PKvsIpAddress pHostIpAddress, PKvsIpAddress pPeerAddress, KV
...
+    if (protocol == KVS_SOCKET_PROTOCOL_TCP && connect(sockfd, peerSockAddr, addrLen) < 0) {
         DLOGW("connect() failed with errno %s", strerror(errno));
-        CHK(FALSE, STATUS_SOCKET_CONNECT_FAILED);
+        CHK(FALSE, STATUS_SOCKET_CONNECT_FAILED);    
+    }
+
+    flags = fcntl(sockfd, F_GETFL, 0);
+    if (flags < 0 || fcntl(sockfd, F_SETFL, flags | O_NONBLOCK) < 0) {
+        DLOGW("failed to set socket flag with errno %s", strerror(errno));
+        CHK(FALSE, STATUS_SET_SOCKET_FLAG_FAILED);
```
MushMal: Doesn't let me comment in the CleanUp clause but seem that we might want to release the socket in case of an error

src/source/Ice/Network.c
```diff
@@ -133,12 +139,15 @@ STATUS createSocket(PKvsIpAddress pHostIpAddress, PKvsIpAddress pPeerAddress, KV
...
-
-    if (connect(sockfd, peerSockAddr, addrLen) < 0) {
+    if (protocol == KVS_SOCKET_PROTOCOL_TCP && connect(sockfd, peerSockAddr, addrLen) < 0) {
         DLOGW("connect() failed with errno %s", strerror(errno));
-        CHK(FALSE, STATUS_SOCKET_CONNECT_FAILED);
+        CHK(FALSE, STATUS_SOCKET_CONNECT_FAILED);    
+    }
+
+    flags = fcntl(sockfd, F_GETFL, 0);
+    if (flags < 0 || fcntl(sockfd, F_SETFL, flags | O_NONBLOCK) < 0) {
```
MushMal: Perhaps I am not understanding this piece. How does this not change the logic of handling the socket? Does this affect connect() API? If so, it was called above.

chehefen:
changes has been merged in another branch

PR #171, opened 2020-01-31
remove code where turnConnection detecting whether it is being used o…

### LINKED ISSUE #162: TURN connection check bug?
Not went deep into the implementation, seems disconnection and clean up would be triggered as follows:
https://github.com/awslabs/amazon-kinesis-video-streams-webrtc-sdk-c/blob/master/src/source/Ice/TurnConnection.c#L1091
https://github.com/awslabs/amazon-kinesis-video-streams-webrtc-sdk-c/blob/master/src/source/Ice/IceAgentStateMachine.c#L101

Imagine the case, both sides are `ICE_TRANSPORT_POLICY_RELAY`, and only one side is actively streaming AV data, would it trigger disconnection detection at the sender side and TURN connection clean up at receiver side? However, the TURN connection is health and should not be reconnected or cleaned up.

When both sides are `ICE_TRANSPORT_POLICY_RELAY`, the connection looks like
```
***********         ***************
* sender *   --->  *TURN server 1*
***********         ***************
                          |
                          |
***********         ***************
* recver *   <---  *TURN server 2*
***********         ***************

```
If we cut any part of the connection, the connection is broken, let me know if I misunderstood anything.

chehefen:
…r not. let iceAgent stop turn explicitly if relayed candidate is not selected

*Issue #, if available:*

fixes #162 

*Description of changes:*

### DIFF
--- a/src/source/Ice/IceAgentStateMachine.c
+++ b/src/source/Ice/IceAgentStateMachine.c
@@ -886,6 +886,11 @@ STATUS executeReadyIceAgentState(UINT64 customData, UINT64 time)
 
     pIceAgent->iceAgentState = ICE_AGENT_STATE_READY;
 
+    if (!IS_CANN_PAIR_SENDING_FROM_RELAYED(pIceAgent->pDataSendingIceCandidatePair) && pIceAgent->pTurnConnection != NULL) {
+        DLOGI("Relayed candidate is not selected. Turn allocation will be freed");
+        CHK_STATUS(turnConnectionStop(pIceAgent->pTurnConnection));
+    }
+
 CleanUp:
 
     if (STATUS_FAILED(retStatus)) {
--- a/src/source/Ice/IceAgentStateMachine.c
+++ b/src/source/Ice/IceAgentStateMachine.c
@@ -887,7 +887,7 @@ STATUS executeReadyIceAgentState(UINT64 customData, UINT64 time)
     pIceAgent->iceAgentState = ICE_AGENT_STATE_READY;
 
     if (!IS_CANN_PAIR_SENDING_FROM_RELAYED(pIceAgent->pDataSendingIceCandidatePair) && pIceAgent->pTurnConnection != NULL) {
-        DLOGI("Relayed candidate is not selected. Turn allocation will be freed");
+        DLOGD("Relayed candidate is not selected. Turn allocation will be freed");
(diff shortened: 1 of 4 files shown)


tst/TurnConnectionFunctionalityTest.cpp
```diff
@@ -169,6 +169,7 @@ namespace com { namespace amazonaws { namespace kinesis { namespace video { name
         }
 
         EXPECT_TRUE(turnReady == TRUE);
+        EXPECT_EQ(STATUS_SUCCESS, turnConnectionStop(pTurnConnection));
```
staowyze: Is this just for function test coverage? Looks like it's not intensively enough to test turn connection stopping without sort of integration test.

src/source/Ice/IceAgentStateMachine.c
```diff
@@ -886,6 +886,11 @@ STATUS executeReadyIceAgentState(UINT64 customData, UINT64 time)
 
     pIceAgent->iceAgentState = ICE_AGENT_STATE_READY;
 
+    if (!IS_CANN_PAIR_SENDING_FROM_RELAYED(pIceAgent->pDataSendingIceCandidatePair) && pIceAgent->pTurnConnection != NULL) {
```
staowyze: `IS_CANN_PAIR_SENDING_FROM_RELAYED` just checks if the candidate pair is relay connection or not, does it mean we are going to stop the connection for all the relay connection, i.e. turn connection?

src/source/Ice/IceAgentStateMachine.c
```diff
@@ -886,6 +886,11 @@ STATUS executeReadyIceAgentState(UINT64 customData, UINT64 time)
...
+        DLOGI("Relayed candidate is not selected. Turn allocation will be freed");
```
MushMal: DLOGD?

src/source/Ice/IceAgentStateMachine.c
chehefen: IS_CANN_PAIR_SENDING_FROM_RELAYED checks whether the local candidate of the selected pair is relay or not. if not, it means we are not using turn, thus we can safely shutdown turnConnection.

tst/TurnConnectionFunctionalityTest.cpp
chehefen: yes, this is just function test coverage.

PR #177, opened 2020-02-04
Properly parse compound RTCP packets

### LINKED ISSUE #45: Support non-rsize RTCP

Sean-Der:
Before we only parsed the first message in the datagram

Resolves #45


src/source/PeerConnection/Rtcp.c
```diff
@@ -6,17 +6,22 @@ STATUS onRtcpPacket(PKvsPeerConnection pKvsPeerConnection, PBYTE pBuff, INT32 pB
...
     CHK(pKvsPeerConnection != NULL && pBuff != NULL, STATUS_NULL_ARG);
-
     MEMSET(&rtcpPacket, 0x00, SIZEOF(rtcpPacket));
-    CHK_STATUS(setRtcpPacketFromBytes(pBuff, pBuffLen, &rtcpPacket));
 
-    if (rtcpPacket.header.packetType == RTCP_PACKET_TYPE_GENERIC_RTP_FEEDBACK
-        && rtcpPacket.header.receptionReportCount == RTCP_FEEDBACK_MESSAGE_TYPE_NACK) {
-        CHK_STATUS(resendPacketOnNack(&rtcpPacket, pKvsPeerConnection));
+    while (currentOffset < pBuffLen) {
+        CHK_STATUS(setRtcpPacketFromBytes(pBuff + currentOffset, pBuffLen - currentOffset, &rtcpPacket));
```
MushMal: This will create a packet from bytes assuming the whole packet can fit in buffLen - currentOffset. Are we handling partial packets OK? The situation when we have the packet to be split across multiple network packets causing onRtcpPacket be called twice for a single Rtcp packet

src/source/PeerConnection/Rtcp.c
```diff
@@ -6,17 +6,22 @@ STATUS onRtcpPacket(PKvsPeerConnection pKvsPeerConnection, PBYTE pBuff, INT32 pB
 {
     STATUS retStatus = STATUS_SUCCESS;
     RtcpPacket rtcpPacket;
+    UINT32 currentOffset = 0;
```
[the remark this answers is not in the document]
Sean-Der: Fixed!

src/source/PeerConnection/Rtcp.c
Sean-Der: RTCP doesn't support the packets being fragmented. They should just fail to parse in this case. This function just deserializes into `rtcpPacket`

PR #180, opened 2020-02-05
[Issue-113] Allow interface filtering while gathering local ICE candidates

### LINKED ISSUE #113: Allow users to filter which interfaces are used by local gathering
This was requested by a customer 

I would just copy how it was done with Pion, have the user set a callback and allow them to exclude an interface by name? https://godoc.org/github.com/pion/webrtc#SettingEngine.SetInterfaceFilter

disa6302:
*Issue #, if available: #113*

*Description of changes:*

- Added provision to set callback to whitelist/blacklist specific network interfaces
- This will give the user more flexibility to choose known interfaces to scan for ICE candidate gathering
- The callback can be set by setting the setFilterFn member of RtcConfiguration to the user's custom filter function

Resolve #113 

### DIFF
--- a/src/include/com/amazonaws/kinesis/video/webrtcclient/Include.h
+++ b/src/include/com/amazonaws/kinesis/video/webrtcclient/Include.h
@@ -458,6 +458,12 @@ typedef VOID (*RtcOnDataChannel)(UINT64, struct __RtcDataChannel*);
 typedef VOID (*RtcOnIceCandidate)(UINT64, PCHAR);
 
 /*
+ * IceSetInterfaceFilterFunc is fired when a callback function to filter network interfaces is assigned.
+ * The callback function is expected to check for specific interface names to be whitelisted/blacklisted
+ */
+typedef BOOL (*IceSetInterfaceFilterFunc) (UINT64, PCHAR);
+
+/*
  * https://www.w3.org/TR/webrtc/#rtcpeerconnectionstate-enum
  */
 typedef enum {
@@ -644,6 +650,8 @@ typedef struct {
     // If unset GENERATED_CERTIFICATE_BITS will be used
     INT32 generatedCertificateBits;
 
+    UINT64 filterCustomData;
+    IceSetInterfaceFilterFunc iceSetInterfaceFilterFunc;
 } KvsRtcConfiguration, *PKvsRtcConfiguration;
 
 /**
--- a/src/source/Ice/IceAgent.c
+++ b/src/source/Ice/IceAgent.c
@@ -364,8 +364,10 @@ STATUS iceAgentGatherLocalCandidate(PIceAgent pIceAgent)
     KvsIpAddress localIpAddresses[MAX_LOCAL_NETWORK_INTERFACE_COUNT];
     UINT32 localIpAddressesCount = MAX_LOCAL_NETWORK_INTERFACE_COUNT, i;
     PSocketConnection pSocketConnection = NULL;
-
-    CHK_STATUS(getLocalhostIpAddresses(localIpAddresses, &localIpAddressesCount));
+    CHK_STATUS(getLocalhostIpAddresses(localIpAddresses,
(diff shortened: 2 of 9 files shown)


src/include/com/amazonaws/kinesis/video/webrtcclient/Include.h
```diff
@@ -644,6 +650,7 @@ typedef struct {
     // If unset GENERATED_CERTIFICATE_BITS will be used
     INT32 generatedCertificateBits;
 
+    IceSetInterfaceFilterFunc setFilterFn;
```
Sean-Der: I would call this `iceSetInterfaceFilterFunc`

disa6302: Typo at both locations. Changing it now.

src/source/Ice/IceAgent.c
```diff
@@ -364,12 +364,24 @@ STATUS iceAgentGatherLocalCandidate(PIceAgent pIceAgent)
     KvsIpAddress localIpAddresses[MAX_LOCAL_NETWORK_INTERFACE_COUNT];
     UINT32 localIpAddressesCount = MAX_LOCAL_NETWORK_INTERFACE_COUNT, i;
     PSocketConnection pSocketConnection = NULL;
+    BOOL filterCallbackSet = TRUE;
 
-    CHK_STATUS(getLocalhostIpAddresses(localIpAddresses, &localIpAddressesCount));
+    // Check if a callback is set for filtering network interfaces
+    if(pIceAgent->kvsRtcConfiguration.setFilterFn == NULL) {
```
Sean-Der: Is it possible to move this all to https://github.com/awslabs/amazon-kinesis-video-streams-webrtc-sdk-c/blob/master/src/source/Ice/Network.c#L27 ?

Then we can just do `continue` and not have to copy stuff around.

disa6302: Done

src/source/Ice/TurnConnection.c
```diff
@@ -832,7 +832,7 @@ STATUS turnConnectionStepState(PTurnConnection pTurnConnection)
         case TURN_STATE_NEW:
             // find a host address to create new socket
             // TODO currently we use the first address found. Need to handle case when VPN is involved.
-            CHK_STATUS(getLocalhostIpAddresses(localhostIps, &localhostIpsLen));
+            CHK_STATUS(getLocalhostIpAddresses(localhostIps, &localhostIpsLen, NULL));
             for(i = 0; i < localhostIpsLen && !hostAddrFound; ++i) {
```
Sean-Der: We need here also

src/include/com/amazonaws/kinesis/video/webrtcclient/Include.h
```diff
@@ -457,6 +457,12 @@ typedef VOID (*RtcOnDataChannel)(UINT64, struct __RtcDataChannel*);
  */
 typedef VOID (*RtcOnIceCandidate)(UINT64, PCHAR);
 
+/*
+ * IceSetInterfaceFilterFunc is fired when a callback function to filter network interfaces is assigned.
+ * The callback function is expected to check for specific interface names to be whitelisted/blacklisted
+ */
+typedef BOOL (*IceSetInterfaceFilterFunc) (PCHAR);
```
chehefen: this callback should also have a UINT64 as its custom data like other callbacks

disa6302: Even if I have a custom data, I am not sure how I will be dealing with it in the APIs that are not exposed to the user. Can you tell me how custom data would be consumed in the API call?

disa6302: I can add that. However, what do I pass in custom data? Any useful piece of information you can think of?

PR #191, opened 2020-02-11
Kvs persistence sample

MushMal:
*Issue #, if available:*

*Description of changes:*


chehefen:
should we put webrtc + persistence into a separate sample? it seems like a well defined use case.

MushMal:
> should we put webrtc + persistence into a separate sample? it seems like a well defined use case.

It's a good question. Moreover, bringing this to webrtc samples implies also compiling libCurl which it doesn't need. I believe that for now we can park this sample as is but we need a proper samples repository which would be separate from these and can be faster pace, and, can include webrtc assets and kvs producer assets. I just need to get this out first.

PR #213, opened 2020-02-24
Pin libwebsockets to v3.2.0 and not the stable branch

### LINKED ISSUE #211: libwebsocket version causes problems
For libwebsocket, this GIT_TAG is used:    v3.2-stable.
It caused problems recently.  Before 1/16, there was no problems
* When running binary on Ubuntu 18.04 ( using system's openssl), peer connection crashes in createCertificateAndKey function.  I found it has a few 509_xx calls.
* When cross-compile for ARM, the build fails with error X509_VERIFY_PARAM_set1_host not defined.
Downgrading libwebsocket to  GIT_TAG =3179323afa81448287a15982755ed0e4a34d80cb solves the problem

Sean-Der:
This branch is for active development, and is seeing API breaking changes

Resolves #211


MushMal:
Do we need to have similar changes in C producer repo?

PR #254, opened 2020-03-10
Don't bail out on an error in re-create in samples.

MushMal:
*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/samples/Common.c
+++ b/samples/Common.c
@@ -692,6 +692,7 @@ STATUS sessionCleanupWait(PSampleConfiguration pSampleConfiguration)
     PSampleStreamingSession pSampleStreamingSession = NULL;
     UINT32 i;
     BOOL locked = FALSE;
+    SIGNALING_CLIENT_STATE signalingClientState;
 
     CHK(pSampleConfiguration != NULL, STATUS_NULL_ARG);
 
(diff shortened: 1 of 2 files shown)


samples/Common.c
```diff
@@ -723,18 +724,24 @@ STATUS sessionCleanupWait(PSampleConfiguration pSampleConfiguration)
...
+                    pSampleConfiguration->pCredentialProvider,
+                    &pSampleConfiguration->signalingClientHandle))) {
             // Re-set the variable again
             ATOMIC_STORE_BOOL(&pSampleConfiguration->recreateSignalingClient, FALSE);
         }
+
+        // Check the signaling client state and connect if needed
+        if (IS_VALID_SIGNALING_CLIENT_HANDLE(pSampleConfiguration->signalingClientHandle)) {
+            CHK_STATUS(signalingClientGetCurrentState(pSampleConfiguration->signalingClientHandle, &signalingClientState));
+            if (signalingClientState == SIGNALING_CLIENT_STATE_READY) {
```
chehefen: why do we have to wait until state ready to call signalingClientConnectSync?

chehefen: in the master sample we just call signalingClientConnectSync

MushMal: Good question. In the startup, we will bail out on creation error. Here, we don't. We keep re-trying. Only when the creation is OK and we are in Ready state, then we will attempt to connect. Otherwise, we will loop in 5 seconds and retry re-creating the client first.

The reason I thought this would be useful is the following. Consider the sample application is running and suddenly we hit a network outage. This causes the signaling client to do it's thing and attempt to reconnect until timeout. This will then call the error callback which would trigger the re-create and re-connect. The network outage still being present will cause the re-create to fail and the application to gracefully terminate.

This is a continuous attempt to re-create/re-connect.

PR #279, opened 2020-03-18
handle EAGAIN and EWOULDBLOCK from sendTo

chehefen:
*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/src/source/Ice/SocketConnection.c
+++ b/src/source/Ice/SocketConnection.c
@@ -144,12 +144,14 @@ STATUS socketConnectionSendData(PSocketConnection pSocketConnection, PBYTE pBuf,
 {
     STATUS retStatus = STATUS_SUCCESS;
     BOOL locked = FALSE, iterate = TRUE;
-    INT32 sslRet, sslErr;
+    INT32 sslRet, sslErr, result;
 
     socklen_t addrLen;
     struct sockaddr *destAddr;
     struct sockaddr_in ipv4Addr;
     struct sockaddr_in6 ipv6Addr;
+    fd_set wfds;
+    struct timeval tv;
 
     CHK(pSocketConnection != NULL && pBuf != NULL, STATUS_NULL_ARG);
     CHK(bufLen != 0 && (pSocketConnection->protocol == KVS_SOCKET_PROTOCOL_TCP || pDestIp != NULL), STATUS_INVALID_ARG);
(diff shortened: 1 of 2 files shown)


src/source/Ice/SocketConnection.c
```diff
@@ -197,7 +199,25 @@ STATUS socketConnectionSendData(PSocketConnection pSocketConnection, PBYTE pBuf,
         }
 
         // sending through UDP never block
-        if (sendto(pSocketConnection->localSocket, pBuf, bufLen, 0, destAddr, addrLen) < 0) {
+        result = sendto(pSocketConnection->localSocket, pBuf, bufLen, 0, destAddr, addrLen);
+        if (result < 0 && (errno == EAGAIN || errno == EWOULDBLOCK)) {
```
MushMal: Combine this if and the next into one

if (result < 0 ) {
  switch (errno) {
case foo:
// Explicit fall-through
case bar:
// handle it
break;
default:
warn();
}
}

src/source/Ice/SocketConnection.c
```diff
@@ -197,7 +199,25 @@ STATUS socketConnectionSendData(PSocketConnection pSocketConnection, PBYTE pBuf,
...
+            FD_ZERO(&wfds);
```
MushMal: Comment about blocking

src/source/Ice/SocketConnection.c
```diff
@@ -197,7 +199,25 @@ STATUS socketConnectionSendData(PSocketConnection pSocketConnection, PBYTE pBuf,
...
+            FD_SET(pSocketConnection->localSocket, &wfds);
+            tv.tv_sec = 0;
+            tv.tv_usec = SOCKET_SEND_RETRY_TIMEOUT_MICRO_SECOND;
+            result = select(pSocketConnection->localSocket + 1, NULL, &wfds, NULL, &tv);
+
+            if (result > 0) {
+                result = sendto(pSocketConnection->localSocket, pBuf, bufLen, 0, destAddr, addrLen);
+            } else if (result == 0) {
+                DLOGD("select() timed out");
```
MushMal: Are these WARNS? I see you have DLOGD here but in the else below you have DLOGE. See whether this is indeed an DLOGE or DLOGW.

Also, in case of DLOGE you could use CHK_ERR.

src/source/Ice/SocketConnection.c
chehefen: EWOULDBLOCK and EAGAIN are the same value and cant be put in a switch state. These two are the only errno we gonna handle.

src/source/Ice/SocketConnection.c
chehefen: I think all these should be DLOGD cuz higher layer should to agnostic to send failures.

src/source/Ice/SocketConnection.c
chehefen: done

src/source/Ice/SocketConnection.c
MushMal: That's what I thought. In that case, please make the other DLOGD

src/source/Ice/SocketConnection.c
MushMal: Perhaps those are platform specific? At any rate, you can still do if/else but combine the outer if together

PR #284, opened 2020-03-19
network endian timestamp and sequence number

zhuker:
*Description of changes:*
on little endian CPUs sequence number of rtp packet is left little endian which is against the rtp spec

### DIFF
--- a/src/source/Rtp/RtpPacket.c
+++ b/src/source/Rtp/RtpPacket.c
@@ -308,7 +308,7 @@ STATUS setBytesFromRtpPacket(PRtpPacket pRtpPacket, PBYTE pRawPacket, UINT32 pac
     pCurPtr += SIZEOF(UINT32);
 
     for (i = 0; i < pHeader->csrcCount; i++, pCurPtr += SIZEOF(UINT32)) {
-        putUnalignedInt32BigEndian((PINT32) pCurPtr, pHeader->csrcArray[i]);
+        putUnalignedInt32BigEndian(pCurPtr, pHeader->csrcArray[i]);
     }
 
     if (pHeader->extension) {
--- a/src/source/Rtp/RtpPacket.c
+++ b/src/source/Rtp/RtpPacket.c
@@ -298,13 +298,13 @@ STATUS setBytesFromRtpPacket(PRtpPacket pRtpPacket, PBYTE pRawPacket, UINT32 pac
     //https://tools.ietf.org/html/rfc7741#page-5
     //All integer fields in the specifications are encoded as
     //   unsigned integers in network octet order.
-    putUnalignedInt16BigEndian((PINT16) pCurPtr, pHeader->sequenceNumber);
+    putUnalignedInt16BigEndian(pCurPtr, pHeader->sequenceNumber);
     pCurPtr += SIZEOF(UINT16);
 
-    putUnalignedInt32BigEndian((PINT32) pCurPtr, pHeader->timestamp);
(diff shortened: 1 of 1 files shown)


src/source/Rtp/RtpPacket.c
```diff
@@ -295,26 +295,29 @@ STATUS setBytesFromRtpPacket(PRtpPacket pRtpPacket, PBYTE pRawPacket, UINT32 pac
     }
     pCurPtr++;
 
-    putInt16((PINT16) pCurPtr, pHeader->sequenceNumber);
+    //https://tools.ietf.org/html/rfc7741#page-5
+    //All integer fields in the specifications are encoded as
+    //   unsigned integers in network octet order.
+    putUnalignedInt16BigEndian((PINT16) pCurPtr, pHeader->sequenceNumber);
```
MushMal: You can drop the casting here as these functions take PVOID. Also, consider if you want to use the macro version PUT_UNALIGNED_BIG_ENDIAN in which case you will need the casting

zhuker: makes sense

MushMal: @zhuker I was wondering whether you are going to make these changes so we can merge the PR?

zhuker: @MushMal done

Sean-Der:
This is really awesome @zhuker thank you so much for the contribution :)

When working with the repo is there anything we can be doing better? Any feedback/improvement suggestions we would love to hear!

zhuker:
@Sean-Der you are welcome
re working with repo: code looks very cleanly written and easily debuggable, kudos

random rant follows, feel free to disregard :)
i am replacing googles webrtc lib with this one and i am struggling to push vp8 to chrome client (custom signalling not kvs) using the lib, I  tweaked the vp8 payloader to act exactly like gstreamer's `rtpvp8pay`  (hence the PR but they also do PIDs in vp8 payload desc)  but still no luck.

what's curious is if i encode frames at really low bit rate they seem to be pushed kind of ok (some decode failures but most of the stream is visible)
chrome reports packets received but wont decode frames

Sean-Der:
@zhuker What is your keyframe interval? I need to add 'Keyframe Request' to the public API, but right now we don't have anyway to request one.

I also threw a VP8 pipeline in the GStreamer sample we have, try that out and see if that works! VP8 has always worked for me though unfortunately, not a lot of debug advice :( H264 is usually the one that makes me suffer :)

Sean-Der:
Had to move to https://github.com/awslabs/amazon-kinesis-video-streams-webrtc-sdk-c/pull/288

PRs are currently broken from outside contributors. We added Doxygen generation, and because that has secure env variables it doesn't allow outside contributors to use them (and the jobs fail). For now I am just gonna cherry-pick peoples commits and land them that way, hopefully find a better solution soon.

zhuker:
> @zhuker What is your keyframe interval? I need to add 'Keyframe Request' to the public API, but right now we don't have anyway to request one.
I had 10 frames keyframe interval but now that you said it i set it to all keyframes and still get all PLIs and no picture in browser :-\
Checked gstreamer example `vp8enc error-resilient=partitions keyframe-max-dist=10 auto-alt-ref=true cpu-used=5 deadline=1` doesnt help either

> 
> I also threw a VP8 pipeline in the GStreamer sample we have, try that out and see if that works! VP8 has always worked for me though unfortunately, not a lot of debug advice :( H264 is usually the one that makes me suffer :)

PR #299, opened 2020-03-26
Adding env variable parsing for logger log level

MushMal:
*Issue #, if available:*

Adding ability to control verbosity using env variable

### DIFF
--- a/samples/Common.c
+++ b/samples/Common.c
@@ -569,7 +569,7 @@ STATUS createSampleConfiguration(PCHAR channelName, SIGNALING_CHANNEL_ROLE_TYPE 
     STATUS retStatus = STATUS_SUCCESS;
     PCHAR pAccessKey, pSecretKey, pSessionToken, pLogLevel;
     PSampleConfiguration pSampleConfiguration = NULL;
-    UINT32 logLevel = LOG_LEVEL_WARN;
+    UINT32 logLevel;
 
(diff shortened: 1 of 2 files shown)


samples/Common.c
```diff
@@ -584,6 +585,15 @@ STATUS createSampleConfiguration(PCHAR channelName, SIGNALING_CHANNEL_ROLE_TYPE
 
     CHK_STATUS(lookForSslCert(&pSampleConfiguration));
 
+    // Set the logger log level
```
disa6302: I am a little confused. When we set the `AWS_KVS_LOG_LEVEL` environment variable, is it not the log level that is used for the entire SDK? 

Also, do we really want to jump to CleanUp section if log level setting fails? There is a default log level the SDK operates at even if the env is not set, correct?

MushMal: Yes, the logging level is used by the default logger in PIC. The failure is only for parsing. If the env variable is not set we default to warn level which is the default anyway

disa6302: Apologies for my verbatim. I meant parsing in my above comment. But do we really need to return from the function if the parsing fails? Is it not enough to log that the parsing failed?

MushMal: Hmm, perhaps you are right. Even when parsed we shouldn't fail. Let me modify the code really quick.

PR #315, opened 2020-03-30
remove socket connection from connection listener when connection is …

chehefen:
…reset by peer

*Issue #, if available:*

*Description of changes:*


src/source/Ice/ConnectionListener.c
```diff
@@ -285,8 +285,16 @@ PVOID connectionListenerReceiveDataRoutine(PVOID arg)
...
-                                DLOGW("recvfrom() failed with errno %s", strerror(errno));
+                            switch (errno) {
+                                case EWOULDBLOCK:
+                                    break;
+                                case ECONNRESET:
+                                    // case when peer closed the connection
+                                    CHK_STATUS(doubleListRemoveNode(pConnectionListener->connectionList, pCurNode));
+                                    // explicit fall-through
+                                default:
+                                    DLOGD("recvfrom() failed with errno %s for socket %d", strerror(errno), pSocketConnection->localSocket);
```
MushMal: Should we also remove in the default case?

PR #331, opened 2020-04-07
[issue-301] H264 FMTP

### LINKED ISSUE #301: When answering only include H264 fmtp line if offerer did
If we receive an offer and it contains H264 (but no profile) we shouldn't include one

disa6302:
*Issue #, if available: #301 *

*Description of changes:*
- Add h.264 fmtp / opus fmtp in the answer only if the offer contains fmtp attribute
- Add default fmtp in case of offering
- Added unit test to test both the scenarios - when fmtp is included in the offer and when it is not included in the offer

Resolves #301 

### DIFF
--- a/tst/SdpApiTest.cpp
+++ b/tst/SdpApiTest.cpp
@@ -317,6 +317,112 @@ TEST_F(SdpApiTest, populateSingleMediaSection_TestTxRecvOnly) {
     freePeerConnection(&offerPc);
 }
 
+TEST_F(SdpApiTest, populateSingleMediaSection_TestPayloadNoFmtp) {
+    auto remoteSessionDescription = R"(v=0
+o=- 7732334361409071710 2 IN IP4 127.0.0.1
+s=-
+t=0 0
+a=group:BUNDLE 0
+a=msid-semantic: WMS
+m=video 16485 UDP/TLS/RTP/SAVPF 96 102
+c=IN IP4 205.251.233.176
+a=rtcp:9 IN IP4 0.0.0.0
+a=ice-ufrag:9YRc
+a=ice-pwd:/ELMEiczRSsx2OEi2ynq+TbZ
+a=ice-options:trickle
+a=fingerprint:sha-256 51:04:F9:20:45:5C:9D:85:AF:D7:AF:FB:2B:F8:DB:24:66:7B:6A:E3:E3:EF:EC:72:93:6E:01:B8:C9:53:A6:31
+a=setup:actpass
+a=mid:1
+a=recvonly
+a=rtcp-mux
+a=rtcp-rsize
+a=rtpmap:96 VP8/90000
+a=rtpmap:102 H264/90000
+)";
+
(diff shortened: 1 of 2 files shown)


tst/SdpApiTest.cpp
```diff
@@ -317,6 +317,111 @@ TEST_F(SdpApiTest, populateSingleMediaSection_TestTxRecvOnly) {
...
+a=ice-pwd:/ELMEiczRSsx2OEi2ynq+TbZ
+a=ice-options:trickle
+a=fingerprint:sha-256 51:04:F9:20:45:5C:9D:85:AF:D7:AF:FB:2B:F8:DB:24:66:7B:6A:E3:E3:EF:EC:72:93:6E:01:B8:C9:53:A6:31
+a=setup:actpass
+a=mid:1
+a=recvonly
+a=rtcp-mux
+a=rtcp-rsize
+a=rtpmap:96 VP8/90000
+a=rtpmap:102 H264/90000
```
unicornss, line 340: are we just defaulting to a specific format if negotiation does not include fmtp? should n't the dynamic payload types (96-127) require media format parameter always?

Sean-Der: The issue is right now we are working with clients that are giving us offers with fmtp. They should though :(

tst/SdpApiTest.cpp
```diff
@@ -317,6 +317,111 @@ TEST_F(SdpApiTest, populateSingleMediaSection_TestTxRecvOnly) {
...
+    rtcMediaStreamTrack.codec = RTC_CODEC_H264_PROFILE_42E01F_LEVEL_ASYMMETRY_ALLOWED_PACKETIZATION_MODE;
+    STRCPY(rtcMediaStreamTrack.streamId, "myKvsVideoStream");
+    STRCPY(rtcMediaStreamTrack.trackId, "myTrack");
+    EXPECT_EQ(addTransceiver(pRtcPeerConnection, &rtcMediaStreamTrack, &rtcRtpTransceiverInit, &pRtcRtpTransceiver), STATUS_SUCCESS);
+
+    STRCPY(rtcSessionDescriptionInit.sdp, remoteSessionDescription);
+    rtcSessionDescriptionInit.type = SDP_TYPE_OFFER;
+    EXPECT_EQ(setRemoteDescription(pRtcPeerConnection, &rtcSessionDescriptionInit), STATUS_SUCCESS);
+    EXPECT_EQ(createAnswer(pRtcPeerConnection, &rtcSessionDescriptionInit), STATUS_SUCCESS);
+    EXPECT_PRED_FORMAT2(testing::IsSubstring, "fmtp", rtcSessionDescriptionInit.sdp);
```
Sean-Der, line 421: Can you assert the entire value? Also maybe do some weird value like `fmtp:102 my-testing-value` just to make sure we aren't just doing the default.

disa6302: Done. Adding two ASSERTS: One with the the entire fmtp string and the other with a `fmtp:102 strange` to add more coverage

PR #332, opened 2020-04-08
optimize getNextNaluLength

chehefen:
*Issue #, if available:*

*Description of changes:*
Streaming using non gstreamer master sample on raspberry pi. CPU usage reported in top before change: 12%. After change: 10%.

### DIFF
--- a/src/source/Rtp/Codecs/RtpH264Payloader.c
+++ b/src/source/Rtp/Codecs/RtpH264Payloader.c
@@ -80,27 +80,46 @@ STATUS getNextNaluLength(PBYTE nalus, UINT32 nalusLength, PUINT32 pStart, PUINT3
     ENTERS();
 
     STATUS retStatus = STATUS_SUCCESS;
-    UINT32 zeroCount = 0;
-    PBYTE pCurPtr = nalus;
-    UINT32 i;
+    UINT32 zeroCount = 0, offset;
+    BOOL naluFound = FALSE;
+    PBYTE pCurrent = NULL;
 
     CHK(nalus != NULL && pStart != NULL && pNaluLength != NULL, STATUS_NULL_ARG);
 
     // Annex-B Nalu will have 0x000000001 or 0x000001 start code, at most 4 bytes
-    for (i = 0; i < 4 && i < nalusLength && nalus[i] == 0; i++);
+    for (offset = 0; offset < 4 && offset < nalusLength && nalus[offset] == 0; offset++);
 
-    CHK(i < nalusLength && i < 4 && i >= 2 && nalus[i] == 1, STATUS_RTP_INVALID_NALU);
-    *pStart = ++i;
+    CHK(offset < nalusLength && offset < 4 && offset >= 2 && nalus[offset] == 1, STATUS_RTP_INVALID_NALU);
+    *pStart = ++offset;
+    pCurrent = nalus + offset;
 
-    for (pCurPtr = nalus + i; i < nalusLength && !(*pCurPtr == 1 && (zeroCount == 2 || zeroCount == 3)); i++, pCurPtr++)
-    {
-        if (*pCurPtr == 0) {
-            zeroCount++;
+    /* Not doing validation on number of consecutive zeros being less than 4 because some device can produce
+     * data with trailing zeros. */
+    while(offset < nalusLength) {
+        if (*pCurrent == 0) {
+            /* Maybe next byte is 1 */
+            offset++;
+            pCurrent++;
+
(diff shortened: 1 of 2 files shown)


src/source/Rtp/Codecs/RtpH264Payloader.c
```diff
@@ -80,27 +80,38 @@ STATUS getNextNaluLength(PBYTE nalus, UINT32 nalusLength, PUINT32 pStart, PUINT3
...
-    *pStart = ++i;
+    for (offset = 0; offset < 4 && offset < nalusLength && nalus[offset] == 0; offset++);
+
+    CHK(offset < nalusLength && offset < 4 && offset >= 2 && nalus[offset] == 1, STATUS_RTP_INVALID_NALU);
+    *pStart = ++offset;
+
+    while(offset < nalusLength) {
+        if (nalus[offset] == 0) {
+            /* Maybe next byte is 1 */
+            offset++;
```
MushMal, line 97: What if you have consequent 0s more than 3 in the row?

src/source/Rtp/Codecs/RtpH264Payloader.c
```diff
@@ -80,27 +80,38 @@ STATUS getNextNaluLength(PBYTE nalus, UINT32 nalusLength, PUINT32 pStart, PUINT3
...
+            offset += 3;
         } else {
-            zeroCount = 0;
+
+            /* Can jump 3 bytes forward */
+            offset += 3;
         }
     }
-    *pNaluLength = i - *pStart - (zeroCount >= 2 && zeroCount < 4 && i < nalusLength? zeroCount : 0);
+    *pNaluLength = MIN(offset, nalusLength) - *pStart - (naluFound? zeroCount : 0);
```
MushMal, line 114: Perhaps we should also check whether we end up with 2 or 3 consequent 0s - that's not legal

src/source/Rtp/Codecs/RtpH264Payloader.c
chehefen: we are gonna incur a cost there for correctness sake. the next character could be 1.

MushMal: There is a slight perf hit by tracking the count of 0s and checking if the count is 3 and checking one next to be 1. However, I doubt that this if comparison will add any observable perf costs.

chehefen: I dont quite get it, even if we are tracking 0s, we cant step more than 1 if current value is 0. what is the benefit compared to what it is right now?

src/source/Rtp/Codecs/RtpH264Payloader.c
chehefen: in that case we may generate false positives if buffer has trailing 0s.

MushMal: NALus should EPB the trailing zeroes. This said, perhaps we are ok to not check this.

src/source/Rtp/Codecs/RtpH264Payloader.c
```diff
@@ -80,27 +80,38 @@ STATUS getNextNaluLength(PBYTE nalus, UINT32 nalusLength, PUINT32 pStart, PUINT3
...
     CHK(nalus != NULL && pStart != NULL && pNaluLength != NULL, STATUS_NULL_ARG);
 
     // Annex-B Nalu will have 0x000000001 or 0x000001 start code, at most 4 bytes
-    for (i = 0; i < 4 && i < nalusLength && nalus[i] == 0; i++);
-
-    CHK(i < nalusLength && i < 4 && i >= 2 && nalus[i] == 1, STATUS_RTP_INVALID_NALU);
-    *pStart = ++i;
+    for (offset = 0; offset < 4 && offset < nalusLength && nalus[offset] == 0; offset++);
+
+    CHK(offset < nalusLength && offset < 4 && offset >= 2 && nalus[offset] == 1, STATUS_RTP_INVALID_NALU);
```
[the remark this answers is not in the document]
chehefen: removing that breaks RtpFunctionalityTest.invalidNaluParse

src/source/Rtp/Codecs/RtpH264Payloader.c
MushMal: Indeed. You are not bounding your "stepping" say I have 100s of 0s

MushMal:
NOTE: We are not doing full Annex-B format validation in this code - just a rudimentary check. For which the code is fine.

PR #333, opened 2020-04-08
Fix sctp crash

chehefen:
*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/src/source/Sctp/Sctp.c
+++ b/src/source/Sctp/Sctp.c
@@ -1,6 +1,8 @@
 #define LOG_CLASS "SCTP"
 #include "../Include_i.h"
 
+static volatile PSctpSessionControl pSctpSessionControl = NULL;
+
 STATUS initSctpAddrConn(PSctpSession pSctpSession, struct sockaddr_conn *sconn)
 {
     ENTERS();
@@ -63,14 +65,32 @@ STATUS initSctpSession()
     // Disable Explicit Congestion Notification
     usrsctp_sysctl_set_sctp_ecn_enable(0);
 
+    pSctpSessionControl = MEMALLOC(SIZEOF(SctpSessionControl));
+    pSctpSessionControl->lock = MUTEX_CREATE(FALSE);
+    CHK_STATUS(hashTableCreate(&pSctpSessionControl->activeSctpSessions));
+
+CleanUp:
+
     return retStatus;
 }
 
 VOID deinitSctpSession()
 {
+    if (pSctpSessionControl != NULL) {
+        MUTEX_LOCK(pSctpSessionControl->lock);
+        hashTableClear(pSctpSessionControl->activeSctpSessions);
+        MUTEX_UNLOCK(pSctpSessionControl->lock);
+    }
+
     // need to block until usrsctp_finish or sctp thread could be calling free objects and cause segfault
     while (usrsctp_finish() != 0) {
         THREAD_SLEEP(DEFAULT_USRSCTP_TEARDOWN_POLLING_INTERVAL);
+    }
+
+    if (pSctpSessionControl != NULL) {
+        MUTEX_FREE(pSctpSessionControl->lock);
+        hashTableFree(pSctpSessionControl->activeSctpSessions);
+        SAFE_MEMFREE(pSctpSessionControl);
     }
 }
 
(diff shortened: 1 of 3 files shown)


src/source/Sctp/Sctp.c
```diff
@@ -1,6 +1,8 @@
 #define LOG_CLASS "SCTP"
 #include "../Include_i.h"
 
+static volatile PSctpSessionControl pSctpSessionControl = NULL;
```
MushMal, line 4: We shouldn't use statics here. This can be part of the object graph. Please change the deinitSctpSession to take this object of a parent object that it can use instead of the global

src/source/Sctp/Sctp.c
```diff
@@ -63,15 +65,33 @@ STATUS initSctpSession()
     // Disable Explicit Congestion Notification
     usrsctp_sysctl_set_sctp_ecn_enable(0);
 
+    pSctpSessionControl = MEMALLOC(SIZEOF(SctpSessionControl));
```
MushMal, line 68: The init and deinit can take an object/return an object to avoid the global. This object then can be part of the overall WebRTC object graph

src/source/Sctp/Sctp.c
```diff
@@ -63,15 +65,33 @@ STATUS initSctpSession()
...
+    pSctpSessionControl->lock = MUTEX_CREATE(FALSE);
+    CHK_STATUS(hashTableCreate(&pSctpSessionControl->activeSctpSessions));
```
MushMal, line 70: Use the parameterized version of the hash table. The default will create 2 * 10K buckets

src/source/Sctp/Sctp.c
```diff
@@ -219,19 +239,25 @@ INT32 onSctpOutboundPacket(PVOID addr, PVOID data, ULONG length, UINT8 tos, UINT
...
     }
 
-    pSctpSession->sctpSessionCallbacks.outboundPacketFunc(pSctpSession->sctpSessionCallbacks.customData, data, length);
+    MUTEX_LOCK(pSctpSessionControl->lock);
+    hashTableContains(pSctpSessionControl->activeSctpSessions, pSctpSession->key, &containKey);
+    MUTEX_UNLOCK(pSctpSessionControl->lock);
+
+    /* If the key is not in activeSctpSessions, then the session must have been freed. Thus do not call callback.
+     * Fixes issue stated here: https://github.com/sctplab/usrsctp/issues/147.
+     * (return -1 otherwise usrsctp_finish() won't finish) */
```
MushMal, line 255: So when do we return -1?

src/source/Sctp/Sctp.h
```diff
@@ -59,11 +53,21 @@ typedef struct {
 } SctpSessionCallbacks, *PSctpSessionCallbacks;
 
 typedef struct {
-    volatile SIZE_T shutdownStatus;
     struct socket *socket;
     SctpSessionCallbacks sctpSessionCallbacks;
+    UINT64 key;
```
MushMal, line 58: So you use store the current time here and this acts as the key in the hash table. Perhaps add a comment to state what you use as a key. Also, we really need to create a hashset object in PIC

src/source/Sctp/Sctp.c
Sean-Der: It will still be a global though right? It is shared among all PeerConnections

Sean-Der:
Sorry for the push back, but I think this will just end up accruing more technical debt :/

* Are you able to easily reproduce? Maybe we can get upstream to help.
* We could also do github.com/sctplab/usrsctp/pull/339 and then have a timer pull from the association? That could fix everything for us!

src/source/Sctp/Sctp.c
chehefen: This thing out lives peerConnection, so we dont have a right parent object for it.

MushMal: I understand now your point. Let's make this a high-level object that's lifespan will be tracked at the application level.

Have createSctpSession and freeSctpSession. The object created then will be passed to the perr connections?

src/source/Sctp/Sctp.c
MushMal: Having mutable globals is a very bad idea at a high-level code. I think we might need to have a high-level object that lives in parallel with peer connection (like the signaling client) that will be passed into peer connection and it's lifecycle managed by the app

niyatim23:
[PR](https://github.com/awslabs/amazon-kinesis-video-streams-webrtc-sdk-c/pull/1236) open in develop branch

PR #336, opened 2020-04-08
generateJSONSafeString randBuffer msan

zhuker:
our memory sanitizer complains about the use of uninitialized memory here.

### DIFF
--- a/src/source/PeerConnection/PeerConnection.c
+++ b/src/source/PeerConnection/PeerConnection.c
@@ -429,7 +429,7 @@ STATUS generateJSONSafeString(PCHAR pDst, UINT32 len)
     ENTERS();
     STATUS retStatus = STATUS_SUCCESS;
     UINT32 i = 0;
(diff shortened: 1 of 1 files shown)


src/source/PeerConnection/PeerConnection.c
```diff
@@ -429,7 +429,7 @@ STATUS generateJSONSafeString(PCHAR pDst, UINT32 len)
     ENTERS();
     STATUS retStatus = STATUS_SUCCESS;
     UINT32 i = 0;
-    BYTE randBuffer[MAX_RAND_BUFFER_SIZE_FOR_NAME];
+    BYTE randBuffer[MAX_RAND_BUFFER_SIZE_FOR_NAME] = {0};
```
MushMal, line 432: Ugh.. This is really a false positive. However, I am OK to push this change. NOTE: Make sure the GCC 4.4 runs OK to validate the older compilers are OK with this construct. I believe C99 is OK with this initializer. If not, use MEMSET

PR #338, opened 2020-04-09
potential IceAgent deadlock

zhuker:
*Issue #, if available:*
Thread sanitizer is unhappy about timer queue executor lock
*Description of changes:*
unlock `iceagent.lock` while submitting `iceAgentStateTransitionTimerCallback` to timer queue

WARNING: i don't really know what i'm doing, so please scrutinize this PR   

### DIFF
--- a/src/source/Ice/IceAgent.c
+++ b/src/source/Ice/IceAgent.c
@@ -490,6 +490,7 @@ STATUS iceAgentSendPacket(PIceAgent pIceAgent, PBYTE pBuffer, UINT32 bufferLen)
     // can be invoking incomingDataHandler and cause deadlock.
     // pIceAgent->lock has to be non-reentrant !!
     MUTEX_UNLOCK(pIceAgent->lock);
+    locked = FALSE;
 
     retStatus = iceUtilsSendData(pBuffer,
                                  bufferLen,
--- a/src/source/Ice/IceAgentStateMachine.c
+++ b/src/source/Ice/IceAgentStateMachine.c
@@ -207,13 +207,36 @@ STATUS executeNewIceAgentState(UINT64 customData, UINT64 time)
 
     pIceAgent->iceAgentState = ICE_AGENT_STATE_NEW;
 
+    //assumes this is called while holding pIceAgent->lock
+    BOOL locked = TRUE;
+
+    //unlocking here to make thread sanitizer happy
+
+    // Thread1: TimerQueue thread
+    // timerQueueExecutor holds pTimerQueue->executorLock
+    //  calls iceAgentStateTransitionTimerCallback -> stepIceAgentStateMachine acquires pIceAgent->lock
+
+    // Thread2: main thread
+    // createPeerConnection -> createIceAgent -> stepIceAgentStateMachine holds pIceAgent->lock
+    // -> stepStateMachine -> executeNewIceAgentState tries to call timerQueueAddTimer which will acquire pTimerQueue->executorLock
+
+    // pTimerQueue->executorLock aka TQL
+    // pIceAgent->lock aka IAL
+    // deadlock graph TQL => IAL => TQL
+
+    MUTEX_UNLOCK(pIceAgent->lock);
+    locked = FALSE;
     CHK_STATUS(timerQueueAddTimer(pIceAgent->timerQueueHandle,
                                   0,
                                   pIceAgent->kvsRtcConfiguration.iceConnectionCheckPollingInterval,
                                   iceAgentStateTransitionTimerCallback,
                                   (UINT64) pIceAgent,
                                   &pIceAgent->iceAgentStateTimerCallback));
+
 CleanUp:
+    if (!locked) {
+        MUTEX_LOCK(pIceAgent->lock);
+    }
 
     LEAVES();
     return retStatus;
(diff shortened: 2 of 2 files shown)


src/source/Ice/IceAgent.c
```diff
@@ -490,6 +490,7 @@ STATUS iceAgentSendPacket(PIceAgent pIceAgent, PBYTE pBuffer, UINT32 bufferLen)
     // can be invoking incomingDataHandler and cause deadlock.
     // pIceAgent->lock has to be non-reentrant !!
     MUTEX_UNLOCK(pIceAgent->lock);
+    locked = FALSE;
```
zhuker, line 493: did someone just forget to set `locked = FALSE` my thread sanitizer complains about unlocking unlocked lock in `Cleanup`

src/source/Ice/IceAgentStateMachine.c
```diff
@@ -207,13 +207,36 @@ STATUS executeNewIceAgentState(UINT64 customData, UINT64 time)
...
+
+    // Thread2: main thread
+    // createPeerConnection -> createIceAgent -> stepIceAgentStateMachine holds pIceAgent->lock
+    // -> stepStateMachine -> executeNewIceAgentState tries to call timerQueueAddTimer which will acquire pTimerQueue->executorLock
+
+    // pTimerQueue->executorLock aka TQL
+    // pIceAgent->lock aka IAL
+    // deadlock graph TQL => IAL => TQL
+
+    MUTEX_UNLOCK(pIceAgent->lock);
```
zhuker, line 227: hacky workaround starts here

src/source/Ice/IceAgentStateMachine.c
```diff
@@ -200,15 +200,16 @@ STATUS executeNewIceAgentState(UINT64 customData, UINT64 time)
...
 
     CHK(pIceAgent != NULL, STATUS_NULL_ARG);
     // return early if we are already in ICE_AGENT_STATE_GATHERING
     CHK(pIceAgent->iceAgentState != ICE_AGENT_STATE_NEW, retStatus);
 
-    pIceAgent->iceAgentState = ICE_AGENT_STATE_NEW;
-
     //assumes this is called while holding pIceAgent->lock
-    BOOL locked = TRUE;
+    locked = TRUE;
```
zhuker, line 210: maybe there should be `BOOL IceAgent.locked` which is updated on every lock/unlock?

Sean-Der:
This is awesome, I am so excited you are digging into these @zhuker :fire:

Are you able to kill any of the suppressions we have now? I dropped in some really heavy handed ones, just wasn't able to fix them with the time I had available :(

zhuker:
@Sean-Der the one i hopefully fixed is a special case of `deadlock:stepIceAgentStateMachine` 
not sure i can remove it from suppression list
concurrency is a very subtle thing, i would triple check
@MushMal you wanna take a look?

src/source/Ice/IceAgentStateMachine.c
```diff
@@ -200,20 +200,42 @@ STATUS executeNewIceAgentState(UINT64 customData, UINT64 time)
...
+    //assumes this is called while holding pIceAgent->lock
+    BOOL locked = TRUE;
 
     CHK(pIceAgent != NULL, STATUS_NULL_ARG);
     // return early if we are already in ICE_AGENT_STATE_GATHERING
     CHK(pIceAgent->iceAgentState != ICE_AGENT_STATE_NEW, retStatus);
 
     pIceAgent->iceAgentState = ICE_AGENT_STATE_NEW;
 
+    //unlocking here to make thread sanitizer happy
```
MushMal, line 212: Ah, an interesting case! Nice job TSAN! Here is what's happening. The NEW state is only stepped into from the thread (the application thread) that calls peer connection create. The path leads to creating a new IceAgent which creates a state machine, sets things up and steps it. Each IceAgent object encapsulates its own state machine. We have a higher level stepping function of the state machine that does a few things and acquires a lock before stepping the state machine. Normally, the timer queue thread will get scheduled after this function returns and unlocks the ice agent lock. The timer queue timer that gets set is a multiple time firing timer, however, it's scheduled to start immediately (the second param is 0). The TSAN schedules the timer callback before the executeNewIceAgentState has a chance to return an unlock the lock.

I believe this needs to be properly fixed.

@chehefen has a massive rework of the Ice agent code. Let me pull him in so we can address this race condition properly.

zhuker: good 
thats exactly how i understood what was going on
and yes it's a stop gap to make thread sanitizer happy and avoid a _potential_ deadlock 
should be fixed properly
i was thinking of separate locks for timer queue and executor
queue gets locked for put and get operations

@chehefen what's your approach?

chehefen: I schedule the timer in iceAgentStartAgent: https://github.com/awslabs/amazon-kinesis-video-streams-webrtc-sdk-c/pull/322/files#diff-011ad0584af4444f14cb6ba52959c0a4R435-R444 instead. Since in the PR I got rid of the gathering state, and wait for credential state, state transition would start with check connection state, and there is no point checking connection until we receive the user name and password in iceAgentStartAgent. I also got rid of stepStateMachine call in createIceAgent for the same reason.

chehefen:
Hi @zhuker, the deadlock should be fixed in #322. Can you check out the latest commit? Thanks.

MushMal:
Closing this pull request

zhuker:
> Hi @zhuker, the deadlock should be fixed in #322. Can you check out the latest commit? Thanks.

i'll take a look next week, wrestling with infrastructure now

PR #339, opened 2020-04-09
Add support for test source in GStreamer sample

disa6302:
*Issue #, if available:*

*Description of changes:*

Adding a command line option to select test source for GStreamer pipeline. It would be useful for testing basic sample functionality on systems with no camera attached to get the frame

### DIFF
--- a/samples/Samples.h
+++ b/samples/Samples.h
@@ -40,6 +40,7 @@ typedef struct {
     volatile ATOMIC_BOOL updatingSampleStreamingSessionList;
     volatile ATOMIC_BOOL recreateSignalingClient;
     volatile SIZE_T streamingSessionListReadingThreadCount;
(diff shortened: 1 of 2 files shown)


disa6302:
@Mushmal, this is for running some tests on an EC2 instance for long run testing of the sample itself. 

And yes, I completely agree to the betterment argument parsing. We should improve it while migrating to the new repository. I have some changes locally that I have not pushed yet for this.

MushMal:
> @MushMal, this is for running some tests on an EC2 instance for long run testing of the sample itself. But yes, this option would be a good getting started for customers as well.
> 
> And yes, I completely agree to the betterment argument parsing. We should improve it while migrating to the new repository. I have some changes locally that I have not pushed yet for this.

Makes sense. Thank you!

Issue #328, opened 2020-04-06
DeleteSignalingChannel Public API call


TomyTomm:
https://docs.aws.amazon.com/kinesisvideostreams/latest/dg/API_DeleteSignalingChannel.html
There's an API call to delete signaling channels, but it would be nice to have that wrapped in a C-style API call.

MushMal:
@TomyTomm could you please describe your scenario? We do have a few backend public APIs that we are not modeling in the SDK as we model the lifecycle of the embedded devices. 

In particular, the DeleteSignalingChannel API is implemented in the SDK but not exposed publicly - we use it in the tests. 

Perhaps, if you have a compelling scenario we could expose it

TomyTomm:
Yeah, so I'm currently working on importing this SDK for use in ROS (Robot Operating System) ecosystem. So I have a TurtleBot3 that connects as Master and creates the signaling channel if it doesn't already exist. For cleanup purposes, I was thinking it would be nice to delete the signaling channel on destruction so resources aren't leaked.

MushMal:
Nice! Thanks. Let's use this issue to track this. Having clear customer scenario makes it easier to expose this publicly.

MushMal:
https://github.com/awslabs/amazon-kinesis-video-streams-webrtc-sdk-c/pull/346 got merged. Resolving

PR #346, opened 2020-04-14
Adding delete signaling channel API

MushMal:
Adding tests and fixing some of the existing tests

Issue #328 

*Description of changes:*

### DIFF
--- a/src/include/com/amazonaws/kinesis/video/webrtcclient/Include.h
+++ b/src/include/com/amazonaws/kinesis/video/webrtcclient/Include.h
@@ -268,6 +268,7 @@ extern "C" {
 #define STATUS_SIGNALING_INVALID_MESSAGE_TTL_VALUE                                  STATUS_SIGNALING_BASE + 0x0000002E
 #define STATUS_SIGNALING_ICE_CONFIG_REFRESH_FAILED                                  STATUS_SIGNALING_BASE + 0x0000002F
 #define STATUS_SIGNALING_RECONNECT_FAILED                                           STATUS_SIGNALING_BASE + 0x00000030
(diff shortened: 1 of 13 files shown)


src/include/com/amazonaws/kinesis/video/webrtcclient/Include.h
```diff
@@ -1557,45 +1565,63 @@ PUBLIC_API STATUS signalingClientGetIceConfigInfoCount(SIGNALING_CLIENT_HANDLE,
...
 
+/**
+ * @brief Deletes the signaling channel referenced by SIGNALING_CLIENT_HANDLE
+ *
+ * NOTE: The function is intended to be used to clean up the backend resources and
+ * as such should be called at the end of the lifecycle of the signaling channel resource.
+ * Attempting to connect to the channel or send a message will result in an
+ * error or an unpredictable results after this call.
+ *
+ * NOTE: The call transitions the signaling client state machine to a terminal state
```
chehefen, line 1615: so what happens when signalingClientDeleteSync fails but the signaling channel is not successfully deleted? Is it still possible to delete it using the sdk?

chehefen: when delete channel is called, can we just terminate the state machine and wait on delete api to be finished? The delete channel operation seems orthogonal to any of the state machine operations. This way in the state machine we dont need to worry if about delete happening.

MushMal: The state is a terminal state. There isn't much left for the application to do but to either issue the delete call again or free the client.

We can't simply terminate the state machine as the delete needs the state machine for a couple of reasons - 1) in case of the stale credentials we will need to iterate to get token state and 2) in case of the delete failing with 400 we might have a bad version in which case we need to call describe.

PR #348, opened 2020-04-14
fix broken turn unit test after turn server forbid using internal ip …

chehefen:
…as peer.

*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/tst/TurnConnectionFunctionalityTest.cpp
+++ b/tst/TurnConnectionFunctionalityTest.cpp
@@ -121,14 +121,15 @@ namespace com { namespace amazonaws { namespace kinesis { namespace video { name
         KvsIpAddress turnPeerAddr;
         PTurnPeer pTurnPeer = NULL;
         PDoubleListNode pDoubleListNode = NULL;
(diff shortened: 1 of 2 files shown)


tst/TurnConnectionFunctionalityTest.cpp
```diff
@@ -127,8 +127,8 @@ namespace com { namespace amazonaws { namespace kinesis { namespace video { name
         turnPeerAddr.port = (UINT16) getInt16(8080);
         turnPeerAddr.family = KVS_IP_FAMILY_TYPE_IPV4;
         turnPeerAddr.isPointToPoint = FALSE;
-        // random peer 10.1.1.1, we are not actually sending anything to it.
-        turnPeerAddr.address[0] = 0x0A;
+        /* random peer 77.1.1.1, we are not actually sending anything to it. */
```
MushMal, line 130: Wonder if we can #define these and use these directly?

chehefen: I dont think we can without making it excessively hard to read.

MushMal: I agree.

PR #352, opened 2020-04-15
Adding release notes file for PR

MushMal:
*Issue #, if available:*

NOTE: This is just the release notes to discuss/comment. This won't actually be pushed in. I am also terminating the build as it's not needed.

### DIFF
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -1,0 +1,34 @@
+
+This is the first release of Amazon Kinesis Video WebRTC C SDK
+
+Platforms tested on:
+- Linux
+- MacOS
+- x64
+- ARMv5
+
+Web Browsers tested on:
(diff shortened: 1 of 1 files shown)


RELEASE_NOTES.md
```diff
@@ -0,0 +1,34 @@
+
```
disa6302, line 1: Need to add release version.

MushMal: Cool - will do

MushMal: BTW, as this is for PR purposes only, I will be collecting the data and then using it. I believe the release version goes into GitHub tool

disa6302: Yes. Agreed. The notes would be going in when we cut the release tag.

PR #355, opened 2020-04-16
start dtlsSession when Ice is connected. Make sure dtleSessionStart d…

chehefen:
…ont get executed again once started.
*Issue #, if available:*

*Description of changes:*
Previously we call dtlsSessionStart along with IceAgentStartAgent, at that time ice is not connected yet, so if dtls need to initiate a handshake, the packet it sends will be lost. Only until dtls handshake timeout is hit will it retransmit handshake again. On my laptop there was noticeable reduction in time it took for picture to show up in JS viewer when running the master sample using relay only.

### DIFF
--- a/src/source/Dtls/Dtls.h
+++ b/src/source/Dtls/Dtls.h
@@ -23,7 +23,9 @@ extern "C" {
 /*
  * DTLS transmission interval timer (in 100ns)
  */
-#define DTLS_TRANSMISSION_INTERVAL (200 * HUNDREDS_OF_NANOS_IN_A_MILLISECOND)
+#define DTLS_TRANSMISSION_INTERVAL          (200 * HUNDREDS_OF_NANOS_IN_A_MILLISECOND)
+
+#define DTLS_SESSION_TIMER_START_DELAY      (100 * HUNDREDS_OF_NANOS_IN_A_MILLISECOND)
 
 #define LOG_OPENSSL_ERROR(s)                    while ((sslErr = ERR_get_error()) != 0) { \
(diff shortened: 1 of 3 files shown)


Sean-Der:
@chehefen this makes sense to me. DTLS ClientHello increase ^2 every time.  So depending on how long ICE takes you could have to wait 2 seconds (or maybe 4 or 8)

IMO it would be nicer if we could just force really quick retries all the time (and always do 1 second). Maybe more network usage, but nicer for us :)

src/source/Dtls/Dtls.h
```diff
@@ -61,13 +61,13 @@ typedef struct {
 } DtlsSessionCertificateInfo, *PDtlsSessionCertificateInfo;
 
 typedef struct {
+    volatile BOOL isStarted;
```
MushMal, line 64: Does this need to be atomic?

chehefen: no it doesnt. it's always accessed within lock.

chehefen:
> @chehefen this makes sense to me. DTLS ClientHello increase ^2 every time. So depending on how long ICE takes you could have to wait 2 seconds (or maybe 4 or 8)
> 
> IMO it would be nicer if we could just force really quick retries all the time (and always do 1 second). Maybe more network usage, but nicer for us :)

It doesnt seem like there is an easy way to do that. The closest thing I can find is BIO_set_ssl_renegotiate_timeout, but it operates on BIOs and we have to revamp our dtls setup to do use it.

PR #361, opened 2020-04-17
two video tracks with same codec

zhuker:
*Issue #, if available:*

*Description of changes:*
i receive offer for two video tracks with the same codec 
i add two transceivers with VP8 tracks
expected answer MUST contain two different ssrc

### DIFF
--- a/src/source/PeerConnection/SessionDescription.c
+++ b/src/source/PeerConnection/SessionDescription.c
@@ -666,13 +666,15 @@ STATUS copyTransceiverWithCodec(PKvsPeerConnection pKvsPeerConnection, RTC_CODEC
     *pDidFindCodec = FALSE;
 
     CHK_STATUS(doubleListGetHeadNode(pKvsPeerConnection->pTransceievers, &pCurNode));
-    while (pCurNode != NULL && pTargetKvsRtpTransceiver == NULL) {
+    while (pCurNode != NULL) {
         CHK_STATUS(doubleListGetNodeData(pCurNode, &data));
-        pCurNode = pCurNode->pNext;
         pKvsRtpTransceiver = (PKvsRtpTransceiver) data;
         if (pKvsRtpTransceiver != NULL && pKvsRtpTransceiver->sender.track.codec == rtcCodec) {
             pTargetKvsRtpTransceiver = pKvsRtpTransceiver;
+            doubleListDeleteNode(pKvsPeerConnection->pTransceievers, pCurNode);
+            break;
         }
+        pCurNode = pCurNode->pNext;
     }
     if (pTargetKvsRtpTransceiver != NULL) {
         CHK_STATUS(doubleListInsertItemTail(pKvsPeerConnection->pTransceievers, (UINT64) pTargetKvsRtpTransceiver));
@@ -757,10 +759,6 @@ STATUS reorderTransceiverByRemoteDescription(PKvsPeerConnection pKvsPeerConnecti
         }
     }
 
-    // delete the unordered part in pKvsPeerConnection->pTransceievers
-    for (i = 0; i < transceieverCount; ++i) {
-        CHK_STATUS(doubleListDeleteHead(pKvsPeerConnection->pTransceievers));
-    }
 
 CleanUp:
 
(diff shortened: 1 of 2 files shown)


src/source/PeerConnection/SessionDescription.c
```diff
@@ -666,13 +666,15 @@ STATUS copyTransceiverWithCodec(PKvsPeerConnection pKvsPeerConnection, RTC_CODEC
...
 
     CHK_STATUS(doubleListGetHeadNode(pKvsPeerConnection->pTransceievers, &pCurNode));
-    while (pCurNode != NULL && pTargetKvsRtpTransceiver == NULL) {
+    while (pCurNode != NULL) {
         CHK_STATUS(doubleListGetNodeData(pCurNode, &data));
-        pCurNode = pCurNode->pNext;
         pKvsRtpTransceiver = (PKvsRtpTransceiver) data;
         if (pKvsRtpTransceiver != NULL && pKvsRtpTransceiver->sender.track.codec == rtcCodec) {
             pTargetKvsRtpTransceiver = pKvsRtpTransceiver;
+            doubleListDeleteNode(pKvsPeerConnection->pTransceievers, pCurNode);
```
MushMal, line 674: This will not delete the underlying object so you will leak memory.

src/source/PeerConnection/SessionDescription.c
```diff
@@ -666,13 +666,15 @@ STATUS copyTransceiverWithCodec(PKvsPeerConnection pKvsPeerConnection, RTC_CODEC
...
+            break;
```
MushMal, line 675: In general, we are trying to avoid break; and especially continue; control flow statements within loops. In case of simple loops. it's OK. However, consider a termination based on the condition check in the while loop like it was before. This is MINOR suggestion and can be safely ignored.

src/source/PeerConnection/SessionDescription.c
```diff
@@ -757,10 +759,6 @@ STATUS reorderTransceiverByRemoteDescription(PKvsPeerConnection pKvsPeerConnecti
         }
     }
 
-    // delete the unordered part in pKvsPeerConnection->pTransceievers
```
MushMal, line 760: Are we sure we need to remove these? If my reading of the code is right (and I am rusty here) the ordered part will be copied at the tail so the front might need to be removed.

MushMal: Also, it doesn't let me comment on LN 697 RTC_CODEC rtcCodec; - static tooling complains that this might be uninitialized. We need to either MEMSET it or initialize with {0}

src/source/PeerConnection/SessionDescription.c
zhuker: the purpose of this method is to sort transceivers, so i delete it here and add to the end of linked list on line 680

src/source/PeerConnection/SessionDescription.c
zhuker: > Are we sure we need to remove these? If my reading of the code is right (and I am rusty here) the ordered part will be copied at the tail so the front might need to be removed.

i deleted these in `copyTransceiverWithCodec` - no need to delete again

before:
1. transceivers = `unsorted: transceiver2, transceiver1, sorted: transceiver1, transceiver2`
2. delete 2 transceivers
3. transceivers = `sorted: transceiver1, transceiver2`

after (with the modifications in this PR):
1. transceivers = `sorted: transceiver1, transceiver2`

zhuker: > Also, it doesn't let me comment on LN 697 RTC_CODEC rtcCodec; - static tooling complains that this might be uninitialized. We need to either MEMSET it or initialize with {0}

i noticed it as well, but there's no `RTC_CODEC_UNKNOWN` enum, i think it should be done in a separate PR.

src/source/PeerConnection/SessionDescription.c
zhuker: i think this loop deserves the `simple loop` badge :)

MushMal: OK, you convince me

PR #362, opened 2020-04-19
Fixing https://github.com/awslabs/amazon-kinesis-video-streams-webrtc…

MushMal:
…-sdk-c/issues/307

It seems that we are directly leaking 95 object with total size of 9120 which evaluates to the actual size of 95 RtpPacket size objects.

Issue #307 

*Description of changes:*


Sean-Der:
@MushMal Do you know why this doesn't happen every time? Are you able to reproduce/have a test for it?

MushMal:
> @MushMal Do you know why this doesn't happen every time? Are you able to reproduce/have a test for it?

I looked at the size of the leaked objects and the ASAN stack. The casual code inspection showed the root cause. Basically, if we ever fail to jitterBufferPush the rtp packet we will only free the payload and not the wrapper. The failure could happen likely when the callback onFrameReadyFn, onFrameDroppedFn or depayPayloadFn fail as other paths should be OK.

I initially suspected the depayPayloadFn must be failing - this is content dependent (currently, we have depayOpusFromRtpPayload, depayG711FromRtpPayload, depayH264FromRtpPayload and depayVP8FromRtpPayload) but the functions are pretty straight forward but nevertheless they depend on the content. onFrameReadyFunc might also cause a failure.

I really don't have much more context to work with here as the ASAN stack only has the thread stack but I believe we could craft a "bad" content that will cause a failure in one of these (say h264 depay APIs) which would cause the leak.

chehefen:
can you also remove the corresponding suppression here: https://github.com/awslabs/amazon-kinesis-video-streams-webrtc-sdk-c/blob/master/tst/suppressions/LSAN.supp#L10? thanks

MushMal:
> can you also remove the corresponding suppression here: https://github.com/awslabs/amazon-kinesis-video-streams-webrtc-sdk-c/blob/master/tst/suppressions/LSAN.supp#L10? thanks

Oh, sure. Let me do that

tst/SdpApiTest.cpp
```diff
@@ -472,6 +472,8 @@ a=end-of-candidates
     }
 
     ASSERT_EQ(2, ssrcLines.size());
+
+    EXPECT_EQ(STATUS_SUCCESS, freePeerConnection(&pRtcPeerConnection));
```
Sean-Der, line 476: This is really concerning. Our instrumented stuff is really nice, but it can't catch the boundaries between libraries (we have had lots of leaks in OpenSSL usage)

I will investigate

PR #378, opened 2020-04-23
In samples, ignore new offer for existing session. In non-trickle mod…

chehefen:
…e, acquire lock before block waiting for ice candidate collection to finish.

*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/samples/Common.c
+++ b/samples/Common.c
@@ -121,7 +121,7 @@ STATUS masterMessageReceived(UINT64 customData, PReceivedSignalingMessage pRecei
     PSampleConfiguration pSampleConfiguration = (PSampleConfiguration) customData;
     PSampleStreamingSession pSampleStreamingSession = NULL;
     UINT32 i;
-    BOOL locked = FALSE;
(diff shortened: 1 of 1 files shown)


samples/Common.c
```diff
@@ -133,6 +133,7 @@ STATUS masterMessageReceived(UINT64 customData, PReceivedSignalingMessage pRecei
     for (i = 0; i < pSampleConfiguration->streamingSessionCount && pSampleStreamingSession == NULL; ++i) {
         if (0 == STRCMP(pReceivedSignalingMessage->signalingMessage.peerClientId, pSampleConfiguration->sampleStreamingSessionList[i]->peerId)) {
             pSampleStreamingSession = pSampleConfiguration->sampleStreamingSessionList[i];
+            sessionExisted = TRUE;
```
MushMal, line 136: break out out of the loop

chehefen: it already does because pSampleStreamingSession != NULL.

PR #379, opened 2020-04-23
Fix the static build

InspectorConstructor:
1. Rename CMake variable "BUILD_STATIC" to "BUILD_STATIC_LIBS" to better reflect what it does.
2. Remove unnecessary "-static" linker flag. It affected our executable cmake targets, which we don't want. The wasn't necessary to build libraries statically, either, as that is specified to the cmake add_library() call.
3. Build libwebsocket without libcap. libwebsockets tries to build with libcap, and this results in needing to specify linking with lcap in top-level executable cmake targets. (_If this is inappropriate, please let me know and I can option()-alize it._)


InspectorConstructor:
Thank you for approving. I don't have write access, so an awslabs member needs to merge, whenever suitable.

Issue #115, opened 2020-01-03
IPv6 Support


Sean-Der:
We need IPv6 support, we should allow the user to opt in/out. In some cases you have IPv6, but all packets are discarded so just causes spam in the logs.

lherman-cs:
Status:
* Added host candidates support: https://github.com/awslabs/amazon-kinesis-video-streams-webrtc-sdk-c/pull/382
* Added reflexive candidates support: https://github.com/awslabs/amazon-kinesis-video-streams-webrtc-sdk-c/pull/417

lherman-cs:
I'm closing this issue we have added essential IPv6 support.

PR #382, opened 2020-04-23
Add IPv6 support for host candidates

lherman-cs:
* Issue #115 

* Description of changes:
  * Add IPv6 support for host candidates
  * Refractor `iceAgentAddRemoteCandidate`

* Todo:
  * Add IPv6 support for reflexive candidates
    * Verify the XOR mapped address for IPv6 implementation follows STUN RFC, https://tools.ietf.org/html/rfc5389#section-15.2 (https://github.com/awslabs/amazon-kinesis-video-streams-webrtc-sdk-c/issues/394)
  * Add IPv6 support for TURN, https://tools.ietf.org/html/rfc6156
  * Update `getIpWithHostName` to use `getaddrinfo` instead of `gethostbyname` since `gethostbyname` doesn't support IPv6 (https://github.com/awslabs/amazon-kinesis-video-streams-webrtc-sdk-c/issues/395)

### DIFF
--- a/src/source/Ice/IceAgent.c
+++ b/src/source/Ice/IceAgent.c
@@ -255,9 +255,9 @@ STATUS iceAgentAddRemoteCandidate(PIceAgent pIceAgent, PCHAR pIceCandidateString
     PIceCandidate pIceCandidate = NULL, pDuplicatedIceCandidate = NULL;
     PCHAR curr, tail, next;
     UINT32 tokenLen, portValue, remoteCandidateCount, i, len;
-    BOOL foundIpAndPort = FALSE, freeIceCandidateIfFail = TRUE;
-    BOOL foundIp6 = FALSE;
-    CHAR ip6Buf[KVS_IP_ADDRESS_STRING_BUFFER_LEN];
+    BOOL freeIceCandidateIfFail = TRUE;
+    BOOL foundIp = FALSE, foundPort = FALSE;
+    CHAR ipBuf[KVS_IP_ADDRESS_STRING_BUFFER_LEN];
     KvsIpAddress candidateIpAddr;
 
     CHK(pIceAgent != NULL && pIceCandidateString != NULL, STATUS_NULL_ARG);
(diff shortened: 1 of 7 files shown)


src/source/Ice/IceAgent.c
```diff
@@ -256,12 +256,15 @@ STATUS iceAgentAddRemoteCandidate(PIceAgent pIceAgent, PCHAR pIceCandidateString
     PCHAR curr, tail, next;
     UINT32 tokenLen, portValue, remoteCandidateCount, i;
     BOOL foundIpAndPort = FALSE, freeIceCandidateIfFail = TRUE;
+    BOOL foundIp6 = FALSE;
+    CHAR ip6Buf[KVS_MAX_IPV6_ADDRESS_STRING_LEN+1];
```
MushMal, line 260: Might need to setup some editor auto-formatting preferences. In this case, the '+' sign should have spaces around it

src/source/Ice/IceAgent.c
```diff
@@ -256,12 +256,15 @@ STATUS iceAgentAddRemoteCandidate(PIceAgent pIceAgent, PCHAR pIceCandidateString
...
     KvsIpAddress candidateIpAddr;
 
     CHK(pIceAgent != NULL && pIceCandidateString != NULL, STATUS_NULL_ARG);
     CHK(!IS_EMPTY_STRING(pIceCandidateString), STATUS_INVALID_ARG);
 
     MEMSET(&candidateIpAddr, 0x00, SIZEOF(KvsIpAddress));
+    MEMSET(ip6Buf, 0x00, KVS_MAX_IPV6_ADDRESS_STRING_LEN+1);
```
MushMal, line 267: ditto. I won't add comments about the formatting.

src/source/Ice/IceAgent.c
```diff
@@ -275,22 +278,25 @@ STATUS iceAgentAddRemoteCandidate(PIceAgent pIceAgent, PCHAR pIceCandidateString
...
 
             candidateIpAddr.port = htons(portValue);
-            candidateIpAddr.family = KVS_IP_FAMILY_TYPE_IPV4;
+            candidateIpAddr.family = foundIp6 ? KVS_IP_FAMILY_TYPE_IPV6 : KVS_IP_FAMILY_TYPE_IPV4;
         } else if (STRNCHR(curr, tokenLen, '.') != NULL) {
             CHK(tokenLen <= KVS_MAX_IPV4_ADDRESS_STRING_LEN, STATUS_ICE_CANDIDATE_STRING_INVALID_IP); // IPv4 is 15 characters at most
             CHK_STATUS(populateIpFromString(&candidateIpAddr, curr));
+        } else {
+            STRNCPY(ip6Buf, curr, KVS_MAX_IPV6_ADDRESS_STRING_LEN);
+            foundIp6 = inet_pton(AF_INET6, ip6Buf, candidateIpAddr.address) == 1 ? TRUE : FALSE;
```
MushMal, line 291: This is actually Windows supported.

src/source/Ice/IceAgent.c
```diff
@@ -275,22 +278,25 @@ STATUS iceAgentAddRemoteCandidate(PIceAgent pIceAgent, PCHAR pIceCandidateString
...
         }
 
         curr = next + 1;
-        foundIpAndPort = (candidateIpAddr.port != 0) && (candidateIpAddr.address[0] != 0);
+        foundIpAndPort = (candidateIpAddr.port != 0) && ((candidateIpAddr.address[0] != 0) || foundIp6);
```
MushMal, line 295: Perhaps I am not following the logic correctly. KvsIpAddress can hold both IPv4 and IPv6 addresses. The inners of the while loop should set this param and you can check this param only. Also, consult with Felix, I am not sure why the check is against 0s. Can't the first byte of the address be 0 and the port be 0? I think an explicit 'found' boolean should be a good sentinel rather than checking for zeroed values.

src/source/Ice/IceAgent.c
```diff
@@ -1047,6 +1055,9 @@ STATUS iceAgentSendSrflxCandidateRequest(PIceAgent pIceAgent)
             switch(pCandidate->iceCandidateType) {
                 case ICE_CANDIDATE_TYPE_SERVER_REFLEXIVE:
                     pIceServer = &(pIceAgent->iceServers[pCandidate->iceServerIndex]);
+                    if (pIceServer->ipAddress.family != pCandidate->ipAddress.family) {
+                      break;
```
MushMal, line 1059: We are trying to minimize the usage of the control flow keywords such as break and continue. It's not a good coding style to use break within a switch all together. Use the negation of the boolean predicate here

if (family == other) {
  // do the stuff
}

src/source/Ice/IceAgent.c
```diff
@@ -1324,7 +1335,9 @@ STATUS iceAgentInitSrflxCandidate(PIceAgent pIceAgent)
         pCurNode = pCurNode->pNext;
         pCandidate = (PIceCandidate) data;
 
-        if (pCandidate->iceCandidateType == ICE_CANDIDATE_TYPE_HOST) {
+        // TODO: Stop skipping IPv6. Stun serialization and deserialization needs to be implemented properly first.
```
MushMal, line 1338: Add a GitHub issue tracking this work

src/source/Ice/IceAgent.c
```diff
@@ -1402,8 +1415,15 @@ STATUS iceAgentInitRelayCandidate(PIceAgent pIceAgent)
             pSocketAddrForTurn = NULL;
             // if an VPN interface (isPointToPoint is TRUE) is found, use that instead
             for(i = 0; i < pIceAgent->localNetworkInterfaceCount; ++i) {
-                if(pIceAgent->localNetworkInterfaces[i].family == pIceAgent->iceServers[j].ipAddress.family &&
-                   (pSocketAddrForTurn == NULL || pIceAgent->localNetworkInterfaces[i].isPointToPoint)) {
+                // TODO: Stop skipping IPv6. Stun serialization and deserialization needs to be implemented properly first.
```
MushMal, line 1418: Same comment about GitHub issue tracking it

src/source/Ice/IceAgent.c
chehefen: can use KVS_IP_ADDRESS_STRING_BUFFER_LEN instead

src/source/Ice/IceAgent.c
chehefen: use SIZEOF(ip6Buf) ?

src/source/Ice/IceAgent.c
lherman-cs: Do we have a standard format config? Something like .clang-format?

Ah, didn't realize that we have `KVS_IP_ADDRESS_STRING_BUFFER_LEN`. I'll update it. Thanks.

src/source/Ice/IceAgent.c
lherman-cs: Yep, makes sense.

src/source/Ice/IceAgent.c
lherman-cs: Got it. I'll update it.

src/source/Ice/IceAgent.c
lherman-cs: I've discussed this offline with @chehefen. He didn't write this part. But, we agreed that it makes more sense to have a boolean to check this instead. So, I'll update the PR to use a boolean instead.

src/source/Ice/IceAgent.c
lherman-cs: https://github.com/awslabs/amazon-kinesis-video-streams-webrtc-sdk-c/issues/394

src/source/Ice/IceAgent.c
lherman-cs: Same comment as above.

Issue #376, opened 2020-04-22
Backend API call caching policy implementation


MushMal:
Implement various backend caching policies. Currently, we don't cache calls and make every call directly to the backend.

The idea is to have the following caching policies

* No caching (current behavior)
* Cache all APIs (All of the APIs are being cached and the TTL for the value is specified)
* Cache all and not block on ICE server candidates (This is similar to the one above with the exception that it will not block on the ICE config but move directly to Ready state and handle ICE server config call async)

NOTE: Currently, If the ChannelInfo specifies ARN already, it will be assumed that the channel is pre-created and describe and create states will be skipped.

chehefen:
whats the reasoning behind making getIceServerConifg async? If ICE needs to use TURN then turn servers must be available before ICE start. So application still need to wait for getIceServerConifg to finish before it can proceed. Also cachings are done in memory. They need to be persisted into disk space ask well to support case where kvs client is terminated.

MushMal:
> whats the reasoning behind making getIceServerConifg async? If ICE needs to use TURN then turn servers must be available before ICE start. So application still need to wait for getIceServerConifg to finish before it can proceed. Also cachings are done in memory. They need to be persisted into disk space ask well to support case where kvs client is terminated.

Client application will still need to connect which can benefit from async ice config retrieval.

disa6302:
Change merged

PR #385, opened 2020-04-24
Implementing front-end API caching policy

MushMal:
* Describe and Get Endpoint API caching policy
* Flag in ChannelInfo to return without blocking for ICE server config retrieval
* Async ICE server config retrieval (opt-in)
* Disconnect public API
* Unit test coverage

Issue #376 

*Description of changes:*


src/source/Signaling/LwsApiCalls.c
```diff
@@ -1117,6 +1081,8 @@ STATUS deleteChannelLws(PSignalingClient pSignalingClient, UINT64 time)
 {
     ENTERS();
     STATUS retStatus = STATUS_SUCCESS;
+    UNUSED_PARAM(time);
```
disa6302, line 1084: Why arent we using the time parameter anymore?

src/include/com/amazonaws/kinesis/video/webrtcclient/Include.h
```diff
@@ -1018,6 +1025,23 @@ typedef struct {
     CHAR description[MAX_MESSAGE_DESCRIPTION_LEN + 1]; //!< Optional description of the message
 } ReceivedSignalingMessage, *PReceivedSignalingMessage;
 
+/**
+ * @brief Type of caching implementation to use with the signaling client
+ */
+typedef enum {
```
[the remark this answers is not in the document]
MushMal: There might be. For example, one of the possible ways would be file-backed caching. In this case, the cache survives the app re-launch. We can do this if we get some customer feedback that this is needed. So far, I don't see it

src/source/Signaling/LwsApiCalls.c
MushMal: The caching aggregate should handle the awaiting now. The caching aggregate then calls the LWS routine to perform the operation if the call actually need to be made. In case of delete, we don't do caching.

tst/suppressions/TSAN.supp
```diff
@@ -358,3 +358,100 @@ race:PeerConnectionFunctionalityTest_freeTurnDueToP2PFoundAfterTurnEstablished
...
+#
+#     Hint: use TSAN_OPTIONS=second_deadlock_stack=1 to get more informative warning message
+#
+#   Mutex M877774268853663472 acquired here while holding mutex M880307543644078960 in thread T572:
+#     #0 pthread_mutex_lock /tmp/final/llvm.src/projects/compiler-rt/lib/tsan/../sanitizer_common/sanitizer_common_interceptors.inc:4071:3 (webrtc_client_test+0x4fcb0e)
+#     #1 defaultLockMutex /home/travis/build/awslabs/amazon-kinesis-video-streams-webrtc-sdk-c/open-source/amazon-kinesis-video-streams-pic/src/utils/src/Mutex.c:202:5 (libkvsWebrtcClient.so+0x6b9b8)
+#     #2 stepSignalingStateMachine /home/travis/build/awslabs/amazon-kinesis-video-streams-webrtc-sdk-c/src/source/Signaling/StateMachine.c:40:5 (libkvsWebrtcSignalingClient.so+0x23815)
+#     #3 refreshIceConfigurationCallback /home/travis/build/awslabs/amazon-kinesis-video-streams-webrtc-sdk-c/src/source/Signaling/Signaling.c:593:9 (libkvsWebrtcSignalingClient.so+0x1ee93)
+#     #4 timerQueueExecutor /home/travis/build/awslabs/amazon-kinesis-video-streams-webrtc-sdk-c/open-source/amazon-kinesis-video-streams-pic/src/utils/src/TimerQueue.c:561:37 (libkvsWebrtcClient.so+0x72745)
```
chehefen, line 399: if timer dont hold execute lock while calling the callback we can avoid this deadlock. The executor thread can keep a local copy of timer tasks and update when necessary. Similar pattern in connectionListener executor

MushMal: I think it's doable. Release/re-acquire lock for the call. We should implement it soon.

PR #392, opened 2020-04-27
disable instrumented allocator

chehefen:
*Issue #, if available:*
disable instrumented allocator due to inconsistent behavior. Sometimes it reports leak like this job: https://travis-ci.com/github/awslabs/amazon-kinesis-video-streams-webrtc-sdk-c/jobs/323627407. But re-running the job again it passes.
*Description of changes:*

### DIFF
--- a/tst/WebRTCClientTestFixture.cpp
+++ b/tst/WebRTCClientTestFixture.cpp
@@ -36,7 +36,9 @@ void WebRtcClientTestBase::SetUp()
     mExpectedFrameCount = 0;
     mExpectedDroppedFrameCount = 0;
 
-    SET_INSTRUMENTED_ALLOCATORS();
+    /* Disable instrumented allocator for now due to inconsistent behavior.
+     * Sometimes it report leaks but running the same test again it doesnt report leak again. */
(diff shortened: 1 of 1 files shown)


tst/WebRTCClientTestFixture.cpp
```diff
@@ -36,7 +36,9 @@ void WebRtcClientTestBase::SetUp()
     mExpectedFrameCount = 0;
     mExpectedDroppedFrameCount = 0;
 
-    SET_INSTRUMENTED_ALLOCATORS();
+    /* Disable instrumented allocator for now due to inconsistent behavior.
```
MushMal, line 39: Don't change the code. Modify tst/CMakeList.txt and comment

add_definitions(-DINSTRUMENTED_ALLOCATORS)

chehefen: it doesnt work. the build fails.

MushMal: Oh, I forgot that we need to fix this in PIC first. OK then. We will deal with this once this is fixed in PIC properly.

PR #407, opened 2020-04-30
add ability to configure TURN protocol.

chehefen:
*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/samples/Common.c
+++ b/samples/Common.c
@@ -352,7 +352,14 @@ STATUS initializePeerConnection(PSampleConfiguration pSampleConfiguration, PRtcP
             CHK_STATUS(signalingClientGetIceConfigInfo(pSampleConfiguration->signalingClientHandle, i, &pIceConfigInfo));
             for (j = 0; j < pIceConfigInfo->uriCount; j++) {
                 CHECK(uriCount < MAX_ICE_SERVERS_COUNT);
-                STRNCPY(configuration.iceServers[uriCount + 1].urls, pIceConfigInfo->uris[j], MAX_ICE_CONFIG_URI_LEN);
+                /* - change url to "turn:ip:port?transport=udp" to do turn through UDP
+                 * - change url to "turn:ip:port?transport=tcp" will do turn through TCP/TLS currently,
+                 *   but later will be changed to plain tcp
+                 * - change url to "turns:ip:port?transport=tcp" to do turn through TCP/TLS
+                 * - "turns:ip:port?transport=udp" will just do UDP.
+                 * Currently if "transport=" is missing, ICE will try both UDP and TCP/TLS
+                 * By default use UDP since it's fastest. */
+                SNPRINTF(configuration.iceServers[uriCount + 1].urls, MAX_ICE_CONFIG_URI_LEN, "%s?transport=udp", pIceConfigInfo->uris[j]);
                 STRNCPY(configuration.iceServers[uriCount + 1].credential, pIceConfigInfo->password, MAX_ICE_CONFIG_CREDENTIAL_LEN);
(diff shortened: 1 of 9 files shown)


samples/Common.c
```diff
@@ -352,7 +352,14 @@ STATUS initializePeerConnection(PSampleConfiguration pSampleConfiguration, PRtcP
...
             for (j = 0; j < pIceConfigInfo->uriCount; j++) {
                 CHECK(uriCount < MAX_ICE_SERVERS_COUNT);
-                STRNCPY(configuration.iceServers[uriCount + 1].urls, pIceConfigInfo->uris[j], MAX_ICE_CONFIG_URI_LEN);
+                /* - change url to "turn:ip:port?transport=udp" to do turn through UDP
+                 * - change url to "turn:ip:port?transport=tcp" will do turn through TCP/TLS currently,
+                 *   but later will be changed to plain tcp
+                 * - change url to "turns:ip:port?transport=tcp" to do turn through TCP/TLS
+                 * - "turns:ip:port?transport=udp" will just do UDP.
+                 * Currently if "transport=" is missing, ICE will try both UDP and TCP/TLS
+                 * By default use UDP since it's fastest. */
```
Sean-Der, line 361: I wouldn't do `%s?transport=udp`, you are going to see a slower success rate. With trickle ICE it should just use UDP TURN if it connects first!

chehefen: this is mainly for the alexa case. Trying to do allocation just once

src/source/Ice/IceAgent.c
```diff
@@ -1460,69 +1436,15 @@ STATUS iceAgentInitRelayCandidate(PIceAgent pIceAgent)
...
-                    pCurNode = pCurNode->pNext;
-                    pCandidate = (PIceCandidate) data;
-
-                    // TODO: Stop skipping IPv6. Since we're allowing IPv6 remote candidates from iceAgentAddRemoteCandidate for host candidates,
-                    // it's possible to have a situation where the turn server uses IPv4 and the remote candidate uses IPv6.
-                    if (IS_IPV4_ADDR(&pCandidate->ipAddress)) {
-                      CHK_STATUS(turnConnectionAddPeer(pCurrentTurnConnectionTracker->pTurnConnection,
-                                                       &pCandidate->ipAddress));
-                    }
+                if (pIceAgent->iceServers[j].transport == KVS_SOCKET_PROTOCOL_UDP ||
```
MushMal, line 1439: Perhaps instead of the else cases you could simply pass the pIceAgent->iceServers[j].transport param directly to iceAgentInitRelayCandidate and handle the NONE case there?

chehefen: I cant, each turnConnection has to be either UDP to TCP. turnConnection cant create two copies of itself for UDP and TCP, has to be done by the caller.

MushMal: What I mean is in else clause here to call CHK_STATUS(iceAgentInitRelayCandidate(pIceAgent, pSocketAddrForTurn, j, pIceAgent->iceServers[j].transport)); directly.

then, inside the iceAgentInitRelayCandidate, do the assumption that if the transport is NONE then assume it UDP - like what you are doing here. 

In general, having this else if/else if with both if statements ORing with PROTOCOL_NONE is not a good idea as both with match the NONE but the first will take precedence.

chehefen: what I want to do here is if transport is NONE, then create both TCP and UDP. if I pass transport to iceAgentInitRelayCandidate and assume UDP if it is NONE, it will create two UDPs which is not what I want.

MushMal: Oh, got it. Didn't notice that the second one is not else if

PR #413, opened 2020-05-01
adjust KVS_ICE_ENTER_STATE_DISCONNECTION_GRACE_PERIOD to two times ke…

### LINKED ISSUE #387: [Question] Recommended Steps Handling Disconnection/Lost Transceiver?
We are testing streaming for as long as we can. The current issue we are encountering is that if for some reason when the master side disconnects and went back online, a naive viewer would not have detected it, and the video freezes indefinitely. One thing we can do on the viewer side can be monitoring the video stats through the (https://developer.mozilla.org/en-US/docs/Web/API/WebRTC_Statistics_API), and retry connection if no video for more than n seconds, and possibly send some message through the data channel to tell the master there is no video. The master then can perform a quick diagnostics on recovering some of the threads or decide to reboot itself, then inform the viewer what the master is going to do. Then the viewer performs some more steps depending on the master's choice.

So our question is would this be a good approach? Or are there some APIs we can call, or some parameters/settings we can tweak to solve this problem?

chehefen:
…ep alive period.

*Issue #, if available:*
fix #387 
*Description of changes:*

### DIFF
--- a/src/source/Ice/IceAgent.h
+++ b/src/source/Ice/IceAgent.h
@@ -26,7 +26,7 @@ extern "C" {
 /* Control the calling rate of iceCandidateGatheringTimerTask. Can affect STUN TURN candidate gathering time */
 #define KVS_ICE_GATHER_CANDIDATE_TIMER_POLLING_INTERVAL                 50 * HUNDREDS_OF_NANOS_IN_A_MILLISECOND
 
-/* Should see at least one keep alive within this period */
+/* ICE should've received at least one keep alive within this period. Since keep alives are send every 15s */
 #define KVS_ICE_ENTER_STATE_DISCONNECTION_GRACE_PERIOD                  2 * KVS_ICE_SEND_KEEP_ALIVE_INTERVAL
 #define KVS_ICE_ENTER_STATE_FAILED_GRACE_PERIOD                         15 * HUNDREDS_OF_NANOS_IN_A_SECOND
 


src/source/Ice/IceAgent.h
```diff
@@ -26,8 +26,8 @@ extern "C" {
 /* Control the calling rate of iceCandidateGatheringTimerTask. Can affect STUN TURN candidate gathering time */
 #define KVS_ICE_GATHER_CANDIDATE_TIMER_POLLING_INTERVAL                 50 * HUNDREDS_OF_NANOS_IN_A_MILLISECOND
 
-/* Should see at least one keep alive within this period */
-#define KVS_ICE_ENTER_STATE_DISCONNECTION_GRACE_PERIOD                  2 * KVS_ICE_SEND_KEEP_ALIVE_INTERVAL
+/* ICE should've received at least one keep alive within this period. Since keep alives are send every 15s */
+#define KVS_ICE_ENTER_STATE_DISCONNECTION_GRACE_PERIOD                  30 * HUNDREDS_OF_NANOS_IN_A_SECOND
```
MushMal, line 30: I don't understand this change.

chehefen: it should be 2 * KVS_ICE_SEND_KEEP_ALIVE_INTERVAL. changed back.

Issue #368, opened 2020-04-21
Request for support of peerConnectionGetLocalDescription()


enobufs:
In my app, there are nodes that do not support trickle ICE, and I'm hoping to send SDP after all local candidates are gathered.

I noticed the current implementation did not have a method equivalent to WebRTC API `localDescription()` which would return pending local description. I tried peerConnectionGetCurrentLocalDescription but it had nothing in it: ```{"type": "answer", "sdp": ""}``` (though it is offerer..)

chehefen:
we do implement support for non trickle ice here: https://github.com/awslabs/amazon-kinesis-video-streams-webrtc-sdk-c/blob/master/samples/Common.c#L583. You can take a look at our viewer sample.

enobufs:
@chehefen Looking at the sample, I see the sample doing basically the same thing, but mine is offerer.

I was looking for the offerer side of code in the sample, but I could not locate lines that are waiting for cvar with candidateGatheringDone being true for non-trickle behavior.

My code does:
1. createOffer() - I see it generates 'desc' without candiates
1. wait for onIceCandidate with null json
1. call peerConnectionGetCurrentLocalDescription() - which returns STATUS_SUCCESS with an empty description like below!

```
{"type": "answer", "sdp": ""}
```

enobufs:
In the browser, for the same pattern, I'd used pc.localDescription() (see the link below) instead of currentLocalDescription(). Therefore, I initially thought the problem was the absence of a method equivalent to pc.localDescription().

Reference: [Pending and current descriptions](https://developer.mozilla.org/en-US/docs/Web/API/WebRTC_API/Connectivity#Pending_and_current_descriptions)

chehefen:
please refer to this pr: https://github.com/awslabs/amazon-kinesis-video-streams-webrtc-sdk-c/pull/390

enobufs:
@chehefen did you get a correct SDP with the PR?

chehefen:
yes

enobufs:
@chehefen I just ran the branch for PR #390.
The offer SDP is not completely empty, but it does not include candidates (`a=candidate` lines) at all.

Here's the offer SDP I just printed out:
```
[...]

PR #416, opened 2020-05-01
Introduce peerConnectionGetLocalDescription() with other bug fixes

enobufs:
Issue #368
Description of changes:
* Introduce peerConnectionGetLocalDescription() to return pending local description.
    - peerConnectionGetCurrentLocalDescription() should be for description that has already offered to the peer. There needs to be another method that can return the "pending" description. (which aligns with W3C WebRTC API)
* Fix deadlock when calling peerConnectionGetLocalDescription() from within the onIceCandidate callback.
* Fix incorrect endianness of the candidate port on the SDP.

With the above fixes, I was able to establish a peer connection in non-trickle-ice procedure.

### DIFF
--- a/src/include/com/amazonaws/kinesis/video/webrtcclient/Include.h
+++ b/src/include/com/amazonaws/kinesis/video/webrtcclient/Include.h
@@ -1297,7 +1297,17 @@ PUBLIC_API STATUS peerConnectionOnDataChannel(PRtcPeerConnection, UINT64, RtcOnD
 PUBLIC_API STATUS peerConnectionOnConnectionStateChange(PRtcPeerConnection, UINT64, RtcOnConnectionStateChange);
 
 /**
- * Load the sdp field of PRtcSessionDescriptionInit with latest local session description
+ * Load the sdp field of PRtcSessionDescriptionInit with pending or current local session description
+ *
+ * @param[in] PRtcPeerConnection Initialized RtcPeerConnection
+ * @param[in,out] PRtcSessionDescriptionInit IN/PRtcSessionDescriptionInit whose sdp field will be modified.
+ *
+ * @return STATUS code of the execution. STATUS_SUCCESS on success
+ */
+PUBLIC_API STATUS peerConnectionGetLocalDescription(PRtcPeerConnection, PRtcSessionDescriptionInit);
+
+/**
+ * Load the sdp field of PRtcSessionDescriptionInit with current local session description
  *
  * @param[in] PRtcPeerConnection Initialized RtcPeerConnection
  * @param[in,out] PRtcSessionDescriptionInit IN/PRtcSessionDescriptionInit whose sdp field will be modified.
--- a/src/source/PeerConnection/PeerConnection.c
+++ b/src/source/PeerConnection/PeerConnection.c
@@ -665,6 +665,36 @@ CleanUp:
     return retStatus;
 }
 
+STATUS peerConnectionGetLocalDescription(PRtcPeerConnection pRtcPeerConnection, PRtcSessionDescriptionInit pRtcSessionDescriptionInit)
+{
+    ENTERS();
+    STATUS retStatus = STATUS_SUCCESS;
+    SessionDescription sessionDescription;
+    UINT32 deserializeLen = 0;
+    PKvsPeerConnection pKvsPeerConnection = (PKvsPeerConnection) pRtcPeerConnection;
+
+    CHK(pRtcPeerConnection != NULL && pRtcSessionDescriptionInit != NULL, STATUS_NULL_ARG);
+
+    MEMSET(&sessionDescription, 0x00, SIZEOF(SessionDescription));
+
+    if (pKvsPeerConnection->isOffer) {
+        pRtcSessionDescriptionInit->type = SDP_TYPE_OFFER;
+    } else {
+        pRtcSessionDescriptionInit->type = SDP_TYPE_ANSWER;
+    }
+
+    CHK_STATUS(populateSessionDescription(pKvsPeerConnection, &(pKvsPeerConnection->remoteSessionDescription), &sessionDescription));
+    CHK_STATUS(deserializeSessionDescription(&sessionDescription, NULL, &deserializeLen));
+    CHK(deserializeLen <= MAX_SESSION_DESCRIPTION_INIT_SDP_LEN, STATUS_NOT_ENOUGH_MEMORY);
+
+    CHK_STATUS(deserializeSessionDescription(&sessionDescription, pRtcSessionDescriptionInit->sdp, &deserializeLen));
+
+CleanUp:
+
+    LEAVES();
+    return retStatus;
+}
+
 STATUS peerConnectionGetCurrentLocalDescription(PRtcPeerConnection pRtcPeerConnection, PRtcSessionDescriptionInit pRtcSessionDescriptionInit)
 {
     ENTERS();


src/include/com/amazonaws/kinesis/video/webrtcclient/Include.h
```diff
@@ -1262,7 +1262,17 @@ PUBLIC_API STATUS peerConnectionOnDataChannel(PRtcPeerConnection, UINT64, RtcOnD
 PUBLIC_API STATUS peerConnectionOnConnectionStateChange(PRtcPeerConnection, UINT64, RtcOnConnectionStateChange);
 
 /**
- * Load the sdp field of PRtcSessionDescriptionInit with latest local session description
+ * Load the sdp field of PRtcSessionDescriptionInit with pending or current local session description
+ *
+ * @param[in] PRtcPeerConnection Initialized RtcPeerConnection
+ * @param[in,out] PRtcSessionDescriptionInit IN/PRtcSessionDescriptionInit whose sdp field will be modified.
```
MushMal, line 1268: It's more of an IN/OUT param

src/include/com/amazonaws/kinesis/video/webrtcclient/Include.h
```diff
@@ -1262,7 +1262,17 @@ PUBLIC_API STATUS peerConnectionOnDataChannel(PRtcPeerConnection, UINT64, RtcOnD
...
+ *
+ * @return STATUS code of the execution. STATUS_SUCCESS on success
+ */
+PUBLIC_API STATUS peerConnectionGetLocalDescription(PRtcPeerConnection, PRtcSessionDescriptionInit);
```
MushMal, line 1272: Why are we introducing a new API here instead of modifying the existing and adding an ENUM?

src/source/Ice/IceAgent.c
```diff
@@ -1902,7 +1902,7 @@ STATUS iceCandidateSerialize(PIceCandidate pIceCandidate, PCHAR pOutputData, PUI
                                  pIceCandidate->ipAddress.address[1],
                                  pIceCandidate->ipAddress.address[2],
                                  pIceCandidate->ipAddress.address[3],
-                                 (UINT16) getInt16(pIceCandidate->ipAddress.port),
+                                 (UINT16) SWAP_INT16(pIceCandidate->ipAddress.port),
```
MushMal, line 1905: Nice!

src/source/Ice/IceAgent.c
```diff
@@ -45,7 +45,7 @@ STATUS createIceAgent(PCHAR username, PCHAR password, PIceAgentCallbacks pIceAge
     pIceAgent->localNetworkInterfaceCount = ARRAY_SIZE(pIceAgent->localNetworkInterfaces);
     pIceAgent->candidateGatheringEndTime = INVALID_TIMESTAMP_VALUE;
 
-    pIceAgent->lock = MUTEX_CREATE(FALSE);
+    pIceAgent->lock = MUTEX_CREATE(TRUE);
```
MushMal, line 48: @chehefen could you comment on this change?

src/source/Ice/IceAgent.c
chehefen, line 1905: is this actually an issue for you? getInt16 will trigger SWAP_INT16 on LE machine.

src/source/Ice/IceAgent.c
chehefen: The idea is to not call our apis directly inside our callbacks. Can you set some flag in onIceCandidate and call peerConnectionGetLocalDescription  in your application thread?

src/source/Ice/IceAgent.c
enobufs: Oh.. I see. I assumed SWAP_INT16 is the one equivalent to ntohs().
Weird.. yes it was definitely swapped because the port number reported by this code path was deferent from what onIceCandidate() reported. By doing this, it worked. (SDP matched the port number it actually listening)

```
$ uname -a
Darwin xxxxx.local 19.4.0 Darwin Kernel Version 19.4.0: Wed Mar  4 22:28:40 PST 2020; root:xnu-6153.101.6~15/RELEASE_X86_64 x86_64
```

src/source/Ice/IceAgent.c
enobufs: It this is an open-source library, you should allow API calls from the callbacks, otherwise:
* It forces the application to use another thread, or periodic checks - the unnecessary cost to the app, and I am certainly not interested in going down this route.
* Error-prone, hence the extra cost for you the supporters. (I am sure I wouldn't be the only one attempting to this.)

In another word, a library should avoid forcing a particular design pattern to the users.

So, I'd suggest:
(1) Unlock mutex before callback
(2) Or use recursive/reentrant mutex

I personally think (1) is relatively safer if unlocking and locking again won't cause any atomicity/ordering issues with the function calling the callback.

Any thoughts?

src/source/Ice/IceAgent.c
MushMal: Maybe not... This might not work on LE machine properly... Needs more though

src/source/Ice/IceAgent.c
MushMal: @enobufs I would recommend looking deeper into the interlocking issue with @chehefen 

I agree that forcing the customer to spin up a thread might not be the best solution for multitude of reasons. Re-entrant locking might be OK too. Unlocking/re-locking needs to deal with consistency issues after re-locking. 

I will let @chehefen to look into this

src/source/Ice/IceAgent.c
MushMal: What do we keep in memory? Is this adapted port or the on-the-wire port? I assume the spec is BE right?

src/include/com/amazonaws/kinesis/video/webrtcclient/Include.h
enobufs: As I mentioned in #386 and the beginning of this PR, it was from W3C WebRTC API which has 3 methods to retrieve local descriptions:

* currentLocalDescription() - kvs supported
* pendingLocalDescription()
* localDescription()

When you need to acquire local SDP that is not offered to the remote yet, you will need to use pendingLocalDescription(), or localDescription(), not currentLocalDescription().

There isn't a good document describes what they are but [this one](https://docs.w3cub.com/dom/webrtc_api/connectivity/#Pending_and_current_descriptions) seems to be the best one.

I believe @Sean-Der knows about this much better than I.

src/source/Ice/IceAgent.c
enobufs: Great! thanks @MushMal

src/source/Ice/IceAgent.c
enobufs: Agreed. As I have limited insight into the internals.  Please suggest code changes and I'd happy to update this PR.

src/source/Ice/IceAgent.c
chehefen: I dont want to make the lock re-entrant because there are places where lock is given up early. If the lock is re-entrant, I dont know if the thread actually given up the lock or not. Unlocking/re-locking needs to deal with consistency issues after re-locking. Ideally we should have an event loop and schedule the callback task there. But since we dont have such construct I would just create a thread in ice to do the callback if you have to make the call within the callback.

enobufs: > I would just create a thread ...
Is the cost justifiable? The cost for that comes from users' budget, you know...

> consistency issues after re-locking
Then, you could arrange the code not to require re-locking, couldn't you? Defer all callbacks and call them after unlocking when they are one-way notifications.

chehefen: can you rebase your PR on top of this commit https://github.com/awslabs/amazon-kinesis-video-streams-webrtc-sdk-c/commit/d9cfcdf0165660969e1eb2cbdee6fabda4dcd9c6? I moved the ice candidate reporting off user thread. You dont need to make the iceAgent lock re-entrant.

src/source/Ice/IceAgent.c
chehefen: can you change it back to getInt16? it should evaluate to SWAP_INT16 in LE machine.

src/source/Ice/IceAgent.c
enobufs: @chehefen I have reverted those two (reentrant and swap), then rebase on d9cfcdf as you suggested.

I see SDP generated from the new method, but all of those port numbers reported on the SDP are not what the library is actually listening on.

SDP returned by GetLocalDescription():
```
{"type": "offer", "sdp": "v=0\r\no=- 114807987 2 IN IP4 127.0.0.1\r\ns=-\r\nt=0 0\r\na=group:BUNDLE 0\r\na=msid-semantic: WMS myKvsVideoStream\r\nm=application 9 UDP/DTLS/SCTP webrtc-datachannel\r\nc=IN IP4 127.0.0.1\r\na=candidate:3 1 udp 658881 2601:0646:C202:12D0:0000:0000:0000:1B60 43481 typ host raddr ::/0 rport 0 generation 0 network-cost 999\r\na=candidate:2 1 udp 658881 2601:0646:C202:12D0:DD30:417A:2D01:64CC 43225 typ host raddr ::/0 rport 0 generation 0 network-cost 999\r\na=candidate:1 1 udp 658881 2601:0646:C202:12D0:0845:6473:64BB:42AB 42969 typ host raddr ::/0 rport 0 generation 0 network-cost 999\r\na=candidate:0 1 udp 658881 10.0.0.135 42713 typ host raddr 0.0.0.0 rport 0 generation 0 network-cost 999\r\na=rtcp:9 IN IP4 0.0.0.0\r\na=ice-ufrag:Hlv6\r\na=ice-pwd:WOfwKoKbs09L07z41jI7FhGX\r\na=fingerprint:sha-256 4D:FF:37:B5:0D:6D:D4:5F:76:FF:DD:38:FF:E4:DC:8F:75:65:97:C2:E6:E6:AA:A6:00:85:62:C4:74:AC:45:6B\r\na=setup:actpass\r\na=mid:0\r\na=sctp-port:5000\r\n"}
```

What netstat says during the connectivity checking:
```
$ netstat  -an -p udp
Active Internet connections (including servers)
Proto Recv-Q Send-Q  Local Address          Foreign Address        (state)
udp6       0      0  2601:646:c202:12.55721 *.*
udp6       0      0  2601:646:c202:12.55720 *.*
udp6       0      0  2601:646:c202:12.55719 *.*
udp4       0      0  10.0.0.135.55718       *.*
```

All these port numbers on SDP are all swapped as bellow:
```
43481 = 0xa9d9, 0xd9a9 = 55721
43225 = 0xa8d9, 0xd9a8 = 55720
42969 = 0xa7d9, 0xd9a7 = 55719
42713 = 0xa6d9, 0xd9a6 = 55718
```

enobufs: My log printed out what onIceCandidate callback reported also. Same situation. All port numbers are swapped:
```
onIceCandidate: sess=0x7f9fad80cc00 candidate={"candidate":"candidate:3 1 udp 658881 2601:0646:C202:12D0:0000:0000:0000:1B60 43481 typ host raddr ::/0 rport 0 generation 0 network-cost 999","sdpMid":"0","sdpMLineIndex":0}
2020-05-02 20:53:07 DEBUG   iceAgentLogNewCandidate(): New local ice candidate discovered. Id: Lob/Felt4. Ip: 2601:0646:c202:12d0:dd30:417a:2d01:64cc:43225. Type: host
onIceCandidate: sess=0x7f9fad80cc00 candidate={"candidate":"candidate:2 1 udp 658881 2601:0646:C202:12D0:DD30:417A:2D01:64CC 43225 typ host raddr ::/0 rport 0 generation 0 network-cost 999","sdpMid":"0","sdpMLineIndex":0}
2020-05-02 20:53:07 DEBUG   iceAgentLogNewCandidate(): New local ice candidate discovered. Id: 92j/NEDW8. Ip: 2601:0646:c202:12d0:0845:6473:64bb:42ab:42969. Type: host
onIceCandidate: sess=0x7f9fad80cc00 candidate={"candidate":"candidate:1 1 udp 658881 2601:0646:C202:12D0:0845:6473:64BB:42AB 42969 typ host raddr ::/0 rport 0 generation 0 network-cost 999","sdpMid":"0","sdpMLineIndex":0}
2020-05-02 20:53:07 DEBUG   iceAgentLogNewCandidate(): New local ice candidate discovered. Id: MFfUSNlpg. Ip: 10.0.0.135:42713. Type: host
onIceCandidate: sess=0x7f9fad80cc00 candidate={"candidate":"candidate:0 1 udp 658881 10.0.0.135 42713 typ host raddr 0.0.0.0 rport 0 generation 0 network-cost 999","sdpMid":"0","sdpMLineIndex":0}
```

enobufs: @chehefen @MushMal Okay, I found an error in my code. `initKvsWebRtc()` wasn't called. Consequently, getInt16 was using the default being BE.

With the currently pushed branch 64fd2d, my problems have all been resolved.

src/source/Ice/IceAgent.c
enobufs: Again `initKvsWebRtc()` wasn't called in my code. I have already reverted the change. Let me resolve this conversation.

enobufs: This change has been reverted. The cause was in my code, not initializing the library.

PR #421, opened 2020-05-02
move ice local candidate reporting from user thread to timer thread

chehefen:
*Issue #, if available:*

*Description of changes:*
by moving ice local candidate reporting to timer thread, user can call ice api inside new local candidate callback.


src/source/Ice/IceAgent.h
```diff
@@ -104,6 +105,9 @@ typedef struct {
      * TurnConnectionTracker this candidate is associated to */
     PTurnConnectionTracker pTurnConnectionTracker;
 
+    /* If candidate is local. Indicate whether candidate
+     * has been reported through IceNewLocalCandidateFunc */
+    BOOL reported;
```
MushMal, line 110: I see this is calloc-ed so it's false. Just want to make sure that I checked all of the locations

PR #422, opened 2020-05-02
Enabling instrumented allocators

MushMal:
*Issue #, if available:*

Enabling the instrumented allocators. As it turns out we are doing the right thing here and we don't end up with two global symbols because they are not marked as statics in their TUs.

The issues I was facing were caused by my incorrect interpretation of the readelf -Ws output from Linux with GCC. The duplicate entires were entries in .dynsym and .symtab tables which are OK and as they point to the exact same export.


MushMal:
Run a few of the tests on Mac and Linux and iterated multiple times (50) with all OK

Issue #364, opened 2020-04-20
Missing partial-reliability option for data channel


enobufs:
I have just noticed that this library [ignores RtcDataChannelInit](https://github.com/awslabs/amazon-kinesis-video-streams-webrtc-sdk-c/blob/d69dca161241e62e67b2c68ad274119529309932/src/source/PeerConnection/DataChannel.c#L12) passed to createDataChannel() method. :(

Would initiating a data channel with a partial reliability option become available any time soon?

disa6302:
@enobufs ,

Can you explain what you mean by partial reliability? And what your use case is?

enobufs:
You can learn what WebRTC DataChannel offers re reliability options with these links:

* https://www.w3.org/TR/webrtc/#methods-11
* https://www.html5rocks.com/en/tutorials/webrtc/datachannels/
* https://tools.ietf.org/html/rfc3758

Over a lossy connection, WebRTC data channel may keep retransmitting lost packets, which would cause large latency with a head-of-line blocking.  Partial reliability option would help you reducing such head-of-line blocking (by giving up the retransmissions with certain conditions you specify) to deliver subsequent packets with less delay.

disa6302:
@enobufs ,

Providing the partial reliability option was not in the plan, but I do not see why not! Adding it should not be difficult. I will keep this ticket open to track this feature request.

disa6302:
@enobufs 

The change is in :D
Let us know if you face any issues with it.

zhiyua-git:
Partial-reliability option for data channel is provided in latest master branch. Closing the issue and feel free to open new issue when you try out the function.

enobufs:
@disa6302 @zhiyua-git The changes you made passes the settings to the remote using DCEP, but it does not seem to write up the settings to usrsctp at all.

> Partial-reliability option for data channel is provided in latest master branch.

What are your criteria for concluding that the new feature was provided? May I expect the feature was tested?

Also, there are couple of questions:

## Enable unreliable transfer with maxRetransmits set to 0.
In WebRTC API (JavaScript), if dc.maxRetransmits is set to null, it would do the normal reliable transfer. If it is set to 0, it would do up to 0 retransmission (no retransmission). But your change does not allow the application to configure "no-retransmission" because you take the value 0 to be "reliable" (normal) ops.

## How can the remote node would know what partial-reliability options were used?
[...]

PR #435, opened 2020-05-07
[Issue-364] Add reliability options for data channel

disa6302:
*Issue #, if available: #364 *

*Description of changes:*
- Added support for partial reliability in data channel
- If Retransmit and packet lifetime are set in `RtcDataChannelInit`, retransmit will be considered in the message
- Support for negotiated parameter is not provided at the moment (https://www.w3.org/TR/webrtc/#dictionary-rtcdatachannelinit-members)

Testing:
- Tested locally to see if the parameters get set correctly based on different input combinations.


src/source/PeerConnection/DataChannel.h
```diff
@@ -14,8 +14,8 @@ typedef struct {
     RtcDataChannel dataChannel;
 
     PRtcPeerConnection pRtcPeerConnection;
+    PRtcDataChannelInit pRtcDataChannelInit;
```
MushMal, line 17: Danger! In this case, you will be storing the pointer to an object that was passed in by the customer. This implies that the object has to survive throughout the duration of the lifetime of your object. This might not be the case when it comes to the customers application. We also specifically do not require it.

Suggestion: Instead of a pointer, have a full structure here. Copy the structure in the createDataChannel API and use the copy whenever needed.

src/source/Sctp/Sctp.c
```diff
@@ -213,11 +212,52 @@ STATUS sctpSessionWriteDcep(PSctpSession pSctpSession, UINT32 streamId, PCHAR pC
...
+     // Set channel type and reliability parameters based on input
+     // SCTP allows fine tuning the channel robustness:
+     // 1. Ordering: The data packets can be sent out in an ordered/unordered fashion
+     // 2. Reliability: This determines how the retransmission of packets is handled.
+     // There are 2 parameters that can be fine tuned to achieve this:
+     //     a. Number of retransmits
+     //     b. Packet lifetime
+     // Default values for the parameters is 0. This falls back to reliable channel
+     if(pRtcDataChannelInit != NULL) {
+         switch(pRtcDataChannelInit->ordered) {
```
MushMal, line 231: use if/else with boolean type not switch

src/source/Sctp/Sctp.c
```diff
@@ -213,11 +212,52 @@ STATUS sctpSessionWriteDcep(PSctpSession pSctpSession, UINT32 streamId, PCHAR pC
...
+         case TRUE:
+             if(pRtcDataChannelInit->maxRetransmits > 0) {
```
MushMal, line 233: Optimize the boolean expressions here - when changing from switch to if/else, make sure you don't repeat the same statements in both clauses.

src/source/Sctp/Sctp.h
```diff
@@ -38,9 +38,14 @@ enum {
     DCEP_DATA_CHANNEL_OPEN = 0x03,
 };
 
-enum {
+typedef enum {
     DCEP_DATA_CHANNEL_RELIABLE = 0x00,
```
MushMal, line 42: Force a byte type in enum (not all compilers will respect). DCEP_DATA_CHANNEL_RELIABLE = (BYTE) 0x00, etc...

src/source/PeerConnection/DataChannel.h
disa6302: Oops. You are right. Fixed.

src/source/Sctp/Sctp.c
disa6302: Changed to if/else.

src/source/Sctp/Sctp.c
disa6302: Not sure if I can avoid not repeating the statements because pPacket[1] varies in all 4 cases. The only place I can think of optimizing is in `putUnalignedInt16BigEndian()`, where I can add the line after the if-else clause based on pPacket[1].

src/source/Sctp/Sctp.h
disa6302: Done.

src/source/PeerConnection/DataChannel.h
```diff
@@ -14,8 +14,8 @@ typedef struct {
     RtcDataChannel dataChannel;
 
     PRtcPeerConnection pRtcPeerConnection;
+    RtcDataChannelInit RtcDataChannelInit;
```
MushMal, line 17: camelCase it = rtcDataChannelInit

src/source/PeerConnection/DataChannel.c
```diff
@@ -24,7 +23,16 @@ STATUS createDataChannel(PRtcPeerConnection pPeerConnection, PCHAR pDataChannelN
...
+    if(pRtcDataChannelInit != NULL) {
+        // Setting negotiated to false. Not supporting at the moment
+        pRtcDataChannelInit->negotiated = FALSE;
+        pKvsDataChannel->RtcDataChannelInit = *pRtcDataChannelInit;
+    }
+    else {
+        // If nothing is set, set default to ordered mode
+        pKvsDataChannel->RtcDataChannelInit.ordered = TRUE;
+    }
+//    DLOGD("%d\n", pKvsDataChannel->RtcDataChannelInit);
```
MushMal, line 35: remove the comment

PR #437, opened 2020-05-07
in connection listener, when removing socket connection, check if the…

chehefen:
… connection has already been freed or not. This can happen because up stream can have a reference of PSocketConnection thats already freed.

*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/src/source/Ice/ConnectionListener.c
+++ b/src/source/Ice/ConnectionListener.c
@@ -6,6 +6,9 @@
 
 STATUS createConnectionListener(PConnectionListener* ppConnectionListener)
 {
+    const UINT32 hashTableBucketCount = 50;
+    const UINT32 hashTableBucketLen = 2;
+
     STATUS retStatus = STATUS_SUCCESS;
     UINT32 allocationSize = SIZEOF(ConnectionListener) + MAX_UDP_PACKET_SIZE;
     PConnectionListener pConnectionListener = NULL;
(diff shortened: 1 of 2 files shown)


src/source/Ice/ConnectionListener.c
```diff
@@ -116,17 +126,28 @@ STATUS connectionListenerAddConnection(PConnectionListener pConnectionListener,
...
     STATUS retStatus = STATUS_SUCCESS;
+    BOOL locked = FALSE, socketConnectionActive = FALSE;
 
     CHK(pConnectionListener != NULL && pSocketConnection != NULL, STATUS_NULL_ARG);
     CHK(!ATOMIC_LOAD_BOOL(&pConnectionListener->terminate), retStatus);
 
+    MUTEX_LOCK(pConnectionListener->lock);
+    locked = TRUE;
+
+    CHK_STATUS(hashTableContains(pConnectionListener->activeSocketConnections, ((UINT64) pSocketConnection), &socketConnectionActive));
```
lherman-cs, line 137: I'm not sure if I missed something, but why can't we just check if the connection is NULL? Since we already nullified it in `freeSocketConnection`, https://github.com/awslabs/amazon-kinesis-video-streams-webrtc-sdk-c/blob/626aae97e93cf05d34b62a4ca44f22435d836ac1/src/source/Ice/SocketConnection.c#L91.

chehefen: yh. Please take a look at this PR: https://github.com/awslabs/amazon-kinesis-video-streams-webrtc-sdk-c/pull/441. I will cancel this one.

PR #441, opened 2020-05-08
make sure connectionListener is not in the middle of calling dataAvai…

chehefen:
…lable callback for the socketConnection once connectionListenerRemoveConnection returns

*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/src/source/Ice/ConnectionListener.c
+++ b/src/source/Ice/ConnectionListener.c
@@ -46,8 +46,6 @@ STATUS freeConnectionListener(PConnectionListener* ppConnectionListener)
 {
     STATUS retStatus = STATUS_SUCCESS;
     PConnectionListener pConnectionListener = NULL;
-    PDoubleListNode pCurNode = NULL;
-    PSocketConnection pSocketConnection = NULL;
 
     CHK(ppConnectionListener != NULL, STATUS_NULL_ARG);
     CHK(*ppConnectionListener != NULL, retStatus);
@@ -66,12 +64,6 @@ STATUS freeConnectionListener(PConnectionListener* ppConnectionListener)
     }
 
     if (pConnectionListener->connectionList != NULL) {
-        CHK_LOG_ERR(doubleListGetHeadNode(pConnectionListener->connectionList, &pCurNode));
-        while(pCurNode != NULL) {
-            pSocketConnection = (PSocketConnection) pCurNode->data;
-            pCurNode = pCurNode->pNext;
-            CHK_STATUS(freeSocketConnection(&pSocketConnection));
-        }
         CHK_LOG_ERR(doubleListClear(pConnectionListener->connectionList, FALSE));
         CHK_LOG_ERR(doubleListFree(pConnectionListener->connectionList));
     }
@@ -124,9 +116,8 @@ CleanUp:
 
 STATUS connectionListenerRemoveConnection(PConnectionListener pConnectionListener, PSocketConnection pSocketConnection)
 {
-    STATUS retStatus = STATUS_SUCCESS;
+    STATUS retStatus = STATUS_SUCCESS, cvarWaitStatus = STATUS_SUCCESS;
     BOOL locked = FALSE;
-    const UINT64 longTimeout = 1 * HUNDREDS_OF_NANOS_IN_AN_HOUR;
 
     CHK(pConnectionListener != NULL && pSocketConnection != NULL, STATUS_NULL_ARG);
     CHK(!ATOMIC_LOAD_BOOL(&pConnectionListener->terminate), retStatus);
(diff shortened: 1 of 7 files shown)


src/source/Ice/ConnectionListener.c
```diff
@@ -116,17 +117,34 @@ STATUS connectionListenerAddConnection(PConnectionListener pConnectionListener,
 STATUS connectionListenerRemoveConnection(PConnectionListener pConnectionListener, PSocketConnection pSocketConnection)
 {
     STATUS retStatus = STATUS_SUCCESS;
+    BOOL locked = FALSE;
+    const UINT64 longTimeout = 1 * HUNDREDS_OF_NANOS_IN_AN_HOUR;
```
MushMal, line 121: Use #define in the header. Const is a C++ keyword

src/source/Ice/ConnectionListener.c
```diff
@@ -116,17 +117,34 @@ STATUS connectionListenerAddConnection(PConnectionListener pConnectionListener,
...
 
+    MUTEX_LOCK(pConnectionListener->lock);
+    locked = TRUE;
+    /* make sure connectionListenerRemoveConnection return after connectionListenerReceiveDataRoutine has picked up
+     * the change. */
+    while(ATOMIC_LOAD_BOOL(&pConnectionListener->listenerRoutineStarted) &&
+          !ATOMIC_LOAD_BOOL(&pConnectionListener->terminate) &&
+          ATOMIC_LOAD_BOOL(&pConnectionListener->connectionListChanged)) {
+        /* use longTimeout because connectionListenerReceiveDataRoutine should wake this up soon enough */
+        CVAR_WAIT(pConnectionListener->removeConnectionComplete, pConnectionListener->lock, longTimeout);
```
MushMal, line 139: I don't understand why 1 hour? Something more sensible perhaps with a handling of a timeout error?

src/source/Ice/ConnectionListener.c
```diff
@@ -136,6 +154,7 @@ STATUS connectionListenerRemoveAllConnection(PConnectionListener pConnectionList
     BOOL locked = FALSE;
     PDoubleListNode pCurNode = NULL;
     PSocketConnection pSocketConnection = NULL;
+    const UINT64 longTimeout = 1 * HUNDREDS_OF_NANOS_IN_AN_HOUR;
```
MushMal, line 157: same - #define it

MushMal: Also, why is this function not calling connectionListenerRemoveConnection underneath it when iterating?

src/source/Ice/SocketConnection.c
```diff
@@ -251,9 +252,10 @@ STATUS socketConnectionReadData(PSocketConnection pSocketConnection, PBYTE pBuf,
         // Unlikely that we will get sslReadRet == 0 here because socketConnectionReadData is only called when
         // socket recevies data. If ssl handshake is not done then -1 is returned.
         if (sslReadRet <= 0) {
-            sslErrorRet = SSL_get_error(pSocketConnection->pSsl, sslReadRet);
-            if (sslErrorRet != SSL_ERROR_WANT_WRITE && sslErrorRet != SSL_ERROR_WANT_READ) {
-                DLOGV("SSL_read failed with %s. Length of data already written %u", ERR_error_string(sslErrorRet, NULL), writtenBytes);
+            while ((sslErrorRet = ERR_get_error()) != 0) {
```
MushMal, line 255: Perhaps deserves a comment, not sure what this returns

src/source/Ice/ConnectionListener.c
chehefen: the point is it should never timeout. the connectionListenerReceiveDataRoutine should wake it up in time.

src/source/Ice/ConnectionListener.c
chehefen: then we have to wait for each connection to be closed before closing the next one.

src/source/Ice/SocketConnection.c
chehefen: This is just the way openssl error is printed.

src/source/Ice/ConnectionListener.c
MushMal: Perhaps, but if it does? What then? The worse could happen if the code "hangs". Perhaps you need to set something small and error out on a timeout.

src/source/Ice/ConnectionListener.c
MushMal: Got it

src/source/Ice/SocketConnection.c
MushMal: This really does look fishy. I don't see anything in the description requiring to iterate: https://www.openssl.org/docs/manmaster/man3/ERR_get_error.html

chehefen: the way it mentioned in the doc doesnt give the proper error.

src/source/Ice/ConnectionListener.c
chehefen: I kind of prefer it hangs so we can get to it right away instead of chasing red herring.

chehefen: reduced to 5 seconds

MushMal: au contraire, on a production device, chasing hangs is impossible. I think you are doing the right thing in reducing and erring out.

src/source/Ice/ConnectionListener.c
```diff
@@ -115,24 +116,46 @@ STATUS connectionListenerAddConnection(PConnectionListener pConnectionListener,
 
 STATUS connectionListenerRemoveConnection(PConnectionListener pConnectionListener, PSocketConnection pSocketConnection)
 {
-    STATUS retStatus = STATUS_SUCCESS;
+    STATUS retStatus = STATUS_SUCCESS, cvarWaitStatus = STATUS_SUCCESS;;
```
MushMal, line 119: double ;

PR #444, opened 2020-05-11
Add PRtcDataChannel to datachannel callback arguments

### LINKED ISSUE #443: Insufficient arguments for callbacks from data channel
## onOpen
```c
typedef VOID (*RtcOnOpen)(UINT64);
PUBLIC_API STATUS dataChannelOnOpen(PRtcDataChannel, UINT64, RtcOnOpen);
```

## onMessage
```c
typedef VOID (*RtcOnMessage)(UINT64, BOOL, PBYTE, UINT32);
PUBLIC_API STATUS dataChannelOnMessage(PRtcDataChannel, UINT64, RtcOnMessage);
```

## Issue
Both RtcOnOpen and RtcOnMessage have only one argument for a custom data. We could pass a pointer to a context of the peer connection, but the callee won't be able to tell which data channel it was called on.

If you pass PRtcDataChannel to the custom data of type UINT64, then the callee wouldn't know which peer connection the data channel belongs to.

I thought about passing a pointer to an object which holds two pointers, but the lifetime of the object can not be managed correctly particularly for onMessage.

## Suggestion
Pass a pointer to the datachannel also:
```c
typedef VOID (*RtcOnOpen)(UINT64, PRtcDataChannel);
typedef VOID (*RtcOnMessage)(UINT64, PRtcDataChannel, BOOL, PBYTE, UINT32);
```

enobufs:
Resolves #443

### DIFF
--- a/samples/Common.c
+++ b/samples/Common.c
@@ -12,9 +12,10 @@ VOID sigintHandler(INT32 sigNum)
     }
 }
 
-VOID onDataChannelMessage(UINT64 customData, BOOL isBinary, PBYTE pMessage, UINT32 pMessageLen)
+VOID onDataChannelMessage(UINT64 customData, PRtcDataChannel pDataChannel, BOOL isBinary, PBYTE pMessage, UINT32 pMessageLen)
 {
     UNUSED_PARAM(customData);
+    UNUSED_PARAM(pDataChannel);
     if (isBinary) {
         DLOGI("DataChannel Binary Message");
     } else {
--- a/src/include/com/amazonaws/kinesis/video/webrtcclient/Include.h
+++ b/src/include/com/amazonaws/kinesis/video/webrtcclient/Include.h
@@ -646,14 +646,15 @@ typedef VOID (*RtcOnPictureLoss)(UINT64);
  *
  * Reference: https://www.w3.org/TR/webrtc/#dom-rtcdatachannel-onmessage
  */
-typedef VOID (*RtcOnMessage)(UINT64, BOOL, PBYTE, UINT32);
+struct __RtcDataChannel;
+typedef VOID (*RtcOnMessage)(UINT64, struct __RtcDataChannel*, BOOL, PBYTE, UINT32);
 
 /**
  * RtcOnOpen is fired when the DataChannel has opened
  *
  * Reference: https://www.w3.org/TR/webrtc/#dom-rtcdatachannel-onopen
  */
-typedef VOID (*RtcOnOpen)(UINT64);
+typedef VOID (*RtcOnOpen)(UINT64, struct __RtcDataChannel*);
 
 /**
  * @brief RtcOnDataChannel is fired when the remote PeerConnection
@@ -661,7 +662,6 @@ typedef VOID (*RtcOnOpen)(UINT64);
  *
  * Reference: https://www.w3.org/TR/webrtc/#dom-rtcpeerconnection-ondatachannel
  */
-struct __RtcDataChannel;
 typedef VOID (*RtcOnDataChannel)(UINT64, struct __RtcDataChannel*);
 
 /**
--- a/src/include/com/amazonaws/kinesis/video/webrtcclient/Include.h
+++ b/src/include/com/amazonaws/kinesis/video/webrtcclient/Include.h
@@ -642,19 +642,27 @@ typedef VOID (*RtcOnBandwidthEstimation)(UINT64, DOUBLE);
 typedef VOID (*RtcOnPictureLoss)(UINT64);
 
 /**
+ * @brief RtcDataChannel represents a bi-directional data channel between two peers.
+ *
+ * Reference: https://www.w3.org/TR/webrtc/#dom-rtcdatachannel
+ */
+typedef struct __RtcDataChannel {
+    CHAR name[MAX_DATA_CHANNEL_NAME_LEN + 1]; //!< Define name of data channel. Max length is 256 characters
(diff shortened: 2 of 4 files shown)


samples/Common.c
```diff
@@ -12,9 +12,10 @@ VOID sigintHandler(INT32 sigNum)
     }
 }
 
-VOID onDataChannelMessage(UINT64 customData, BOOL isBinary, PBYTE pMessage, UINT32 pMessageLen)
+VOID onDataChannelMessage(UINT64 customData, PRtcDataChannel dc, BOOL isBinary, PBYTE pMessage, UINT32 pMessageLen)
```
MushMal, line 15: Though this is a sample, we still try to keep the coding standards in part with the main SDK codebase. Recommend changing dc -> pDataChannel

src/include/com/amazonaws/kinesis/video/webrtcclient/Include.h
```diff
@@ -644,22 +644,22 @@ typedef VOID (*RtcOnPictureLoss)(UINT64);
  *
  * Reference: https://www.w3.org/TR/webrtc/#dom-rtcdatachannel-onmessage
  */
-typedef VOID (*RtcOnMessage)(UINT64, BOOL, PBYTE, UINT32);
+struct __RtcDataChannel;
```
MushMal, line 647: Do we really need this forward declaration? I see we have this elsewhere in the header  LN 664 but I believe this should not be necessary as the declaration of the structs fkr RtcDataChannel and RtcDataChannelInit should be moved before the function declarations

src/include/com/amazonaws/kinesis/video/webrtcclient/Include.h
```diff
@@ -644,22 +644,22 @@ typedef VOID (*RtcOnPictureLoss)(UINT64);
...
+typedef VOID (*RtcOnMessage)(UINT64, struct __RtcDataChannel*, BOOL, PBYTE, UINT32);
 
 /**
  * RtcOnOpen is fired when the DataChannel has opened
  *
  * Reference: https://www.w3.org/TR/webrtc/#dom-rtcdatachannel-onopen
  */
-typedef VOID (*RtcOnOpen)(UINT64);
+typedef VOID (*RtcOnOpen)(UINT64, struct __RtcDataChannel*);
```
MushMal, line 655: My main concern here is the backward compatibility. If we implement this change then we need to rev the major version. We could have a separate callback with a different name that would ensure the compile-time compatibility but it would really be ugly but I just don't see any other solution. 

Another proposal is to redefine the minor version update to indicate that in case of the minor version update there might be some compile time compatibility issues that can be trivial to solve in the application code. I would vote for the second. The reason is that this is still a very active development codebase on a dynamic technology standard that morphs continuously and vamping a major version with small incremental implementation/changes like these makes no sense.

Please discuss/debage.

enobufs: Yeah, I hear you. It's a breaking change. As I could not come up with another workaround, I will stick with my fork with the hope that you would role out major version update.

If I were you, I wouldn't be so concerned about advancing the major version rapidly. I'd rather communicate what you are releasing to users by strictly following semantic versioning. (2c)

MushMal: I agree, revving the major version is not needed in this case. Let us make  few minor changes per the comments and we will push another pr as co-authors

samples/Common.c
enobufs: Updated in dc7af43 as suggested.

src/include/com/amazonaws/kinesis/video/webrtcclient/Include.h
enobufs: It was there and I don't know exactly what to do. Can you suggest a change?

MushMal: Oh sure, __RtcDataChannel is declared below the function calls that uses it, hence the need for the forward declaration. You could move the declaration above the callback declaration.

This is really a minor change so you can chose to ignore it - no worries

src/include/com/amazonaws/kinesis/video/webrtcclient/Include.h
enobufs: That's not what I meant tho. I was saying, "don't be afraid to advance the major version number, and you should increase the version number when there's a breaking change so that users will know what he/she can take appropriate action".  But, It's your call.

MushMal: The issue is that this codebase is still undergoing active development. For example, we will be adding many new structures to the code as we progress and perhaps changing some minor callbacks. Revving up a major version is just going to be more confusing. I think for these types of changes my vote would be to rev the minor version (not the revision though). At any rate, let's get these changes in first. When cutting the release we can properly word the release notes with this.

MushMal:
Seems there are build breaks: https://github.com/awslabs/amazon-kinesis-video-streams-webrtc-sdk-c/pull/444/checks?check_run_id=726848715

Let's get this fixed ASAP and push this change in

src/include/com/amazonaws/kinesis/video/webrtcclient/Include.h
enobufs: > You could move the declaration above the callback declaration.
I did in 1666077

MushMal:
Who hoo!!!!

PR #448, opened 2020-05-13
WebRTC Stats framework

disa6302:
*Issue #, if available:*

*Description of changes:*
  
- Initial draft for WebRTC stats
- Addition of a new header file (Stats.h) exclusively for Stats to keep them separated. 
- Grouped the stats together based on the part it focusses on in the WebRTC Stack
- Added all members of WebRTC Stats, which can be removed in the future based on use case

To do:
- Add the guts of the metrics in the code. WIP meanwhile

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -197,7 +197,8 @@ file(
   "src/source/Sdp/*.c"
   "src/source/Srtp/*.c"
   "src/source/Stun/*.c"
-  "src/source/Sctp/*.c")
+  "src/source/Sctp/*.c"
+  "src/source/WebRtcMetrics/*.c")
 
 file(GLOB WEBRTC_SIGNALING_CLIENT_SOURCE_FILES "src/source/Signaling/*.c")
 
--- a/samples/Common.c
+++ b/samples/Common.c
@@ -208,7 +208,7 @@ STATUS handleOffer(PSampleConfiguration pSampleConfiguration, PSampleStreamingSe
     RtcSessionDescriptionInit offerSessionDescriptionInit;
     BOOL locked = FALSE;
     NullableBool canTrickle;
-
+//    RTCStats rtcMetrics;
     CHK(pSampleConfiguration != NULL && pSignalingMessage != NULL, STATUS_NULL_ARG);
 
     MEMSET(&offerSessionDescriptionInit, 0x00, SIZEOF(RtcSessionDescriptionInit));
@@ -264,7 +264,9 @@ STATUS handleOffer(PSampleConfiguration pSampleConfiguration, PSampleStreamingSe
         THREAD_CREATE(&pSampleStreamingSession->receiveAudioVideoSenderTid, pSampleConfiguration->receiveAudioVideoSource,
                       (PVOID) pSampleStreamingSession);
     }
-
+//    rtcMetrics.requestedTypeOfStats = RTC_STATS_TYPE_CANDIDATE_PAIR;
+//    getRtcPeerConnectionStats(pSampleStreamingSession->pPeerConnection, &rtcMetrics);
+//    printf("Requests Received:%lf\n", rtcMetrics.rtcStatsObject.iceCandidatePairStats.currentRoundTripTime);
 CleanUp:
 
     CHK_LOG_ERR(retStatus);
--- a/src/include/com/amazonaws/kinesis/video/webrtcclient/Include.h
+++ b/src/include/com/amazonaws/kinesis/video/webrtcclient/Include.h
@@ -18,6 +18,7 @@ extern "C" {
 #include <com/amazonaws/kinesis/video/client/Include.h>
 #include <com/amazonaws/kinesis/video/common/Include.h>
 #include <com/amazonaws/kinesis/video/webrtcclient/NullableDefs.h>
+#include <com/amazonaws/kinesis/video/webrtcclient/Stats.h>
 
 #pragma clang diagnostic pop
 
@@ -475,6 +476,22 @@ extern "C" {
  * Version of SignalingMessage structure
  */
 #define SIGNALING_MESSAGE_CURRENT_VERSION                                           0
+
+/**
+ * Version of RtcIceMetrics structure
+ */
+#define RTC_ICE_METRICS_CURRENT_VERSION                                             0
+
+/**
+ * Version of RtcStreamMetrics structure
+ */
+#define RTC_STREAM_METRICS_CURRENT_VERSION                                          0
+
+/**
+ * Version of SignalingClientMetrics structure
+ */
+#define SIGNALING_CLIENT_METRICS_CURRENT_VERSION                                    0
+
 /*!@} */
 
 /*===========================================================================================*/
@@ -1242,6 +1259,55 @@ typedef struct {
                      //!< RTCDataChannel object with the same id as the other peer.
 } RtcDataChannelInit, *PRtcDataChannelInit;
 
+
+/////////////////////////////////////////////////////
+/// Metrics/Stats Related structures
+/////////////////////////////////////////////////////
+
+/**
+ * @brief Collection of ICE related stats
+ * Reference: https://www.w3.org/TR/webrtc-stats/#ice-server-dict*
+ * Reference: https://www.w3.org/TR/webrtc-stats/#icecandidate-dict*
+ * Reference: https://www.w3.org/TR/webrtc-stats/#candidatepair-dict*
+ */
+typedef struct {
+    UINT32 version; //!< Structure version
+    PRtcIceServerStats pRTCIceServerStats; //!< Server related stats. Reference in Stats.h
+    PRtcIceCandidateStats pRTCIceCandidateStats; //!< Single candidate stats. Reference in Stats.h
+    PRtcIceCandidatePairStats pRTCIceCandidatePairStats; //!< Candidate pair stats. Reference in Stats.h
+} RtcIceMetrics, *PRtcIceMetrics;
+
+/**
+ * @brief Collection of RTP stream related stats
+ * Reference: https://www.w3.org/TR/webrtc-stats/#remoteinboundrtpstats-dict*
+ * Reference: https://www.w3.org/TR/webrtc-stats/#outboundrtpstats-dict*
+ * Reference: https://www.w3.org/TR/webrtc-stats/#transportstats-dict*
+ */
+typedef struct {
+    UINT32 version; //!< Structure version
+    PRtcRemoteInboundRtpStreamStats pRTCInboundStats; //!< Inbound RTP Stats. Reference in Stats.h
+    PRtcOutboundRtpStreamStats pRTCOutboundStats; //!< Outbound RTP Stats. Reference in Stats.h
+    PRtcTransportStats pRTCTransportStats; //!< Transport stats. Reference in Stats.h
+} RtcStreamMetrics, *PRtcStreamMetrics;
+
+/**
+ * @brief SignalingStats Collection of signaling related stats. Can be expanded in the future
+ */
+typedef struct {
+    UINT32 version; //!< Structure version
+    PSignalingClientStats pSignalingClientMetrics; //!< Signaling client metrics stats. Reference in Stats.h
+} SignalingClientMetrics, *PSignalingClientMetrics;
+
+/**
+ * @brief The stats object is populated based on RTCStatsType request
+ *
+ */
+typedef struct {
+    UINT64 timestamp; //!< Timestamp of request for stats
+    RTC_STATS_TYPE requestedTypeOfStats; //!< Type of stats requested. Set to RTC_ALL to get all supported stats
+    RtcStatsObject rtcStatsObject; //!< Object that is populated by the SDK on request
+} RTCStats, *PRTCStats;
+
 ////////////////////////////////////////////////////
 // Public functions
 ////////////////////////////////////////////////////
@@ -1679,7 +1745,7 @@ PUBLIC_API STATUS signalingClientConnectSync(SIGNALING_CLIENT_HANDLE);
  */
 PUBLIC_API STATUS signalingClientDisconnectSync(SIGNALING_CLIENT_HANDLE);
 
-/*
+/**
  * @brief Gets the Signaling client current state.
  *
  * @param[in] SIGNALING_CLIENT_HANDLE Signaling client handle
@@ -1689,7 +1755,7 @@ PUBLIC_API STATUS signalingClientDisconnectSync(SIGNALING_CLIENT_HANDLE);
  */
 PUBLIC_API STATUS signalingClientGetCurrentState(SIGNALING_CLIENT_HANDLE, PSIGNALING_CLIENT_STATE);
 
-/*
+/**
  * Gets a literal string representing a Signaling client state.
  *
  * @param[in] SIGNALING_CLIENT_HANDLE Signaling client handle
@@ -1716,6 +1782,23 @@ PUBLIC_API STATUS signalingClientGetStateString(SIGNALING_CLIENT_STATE, PCHAR*);
  * @return STATUS code of the execution. STATUS_SUCCESS on success
  */
 PUBLIC_API STATUS signalingClientDeleteSync(SIGNALING_CLIENT_HANDLE);
+
+/**
+ * @brief Get the relevant/all metrics based on the RTCStatsType field. This does not include
+ * any signaling related metrics
+ *
+ * @param PRtcPeerConnection Peer connection for which the stats need to be collected
+ * @param[in/out] PRtcStats The stats object with the RTCStatsType field populated
+ */
+PUBLIC_API STATUS getRtcPeerConnectionStats(PRtcPeerConnection, PRTCStats);
+
+/**
+ * @brief Get signaling related metrics
+ *
+ * @param[in] SIGNALING_CLIENT_HANDLE Signaling client handle
+ * @param[in/out] PSignalingClientMetrics Signaling stats
+ */
+PUBLIC_API STATUS getSignalingClientStats(SIGNALING_CLIENT_HANDLE, PSignalingClientMetrics);
 
 #ifdef  __cplusplus
 }
--- a/src/include/com/amazonaws/kinesis/video/webrtcclient/Include.h
+++ b/src/include/com/amazonaws/kinesis/video/webrtcclient/Include.h
@@ -19,7 +19,6 @@ extern "C" {
 #include <com/amazonaws/kinesis/video/common/Include.h>
 #include <com/amazonaws/kinesis/video/webrtcclient/NullableDefs.h>
 #include <com/amazonaws/kinesis/video/webrtcclient/Stats.h>
-
 #pragma clang diagnostic pop
 
 /*===========================================================================================*/
@@ -1272,9 +1271,9 @@ typedef struct {
  */
 typedef struct {
     UINT32 version; //!< Structure version
-    PRtcIceServerStats pRTCIceServerStats; //!< Server related stats. Reference in Stats.h
-    PRtcIceCandidateStats pRTCIceCandidateStats; //!< Single candidate stats. Reference in Stats.h
-    PRtcIceCandidatePairStats pRTCIceCandidatePairStats; //!< Candidate pair stats. Reference in Stats.h
+    RtcIceServerStats rtcIceServerStats; //!< Server related stats. Reference in Stats.h
+    RtcIceCandidateStats rtcIceCandidateStats; //!< Single candidate stats. Reference in Stats.h
+    RtcIceCandidatePairStats rtcIceCandidatePairStats; //!< Candidate pair stats. Reference in Stats.h
 } RtcIceMetrics, *PRtcIceMetrics;
 
 /**
@@ -1285,9 +1284,9 @@ typedef struct {
  */
 typedef struct {
     UINT32 version; //!< Structure version
-    PRtcRemoteInboundRtpStreamStats pRTCInboundStats; //!< Inbound RTP Stats. Reference in Stats.h
-    PRtcOutboundRtpStreamStats pRTCOutboundStats; //!< Outbound RTP Stats. Reference in Stats.h
-    PRtcTransportStats pRTCTransportStats; //!< Transport stats. Reference in Stats.h
+    RtcRemoteInboundRtpStreamStats rtcInboundStats; //!< Inbound RTP Stats. Reference in Stats.h
+    RtcOutboundRtpStreamStats rtcOutboundStats; //!< Outbound RTP Stats. Reference in Stats.h
+    RtcTransportStats rtcTransportStats; //!< Transport stats. Reference in Stats.h
 } RtcStreamMetrics, *PRtcStreamMetrics;
 
 /**
@@ -1295,7 +1294,7 @@ typedef struct {
  */
 typedef struct {
     UINT32 version; //!< Structure version
-    PSignalingClientStats pSignalingClientMetrics; //!< Signaling client metrics stats. Reference in Stats.h
+    SignalingClientStats signalingClientMetrics; //!< Signaling client metrics stats. Reference in Stats.h
 } SignalingClientMetrics, *PSignalingClientMetrics;
 
 /**
--- a/src/include/com/amazonaws/kinesis/video/webrtcclient/Stats.h
+++ b/src/include/com/amazonaws/kinesis/video/webrtcclient/Stats.h
@@ -59,12 +59,12 @@ typedef enum {
  * Reference: https://www.w3.org/TR/webrtc-stats/#rtcstatsicecandidatepairstate-enum
  */
 typedef enum {
-    RTC_STATS_ICE_CANDIDATE_PAIR_STATE_FROZEN,    //!< A check for the pair hasn't been performed because it is waiting for other checks to complete
-    RTC_STATS_ICE_CANDIDATE_PAIR_STATE_WAITING,   //!< A check has not been performed for this pair
-    RTC_STATS_ICE_CANDIDATE_PAIR_STATE_IN_PROGRESS,//!< A check has been sent for this pair
-    RTC_STATS_ICE_CANDIDATE_PAIR_STATE_FAILED,    //!< A check for the pair was already done and failed
-    RTC_STATS_ICE_CANDIDATE_PAIR_STATE_SUCCEEDED  //!< A check for the pair was already done and produced a successful result
-} RTC_STATS_ICE_CANDIDATE_PAIR_STATE;
+    ICE_CANDIDATE_PAIR_STATE_FROZEN         = 0,
+    ICE_CANDIDATE_PAIR_STATE_WAITING        = 1,
+    ICE_CANDIDATE_PAIR_STATE_IN_PROGRESS    = 2,
+    ICE_CANDIDATE_PAIR_STATE_SUCCEEDED      = 3,
+    ICE_CANDIDATE_PAIR_STATE_FAILED         = 4,
+} ICE_CANDIDATE_PAIR_STATE;
 
 /**
  * @brief Set details of the IceAgent based on STUN_ATTRIBUTE_TYPE_USE_CANDIDATE flag
@@ -81,12 +81,12 @@ typedef enum {
  * Reference: https://www.w3.org/TR/webrtc/#rtcdtlstransport-interface
  */
 typedef enum {
-    RTC_DTLS_TRANSPORT_STATE_NEW,       //!< DTLS has not started negotiating yet
-    RTC_DTLS_TRANSPORT_STATE_CONNECTING,//!< DTLS is in the process of negotiating a secure connection
-    RTC_DTLS_TRANSPORT_STATE_CONNECTED, //!< DTLS has completed negotiation of a secure connection
-    RTC_DTLS_TRANSPORT_STATE_CLOSED,    //!< The transport has been closed intentionally
-    RTC_DTLS_TRANSPORT_STATE_FAILED     //!< The transport has failed as the result of an error
-} RTC_DTLS_TRANSPORT_STATE;
+    RTC_DTLS_TRANSPORT_STATE_STATS_NEW,       //!< DTLS has not started negotiating yet
+    RTC_DTLS_TRANSPORT_STATE_STATS_CONNECTING,//!< DTLS is in the process of negotiating a secure connection
+    RTC_DTLS_TRANSPORT_STATE_STATS_CONNECTED, //!< DTLS has completed negotiation of a secure connection
+    RTC_DTLS_TRANSPORT_STATE_STATS_CLOSED,    //!< The transport has been closed intentionally
+    RTC_DTLS_TRANSPORT_STATE_STATS_FAILED     //!< The transport has failed as the result of an error
+} RTC_DTLS_TRANSPORT_STATE_STATS;
 
 /**
  * @brief Defines reasons for quality limitation for sending streams
@@ -118,15 +118,6 @@ typedef struct {
 } DscpPacketsSentRecord, *PDscpPacketsSentRecord;
 
 /**
- * @brief Since C does not have an explicit undefined value like JS does, this structure uses a bool flag to
- * determine if the value has a definite value. By default, the value is undefined (isDefined is False), with val = 0
- */
-typedef struct {
-    BOOL isDefined; //!< If set to TRUE, the value is defined, else value is treated to be undefined
-    UINT32 val; //!< Value of the item. A value of 0 could be treated as defined/undefined based on the bool flag
-} DefineableUint32;
-
-/**
  * @brief RtcIceCandidatePairStats Stats related to the local-remote ICE candidate pair
  *
  * Reference: https://www.w3.org/TR/webrtc-stats/#candidatepair-dict*
@@ -136,10 +127,10 @@ typedef struct {
     CHAR transportId[MAX_STRING_LENGTH + 1]; //!< ID of object that was inspected for RTCTransportStats
     CHAR localCandidateId[MAX_CANDIDATE_ID_LENGTH + 1]; //!< Local candidate that is inspected in RTCIceCandidateStats
     CHAR remoteCandidateId[MAX_CANDIDATE_ID_LENGTH + 1]; //!< Remote candidate that is inspected in RTCIceCandidateStats
-    RTC_STATS_ICE_CANDIDATE_PAIR_STATE state; //!< State of checklist for the local-remote candidate pair
+    ICE_CANDIDATE_PAIR_STATE state; //!< State of checklist for the local-remote candidate pair
     BOOL nominated; //!< Flag is TRUE if the agent is a controlling agent and FALSE otherwise. The agent role is based on the
                     //!< STUN_ATTRIBUTE_TYPE_USE_CANDIDATE flag
-    DefineableUint32 circuitBreakerTriggerCount; //!< Represents number of times circuit breaker is triggered during media transmission
+    NullableUint32 circuitBreakerTriggerCount; //!< Represents number of times circuit breaker is triggered during media transmission
                                        //!< It is undefined if the user agent does not use this
     UINT32 packetsDiscardedOnSend; //!< Total number of packets discarded for candidate pair due to socket errors,
     UINT64 packetsSent; //!< Total number of packets sent on this candidate pair;
@@ -155,14 +146,14 @@ typedef struct {
                                  //!< The average interval between two consecutive connectivity checks sent can be calculated:
                                  //! (lastRequestTimestamp - firstRequestTimestamp) / requestsSent.
     UINT64 lastResponseTimestamp; //!< Represents the timestamp at which the last STUN response was received on this particular candidate pair.
-    UINT64 totalRoundTripTime; //!< The sum of all round trip time (seconds) since the beginning of the session, based
+    DOUBLE totalRoundTripTime; //!< The sum of all round trip time (seconds) since the beginning of the session, based
                                //!< on STUN connectivity check responses (responsesReceived), including those that reply to requests
                                //!< that are sent in order to verify consent. The average round trip time can be computed from
                                //!< totalRoundTripTime by dividing it by responsesReceived.
-    UINT64 currentRoundTripTime; //!< Latest round trip time (seconds)
-    UINT64 availableOutgoingBitrate; //!< Total available bit rate for all the outgoing RTP streams on this candidate pair. Calculated by underlying
+    DOUBLE currentRoundTripTime; //!< Latest round trip time (seconds)
+    DOUBLE availableOutgoingBitrate; //!< Total available bit rate for all the outgoing RTP streams on this candidate pair. Calculated by underlying
                                      //!< congestion control
-    UINT64 availableIncomingBitrate; //!< Total available bit rate for all the outgoing RTP streams on this candidate pair. Calculated by underlying
+    DOUBLE availableIncomingBitrate; //!< Total available bit rate for all the outgoing RTP streams on this candidate pair. Calculated by underlying
                                      //!< congestion control
     UINT64 requestsReceived; //!< Total number of connectivity check requests received (including retransmission)
     UINT64 requestsSent; //!< The total number of connectivity check requests sent (without retransmissions).
@@ -225,7 +216,7 @@ typedef struct {
     CHAR tlsGroup[MAX_TLS_GROUP_LENGHTH + 1]; //!< Descriptive name of the group used for the encryption
                                               //!< Acceptable values: https://www.iana.org/assignments/tls-parameters/tls-parameters.xhtml#tls-parameters-8
     RTC_ICE_ROLE iceRole; //!< Set to the current value of the "role" attribute of the underlying RTCDtlsTransport's "transport"
-    RTC_DTLS_TRANSPORT_STATE dtlsState; //!< Set to the current value of the "state" attribute of the underlying RTCDtlsTransport
+    RTC_DTLS_TRANSPORT_STATE_STATS dtlsState; //!< Set to the current value of the "state" attribute of the underlying RTCDtlsTransport
     UINT64 packetsSent; //!< Total number of packets sent over the transport
     UINT64 packetsReceived; //!< Total number of packets received over the transport
     UINT64 bytesSent; //!< The total number of payload bytes sent on this PeerConnection (excluding header and padding)
--- a/src/include/com/amazonaws/kinesis/video/webrtcclient/Stats.h
+++ b/src/include/com/amazonaws/kinesis/video/webrtcclient/Stats.h
@@ -1,0 +1,331 @@
+/*
+ * Profiler public include file
+ */
+#ifndef __KINESIS_VIDEO_WEBRTCCLIENT_STATS_INCLUDE__
+#define __KINESIS_VIDEO_WEBRTCCLIENT_STATS_INCLUDE__
+
+#pragma once
+
+#ifdef  __cplusplus
+extern "C" {
+#endif
+
+////////////////////////////////////////////////////
+// Public headers
+////////////////////////////////////////////////////
+
+#define MAX_CANDIDATE_ID_LENGTH                   8U
+#define MAX_ADDRESS_LENGTH                       16U
+#define MAX_RELAY_PROTOCOL_LENGTH                 8U
+#define MAX_TLS_VERSION_LENGTH                    8U
+#define MAX_DTLS_CIPHER_LENGTH                    64U
+#define MAX_SRTP_CIPHER_LENGTH                    64U
+#define MAX_TLS_GROUP_LENGHTH                     32U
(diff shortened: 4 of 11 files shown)


src/include/com/amazonaws/kinesis/video/webrtcclient/Include.h
```diff
@@ -1235,6 +1236,46 @@ typedef struct {
                      //!< RTCDataChannel object with the same id as the other peer.
 } RtcDataChannelInit, *PRtcDataChannelInit;
 
+
+/////////////////////////////////////////////////////
+/// Metrics/Stats Related structures
+/////////////////////////////////////////////////////
+
+/**
+ * @brief Collection of ICE related stats
```
MushMal, line 1245: Could you perhaps include the references for RFC where these are defined?

src/include/com/amazonaws/kinesis/video/webrtcclient/Include.h
```diff
@@ -1235,6 +1236,46 @@ typedef struct {
...
+    PRTCRemoteInboundRtpStreamStats pInboundStats; //!< Inbound RTP Stats. Reference in Stats.h
+    PRTCOutboundRtpStreamStats pOutboundStats; //!< Outbound RTP Stats. Reference in Stats.h
+    PRTCTransportStats pTransportStats; //!< Transport stats. Reference in Stats.h
+} StreamStats, *PStreamStats;
+
+/**
+ * @brief SignalingStats Collection of signaling related stats. Can be expanded in the future
+ */
+typedef struct {
+    SignalingClientStats signalingClientMetrics; //!< Signaling client metrics stats. Reference in Stats.h
```
MushMal, line 1266: Not sure if we need two levels here - perhaps flat signaling stats might be enough

src/include/com/amazonaws/kinesis/video/webrtcclient/Include.h
```diff
@@ -1235,6 +1236,46 @@ typedef struct {
...
+
+/////////////////////////////////////////////////////
+/// Metrics/Stats Related structures
+/////////////////////////////////////////////////////
+
+/**
+ * @brief Collection of ICE related stats
+ */
+typedef struct {
+    PRTCIceServerStats pIceServerStats; //!< Server related stats. Reference in Stats.h
```
MushMal, line 1248: Use RtcBlahBlah - capitalization here and everywhere else

src/include/com/amazonaws/kinesis/video/webrtcclient/Include.h
```diff
@@ -1674,6 +1715,23 @@ PUBLIC_API STATUS signalingClientGetStateString(SIGNALING_CLIENT_STATE, PCHAR*);
...
 PUBLIC_API STATUS signalingClientDeleteSync(SIGNALING_CLIENT_HANDLE);
 
+/**
+ * @brief Get the relevant/all metrics based on the RTCStatsType field. This does not include
+ * any signaling related metrics
+ *
+ * @param PRtcPeerConnection Peer connection for which the stats need to be collected
+ * @param[in/out] PRtcStats The stats object with the RTCStatsType field populated
+ */
+PUBLIC_API STATUS getStats(PRtcPeerConnection, PRTCStats);
```
MushMal, line 1725: As we are using C, all the names are in the same namespace increasing a chance of random name collisions with the customers workspace. Please use a little more descriptive name. For example getRtcPeerConnectionStats

src/include/com/amazonaws/kinesis/video/webrtcclient/Include.h
```diff
@@ -1674,6 +1715,23 @@ PUBLIC_API STATUS signalingClientGetStateString(SIGNALING_CLIENT_STATE, PCHAR*);
...
+
+/**
+ * @brief Get signaling related metrics
+ *
+ * @param[in] SIGNALING_CLIENT_HANDLE Signaling client handle
+ * @param[in/out] PSignalingStats Signaling stats
+ */
+PUBLIC_API STATUS getSignalingStats(SIGNALING_CLIENT_HANDLE, PSignalingStats);
```
MushMal, line 1733: Signaling APIs are prefixed with signalingClient prefix. Use something like signalingClientGetStats

src/include/com/amazonaws/kinesis/video/webrtcclient/Stats.h
```diff
@@ -0,0 +1,321 @@
...
+#define MAX_PROTOCOL_LENGTH                       8U
+
+#define MAX_STRING_LENGTH                         255U
+
+/**
+ * @brief Type of Stats object requested by the application
+ * Reference https://www.w3.org/TR/webrtc-stats/#rtcstatstype-str*
+ */
+typedef enum {
+    CANDIDATE_PAIR, //!< ICE candidate pair statistics related to the RTCIceTransport objects
```
MushMal, line 33: We normally prefix the enum values with the enum name for the same reason - name collisions. So for example this one will be RTS_STAT_TYPE_CANDIDATE_PAIR. Same for all of the entries

src/include/com/amazonaws/kinesis/video/webrtcclient/Stats.h
```diff
@@ -0,0 +1,321 @@
...
+    TRANSCEIVER, //!< Statistics related to a specific RTCRtpTransceiver
+    RTC_ALL //!< Report all supported stats
+} RTC_STATS_TYPE;
+
+/**
+ * @brief Indicates computing states in the checklist
+ * Reference: https://www.w3.org/TR/webrtc-stats/#rtcstatsicecandidatepairstate-enum
+ */
+typedef enum {
+    FROZEN,    //!< A check for the pair hasn't been performed because it is waiting for other checks to complete
```
MushMal, line 62: Same here. Think of how you can shorten the name of the enum. For example ICE_PAIR_STATE and then this entry will be ICE_PAIR_STATE_FROZEN

src/include/com/amazonaws/kinesis/video/webrtcclient/Stats.h
```diff
@@ -0,0 +1,321 @@
...
+
+/**
+ * @brief Set details of the IceAgent based on STUN_ATTRIBUTE_TYPE_USE_CANDIDATE flag
+ * Reference: https://www.w3.org/TR/webrtc/#rtcicerole
+ */
+typedef enum {
+    UNKNOWN,    //!< An agent whose role is undetermined. Initial state
+    CONTROLLING,//!< A controlling agent. The ICE agent that is responsible for selecting the final choice of candidate pairs
+    CONTROLLED  //!< A controlled agent. The iCE agent waits for the controlling agent to select the final choice of candidate pairs
+} RTCIceRole;
```
MushMal, line 77: All enums are capitalized and underscored. RTC_ICE_ROLE.

src/include/com/amazonaws/kinesis/video/webrtcclient/Stats.h
```diff
@@ -0,0 +1,321 @@
...
+}qualityLimitationDurationsRecord, PQualityLimitationDurationsRecord;
+
+/**
+ * @brief Record of total number of packets sent per DSCP. Used by RTCOutboundRtpStreamStats
+ * object
+ */
+typedef struct {
+    CHAR dscp[MAX_STRING_LENGTH + 1]; //!< DSCP String
+    UINT64 totalNumberOfPacketsSent; //!< Number of packets sent
+} perDscpPacketsSentRecord, *PperDscpPacketsSentRecord;
```
MushMal, line 118: Capitalize the struct names. Perhaps drop the 'per' prefix - something like DscpPacketsSentRecord

src/include/com/amazonaws/kinesis/video/webrtcclient/Stats.h
```diff
@@ -0,0 +1,321 @@
...
+    UINT64 requestsReceived; //!< Total number of connectivity check requests received (including retransmission)
+    UINT64 requestsSent; //!< The total number of connectivity check requests sent (without retransmissions).
+    UINT64 responsesReceived; //!< The total number of connectivity check responses received.
+    UINT64 responsesSent; //!< The total number of connectivity check responses sent.
+    UINT64 retransmissionsReceived; //!< The total number of connectivity check request retransmissions received
+    UINT64 retransmissionsSent; //!< The total number of connectivity check request retransmissions sent.
+    UINT64 consentRequestsSent; //!< The total number of consent requests sent.
+    UINT64 consentExpiredTimestamp; //!< The timestamp at which the latest valid STUN binding response expired
+    UINT64 bytesDiscardedOnSend; //!< Total number of bytes for this candidate pair discarded due to socket errors
+} RTCIceCandidatePairStats, *PRTCIceCandidatePairStats;
```
MushMal, line 167: RtcIceCandidatePairStats

src/include/com/amazonaws/kinesis/video/webrtcclient/Stats.h
```diff
@@ -0,0 +1,321 @@
...
+ * Reference: https://www.w3.org/TR/webrtc-stats/#ice-server-dict*
+ */
+typedef struct {
+    CHAR url[MAX_STRING_LENGTH + 1]; //!< STUN/TURN server URL
+    CHAR protocol[MAX_STRING_LENGTH + 1]; //!< Valid values: UDP, TCP
+    INT32 port; //!< Port number used by client
+    UINT64 totalRequestsSent; //!< Total amount of requests that have been sent to the server
+    UINT64 totalResponsesReceived; //!< Total number of responses received from the server
+    UINT64 totalRoundTripTime; //!< Sum of RTTs of all the requests for which response has been received
+} RTCIceServerStats, *PRTCIceServerStats;
```
MushMal, line 181: same and below

src/source/Ice/IceAgent.h
```diff
@@ -200,6 +200,9 @@ struct __IceAgent {
 
     // store transaction ids for stun binding request.
     PTransactionIdStore pStunBindingRequestTransactionIdStore;
+
+    // Add ICE related Stats
+    PIceStats pIceStats;
```
MushMal, line 205: Don't include these. We will be adding them later - not sure if we will keep the public struct or an internal one that can carry more info needed for tracking. Also, you are including a pointer here.

src/source/PeerConnection/PeerConnection.h
```diff
@@ -89,6 +89,8 @@ typedef struct {
     RTC_PEER_CONNECTION_STATE previousConnectionState;
 
     UINT16 MTU;
+    // Get RTP Stream Stats
+    PStreamStats pStreamStats;
```
MushMal, line 93: same

src/source/Signaling/Signaling.h
```diff
@@ -243,6 +243,7 @@ typedef struct {
     UINT64 getIceConfigTime;
     UINT64 deleteTime;
     UINT64 connectTime;
+    PSignalingStats pSignalingStats;
```
MushMal, line 246: ditto. Each individual owner will drive this

src/include/com/amazonaws/kinesis/video/webrtcclient/Include.h
disa6302: I have added them in Stats.h where the structures are defined. I will also add it here for reference.

src/include/com/amazonaws/kinesis/video/webrtcclient/Include.h
disa6302: Done

src/include/com/amazonaws/kinesis/video/webrtcclient/Include.h
disa6302: Yes it is. Just having it in 2 levels incase we have any new signaling related structure in the future. That is the reason I mentioned can be expanded in the future

src/include/com/amazonaws/kinesis/video/webrtcclient/Include.h
disa6302: Done

src/include/com/amazonaws/kinesis/video/webrtcclient/Include.h
disa6302: Done

src/include/com/amazonaws/kinesis/video/webrtcclient/Stats.h
disa6302: Got it. The reason I named the enums this way is to have it consistent with the spec. Let me rename all the enums with the prefix.

src/include/com/amazonaws/kinesis/video/webrtcclient/Stats.h
disa6302: Done

src/include/com/amazonaws/kinesis/video/webrtcclient/Stats.h
disa6302: Done

src/include/com/amazonaws/kinesis/video/webrtcclient/Stats.h
disa6302: Done

src/include/com/amazonaws/kinesis/video/webrtcclient/Stats.h
disa6302: Done

src/include/com/amazonaws/kinesis/video/webrtcclient/Stats.h
disa6302: Done

src/source/PeerConnection/PeerConnection.h
disa6302: Done

src/source/Signaling/Signaling.h
disa6302: Got it

src/source/Ice/IceAgent.h
disa6302: Done

disa6302:
> Great start!!!
> 
> My comments inline are more about the naming and style.
> 
> One other important question you might want to think through is whether all of the public structures need UINT32 version fields. Think about your call pattern. The user code will be calling the getXXXStats method and provide a pointer that you know is pre-allocated. Having a version field allows you to be forward and backward compat. For example, in other cases, the public structures are versioned in case if the customers application is using older header files with new library. Not a big issue but have been burned before. I don't feel strongly about the version fields now as we have 1:1 tight versioning and explicit handling of major, minor and revision versions. Certainly open for discussions.

I have no strong opinion against versioning. I can certainly add a version field to the public structures - RtcIceMetrics, RtcStreamMetrics, SignalingClientMetrics. I feel we should not just depend on the 1:1 tight versioning and do our due diligence in the code base itself. Let me add those fields in.

src/include/com/amazonaws/kinesis/video/webrtcclient/Stats.h
```diff
@@ -0,0 +1,321 @@
...
+    RTC_STATS_TYPE_PEER_CONNECTION, //!< Statistics related to the RTCPeerConnection object.
+    RTC_STATS_TYPE_RECEIVER, //!< Statistics related to a specific receiver and the corresponding media-level metrics
+    RTC_STATS_TYPE_REMOTE_CANDIDATE, //!< ICE remote candidate statistics related to the RTCIceTransport objects
+    RTC_STATS_TYPE_REMOTE_INBOUND_RTP, //!< Statistics for the remote endpoint's inbound RTP stream corresponding to an outbound stream that
+                        //!< is currently sent with this RTCPeerConnection object
+    RTC_STATS_TYPE_REMOTE_OUTBOUND_RTP, //!< Statistics for the remote endpoint's outbound RTP stream corresponding to an inbound stream that
+                         //!< is currently sent with this RTCPeerConnection object
+    RTC_STATS_TYPE_SENDER, //!< Statistics related to a specific RTCRtpSender and the corresponding media-level metrics
+    RTC_STATS_TYPE_TRACK, //!< Statistics related to a specific MediaStreamTrack's attachment to an RTCRtpSender
+    RTC_STATS_TYPE_, //!< Transport statistics related to the RTCPeerConnection object.
```
chehefen, line 51: is this complete?

src/include/com/amazonaws/kinesis/video/webrtcclient/Stats.h
```diff
@@ -0,0 +1,321 @@
...
+ */
+
+typedef struct {
+    CHAR transportId[MAX_STRING_LENGTH + 1]; //!< ID of object that was inspected for RTCTransportStats
+    CHAR localCandidateId[MAX_CANDIDATE_ID_LENGTH + 1]; //!< Local candidate that is inspected in RTCIceCandidateStats
+    CHAR remoteCandidateId[MAX_CANDIDATE_ID_LENGTH + 1]; //!< Remote candidate that is inspected in RTCIceCandidateStats
+    RTC_STATS_ICE_CANDIDATE_PAIR_STATE state; //!< State of checklist for the local-remote candidate pair
+    BOOL nominated; //!< Flag is TRUE if the agent is a controlling agent and FALSE otherwise. The agent role is based on the
+                    //!< STUN_ATTRIBUTE_TYPE_USE_CANDIDATE flag
+    UINT32 circuitBreakerTriggerCount; //!< Represents number of times circuit breaker is triggered during media transmission
```
chehefen, line 133: this is undefined for us. should we implement data types that can be undefined?

chehefen:
some of the metrics wont be available right away. I wonder whats a good way to communicate that to the user. If we just set it as undefined, it's hard to tell if the value is actually undefined or not implemented.

disa6302:
> some of the metrics wont be available right away. I wonder whats a good way to communicate that to the user. If we just set it as undefined, it's hard to tell if the value is actually undefined or not implemented.

There are 2 ways to go about this: We can remove the unused ones and add it back in the future when we use it. Or, we can have a sentinel (maybe set it to MAX value by default, with the assumption that we would not be hitting MAX at all. When the user passes in the structure object, we can reset the value to MAX and add a log maybe, could be messy though). 

I do not feel strongly against either. We can always add these back in the future, but, the reason I defined the whole structure now is to avoid the hassle of adding new items later and make it readily available, that is all.

src/include/com/amazonaws/kinesis/video/webrtcclient/Stats.h
disa6302: Fixed. Missed pushing in the change

chehefen:
> > some of the metrics wont be available right away. I wonder whats a good way to communicate that to the user. If we just set it as undefined, it's hard to tell if the value is actually undefined or not implemented.
> 
> There are 2 ways to go about this: We can remove the unused ones and add it back in the future when we use it. Or, we can have a sentinel (maybe set it to MAX value by default, with the assumption that we would not be hitting MAX at all. When the user passes in the structure object, we can reset the value to MAX and add a log maybe, could be messy though).
> 
> I do not feel strongly against either. We can always add these back in the future, but, the reason I defined the whole structure now is to avoid the hassle of adding new items later and make it readily available, that is all.

Yh I see your point. The problem is sentinel is that there are different data types like string and uint and bool, combining it with the fact that in the spec these attributes can be undefined, it becomes hard to tell whether one attribute is available or not. what do you think if we start with all of them commented and uncomment them as we fill them in?

disa6302:
> > > some of the metrics wont be available right away. I wonder whats a good way to communicate that to the user. If we just set it as undefined, it's hard to tell if the value is actually undefined or not implemented.
> > 
> > 
> > There are 2 ways to go about this: We can remove the unused ones and add it back in the future when we use it. Or, we can have a sentinel (maybe set it to MAX value by default, with the assumption that we would not be hitting MAX at all. When the user passes in the structure object, we can reset the value to MAX and add a log maybe, could be messy though).
> > I do not feel strongly against either. We can always add these back in the future, but, the reason I defined the whole structure now is to avoid the hassle of adding new items later and make it readily available, that is all.
> 
> Yh I see your point. The problem is sentinel is that there are different data types like string and uint and bool, combining it with the fact that in the spec these attributes can be undefined, it becomes hard to tell whether one attribute is available or not. what do you think if we start with all of them commented and uncomment them as we fill them in?

All the attributes that has the potential to be undefined? Sure! We can do that! Another way is add a comment in the attribute itself saying it is UNUSED for now. Either ways, these values are computed and returned to the application. So, I do not foresee much issue with them not being used.

MushMal:
> > > > some of the metrics wont be available right away. I wonder whats a good way to communicate that to the user. If we just set it as undefined, it's hard to tell if the value is actually undefined or not implemented.
> > > 
> > > 
> > > There are 2 ways to go about this: We can remove the unused ones and add it back in the future when we use it. Or, we can have a sentinel (maybe set it to MAX value by default, with the assumption that we would not be hitting MAX at all. When the user passes in the structure object, we can reset the value to MAX and add a log maybe, could be messy though).
> > > I do not feel strongly against either. We can always add these back in the future, but, the reason I defined the whole structure now is to avoid the hassle of adding new items later and make it readily available, that is all.
> > 
> > 
> > Yh I see your point. The problem is sentinel is that there are different data types like string and uint and bool, combining it with the fact that in the spec these attributes can be undefined, it becomes hard to tell whether one attribute is available or not. what do you think if we start with all of them commented and uncomment them as we fill them in?
> 
> All the attributes that has the potential to be undefined? Sure! We can do that! Another way is add a comment in the attribute itself saying it is UNUSED for now. Either ways, these values are computed and returned to the application. So, I do not foresee much issue with them not being used.

So, to recap - what's the main issue with the "white list" approach? We add the enums as we implement them?

disa6302:
> > > > > some of the metrics wont be available right away. I wonder whats a good way to communicate that to the user. If we just set it as undefined, it's hard to tell if the value is actually undefined or not implemented.
> > > > 
> > > > 
> > > > There are 2 ways to go about this: We can remove the unused ones and add it back in the future when we use it. Or, we can have a sentinel (maybe set it to MAX value by default, with the assumption that we would not be hitting MAX at all. When the user passes in the structure object, we can reset the value to MAX and add a log maybe, could be messy though).
> > > > I do not feel strongly against either. We can always add these back in the future, but, the reason I defined the whole structure now is to avoid the hassle of adding new items later and make it readily available, that is all.
> > > 
> > > 
> > > Yh I see your point. The problem is sentinel is that there are different data types like string and uint and bool, combining it with the fact that in the spec these attributes can be undefined, it becomes hard to tell whether one attribute is available or not. what do you think if we start with all of them commented and uncomment them as we fill them in?
> > 
> > 
> > All the attributes that has the potential to be undefined? Sure! We can do that! Another way is add a comment in the attribute itself saying it is UNUSED for now. Either ways, these values are computed and returned to the application. So, I do not foresee much issue with them not being used.
> 
> So, to recap - what's the main issue with the "white list" approach? We add the enums as we implement them?

Yes. No issues. Just that we need to add them as and when we implement => multiple changes to a public header file. That is all.

MushMal:
> > > > > > some of the metrics wont be available right away. I wonder whats a good way to communicate that to the user. If we just set it as undefined, it's hard to tell if the value is actually undefined or not implemented.
> > > > > 
> > > > > 
> > > > > There are 2 ways to go about this: We can remove the unused ones and add it back in the future when we use it. Or, we can have a sentinel (maybe set it to MAX value by default, with the assumption that we would not be hitting MAX at all. When the user passes in the structure object, we can reset the value to MAX and add a log maybe, could be messy though).
> > > > > I do not feel strongly against either. We can always add these back in the future, but, the reason I defined the whole structure now is to avoid the hassle of adding new items later and make it readily available, that is all.
> > > > 
> > > > 
> > > > Yh I see your point. The problem is sentinel is that there are different data types like string and uint and bool, combining it with the fact that in the spec these attributes can be undefined, it becomes hard to tell whether one attribute is available or not. what do you think if we start with all of them commented and uncomment them as we fill them in?
> > > 
> > > 
> > > All the attributes that has the potential to be undefined? Sure! We can do that! Another way is add a comment in the attribute itself saying it is UNUSED for now. Either ways, these values are computed and returned to the application. So, I do not foresee much issue with them not being used.
> > 
> > 
> > So, to recap - what's the main issue with the "white list" approach? We add the enums as we implement them?
> 
> Yes. No issues. Just that we need to add them as and when we implement => multiple changes to a public header file. That is all.

I got it, the main issue is ABI - you can always add an enum field without breaking back compat. I recommend focusing on the structures themselves. It’s ok to add new structures but harder to modify the ones that are public without the version field

disa6302:
> > > > > > > some of the metrics wont be available right away. I wonder whats a good way to communicate that to the user. If we just set it as undefined, it's hard to tell if the value is actually undefined or not implemented.
> > > > > > 
> > > > > > 
> > > > > > There are 2 ways to go about this: We can remove the unused ones and add it back in the future when we use it. Or, we can have a sentinel (maybe set it to MAX value by default, with the assumption that we would not be hitting MAX at all. When the user passes in the structure object, we can reset the value to MAX and add a log maybe, could be messy though).
> > > > > > I do not feel strongly against either. We can always add these back in the future, but, the reason I defined the whole structure now is to avoid the hassle of adding new items later and make it readily available, that is all.
> > > > > 
> > > > > 
> > > > > Yh I see your point. The problem is sentinel is that there are different data types like string and uint and bool, combining it with the fact that in the spec these attributes can be undefined, it becomes hard to tell whether one attribute is available or not. what do you think if we start with all of them commented and uncomment them as we fill them in?
> > > > 
> > > > 
> > > > All the attributes that has the potential to be undefined? Sure! We can do that! Another way is add a comment in the attribute itself saying it is UNUSED for now. Either ways, these values are computed and returned to the application. So, I do not foresee much issue with them not being used.
> > > 
> > > 
> > > So, to recap - what's the main issue with the "white list" approach? We add the enums as we implement them?
> > 
> > 
> > Yes. No issues. Just that we need to add them as and when we implement => multiple changes to a public header file. That is all.
> 
> I got it, the main issue is ABI - you can always add an enum field without breaking back compat. I recommend focusing on the structures themselves. It’s ok to add new structures but harder to modify the ones that are public without the version field

Agreed. The enums are fine. It is the structures that is under discussion. I have added a version field to the structures in Include.h, but not to the structures in Stats.h because I am referencing the spec directly. So, I am unsure on whether it is a good idea to add the version field in there.

MushMal:
> > > > > > > > some of the metrics wont be available right away. I wonder whats a good way to communicate that to the user. If we just set it as undefined, it's hard to tell if the value is actually undefined or not implemented.
> > > > > > > 
> > > > > > > 
> > > > > > > There are 2 ways to go about this: We can remove the unused ones and add it back in the future when we use it. Or, we can have a sentinel (maybe set it to MAX value by default, with the assumption that we would not be hitting MAX at all. When the user passes in the structure object, we can reset the value to MAX and add a log maybe, could be messy though).
> > > > > > > I do not feel strongly against either. We can always add these back in the future, but, the reason I defined the whole structure now is to avoid the hassle of adding new items later and make it readily available, that is all.
> > > > > > 
> > > > > > 
> > > > > > Yh I see your point. The problem is sentinel is that there are different data types like string and uint and bool, combining it with the fact that in the spec these attributes can be undefined, it becomes hard to tell whether one attribute is available or not. what do you think if we start with all of them commented and uncomment them as we fill them in?
> > > > > 
> > > > > 
> > > > > All the attributes that has the potential to be undefined? Sure! We can do that! Another way is add a comment in the attribute itself saying it is UNUSED for now. Either ways, these values are computed and returned to the application. So, I do not foresee much issue with them not being used.
> > > > 
> > > > 
> > > > So, to recap - what's the main issue with the "white list" approach? We add the enums as we implement them?
> > > 
> > > 
> > > Yes. No issues. Just that we need to add them as and when we implement => multiple changes to a public header file. That is all.
> > 
> > 
> > I got it, the main issue is ABI - you can always add an enum field without breaking back compat. I recommend focusing on the structures themselves. It’s ok to add new structures but harder to modify the ones that are public without the version field
> 
> Agreed. The enums are fine. It is the structures that is under discussion. I have added a version field to the structures in Include.h, but not to the structures in Stats.h because I am referencing the spec directly. So, I am unsure on whether it is a good idea to add the version field in there.

The rfc shouldn’t really be treated verbatim. Adding a version field is an OK thing to do. However, the public rfc structures will be more static so as long as we fill in the fields we should be ok. We can only publish the structures that we will support right away. We can add more as we progress

src/include/com/amazonaws/kinesis/video/webrtcclient/Stats.h
```diff
@@ -0,0 +1,331 @@
...
+ * Reference: https://www.w3.org/TR/webrtc-stats/#outboundrtpstats-dict*
+ */
+typedef struct {
+    BOOL voiceActivityFlag; //!< Only valid for audio. Whether the last RTP packet sent contained voice activity or not based on the presence
+                            //!< of the V bit in the extension header
+    CHAR trackId[MAX_STRING_LENGTH + 1]; //!< ID representing current track attached to the sender of the stream
+    CHAR mediaSourceId[MAX_STRING_LENGTH + 1]; //!< ID representing the current media source
+    CHAR senderId[MAX_STRING_LENGTH + 1]; //!< The stats ID used to look up the RTCAudioSenderStats or RTCVideoSenderStats object sending this stream
+    CHAR remoteId[MAX_STRING_LENGTH + 1]; //!< ID to look up the remote RTCRemoteInboundRtpStreamStats object for the same SSRC
+    CHAR rid[MAX_STRING_LENGTH + 1]; //!< Exposes the rid encoding parameter of this RTP stream if it has been set, otherwise it is undefined
```
chehefen, line 248: this can be undefined.

src/include/com/amazonaws/kinesis/video/webrtcclient/Stats.h
```diff
@@ -0,0 +1,331 @@
...
+                                 //!< The average interval between two consecutive connectivity checks sent can be calculated:
+                                 //! (lastRequestTimestamp - firstRequestTimestamp) / requestsSent.
+    UINT64 lastResponseTimestamp; //!< Represents the timestamp at which the last STUN response was received on this particular candidate pair.
+    UINT64 totalRoundTripTime; //!< The sum of all round trip time (seconds) since the beginning of the session, based
+                               //!< on STUN connectivity check responses (responsesReceived), including those that reply to requests
+                               //!< that are sent in order to verify consent. The average round trip time can be computed from
+                               //!< totalRoundTripTime by dividing it by responsesReceived.
+    UINT64 currentRoundTripTime; //!< Latest round trip time (seconds)
+    UINT64 availableOutgoingBitrate; //!< Total available bit rate for all the outgoing RTP streams on this candidate pair. Calculated by underlying
```
chehefen, line 163: this can be undefined. perhaps we should make them all use definable data types?

CMakeLists.txt
```diff
@@ -197,7 +197,8 @@ file(
   "src/source/Sdp/*.c"
   "src/source/Srtp/*.c"
   "src/source/Stun/*.c"
-  "src/source/Sctp/*.c")
+  "src/source/Sctp/*.c"
+  "src/source/WebRtcMetrics/*.c")
```
MushMal, line 201: Perhaps a simple Metrics/ directory name is good enough?

samples/Common.c
```diff
@@ -208,7 +208,7 @@ STATUS handleOffer(PSampleConfiguration pSampleConfiguration, PSampleStreamingSe
     RtcSessionDescriptionInit offerSessionDescriptionInit;
     BOOL locked = FALSE;
     NullableBool canTrickle;
-
+//    RTCStats rtcMetrics;
```
MushMal, line 211: Pre-prod stuff? Here and below

src/include/com/amazonaws/kinesis/video/webrtcclient/Include.h
MushMal: I don't think this indirection really is necessary even for the future expansion. The version will provide means to expand and still support backward compat.

src/include/com/amazonaws/kinesis/video/webrtcclient/Include.h
```diff
@@ -1242,6 +1258,55 @@ typedef struct {
...
+
+/**
+ * @brief The stats object is populated based on RTCStatsType request
+ *
+ */
+typedef struct {
+    UINT64 timestamp; //!< Timestamp of request for stats
+    RTC_STATS_TYPE requestedTypeOfStats; //!< Type of stats requested. Set to RTC_ALL to get all supported stats
+    RtcStatsObject rtcStatsObject; //!< Object that is populated by the SDK on request
+} RTCStats, *PRTCStats;
```
MushMal, line 1308: RtcStats, PRtsStats

src/include/com/amazonaws/kinesis/video/webrtcclient/Include.h
```diff
@@ -1717,6 +1782,23 @@ PUBLIC_API STATUS signalingClientGetStateString(SIGNALING_CLIENT_STATE, PCHAR*);
...
+ */
+PUBLIC_API STATUS getRtcPeerConnectionStats(PRtcPeerConnection, PRTCStats);
+
+/**
+ * @brief Get signaling related metrics
+ *
+ * @param[in] SIGNALING_CLIENT_HANDLE Signaling client handle
+ * @param[in/out] PSignalingClientMetrics Signaling stats
+ */
+PUBLIC_API STATUS getSignalingClientStats(SIGNALING_CLIENT_HANDLE, PSignalingClientMetrics);
```
MushMal, line 1800: To be consistent with naming:

* Please move this API together with the signaling APIs
* Rename to signalingClientGetMetrics

src/include/com/amazonaws/kinesis/video/webrtcclient/Stats.h
```diff
@@ -0,0 +1,322 @@
...
+ */
+
+typedef struct {
+    CHAR transportId[MAX_STRING_LENGTH + 1]; //!< ID of object that was inspected for RTCTransportStats
+    CHAR localCandidateId[MAX_CANDIDATE_ID_LENGTH + 1]; //!< Local candidate that is inspected in RTCIceCandidateStats
+    CHAR remoteCandidateId[MAX_CANDIDATE_ID_LENGTH + 1]; //!< Remote candidate that is inspected in RTCIceCandidateStats
+    ICE_CANDIDATE_PAIR_STATE state; //!< State of checklist for the local-remote candidate pair
+    BOOL nominated; //!< Flag is TRUE if the agent is a controlling agent and FALSE otherwise. The agent role is based on the
+                    //!< STUN_ATTRIBUTE_TYPE_USE_CANDIDATE flag
+    NullableUint32 circuitBreakerTriggerCount; //!< Represents number of times circuit breaker is triggered during media transmission
```
MushMal, line 133: Does this need to be Nullable type? Odd

src/include/com/amazonaws/kinesis/video/webrtcclient/Stats.h
```diff
@@ -0,0 +1,322 @@
...
+    UINT64 lastPacketSentTimestamp; //!< Represents the timestamp at which the last packet was sent on this particular
+                                    //!< candidate pair, excluding STUN packets.
+    UINT64 lastPacketReceivedTimestamp; //!< Represents the timestamp at which the last packet was sent on this particular
+                                        //!< candidate pair, excluding STUN packets.
+    UINT64 firstRequestTimestamp; //!< Represents the timestamp at which the first STUN request was sent on this particular candidate pair.
+    UINT64 lastRequestTimestamp; //!< Represents the timestamp at which the last STUN request was sent on this particular candidate pair.
+                                 //!< The average interval between two consecutive connectivity checks sent can be calculated:
+                                 //! (lastRequestTimestamp - firstRequestTimestamp) / requestsSent.
+    UINT64 lastResponseTimestamp; //!< Represents the timestamp at which the last STUN response was received on this particular candidate pair.
+    DOUBLE totalRoundTripTime; //!< The sum of all round trip time (seconds) since the beginning of the session, based
```
MushMal, line 149: Generally, I discourage usage of double or float types measuring time for example. Is this spec-ed? The entire SDK is based on the standard time units - 100ns and all our time and period measurements are in UINT64

MushMal: Same for other instances of double

src/include/com/amazonaws/kinesis/video/webrtcclient/Stats.h
```diff
@@ -0,0 +1,322 @@
...
+                               //!< on STUN connectivity check responses (responsesReceived), including those that reply to requests
+                               //!< that are sent in order to verify consent. The average round trip time can be computed from
+                               //!< totalRoundTripTime by dividing it by responsesReceived.
+    DOUBLE currentRoundTripTime; //!< Latest round trip time (seconds)
+    DOUBLE availableOutgoingBitrate; //!< Total available bit rate for all the outgoing RTP streams on this candidate pair. Calculated by underlying
+                                     //!< congestion control
+    DOUBLE availableIncomingBitrate; //!< Total available bit rate for all the outgoing RTP streams on this candidate pair. Calculated by underlying
```
MushMal, line 156: Why double? Can easily be int type - say UINT32

MushMal: LOL, might make UINT64 just in case

src/include/com/amazonaws/kinesis/video/webrtcclient/Stats.h
```diff
@@ -0,0 +1,322 @@
...
+
+/**
+ * @brief: RtcIceServerStats Stats related to the ICE Server
+ *
+ * Reference: https://www.w3.org/TR/webrtc-stats/#ice-server-dict*
+ */
+typedef struct {
+    CHAR url[MAX_STRING_LENGTH + 1]; //!< STUN/TURN server URL
+    CHAR protocol[MAX_STRING_LENGTH + 1]; //!< Valid values: UDP, TCP
+    INT32 port; //!< Port number used by client
```
MushMal, line 177: Should this be unsigned?

src/include/com/amazonaws/kinesis/video/webrtcclient/Stats.h
```diff
@@ -0,0 +1,322 @@
...
+
+typedef struct {
+    CHAR url[MAX_STRING_LENGTH + 1]; //!< For local candidates this is the URL of the ICE server from which the candidate was obtained
+    CHAR transportId[MAX_STRING_LENGTH + 1]; //!< ID of object that was inspected for RTCTransportStats
+    BYTE address[MAX_ADDRESS_LENGTH + 1]; //!< IPv4 or IPv6 address of the candidate
+    CHAR protocol[MAX_PROTOCOL_LENGTH + 1]; //!< Valid values: UDP, TCP
+    CHAR relayProtocol[MAX_RELAY_PROTOCOL_LENGTH + 1]; //!< Protocol used by endpoint to communicate with TURN server.
+                                                       //!< Valid values: UDP, TCP, TLS
+    INT32 priority; //!< Computed using the formula in https://tools.ietf.org/html/rfc5245#section-15.1
+    INT32 port; //!< Port number of the candidate
```
MushMal, line 197: here too

src/include/com/amazonaws/kinesis/video/webrtcclient/Stats.h
```diff
@@ -0,0 +1,322 @@
...
+    UINT64 framesPerSecond; //!< Only valid for video. The number of encoded frames during the last second
+    UINT64 qpSum; //!< Only valid for video. The sum of the QP values of frames encoded by this sender. QP value depends on the codec
+    UINT64 totalSamplesSent; //!< Only valid for audio. The total number of samples that have been sent over this RTP stream
+    UINT64 samplesEncodedWithSilk; //!< Only valid for audio and when the audio codec is Opus. Represnets only SILK portion of codec
+    UINT64 samplesEncodedWithCelt; //!< Only valid for audio and when the audio codec is Opus. Represnets only CELT portion of codec
+    UINT64 totalEncodeTime; //!< Total number of seconds that has been spent encoding the framesEncoded frames of the stream
+    UINT64 totalPacketSendDelay; //!< Total time (seconds) packets have spent buffered locally before being transmitted onto the network
+    UINT64 averageRtcpInterval; //!< The average RTCP interval between two consecutive compound RTCP packets
+    PQualityLimitationDurationsRecord qualityLimitationDurations; //!< Total time (seconds) spent in each reason state
```
MushMal, line 272: This is a pointer? Why

src/include/com/amazonaws/kinesis/video/webrtcclient/Stats.h
```diff
@@ -0,0 +1,322 @@
...
+typedef struct {
+    UINT64 apiCallLatency; //!< Latency (milliseconds) incurred per backend API call
+    UINT64 signalingClientUptime; //!< Client uptime (milliseconds). Timestamp will be recorded at every SIGNALING_CLIENT_STATE_CONNECTED
+    UINT64 connectionDuration; //!< Duration of connection (milliseconds)
+    UINT64 numberOfMessagesSent; //!< Number of messages sent by the signaling client
+    UINT64 numberOfMessagesReceived; //!< Number of messages received by the signaling client
+    UINT64 iceRefreshCount; //!< Number of times the ICE is refreshed
+    UINT64 numberOfRecoverableErrors; //!< Number of recoverable errors. Will be a static count returned when requested
+    UINT64 numberOfReconnects; //!< Number of reconnects in the session
+} SignalingClientStats, PSignalingClientStats;
```
MushMal, line 303: I see why you are making the indirection.. OK, now it makes sense

src/include/com/amazonaws/kinesis/video/webrtcclient/Stats.h
```diff
@@ -0,0 +1,322 @@
...
+ * be populated internally
+ */
+typedef struct {
+    RtcIceCandidatePairStats iceCandidatePairStats; //!< ICE Candidate Pair  stats object
+    RtcIceCandidateStats iceCandidateStats; //!< ICE Candidate stats object
+    RtcIceServerStats iceServerStats; //!< ICE Server Pair stats object
+    RtcTransportStats transportStats; //!< Transport stats object
+    RtcOutboundRtpStreamStats remoteOutboundRtpStreamStats; //!< Outbound RTP Stream stats object
+    RtcRemoteInboundRtpStreamStats remoteInboundRtpStreamStats; //!< Inbound RTP Stream stats object
+    SignalingClientStats signalingClientStats; //!< Signaling Client stats object
```
MushMal, line 316: Don't think this should belong here

src/include/com/amazonaws/kinesis/video/webrtcclient/Stats.h
```diff
@@ -0,0 +1,322 @@
...
+#ifdef  __cplusplus
+extern "C" {
+#endif
+
+////////////////////////////////////////////////////
+// Public headers
+////////////////////////////////////////////////////
+
+#define MAX_CANDIDATE_ID_LENGTH                   8U
+#define MAX_ADDRESS_LENGTH                       16U
```
MushMal, line 18: Very generic name - might conflict

src/include/com/amazonaws/kinesis/video/webrtcclient/Stats.h
```diff
@@ -0,0 +1,322 @@
...
+#define MAX_RELAY_PROTOCOL_LENGTH                 8U
+#define MAX_TLS_VERSION_LENGTH                    8U
+#define MAX_DTLS_CIPHER_LENGTH                    64U
+#define MAX_SRTP_CIPHER_LENGTH                    64U
+#define MAX_TLS_GROUP_LENGHTH                     32U
+#define MAX_PROTOCOL_LENGTH                       8U
```
MushMal, line 24: Same

src/include/com/amazonaws/kinesis/video/webrtcclient/Stats.h
```diff
@@ -0,0 +1,322 @@
...
+
+#define MAX_STRING_LENGTH                         255U
```
MushMal, line 26: Definitely will conflict. Select longer and more localized names. For example, here a better naming might be MAX_STAT_NAME_STRING_LENGTH perhaps..

Also, don't need to specify the U after the constant

src/source/Ice/IceAgent.c
```diff
@@ -2244,6 +2245,8 @@ STATUS handleStunPacket(PIceAgent pIceAgent, PBYTE pBuffer, UINT32 bufferLen, PS
 
     switch (stunPacketType) {
         case STUN_PACKET_TYPE_BINDING_REQUEST:
+            pIceAgent->rtcIceMetrics.rtcIceCandidatePairStats.requestsReceived++;
+            // printf("Requests received:%d\n", pIceAgent->pRtcIceMetrics->pRTCIceCandidatePairStats->requestsReceived++);
```
MushMal, line 2249: remove

src/source/Ice/IceAgent.c
```diff
@@ -2363,6 +2367,11 @@ STATUS handleStunPacket(PIceAgent pIceAgent, PBYTE pBuffer, UINT32 bufferLen, PS
             break;
     }
 
+    pIceAgent->rtcIceMetrics.rtcIceCandidatePairStats.currentRoundTripTime = (DOUBLE) (pIceCandidatePair->roundTripTime / 1000.0);
```
MushMal, line 2370: Use standard 100ns

PR #463, opened 2020-05-16
add canTrickleIceCandidates api. Modify master sample to do trickle i…

### LINKED ISSUE #445: Add support for trickle ICE attribute
We should honour trickle ICE attribute in SDP offer (a=trickle-ice).

chehefen:
…ce based on remote description.

*Issue #, if available:*
fix #445 
*Description of changes:*

### DIFF
--- a/samples/Common.c
+++ b/samples/Common.c
@@ -109,7 +109,7 @@ STATUS signalingClientError(UINT64 customData, STATUS status, PCHAR msg, UINT32 
     // We will force re-create the signaling client on the following errors
     if (status == STATUS_SIGNALING_ICE_CONFIG_REFRESH_FAILED || status == STATUS_SIGNALING_RECONNECT_FAILED) {
         ATOMIC_STORE_BOOL(&pSampleConfiguration->recreateSignalingClient, TRUE);
-        CVAR_BROADCAST(gSampleConfiguration->cvar);
+        CVAR_BROADCAST(pSampleConfiguration->cvar);
     }
 
     return STATUS_SUCCESS;
@@ -206,6 +206,7 @@ STATUS handleOffer(PSampleConfiguration pSampleConfiguration, PSampleStreamingSe
     STATUS retStatus = STATUS_SUCCESS;
     RtcSessionDescriptionInit offerSessionDescriptionInit;
     BOOL locked = FALSE;
+    NullableBool canTrickle;
 
     CHK(pSampleConfiguration != NULL && pSignalingMessage != NULL, STATUS_NULL_ARG);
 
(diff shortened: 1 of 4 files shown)


samples/Common.c
```diff
@@ -214,6 +215,10 @@ STATUS handleOffer(PSampleConfiguration pSampleConfiguration, PSampleStreamingSe
 
     CHK_STATUS(deserializeSessionDescriptionInit(pSignalingMessage->payload, pSignalingMessage->payloadLen, &offerSessionDescriptionInit));
     CHK_STATUS(setRemoteDescription(pSampleStreamingSession->pPeerConnection, &offerSessionDescriptionInit));
+    canTrickle = canTrickleIceCandidates(pSampleStreamingSession->pPeerConnection);
+    /* cannot be null after setRemoteDescription */
+    CHECK(!NULLABLE_CHECK_EMPTY(canTrickle));
```
MushMal, line 220: Are we trying assert here or a negative flow check? Based on the comment it's an assert so CHECK is OK

src/include/com/amazonaws/kinesis/video/webrtcclient/Include.h
```diff
@@ -1337,6 +1337,19 @@ PUBLIC_API STATUS peerConnectionGetCurrentLocalDescription(PRtcPeerConnection, P
...
+ * The value is determined based on whether a remote description indicates support for trickle ICE. Prior to the completion
+ * of setRemoteDescription, this value is null.
+ *
+ * Reference: https://www.w3.org/TR/webrtc/#dom-rtcpeerconnection-cantrickleicecandidates
+ *
+ * @param[in] PRtcPeerConnection Initialized RtcPeerConnection
+ *
+ * @return NullableBool if not null, indicate whether remote support trickle ICE.
+ */
+PUBLIC_API NullableBool canTrickleIceCandidates(PRtcPeerConnection);
```
MushMal, line 1351: NullableBool is a structure and not an intrinsic type. Returning a structure is by copy is not a good idea in functions - in fact, if you search the entire codebase we have no occurrences. I recommend taking in a pointer and setting it perhaps.

src/source/PeerConnection/PeerConnection.c
```diff
@@ -1008,6 +1015,26 @@ STATUS restartIce(PRtcPeerConnection pPeerConnection)
     return retStatus;
 }
 
+PUBLIC_API NullableBool canTrickleIceCandidates(PRtcPeerConnection pPeerConnection)
+{
+    NullableBool canTrickle = {
```
MushMal, line 1020: This is not C syntax. Please use ANSI C syntax {FALSE, FALSE}

chehefen: which c standard are we complying with? this is compliant with c99 https://gcc.gnu.org/onlinedocs/gcc/Designated-Inits.html.

src/include/com/amazonaws/kinesis/video/webrtcclient/Include.h
chehefen: these are small structures whats bad about copying them? Im trying to stick to the public interfaces in w3c.

MushMal: It shouldn't be taken verbatim in my opinion. I would prefer the functions not to return entire structures. Though, in this case, I see your point.

src/source/PeerConnection/PeerConnection.c
MushMal: I would prefer C89 whenever possible. There is no reason to deviate from it. All of the initializers are anonymous in the codebase.

chehefen: but aws-c-io and aws-c-common is only c99. Once we move to them any effort we spent on c89 is wasted.

MushMal: I don't think this is an effort or we are wasting anything. I also don't think in this case you are actually getting any advantage of the named initialization params.

chehefen: isnt it more readable?

MushMal: There are many syntactic sugars that higher languages provide but I would try to be as conservative as possible - especially since it doesn't server any purpose. There is little or none in the codebase that won't allow compiling with c89 only - at least the core.

chehefen: but once we move to aws sdk, it wont be c89 anymore.

chehefen: fixed

Issue #434, opened 2020-05-07
Windows link error


wolfviking0:
Hi, I was able to build a version for kinesis for windows including all dependencies, ssl, web socket, ...

All my build are static libraries, but when I try to create my sample I have two undefined symbol :

```
kvsWebrtcClient.lib(Network.c.obj) : error LNK2019: unresolved external symbol getifaddrs referenced in function getLocalhostIpAddresses
kvsWebrtcClient.lib(Network.c.obj) : error LNK2019: unresolved external symbol freeifaddrs referenced in function getLocalhostIpAddresses
```

I have a feeling I am close to have everything but this two last symbol block me, any idea how I can fix that ? 
I know inside web socket there is some implementation of both function should I try to use them ?

Thanks

MushMal:
Niiiice!!! While the Windows platform is not officially supported in the KVS Embedded WebRTC SDK yet, most of the code is platform agnostic so you should be able to get it working with some elbow grease.

The first function does not exist for Windows platforms so you might consider using the equivalent https://docs.microsoft.com/en-us/windows/win32/api/iphlpapi/nf-iphlpapi-getadaptersaddresses?redirectedfrom=MSDN
in https://github.com/awslabs/amazon-kinesis-video-streams-webrtc-sdk-c/blob/10b2d49960384c6c5e1e4d97ea6f14a7bb11ed6f/src/source/Ice/Network.c#L20

The second one is used here https://github.com/awslabs/amazon-kinesis-video-streams-webrtc-sdk-c/blob/10b2d49960384c6c5e1e4d97ea6f14a7bb11ed6f/src/source/Ice/Network.c#L72

The usage of the API is such that the ownership of the buffer is on the caller so you could use MEMALLOC to allocate the buffer for the API and then release it with MEMFREE.

It would be awesome to see how you get this working and perhaps push a PR that we can incorporate?

wolfviking0:
@MushMal I cannot make any promise, I am not a window guy, but definitively I keep you posted as soon I have it running.

MushMal:
@wolfviking0 we are here to cooperate :) Might take a few iterations. Likely, we will need to massage the entire thing well before pushing into the master branch but it makes sense to preserve your work so after all we can incorporate the bits of it if needed

Sean-Der:
[...]

PR #465, opened 2020-05-16
[WIP] Building WEBRTC-SDK-C on Windows

wolfviking0:
Issue #434 

Few important note:
1 - The build on windows is static only
2 - On windows I install PgkConfig using msys pacman but I am not using it
3 - I updated some version of libraries to have a better compatibility with CMake
4 - In order to build fine I have manually change one line on the submodule producer
      the check package of websocket failed so I changed it from REQUIRE to QUIET


src/source/PeerConnection/PeerConnection.h
```diff
@@ -21,9 +21,9 @@ extern "C" {
 // A non-comprehensive list of valid JSON characters
 #define VALID_CHAR_SET_FOR_JSON         "0123456789ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz+/"
 
-#define ICE_CANDIDATE_JSON_TEMPLATE     (PCHAR) "{\"candidate\":\"candidate:%s\",\"sdpMid\":\"0\",\"sdpMLineIndex\":0}"
-
-#define MAX_ICE_CANDIDATE_JSON_LEN      (MAX_SDP_ATTRIBUTE_VALUE_LENGTH + STRLEN(ICE_CANDIDATE_JSON_TEMPLATE) + 1)
+#define ICE_CANDIDATE_JSON_TEMPLATE     (PCHAR) "{\"candidate\":\"candidate:%s\",\"sdpMid\":\"0\",\"sdpMLineIndex\":0}" // SIZE 59
+ 
+#define MAX_ICE_CANDIDATE_JSON_LEN      (MAX_SDP_ATTRIBUTE_VALUE_LENGTH + 59 + 1) // (MAX_SDP_ATTRIBUTE_VALUE_LENGTH + STRLEN(ICE_CANDIDATE_JSON_TEMPLATE) + 1)
```
Sean-Der, line 26: Are you worried about the cost of `STRLEN` at runtime? Would it be possible to do a SIZEOF instead, just so we don't need to hardcode the 59.

Sean-Der:
Really fantastic work @wolfviking0! Are you able to work on this PR, or would you like us to take it and fix things? Either works for us!

src/source/PeerConnection/PeerConnection.h
wolfviking0: Not at all but Windows is not happy using STRLEN because of is a not const expression

wolfviking0:
@Sean-Der I will not able to work too much on the pull request (mainly the week end only :()
In my original version I change a lot of the cmake because all the dependencies are separated, its why he took me sometime to have something buildable without changing too much the cmake.
If you can take it and fix things that will be genial, but because I am doing some test too if I found something I will just keep you in touch.

src/source/PeerConnection/PeerConnection.h
wolfviking0: https://github.com/wolfviking0/amazon-kinesis-video-streams-webrtc-sdk-c/blob/191754a7b0a15edeb5dd242523ebe9e3f73bd3d8/src/source/PeerConnection/PeerConnection.c#L317
This line failed on Windows if you don't remove the strlen

Sean-Der:
@wolfviking0 perfect that is what we will do :)

Starting Monday I will pull all your commits in (and squash them) and keep you as the author! Then is a separate commit will do all my work. Thank you so much for doing this, it is really exciting.

I also will add a builder to Travis to make sure this never regresses.

samples/Common.c
```diff
@@ -563,7 +563,9 @@ STATUS lookForSslCert(PSampleConfiguration* ppSampleConfiguration)
     STATUS retStatus = STATUS_SUCCESS;
     struct stat pathStat;
     UINT32 length = 0;
+#ifndef _WIN32
```
MushMal, line 566: We could use PIC primitives that are platform independent and can iterate over the directory. Not for this change though.

samples/kvsWebRTCClientMaster.c
```diff
@@ -1,4 +1,5 @@
 #include "Samples.h"
+#include <stdint.h>
```
MushMal, line 2: Please do all includes in a single header. In this case Samples.h

samples/kvsWebRTCClientViewer.c
```diff
@@ -1,4 +1,5 @@
 #include "Samples.h"
+#include <stdint.h>
```
MushMal, line 2: ditto

wolfviking0:
Another point to keep in mind, I am not a strong guy in windows but I think the socket library is a little bit different:
https://github.com/wolfviking0/amazon-kinesis-video-streams-webrtc-sdk-c/blob/191754a7b0a15edeb5dd242523ebe9e3f73bd3d8/src/source/Ice/SocketConnection.c#L388

I don't know if for windows errno vs WSAGetLastError are returning the same value:
https://docs.microsoft.com/en-us/windows/win32/winsock/error-codes-errno-h-errno-and-wsagetlasterror-2

src/source/Ice/Network.c
```diff
@@ -11,15 +11,77 @@ STATUS getLocalhostIpAddresses(PKvsIpAddress destIpList, PUINT32 pDestIpListLen,
     UINT32 ipCount = 0, destIpListLen;
     BOOL filterSet = TRUE;
 
+#ifdef _WIN32
+    DWORD rv, size;
```
MushMal, line 15: Opt for more descriptive names

src/source/Ice/Network.c
```diff
@@ -11,15 +11,77 @@ STATUS getLocalhostIpAddresses(PKvsIpAddress destIpList, PUINT32 pDestIpListLen,
...
     CHK(destIpList != NULL && pDestIpListLen != NULL, STATUS_NULL_ARG);
     CHK(*pDestIpListLen != 0, STATUS_INVALID_ARG);
-    CHK(getifaddrs(&ifaddr) != -1, STATUS_GET_LOCAL_IP_ADDRESSES_FAILED);
 
     destIpListLen = *pDestIpListLen;
+#ifdef _WIN32
+    rv = GetAdaptersAddresses(AF_UNSPEC, GAA_FLAG_INCLUDE_PREFIX, NULL, NULL, &size);
+    CHK(rv == ERROR_BUFFER_OVERFLOW, STATUS_GET_LOCAL_IP_ADDRESSES_FAILED);
+
+    adapter_addresses = (PIP_ADAPTER_ADDRESSES)malloc(size);
```
MushMal, line 32: MEMALLOC

src/source/Ice/Network.c
```diff
@@ -11,15 +11,77 @@ STATUS getLocalhostIpAddresses(PKvsIpAddress destIpList, PUINT32 pDestIpListLen,
...
+
+    rv = GetAdaptersAddresses(AF_UNSPEC, GAA_FLAG_INCLUDE_PREFIX, NULL, adapter_addresses, &size);
+    CHK(rv == ERROR_SUCCESS, STATUS_GET_LOCAL_IP_ADDRESSES_FAILED);
+    
+    for (aa = adapter_addresses; aa != NULL && ipCount < destIpListLen; aa = aa->Next) {
+
+        char ifa_name[BUFSIZ];
+        memset(ifa_name, 0, BUFSIZ);
```
MushMal, line 40: MEMSET

src/source/Ice/Network.c
```diff
@@ -11,15 +11,77 @@ STATUS getLocalhostIpAddresses(PKvsIpAddress destIpList, PUINT32 pDestIpListLen,
...
+    CHK(rv == ERROR_BUFFER_OVERFLOW, STATUS_GET_LOCAL_IP_ADDRESSES_FAILED);
+
+    adapter_addresses = (PIP_ADAPTER_ADDRESSES)malloc(size);
+
+    rv = GetAdaptersAddresses(AF_UNSPEC, GAA_FLAG_INCLUDE_PREFIX, NULL, adapter_addresses, &size);
+    CHK(rv == ERROR_SUCCESS, STATUS_GET_LOCAL_IP_ADDRESSES_FAILED);
+    
+    for (aa = adapter_addresses; aa != NULL && ipCount < destIpListLen; aa = aa->Next) {
+
+        char ifa_name[BUFSIZ];
```
MushMal, line 39: CHAR ifaName. Importantly, move this to the top with the other local variables declarations. Some compilers will choke on this as CHK (goto) can jump over the definitions

src/source/Ice/Network.c
```diff
@@ -11,15 +11,77 @@ STATUS getLocalhostIpAddresses(PKvsIpAddress destIpList, PUINT32 pDestIpListLen,
...
+        memset(ifa_name, 0, BUFSIZ);
+        WideCharToMultiByte(CP_ACP, 0, aa->FriendlyName, wcslen(aa->FriendlyName), ifa_name, BUFSIZ, NULL, NULL);
+
+        for (ua = aa->FirstUnicastAddress; ua != NULL; ua = ua->Next) {
+
+            if (filter != NULL) {
+                DLOGI("Callback set to allow network interface filtering");
```
MushMal, line 46: DLOGD

src/source/Ice/Network.c
```diff
@@ -11,15 +11,77 @@ STATUS getLocalhostIpAddresses(PKvsIpAddress destIpList, PUINT32 pDestIpListLen,
...
+                // The callback evaluates to a FALSE if the application is interested in black listing an interface
+                if (filter(customData, ifa_name) == FALSE) {
+                    filterSet = FALSE;
+                } else {
+                    filterSet = TRUE;
+                }
```
MushMal, line 52: filterSet = filter(customData, ifaName);

MushMal: In fact, you don't even need filterSet variable as its used only in the next statement so

if (filter(blah)) {
}

src/source/Ice/Network.c
```diff
@@ -11,15 +11,77 @@ STATUS getLocalhostIpAddresses(PKvsIpAddress destIpList, PUINT32 pDestIpListLen,
...
+            }
+
+            // If filter is set, ensure the details are collected for the interface
+            if (filterSet == TRUE) {
+                int family = ua->Address.lpSockaddr->sa_family;
```
MushMal, line 57: INT32 - declrate up top

src/source/Ice/Network.c
```diff
@@ -11,15 +11,77 @@ STATUS getLocalhostIpAddresses(PKvsIpAddress destIpList, PUINT32 pDestIpListLen,
...
+                    MEMCPY(destIpList[ipCount].address, &pIpv4Addr->sin_addr, IPV4_ADDRESS_LENGTH);
+                } else {
+                    destIpList[ipCount].family = KVS_IP_FAMILY_TYPE_IPV6;
+                    destIpList[ipCount].port = 0;
+
+                    pIpv6Addr = (struct sockaddr_in6*)(ua->Address.lpSockaddr);
+                    // Ignore link local: not very useful and will add work unnecessarily
+                    // Ignore site local: https://tools.ietf.org/html/rfc8445#section-5.1.1.1
+                    if (IN6_IS_ADDR_LINKLOCAL(&pIpv6Addr->sin6_addr) || IN6_IS_ADDR_SITELOCAL(&pIpv6Addr->sin6_addr)) {
+                        continue;
```
MushMal, line 73: Re-structure the loop to avoid control-flow keywords, especially continue; these are dangerous to use in more complex scenarios. For example, here you could check if (!blah) { MEMCPY}

src/source/Ice/Network.c
```diff
@@ -65,12 +127,19 @@ STATUS getLocalhostIpAddresses(PKvsIpAddress destIpList, PUINT32 pDestIpListLen,
...
+#endif
 
 CleanUp:
 
-    if (ifaddr != NULL) {
-        freeifaddrs(ifaddr);
+#ifdef _WIN32
+    if (adapter_addresses != NULL) {
+        free(adapter_addresses);
     }
```
MushMal, line 137: Initialize pAdapterAddresses to NULL, here, use SAFE_MEMFREE()

src/source/Ice/Network.c
```diff
@@ -11,15 +11,77 @@ STATUS getLocalhostIpAddresses(PKvsIpAddress destIpList, PUINT32 pDestIpListLen,
     UINT32 ipCount = 0, destIpListLen;
     BOOL filterSet = TRUE;
 
+#ifdef _WIN32
+    DWORD rv, size;
+    PIP_ADAPTER_ADDRESSES adapter_addresses, aa = NULL;
```
MushMal, line 16: we use camelCased names - we use underscored_names in C++

src/source/Ice/Network.c
```diff
@@ -164,10 +233,15 @@ STATUS createSocket(PKvsIpAddress pHostIpAddress, PKvsIpAddress pPeerAddress, KV
     pHostIpAddress->port = (UINT16) pHostIpAddress->family == KVS_IP_FAMILY_TYPE_IPV4 ? ipv4Addr.sin_port : ipv6Addr.sin6_port;
     *pSockFd = (INT32) sockfd;
 
+#ifdef _WIN32
+    unsigned long nonblock = 1;
```
MushMal, line 237: Use UINT32 and move the declaration up top with others.

src/source/PeerConnection/PeerConnection.h
```diff
@@ -21,9 +21,9 @@ extern "C" {
 // A non-comprehensive list of valid JSON characters
 #define VALID_CHAR_SET_FOR_JSON         "0123456789ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz+/"
 
-#define ICE_CANDIDATE_JSON_TEMPLATE     (PCHAR) "{\"candidate\":\"candidate:%s\",\"sdpMid\":\"0\",\"sdpMLineIndex\":0}"
-
-#define MAX_ICE_CANDIDATE_JSON_LEN      (MAX_SDP_ATTRIBUTE_VALUE_LENGTH + STRLEN(ICE_CANDIDATE_JSON_TEMPLATE) + 1)
+#define ICE_CANDIDATE_JSON_TEMPLATE     (PCHAR) "{\"candidate\":\"candidate:%s\",\"sdpMid\":\"0\",\"sdpMLineIndex\":0}" // SIZE 59
```
MushMal, line 24: Please don't use line comments after the macro definitions. Move up on top of the macro

src/source/PeerConnection/PeerConnection.h
MushMal: Usage of ARRAY_SIZE() is a compile-time directive which is the right way to go. NOTE: ARRAY_SIZE will return the null terminator so one more char to account for

src/source/PeerConnection/PeerConnection.h
wolfviking0: If I use the original code here windows give me this error:
```
Error (active)	E0059	function call is not allowed in a constant expression	kvsWebrtcClient	C:\Exp\aws\src\source\PeerConnection\PeerConnection.c	317	
```
That the reason why I hardcode the 59 base on the value of STRLEN

samples/Common.c
wolfviking0: For this one I am not sure what is the best way, we can use c++17 who provide filesystem API, or another way will be to use http://www.two-sdg.demon.co.uk/curbralan/code/dirent/dirent.c
who implement the functionality we need.
I think the second approach will be faster to do, @MushMal ?

wolfviking0:
Here the different configuration tested:
OSX High Sierra : STATIC / SHARED
OSX Catalina : STATIC / SHARED
LINUX 18.04 : STATIC / SHARED
LINUX 20.04 : STATIC / SHARED
WINDOWS : STATIC

I think there is maybe a small issue on the cmake on unix platform, it seems when I call cmake the first time the variable for the libwebsocket is not set properly
If I run a second time its set and the build succeed

Sean-Der:
Thank you so much for all your work on this @wolfviking0 I have cherry-picked your changes https://github.com/awslabs/amazon-kinesis-video-streams-webrtc-sdk-c/pull/472

It will probably be another week until this is done. I have hit quite a few failures in our test suite, so need to work through them.

thanks!

PR #471, opened 2020-05-20
Update lookForSslCert to be cross-platform

Sean-Der:
Use traverseDirectory from PiC so we are portable. This is part of the
Windows porting.

### DIFF
--- a/samples/Common.c
+++ b/samples/Common.c
@@ -563,16 +563,32 @@ CleanUp:
     return retStatus;
 }
 
+STATUS traverseDirectoryPEMFileScan(UINT64 customData, DIR_ENTRY_TYPES entryType, PCHAR fullPath, PCHAR fileName) {
+    UNUSED_PARAM(entryType);
+    UNUSED_PARAM(fullPath);
+
+    PCHAR certName = (PCHAR) customData;
+    UINT32 fileNameLen = STRLEN(fileName);
+
+    if(fileNameLen > ARRAY_SIZE(CA_CERT_PEM_FILE_EXTENSION) + 1 &&
+       (STRCMPI(CA_CERT_PEM_FILE_EXTENSION, &fileName[fileNameLen - ARRAY_SIZE(CA_CERT_PEM_FILE_EXTENSION) + 1]) == 0))
+    {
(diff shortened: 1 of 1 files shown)


Sean-Der:
If I have `/home/sean/foo/test.pem` the follow all work

```
AWS_KVS_CACERT_PATH=/home/sean/foo
AWS_KVS_CACERT_PATH=/home/sean/foo/test.pem
```

samples/Common.c
```diff
@@ -563,16 +563,29 @@ STATUS handleRemoteCandidate(PSampleStreamingSession pSampleStreamingSession, PS
     return retStatus;
 }
 
+STATUS traverseDirectoryPEMFileScan(UINT64 customData, DIR_ENTRY_TYPES entryType, PCHAR fullPath, PCHAR fileName) {
+    PCHAR certName = (PCHAR) customData;
+    INT32 fileNameLen = STRLEN(fileName);
```
MushMal, line 568: Strlen returns size_t which will generate warning n some compilers. Use uint32 with a cast perhaps? Or size_t

samples/Common.c
```diff
@@ -563,16 +563,29 @@ STATUS handleRemoteCandidate(PSampleStreamingSession pSampleStreamingSession, PS
...
+
+    if(fileNameLen > ARRAY_SIZE(CA_CERT_PEM_FILE_EXTENSION) + 1 &&
+       (STRCMPI(CA_CERT_PEM_FILE_EXTENSION, &fileName[fileNameLen - ARRAY_SIZE(CA_CERT_PEM_FILE_EXTENSION) + 1]) == 0))
+    {
+        certName[0] = FPATHSEPARATOR;
+        certName++;
+        STRCPY(certName, fileName);
```
MushMal, line 575: I think there is a way to terminate the iterative - was it returning error or something

samples/Common.c
Sean-Der: Fixed!

samples/Common.c
Sean-Der: I double checked and just a `CHK_STATUS` so I don't believe so? happy to make a backlog item out of it though!

PR #472, opened 2020-05-20
Windows Support

Sean-Der:
Code and build system changes needed


tst/SignalingApiFunctionalityTest.cpp
```diff
@@ -965,92 +965,92 @@ TEST_F(SignalingApiFunctionalityTest, iceRefreshEmulation)
     EXPECT_EQ(STATUS_SUCCESS, freeSignalingClient(&signalingHandle));
 }
 
-TEST_F(SignalingApiFunctionalityTest, iceRefreshEmulationAuthExpiration)
```
Sean-Der, line 968: @MushMal This uses a private API, what is the best thing to do here?

CMake/Dependencies/libopenssl-CMakeLists.txt
```diff
@@ -20,7 +24,7 @@ ExternalProject_Add(project_libopenssl
     GIT_TAG           OpenSSL_1_1_1g
     PREFIX            ${CMAKE_CURRENT_BINARY_DIR}/build
     CONFIGURE_COMMAND ${CONFIGURE_COMMAND}
-    BUILD_COMMAND     ${MAKE_EXE} -j 4
+    BUILD_COMMAND     ${MAKE_EXE}
```
MushMal, line 27: This might actually help with spontaneous build failures.

PR #477, opened 2020-05-21
Setting the global instrumented memory allocator size to 0 for each r…

MushMal:
…un of the tests so a single leak will not "contaminate" all of the consequent tests

*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/tst/WebRTCClientTestFixture.cpp
+++ b/tst/WebRTCClientTestFixture.cpp
@@ -36,6 +36,13 @@ void WebRtcClientTestBase::SetUp()
     mExpectedFrameCount = 0;
     mExpectedDroppedFrameCount = 0;
 
+    // We need to reset the current for every run of the test.
+    // This is done so a failed test that leaked will not cause
+    // failures for every other test that follows as the accounting
+    // is done in a global that has a lifespan tied with the process
+    // rather than the particular test run
+    gInstrumentedAllocatorsTotalAllocationSize = 0;
+    
     SET_INSTRUMENTED_ALLOCATORS();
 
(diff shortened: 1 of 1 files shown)


tst/WebRTCClientTestFixture.cpp
```diff
@@ -36,6 +36,13 @@ void WebRtcClientTestBase::SetUp()
     mExpectedFrameCount = 0;
     mExpectedDroppedFrameCount = 0;
 
+    // We need to reset the current for every run of the test.
+    // This is done so a failed test that leaked will not cause
+    // failures for every other test that follows as the accounting
+    // is done in a global that has a lifespan tied with the process
+    // rather than the particular test run
+    gInstrumentedAllocatorsTotalAllocationSize = 0;
```
chehefen, line 44: is this layer violation to directly change the variable like this.

MushMal: Yes, certainly a blatant violation.

As this is more or less unique situation with the consequent run of tests and it's unlike an actual application where we set the instrumented allocators at the start and reset right at the end, my thinking is that it doesn't warrant a public API and macro to reset the total allocation size. Though, I am open for suggestions. If you think we might have an actual use case where a public API is warranted then I can go ahead and change PIC

chehefen: I think resetInstrumentedAllocators should reset the gInstrumentedAllocatorsTotalAllocationSize to 0. If we can do this in the test, what can stop us from setting any other globals in the test.

MushMal: Indeed. I will modify PIC then

PR #479, opened 2020-05-21
SIGNALING_CONNECT_TIMEOUT increased from 5 to 60 seconds

mateuszboryn:
*Issue #, if available:*

*Description of changes:*


src/include/com/amazonaws/kinesis/video/webrtcclient/Include.h
```diff
@@ -522,7 +522,7 @@ extern "C" {
 /**
  * Default signaling connection establishment timeout
  */
-#define SIGNALING_CONNECT_TIMEOUT                                                   (5 * HUNDREDS_OF_NANOS_IN_A_SECOND)
+#define SIGNALING_CONNECT_TIMEOUT                                                   (60 * HUNDREDS_OF_NANOS_IN_A_SECOND)
```
MushMal, line 525: The main concern I have are with this change is that the connect APIs are synchronous and will block for 60 seconds now. This might not be desirable for most of the applications.

Perhaps a better approach is to assume the defaults (5 seconds) and have the application to be able to change those.

We also need to change the complete retry timeout. Currently, we have connect timeout set as 5 seconds (which is similar to what we have for RESTful API) and we have complete timeout set to 15 seconds.

Let me know what you think about configurability of these timeouts (this is similar to the Producer SDK where the application can change the default timeouts).

Matthammond96: Thank you, this resolved the issue I was facing!

doraemon-hub-art:
@mateuszboryn  
Thank you, sir, this method still works today.

After verification, due to network conditions in some regions, adjusting the timeout and re-verifying showed that the total signaling connection duration was approximately 5.2s to 5.4s. It seems the official default of 5s is still too strict.

PR #481, opened 2020-05-21
as master, detect DTLS close alert and issue peerConnectionState chan…

chehefen:
…ge to closed state.

*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/src/source/Dtls/Dtls.c
+++ b/src/source/Dtls/Dtls.c
@@ -54,6 +54,7 @@ STATUS dtlsTransmissionTimerCallback(UINT32 timerID, UINT64 currentTime, UINT64 
     if (SSL_is_init_finished(pDtlsSession->pSsl)) {
         DLOGD("DTLS init completed. Time taken %" PRIu64 " ms",
               (GETTIME() - pDtlsSession->dtlsSessionStartTime) / HUNDREDS_OF_NANOS_IN_A_MILLISECOND);
+        CHK_STATUS(dtlsSessionChangeState(pDtlsSession, CONNECTED));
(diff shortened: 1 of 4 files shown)


src/source/Dtls/Dtls.c
```diff
@@ -532,6 +561,32 @@ STATUS dtlsSessionPutApplicationData(PDtlsSession pDtlsSession, PBYTE pData, INT
...
+STATUS dtlsSessionOnStateChange(PDtlsSession pDtlsSession, UINT64 customData, DtlsSessionOnStateChange callbackFn)
+{
+    STATUS retStatus = STATUS_SUCCESS;
+
+    CHK(pDtlsSession != NULL && callbackFn != NULL, STATUS_NULL_ARG);
+
+    pDtlsSession->dtlsSessionCallbacks.stateChangeFn = callbackFn;
+    pDtlsSession->dtlsSessionCallbacks.stateChangeFnCustomData = customData;
+
+    CleanUp:
```
MushMal, line 586: extra \t

src/source/Dtls/Dtls.h
```diff
@@ -38,12 +38,25 @@ typedef enum {
    SRTP_PROFILE_AES128_CM_HMAC_SHA1_32 = SRTP_AES128_CM_SHA1_32,
 } SRTP_PROFILE;
 
-// Callback that is fired when Dtls Server wishes to send packet
+typedef enum {
+    NEW,
```
MushMal, line 42: Are we going to be exposing/handling any other state other than CLOSED? In which case we need some mapping between these states and the PC states.

chehefen: the mapping is done in the dtls state changed callback.

PR #483, opened 2020-05-22
Add SDP logging support through environment variable

disa6302:
*Issue #, if available:*

*Description of changes:*
- Added an env `DEBUG_LOG_SDP` option to get SDPs displayed
- This will useful when customers want to view the SDPs on the local machine without the need to add debug lines on their own
- Will also be useful for us to debug SDP related issues

### DIFF
--- a/src/source/PeerConnection/PeerConnection.c
+++ b/src/source/PeerConnection/PeerConnection.c
@@ -844,6 +844,10 @@ STATUS setRemoteDescription(PRtcPeerConnection pPeerConnection, PRtcSessionDescr
     CHK_STATUS(setTransceiverPayloadTypes(pKvsPeerConnection->pCodecTable, pKvsPeerConnection->pRtxTable, pKvsPeerConnection->pTransceievers));
     CHK_STATUS(setReceiversSsrc(pSessionDescription, pKvsPeerConnection->pTransceievers));
 
+    if (NULL != getenv(DEBUG_LOG_SDP)) {
+        DLOGD("REMOTE_SDP:%s\n", pSessionDescriptionInit->sdp);
+    }
+
 CleanUp:
 
     LEAVES();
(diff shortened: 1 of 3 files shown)


src/source/PeerConnection/PeerConnection.c
```diff
@@ -904,12 +910,19 @@ STATUS setLocalDescription(PRtcPeerConnection pPeerConnection, PRtcSessionDescri
 {
     ENTERS();
     STATUS retStatus = STATUS_SUCCESS;
+    PCHAR sdpPrint = getenv(DEBUG_LOG_SDP);
```
MushMal, line 913: Remove

src/source/PeerConnection/PeerConnection.c
```diff
@@ -844,6 +844,12 @@ STATUS setRemoteDescription(PRtcPeerConnection pPeerConnection, PRtcSessionDescr
     CHK_STATUS(setTransceiverPayloadTypes(pKvsPeerConnection->pCodecTable, pKvsPeerConnection->pRtxTable, pKvsPeerConnection->pTransceievers));
     CHK_STATUS(setReceiversSsrc(pSessionDescription, pKvsPeerConnection->pTransceievers));
 
+    if (getenv(DEBUG_LOG_SDP)) {
```
MushMal, line 847: Recommend if (NULL != getenv(DEBUG_LOG_SDP)

Also, think whether you can make a single macro that does the ifs and printouts. It should obviously, take two params then - the string denoting whether it's local or remote and the string that's the sdp

src/source/PeerConnection/PeerConnection.c
```diff
@@ -844,6 +844,12 @@ STATUS setRemoteDescription(PRtcPeerConnection pPeerConnection, PRtcSessionDescr
...
+        DLOGD("******************REMOTE SDP********************");
+        DLOGD("%s\n", pSessionDescriptionInit->sdp);
+        DLOGD("************************************************");
```
zhiyua-git, line 850: Is it possible that we will have multi-thread case causing the log lines to be interacting each other? For example:
```
******************REMOTE SDP********************
******************LOCAL SDP********************
(local SDP)
(remote SDP)
************************************************
************************************************
```

src/source/PeerConnection/PeerConnection.c
disa6302: Done

src/source/PeerConnection/PeerConnection.c
disa6302: MACRO is fine. But, I dont see a need for it here. I will let it be an if statement

src/source/PeerConnection/PeerConnection.c
disa6302: Potentially. I changed it to print LOCAL_SDP,...etc along with the sdp printing line.

PR #484, opened 2020-05-22
add closePeerConnection api to trigger DTLS close.

chehefen:
*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/src/source/Dtls/Dtls.c
+++ b/src/source/Dtls/Dtls.c
@@ -55,6 +55,7 @@ STATUS dtlsTransmissionTimerCallback(UINT32 timerID, UINT64 currentTime, UINT64 
         DLOGD("DTLS init completed. Time taken %" PRIu64 " ms",
               (GETTIME() - pDtlsSession->dtlsSessionStartTime) / HUNDREDS_OF_NANOS_IN_A_MILLISECOND);
         CHK_STATUS(dtlsSessionChangeState(pDtlsSession, CONNECTED));
+        ATOMIC_STORE_BOOL(&pDtlsSession->sslInitFinished, TRUE);
         CHK(FALSE, STATUS_TIMER_QUEUE_STOP_SCHEDULING);
     }
 
@@ -309,6 +310,7 @@ STATUS createDtlsSession(PDtlsSessionCallbacks pDtlsSessionCallbacks, TIMER_QUEU
     pDtlsSession->sslLock = MUTEX_CREATE(TRUE);
     pDtlsSession->state = NEW;
     ATOMIC_STORE_BOOL(&pDtlsSession->isStarted, FALSE);
+    ATOMIC_STORE_BOOL(&pDtlsSession->sslInitFinished, FALSE);
 
     pDtlsSession->dtlsSessionCallbacks = *pDtlsSessionCallbacks;
 
@@ -480,7 +482,7 @@ STATUS dtlsSessionProcessPacket(PDtlsSession pDtlsSession, PBYTE pData, PINT32 p
 {
     ENTERS();
     STATUS retStatus = STATUS_SUCCESS;
-    BOOL locked = FALSE;
+    BOOL locked = FALSE, isClosed = FALSE;
     INT32 sslRet = 0, sslErr;
     INT32 dataLen = 0;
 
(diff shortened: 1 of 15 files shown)


src/source/Dtls/Dtls.c
```diff
@@ -539,6 +546,7 @@ STATUS dtlsSessionPutApplicationData(PDtlsSession pDtlsSession, PBYTE pData, INT
     SIZE_T pending;
 
     MUTEX_LOCK(pDtlsSession->sslLock);
+    CHK(!ATOMIC_LOAD_BOOL(&pDtlsSession->shutdown), retStatus);
```
MushMal, line 549: Move this above the lock

src/source/Dtls/Dtls.c
```diff
@@ -561,6 +569,33 @@ STATUS dtlsSessionPutApplicationData(PDtlsSession pDtlsSession, PBYTE pData, INT
...
 
+STATUS dtlsSessionShutdown(PDtlsSession pDtlsSession)
+{
+    STATUS retStatus = STATUS_SUCCESS;
+    BOOL locked = FALSE;
+
+    CHK(pDtlsSession != NULL, STATUS_NULL_ARG);
+
+    MUTEX_LOCK(pDtlsSession->sslLock);
+    CHK(!ATOMIC_LOAD_BOOL(&pDtlsSession->shutdown), retStatus);
```
MushMal, line 580: Same - move above the lock acquisition

src/source/Dtls/Dtls.c
```diff
@@ -561,6 +569,33 @@ STATUS dtlsSessionPutApplicationData(PDtlsSession pDtlsSession, PBYTE pData, INT
...
+    CHK(ATOMIC_LOAD_BOOL(&pDtlsSession->sslInitFinished), retStatus);
```
MushMal, line 581: Perhaps this one also needs to be moved above the lock

src/source/Dtls/Dtls.c
```diff
@@ -561,6 +569,33 @@ STATUS dtlsSessionPutApplicationData(PDtlsSession pDtlsSession, PBYTE pData, INT
...
+
+    SSL_shutdown(pDtlsSession->pSsl);
+    CHK_STATUS(dtlsCheckOutgoingDataBuffer(pDtlsSession));
+
+    ATOMIC_STORE_BOOL(&pDtlsSession->shutdown, TRUE);
```
MushMal, line 586: I think this should go before CHK_STATUS(dtlsCheckOutgoingDataBuffer()) call

src/source/Dtls/Dtls.c
```diff
@@ -561,6 +569,33 @@ STATUS dtlsSessionPutApplicationData(PDtlsSession pDtlsSession, PBYTE pData, INT
...
+
+    CHK_STATUS(dtlsSessionChangeState(pDtlsSession, CLOSED));
```
MushMal, line 588: Are we calling this API with the lock acquired? This is internal so it should be OK

src/source/PeerConnection/PeerConnection.c
```diff
@@ -1045,6 +1045,25 @@ STATUS restartIce(PRtcPeerConnection pPeerConnection)
...
 
+STATUS closePeerConnection(PRtcPeerConnection pPeerConnection)
+{
+    ENTERS();
+    STATUS retStatus = STATUS_SUCCESS;
+    PKvsPeerConnection pKvsPeerConnection = (PKvsPeerConnection) pPeerConnection;
+
+    CHK(pKvsPeerConnection != NULL, STATUS_NULL_ARG);
+
+    CHK_STATUS(dtlsSessionShutdown(pKvsPeerConnection->pDtlsSession));
```
MushMal, line 1056: Should this be CHK_LOG_ERR?

src/source/Dtls/Dtls.c
chehefen: I cant, what could happen is the check passes and the thread block at the lock then shutdown happens.

src/source/Dtls/Dtls.c
chehefen: same reason as before

src/source/Dtls/Dtls.c
chehefen: same reason.

src/source/PeerConnection/PeerConnection.c
chehefen: sry, iceAgentShutdown should be CHK_STATUS

src/source/Dtls/Dtls.c
MushMal: I see, no matter how you look at it, it's impossible to properly handle this for all cases. The situation is that if another thread is already shutting down the lock will be released after the shutdown. The main issue is to guard at the higher-layer logic against the object being freed before the check for the atomic.

PR #473, opened 2020-05-20
WIP

lherman-cs:
* Add a custom test runner to retry on failed tests
* Relax travis timeout to accommodate retries

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.

PR #495, opened 2020-05-26
Guard PRs from forks

lherman-cs:
Travis doesn't allow sharing private keys to PR that are from forks. Since our tests rely on these private keys, there's no need to run tests from forks. 

By exiting early, this will allow more bandwidth to run other tests.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -14,6 +14,15 @@ addons:
     packages:
       - gdb
 
+before_install:
+  # Guard PRs from forks because they won't be able to get credentials, which will cause the tests to always fail
+  - >
+      [[ -z $AWS_ACCESS_KEY_ID || -z $AWS_SECRET_ACCESS_KEY ]] && 
+      echo "Creating a PR from a fork is not allowed. Travis doesn't allow sharing secret keys with fork builds." &&
+      travis_terminate 1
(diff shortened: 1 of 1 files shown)


Sean-Der:
@lherman-cs I don't think we need this (but I could be wrong)

In all our tests I added a guard if access/secret key is not set, and then will skip those tests. The only downside is lower test coverage, but as long as users aren't editing code in those areas I am not worried.

lherman-cs:
@Sean-Der Ah I see. I actually just had this issue recently in this PR, https://github.com/awslabs/amazon-kinesis-video-streams-webrtc-sdk-c/pull/473. So, we might have missed some of them?

MushMal:
> @lherman-cs I don't think we need this (but I could be wrong)
> 
> In all our tests I added a guard if access/secret key is not set, and then will skip those tests. The only downside is lower test coverage, but as long as users aren't editing code in those areas I am not worried.

Yeah, this is one thing we discussed earlier. We should be due diligent when looking at public PRs in terms of test passing. The tests should be designed to work with and without the credentials. If we feel that the changes might impact something that will require credentials then the best way would be to re-submit the job to make sure the tests execute with the credentials

lherman-cs:
That makes sense. In that case, this PR is not necessary then. I'm closing it.

PR #516, opened 2020-06-03
do not block media thread while doing freeStreamingSession

chehefen:
*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/samples/Common.c
+++ b/samples/Common.c
@@ -785,8 +785,8 @@ STATUS sessionCleanupWait(PSampleConfiguration pSampleConfiguration)
                 // swap with last element and decrement count
                 pSampleConfiguration->streamingSessionCount--;
                 pSampleConfiguration->sampleStreamingSessionList[i] = pSampleConfiguration->sampleStreamingSessionList[pSampleConfiguration->streamingSessionCount];
-                CHK_STATUS(freeSampleStreamingSession(&pSampleStreamingSession));
                 ATOMIC_STORE_BOOL(&pSampleConfiguration->updatingSampleStreamingSessionList, FALSE);
+                CHK_STATUS(freeSampleStreamingSession(&pSampleStreamingSession));
             }
         }
 


.travis.yml
```diff
@@ -15,10 +15,10 @@ addons:
       - gdb
 
 script:
-  - export AWS_KVS_LOG_LEVEL=3
+  - export AWS_KVS_LOG_LEVEL=2
```
MushMal, line 18: Unfortunately, we are runnign against max log file size with Travis.

.travis.yml
```diff
@@ -15,10 +15,10 @@ addons:
...
   - make
   - ulimit -c unlimited -S
-  - timeout --signal=SIGABRT 60m ./tst/webrtc_client_test
+  - timeout --signal=SIGABRT 60m ./tst/webrtc_client_test --gtest_filter="PeerConnectionFunctionalityTest.*"
```
MushMal, line 21: Oh, I see, I think this is for testing purposes and you are going to revert this right/

src/source/Dtls/Dtls.c
```diff
@@ -508,7 +508,7 @@ STATUS dtlsSessionProcessPacket(PDtlsSession pDtlsSession, PBYTE pData, PINT32 p
         LOG_OPENSSL_ERROR("SSL_read");
     }
 
-    if (!ATOMIC_LOAD_BOOL(&pDtlsSession->sslInitFinished)) {
+    if (ATOMIC_LOAD_BOOL(&pDtlsSession->sslInitFinished)) {
```
MushMal: The if statement looks weird. I am not sure I understand this. So, we are saying that if the initialization is finished we check the outgoing data buffer. However, if it's not then we set the data length? Can you clarify the logic here?

src/source/Dtls/Dtls.c
```diff
@@ -510,11 +510,11 @@ STATUS dtlsSessionProcessPacket(PDtlsSession pDtlsSession, PBYTE pData, PINT32 p
 
     if (!ATOMIC_LOAD_BOOL(&pDtlsSession->sslInitFinished)) {
         CHK_STATUS(dtlsCheckOutgoingDataBuffer(pDtlsSession));
-    } else {
-        // if dtls handshake is done, and SSL_read did not fail, then sslRet and number of sctp bytes read
-        dataLen = sslRet < 0 ? 0 : sslRet;
     }
 
+    // if dtls handshake is done, and SSL_read did not fail, then sslRet and number of sctp bytes read
```
MushMal, line 515: OK, my previous comment is void with the new change

Issue #508, opened 2020-06-02
[BUG] Regression - Data Channel No Longer Receiving Incoming Messages


suggestedfixes:
**Describe the bug**
In Common.c, we can usually print the message from channel message callback
https://github.com/awslabs/amazon-kinesis-video-streams-webrtc-sdk-c/blob/master/samples/Common.c#L21
However, when using https://awslabs.github.io/amazon-kinesis-video-streams-webrtc-sdk-js/examples/index.html, we can no longer receive data channel messages.

**To Reproduce**
1. Build KVSMaster fb2c393
2. Set Log Level to INFO
3. Use https://awslabs.github.io/amazon-kinesis-video-streams-webrtc-sdk-js/examples/index.html to test messages
4. No messages show up in log 

**Expected behavior**
Master should receive data channel messages

**Desktop (please complete the following information):**
 - OS: win 10
 - Browser chrome
 - Version 83.0.4103.61

**Additional context**

MushMal:
Thanks for reporting and the suggestions. This is the higher priority issue for us to look.

PR #519, opened 2020-06-04
add unit test for sending data channel message after DTLS handshake c…

chehefen:
…ompletion

*Issue #, if available:* catch #508

*Description of changes:*

### DIFF
--- a/samples/Common.c
+++ b/samples/Common.c
@@ -230,7 +230,7 @@ STATUS handleOffer(PSampleConfiguration pSampleConfiguration, PSampleStreamingSe
         while (!ATOMIC_LOAD_BOOL(&pSampleStreamingSession->candidateGatheringDone)) {
             CHK_WARN(!ATOMIC_LOAD_BOOL(&pSampleStreamingSession->terminateFlag), STATUS_OPERATION_TIMED_OUT,
                      "application terminated and candidate gathering still not done");
-            CVAR_WAIT(pSampleConfiguration->cvar, pSampleConfiguration->sampleConfigurationObjLock, 5 * HUNDREDS_OF_NANOS_IN_A_SECOND);
+            CVAR_WAIT(pSampleConfiguration->cvar, pSampleConfiguration->sampleConfigurationObjLock, 500 * HUNDREDS_OF_NANOS_IN_A_MILLISECOND);
         }
 
         MUTEX_UNLOCK(pSampleConfiguration->sampleConfigurationObjLock);
--- a/tst/DataChannelFunctionalityTest.cpp
+++ b/tst/DataChannelFunctionalityTest.cpp
@@ -92,6 +92,78 @@ TEST_F(DataChannelFunctionalityTest, createDataChannel_Disconnected)
     ASSERT_EQ(1u, remoteOpen.channels.at("Answer PeerConnection"));
 }
 
+TEST_F(DataChannelFunctionalityTest, dataChannelSendRecvMessageAfterDtlsCompleted)
+{
+    RtcConfiguration configuration;
+    PRtcPeerConnection offerPc = NULL, answerPc = NULL;
+    PRtcDataChannel pOfferDataChannel = nullptr, pAnswerDataChannel = nullptr;
+    SIZE_T pOfferRemoteDataChannel = 0, pAnswerRemoteDataChannel = 0;
+    SIZE_T datachannelLocalOpenCount = 0, msgCount = 0;
+    BOOL dtlsCompleted = FALSE;
+
+    MEMSET(&configuration, 0x00, SIZEOF(RtcConfiguration));
+
+    auto onDataChannel = [](UINT64 customData, PRtcDataChannel pRtcDataChannel) {
+        ATOMIC_STORE((PSIZE_T) customData, (SIZE_T) pRtcDataChannel);
+    };
+
+    auto dataChannelOnOpenCallback = [](UINT64 customData, PRtcDataChannel pDataChannel) {
+        UNUSED_PARAM(pDataChannel);
+        ATOMIC_INCREMENT((PSIZE_T) customData);
+    };
+
+    auto dataChannelOnMessageCallback = [](UINT64 customData, PRtcDataChannel pDataChannel, BOOL isBinary, PBYTE pMsg, UINT32 pMsgLen) {
+        UNUSED_PARAM(pDataChannel);
+        UNUSED_PARAM(isBinary);
+        if (STRNCMP((PCHAR) pMsg, TEST_DATA_CHANNEL_MESSAGE, pMsgLen) == 0) {
+            ATOMIC_INCREMENT((PSIZE_T) customData);
+        }
+    };
+
+    EXPECT_EQ(createPeerConnection(&configuration, &offerPc), STATUS_SUCCESS);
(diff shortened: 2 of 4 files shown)


tst/DataChannelFunctionalityTest.cpp
```diff
@@ -92,6 +92,78 @@ TEST_F(DataChannelFunctionalityTest, createDataChannel_Disconnected)
     ASSERT_EQ(1u, remoteOpen.channels.at("Answer PeerConnection"));
 }
 
+TEST_F(DataChannelFunctionalityTest, dataChannelSendRecvMessageAfterDtlsCompleted)
+{
+    RtcConfiguration configuration;
+    PRtcPeerConnection offerPc = NULL, answerPc = NULL;
+    PRtcDataChannel pOfferDataChannel = nullptr, pAnswerDataChannel = nullptr;
+    SIZE_T pOfferRemoteDataChannel = 0, pAnswerRemoteDataChannel = 0;
```
MushMal, line 100: Naming: these aren't pointers - suggest dropping 'p' prefix

tst/DataChannelFunctionalityTest.cpp
```diff
@@ -92,6 +92,78 @@ TEST_F(DataChannelFunctionalityTest, createDataChannel_Disconnected)
...
+
+    auto dataChannelOnMessageCallback = [](UINT64 customData, PRtcDataChannel pDataChannel, BOOL isBinary, PBYTE pMsg, UINT32 pMsgLen) {
+        UNUSED_PARAM(pDataChannel);
+        UNUSED_PARAM(isBinary);
+        if (STRNCMP((PCHAR) pMsg, TEST_DATA_CHANNEL_MESSAGE, pMsgLen) == 0) {
+            ATOMIC_INCREMENT((PSIZE_T) customData);
+        }
+    };
+
+    EXPECT_EQ(createPeerConnection(&configuration, &offerPc), STATUS_SUCCESS);
```
MushMal, line 123: Minor: These macros expect the first param to be the expected value and the second the evaluated one

tst/DataChannelFunctionalityTest.cpp
```diff
@@ -92,6 +92,78 @@ TEST_F(DataChannelFunctionalityTest, createDataChannel_Disconnected)
...
+    EXPECT_EQ(dataChannelOnOpen(pAnswerDataChannel, (UINT64) &datachannelLocalOpenCount, dataChannelOnOpenCallback), STATUS_SUCCESS);
+
+    EXPECT_EQ(dataChannelOnMessage(pOfferDataChannel, (UINT64) &msgCount, dataChannelOnMessageCallback), STATUS_SUCCESS);
+    EXPECT_EQ(dataChannelOnMessage(pAnswerDataChannel, (UINT64) &msgCount, dataChannelOnMessageCallback), STATUS_SUCCESS);
+
+    EXPECT_EQ(connectTwoPeers(offerPc, answerPc), TRUE);
+
+    // Busy wait until remote channel open and dtls completed
+    for (auto i = 0; i <= 100 && (dtlsSessionIsInitFinished(((PKvsPeerConnection) offerPc)->pDtlsSession, &dtlsCompleted) || ATOMIC_LOAD(&pOfferRemoteDataChannel) == 0); i++) {
+        THREAD_SLEEP(HUNDREDS_OF_NANOS_IN_A_SECOND);
```
MushMal, line 143: Too long? Perhaps wait for a few millis?

tst/DataChannelFunctionalityTest.cpp
```diff
@@ -92,6 +92,78 @@ TEST_F(DataChannelFunctionalityTest, createDataChannel_Disconnected)
...
+
+    EXPECT_EQ(dataChannelSend(
+            (PRtcDataChannel) ATOMIC_LOAD(&pOfferRemoteDataChannel),
+            FALSE,
+            (PBYTE) TEST_DATA_CHANNEL_MESSAGE,
+            STRLEN(TEST_DATA_CHANNEL_MESSAGE)), STATUS_SUCCESS);
+
+    /* wait until the channel message is received */
+    for (auto i = 0; i <= 5 && ATOMIC_LOAD(&msgCount) == 0; i++) {
+        THREAD_SLEEP(HUNDREDS_OF_NANOS_IN_A_SECOND);
```
MushMal, line 157: same

tst/DataChannelFunctionalityTest.cpp
chehefen: they are actually storing pointers. using size_t because of atomic operation.

samples/Common.c
```diff
@@ -230,7 +230,7 @@ STATUS handleOffer(PSampleConfiguration pSampleConfiguration, PSampleStreamingSe
         while (!ATOMIC_LOAD_BOOL(&pSampleStreamingSession->candidateGatheringDone)) {
             CHK_WARN(!ATOMIC_LOAD_BOOL(&pSampleStreamingSession->terminateFlag), STATUS_OPERATION_TIMED_OUT,
                      "application terminated and candidate gathering still not done");
-            CVAR_WAIT(pSampleConfiguration->cvar, pSampleConfiguration->sampleConfigurationObjLock, 5 * HUNDREDS_OF_NANOS_IN_A_SECOND);
+            CVAR_WAIT(pSampleConfiguration->cvar, pSampleConfiguration->sampleConfigurationObjLock, 50 * HUNDREDS_OF_NANOS_IN_A_MILLISECOND);
```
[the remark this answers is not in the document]
chehefen: yes it happens that candidate gathering is complete and signal the cvar before we even go block here.

tst/DataChannelFunctionalityTest.cpp
MushMal: yeah, understood that, though the type is size-t; really a nit

tst/DataChannelFunctionalityTest.cpp
chehefen: it needs to go all the way to dtls complete. It would be around 1s even in the good case.

tst/DataChannelFunctionalityTest.cpp
chehefen: I dont really think it's gonna make much difference.

samples/Common.c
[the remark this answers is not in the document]
MushMal: Oh, understand

tst/DataChannelFunctionalityTest.cpp
chehefen: this is already prevalent in the test, would be better to fix them all at once in a separate pr.

tst/DataChannelFunctionalityTest.cpp
MushMal: hmm... there is no network involved but seems that some devices would need enough time to generate the asymmetric crypto. OK

MushMal: Still, though, you are busy looping/iterating so can pick up earlier if it resolves earlier.

tst/DataChannelFunctionalityTest.cpp
MushMal: Not for the test

tst/DataChannelFunctionalityTest.cpp
MushMal: yeah of course

PR #525, opened 2020-06-05
malloc video and audio buffer first before realloc them

chehefen:
*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/samples/kvsWebRTCClientMaster.c
+++ b/samples/kvsWebRTCClientMaster.c
@@ -2,6 +2,9 @@
 
 extern PSampleConfiguration gSampleConfiguration;
 
+#define DEFAULT_AUDIO_BUFFER_SIZE               500
+#define DEFAULT_VIDEO_BUFFER_SIZE               120 * 1024
+
 // #define VERBOSE
 
 INT32 main(INT32 argc, CHAR *argv[])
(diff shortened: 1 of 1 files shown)


samples/kvsWebRTCClientMaster.c
```diff
@@ -35,6 +38,10 @@ INT32 main(INT32 argc, CHAR *argv[])
     pSampleConfiguration->receiveAudioVideoSource = sampleReceiveAudioFrame;
     pSampleConfiguration->onDataChannel = onDataChannel;
     pSampleConfiguration->mediaType = SAMPLE_STREAMING_AUDIO_VIDEO;
+    pSampleConfiguration->videoBufferSize = DEFAULT_VIDEO_BUFFER_SIZE;
+    pSampleConfiguration->audioBufferSize = DEFAULT_AUDIO_BUFFER_SIZE;
+    pSampleConfiguration->pVideoFrameBuffer = malloc(DEFAULT_VIDEO_BUFFER_SIZE);
```
MushMal, line 43: If we use malloc we won't be able to use instrumentation. What's our samples strategy around the allocation etc? I know we tried to de-PIC the samples which I don't think is a good idea when it comes to issues like these

MushMal: Importantly, the realloc can alloc - just make sure the buffers are set to NULL

PR #548, opened 2020-06-14
Introducing clang formatting in the sdk

### LINKED ISSUE #120: Add clang-format to travis
Also add a simple `.clang-format` the teams agrees upon.

disa6302:
*Issue #, if available:*
#120 
*Description of changes:*
- Added clang format spec
- Included a script to clang format a file, all files in project, a directory and git diff output only
- Included a script to run clang format check locally.

To-do:
Later, https://github.com/marketplace/actions/clang-format-check can be used to enforce the check for every commit 

Resolves #120 


MushMal:
It's hard to see what this evaluates to :). I assume this runs clean on the existing codebase?

zhuker:
just my two cents
clang format is needed
+1 to this PR

disa6302:
> It's hard to see what this evaluates to :). I assume this runs clean on the existing codebase?

This does not run clean because a few things are not consistent across the codebase. I have added the formatted files as well. A pretty huge PR because I ran the format on all the files.

disa6302:
> just my two cents
> clang format is needed
> +1 to this PR

@zhuker , you are right. It is needed. Sorry this went stale for a while.

disa6302:
> @disa6302 I think we should include the format check in this PR. Otherwise, it would be less meaningful since we might forget to run it before submitting a PR.

@lherman-cs , I will add the travis-ci portion in another PR once we confirm that the options are acceptable.

zhuker:
> This does not run clean because a few things are not consistent across the codebase. I have added the formatted files as well. A pretty huge PR because I ran the format on all the files.

@disa6302 @MushMal 

Yeah i tried formatting and got the same - style is inconsistent across the codebase
I would modify clang format to accommodate for the most common case, meaning - `.clangformat` which produces the smallest formatting PR should win

disa6302:
> > This does not run clean because a few things are not consistent across the codebase. I have added the formatted files as well. A pretty huge PR because I ran the format on all the files.
> 
> @disa6302 @MushMal
> 
> Yeah i tried formatting and got the same - style is inconsistent across the codebase
> I would modify clang format to accommodate for the most common case, meaning - `.clangformat` which produces the smallest formatting PR should win

Agreed. The clang format in this PR leads to the minimal changes I can see.

.clang-format
```diff
@@ -0,0 +1,77 @@
...
+AllowShortLoopsOnASingleLine: false
+AlwaysBreakAfterDefinitionReturnType: None
+AlwaysBreakAfterReturnType: None
+AlwaysBreakBeforeMultilineStrings: false
+BinPackArguments: true
+BinPackParameters: true
+BraceWrapping:
+  AfterCaseLabel:  false
+  AfterClass:      false
+  AfterControlStatement: false
```
zhuker, line 25: my CLion complains that it should be `  AfterControlStatement: "false"` (string not bool)

disa6302: Interesting. Let me look into it.

samples/discoverNatBehavior.c
```diff
@@ -1,36 +1,38 @@
...
             interfaceName = argv[++i];
             i++;
-        } else if (STRCMP(argv[i], STUN_HOSTNAME_PARAM) == 0) {
+        }
+        else if (STRCMP(argv[i], STUN_HOSTNAME_PARAM) == 0) {
             stunHostname = argv[++i];
             i++;
-        } else {
+        }
+        else {
```
zhuker, line 35: if you set `BreakBeforeBraces: WebKit` in .clang-format this `else` will not be formatted

.clang-format
```diff
@@ -0,0 +1,77 @@
...
+  AfterFunction:   false
+  AfterStruct:     false
+  AfterUnion:      false
+  AfterExternBlock: false
+  BeforeCatch:     false
+  BeforeElse:      false
+  IndentBraces:    false
+  SplitEmptyFunction: false
+BreakBeforeBinaryOperators: None
+BreakBeforeBraces: Stroustrup
```
zhuker, line 36: less formatting on existing code base if BreakBeforeBraces: WebKit

.clang-format
disa6302: @zhuker , 

I do not see any problem on my CLion.

.clang-format
disa6302: Done! Thanks!

samples/Samples.h
```diff
@@ -6,31 +6,31 @@ Shared include file for the samples
...
 #include <com/amazonaws/kinesis/video/webrtcclient/Include.h>
 
-#define NUMBER_OF_H264_FRAME_FILES                                              403
-#define NUMBER_OF_OPUS_FRAME_FILES                                              618
-#define DEFAULT_FPS_VALUE                                                       25
-#define DEFAULT_MAX_CONCURRENT_STREAMING_SESSION                                10
+#define NUMBER_OF_H264_FRAME_FILES               403
+#define NUMBER_OF_OPUS_FRAME_FILES               618
+#define DEFAULT_FPS_VALUE                        25
+#define DEFAULT_MAX_CONCURRENT_STREAMING_SESSION 10
```
zhuker, line 18: `AlignConsecutiveAssignments: true` may prevent this formatting

disa6302: It will. But I am going to let it stay.

.travis.yml
```diff
@@ -176,6 +176,20 @@ matrix:
         - .github/build_windows.bat
         - cd build/tst && ./webrtc_client_test.exe --gtest_filter="-DataChannelFunctionalityTest.*:IceApiTest.*:IceFunctionalityTest.*:PeerConnectionFunctionalityTest.*:SignalingApiFunctionalityTest.*:TurnConnectionFunctionalityTest.*:RtpFunctionalityTest.marshallUnmarshallH264Data:RtpFunctionalityTest.packingUnpackingVerifySameH264Frame"
 
+    - name: clang-format Check
```
lherman-cs, line 179: why are there 2 of this?

clang-format.sh
```diff
@@ -0,0 +1,39 @@
+#!/bin/bash
```
lherman-cs, line 1: I think I'm missing something. But, why do we need `check-clang.sh` when we have this script? If `check-clang.sh` is meant to be for checking `clang-format` existence, would it be better to put the script here?

disa6302: I am thinking the same. I am going to see which works out better. So testing it out.

Issue #121, opened 2020-01-04
Mbedtls Support?


fire:
Support using mbedtls? Also see https://github.com/ARMmbed/mbedtls/pull/1813.

MushMal:
I don't see any technical reason why this can't be done. The API is compatible so it should be doable. Perhaps we can come up with a small sample at some stage?

MushMal:
C Producer SDK has been enhanced to use mbedTLS already. We should adapt it for webrtc too

PR #552, opened 2020-06-15
Decouple SocketConnection from OpenSSL

lherman-cs:
The goal for this PR is to decouple Ice from OpenSSL so that we can easily swap the TLS implementation with mbedTLS (https://github.com/awslabs/amazon-kinesis-video-streams-webrtc-sdk-c/issues/121).

`TlsSession` is intentionally designed to be similar to our current `DtlsSession`. The idea is that if the interfaces are similar, we can merge these 2 easily in the future, and have a single proxy class that can swap between DTLS and TLS with a single flag.

### DIFF
--- a/src/source/Ice/SocketConnection.c
+++ b/src/source/Ice/SocketConnection.c
@@ -31,10 +31,8 @@ STATUS createSocketConnection(PKvsIpAddress pHostIpAddr, PKvsIpAddress pPeerIpAd
     }
     ATOMIC_STORE_BOOL(&pSocketConnection->connectionClosed, FALSE);
     ATOMIC_STORE_BOOL(&pSocketConnection->receiveData, FALSE);
-    pSocketConnection->freeBios = TRUE;
     pSocketConnection->dataAvailableCallbackCustomData = customData;
     pSocketConnection->dataAvailableCallbackFn = dataAvailableFn;
-    pSocketConnection->tlsHandshakeStartTime = INVALID_TIMESTAMP_VALUE;
 
 CleanUp:
 
@@ -68,20 +66,8 @@ STATUS freeSocketConnection(PSocketConnection* ppSocketConnection)
         MUTEX_FREE(pSocketConnection->lock);
     }
 
-    if (pSocketConnection->pSslCtx != NULL) {
-        SSL_CTX_free(pSocketConnection->pSslCtx);
-    }
-
-    if (pSocketConnection->freeBios && pSocketConnection->pReadBio != NULL) {
-        BIO_free(pSocketConnection->pReadBio);
-    }
-
-    if (pSocketConnection->freeBios && pSocketConnection->pWriteBio != NULL) {
-        BIO_free(pSocketConnection->pWriteBio);
-    }
-
-    if (pSocketConnection->pSsl != NULL) {
-        SSL_free(pSocketConnection->pSsl);
+    if (pSocketConnection->pTlsSession != NULL) {
+        freeTlsSession(&pSocketConnection->pTlsSession);
     }
 
     close(pSocketConnection->localSocket);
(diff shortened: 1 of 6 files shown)


src/source/Ice/SocketConnection.c
```diff
@@ -96,54 +82,67 @@ STATUS freeSocketConnection(PSocketConnection* ppSocketConnection)
     return retStatus;
 }
 
-STATUS socketConnectionInitSecureConnection(PSocketConnection pSocketConnection, BOOL isServer)
+static STATUS tlsSessionOutBoundPacket(UINT64 customData, PBYTE pBuffer, UINT32 bufferLen)
```
MushMal, line 85: remove static keyword - not needed for function in C

src/source/Ice/SocketConnection.c
```diff
@@ -96,54 +82,67 @@ STATUS freeSocketConnection(PSocketConnection* ppSocketConnection)
...
-    CHK(pSocketConnection != NULL, STATUS_NULL_ARG);
-
-    pSocketConnection->pSslCtx = SSL_CTX_new(SSLv23_method());
-
-    CHK(pSocketConnection->pSslCtx != NULL, STATUS_SSL_CTX_CREATION_FAILED);
-
-    SSL_CTX_set_read_ahead(pSocketConnection->pSslCtx, 1);
-    SSL_CTX_set_verify(pSocketConnection->pSslCtx, SSL_VERIFY_PEER | SSL_VERIFY_FAIL_IF_NO_PEER_CERT, certificateVerifyCallback);
+    pSocketConnection = (PSocketConnection) customData;
+    retStatus = socketSendDataWithRetry(pSocketConnection, pBuffer, bufferLen, NULL, NULL);
```
MushMal, line 92: simply do CHK_STATUS(socketSendData());

Next time we add more data below could intro a bug

src/source/Ice/SocketConnection.c
```diff
@@ -96,54 +82,67 @@ STATUS freeSocketConnection(PSocketConnection* ppSocketConnection)
...
 
-    CHK(SSL_CTX_set_cipher_list(pSocketConnection->pSslCtx, "HIGH:!aNULL:!MD5:!RC4"), STATUS_SSL_CTX_CREATION_FAILED);
+CleanUp:
+    return retStatus;
+}
 
-    pSocketConnection->pSsl = SSL_new(pSocketConnection->pSslCtx);
-    CHK(pSocketConnection->pSsl != NULL, STATUS_CREATE_SSL_FAILED);
+static VOID tlsSessionOnStateChange(UINT64 customData, TLS_SESSION_STATE state)
```
MushMal, line 98: No need for the static

src/source/Ice/SocketConnection.c
```diff
@@ -96,54 +82,67 @@ STATUS freeSocketConnection(PSocketConnection* ppSocketConnection)
...
+    STATUS retStatus = STATUS_SUCCESS;
+    
+    CHK(pSocketConnection != NULL, STATUS_NULL_ARG);
+    CHK(pSocketConnection->pTlsSession == NULL, STATUS_INVALID_ARG);
 
-    BIO_set_mem_eof_return(pSocketConnection->pReadBio, -1);
-    BIO_set_mem_eof_return(pSocketConnection->pWriteBio, -1);
-    SSL_set_bio(pSocketConnection->pSsl, pSocketConnection->pReadBio, pSocketConnection->pWriteBio);
-    pSocketConnection->freeBios = FALSE;
+    TlsSessionCallbacks callbacks;
```
MushMal, line 134: Move all local stack declarations before any code that can "jump" over it. This will generate an error on pure c compilers

src/source/Ice/Tls.c
```diff
@@ -0,0 +1,267 @@
+/**
+ * Kinesis Video Tcp
+ */
+#define LOG_CLASS "Tls"
+#include "../Include_i.h"
+
+STATUS createTlsSession(TlsSessionCallbacks callbacks, PTlsSession *ppTlsSession)
```
MushMal, line 7: I would prefer this to be a pointer instead of an entire struct being passed as a copy. In case of the pointer, change the name to pCallbacks and validate it together with ppTlsSession not being NULL

src/source/Ice/Tls.c
```diff
@@ -0,0 +1,267 @@
...
+{
+    ENTERS();
+    STATUS retStatus = STATUS_SUCCESS;
+    PTlsSession pTlsSession = NULL;
+
+    CHK(ppTlsSession != NULL, STATUS_NULL_ARG);
+    CHK(callbacks.outboundPacketFn != NULL, STATUS_NULL_ARG);
```
MushMal, line 14: Invalid arg is a better err here

src/source/Ice/Tls.c
```diff
@@ -0,0 +1,267 @@
...
+
+    if (pTlsSession->pSslCtx != NULL) {
+        SSL_CTX_free(pTlsSession->pSslCtx);
+    }
+
+    if (pTlsSession->pSsl != NULL) {
+        SSL_free(pTlsSession->pSsl);
+    }
+
+    CHK_STATUS(tlsSessionShutdown(pTlsSession));
```
MushMal, line 54: Can't do CHK_STATUS here. In case of an error it will skip the mem freeing a line below. Store the return value perhaps and return it later

src/source/Ice/Tls.c
```diff
@@ -0,0 +1,267 @@
...
+    SAFE_MEMFREE(*ppTlsSession);
+
+CleanUp:
+    return retStatus;
+}
+
+// https://www.openssl.org/docs/man1.0.2/man3/SSL_CTX_set_verify.html
+static INT32 certificateVerifyCallback(INT32 preverify_ok, X509_STORE_CTX *ctx)
```
MushMal, line 62: no need for static decl type

src/source/Ice/Tls.c
```diff
@@ -0,0 +1,267 @@
...
+{
+    UNUSED_PARAM(preverify_ok);
+    UNUSED_PARAM(ctx);
+    return 1;
+}
+
```
MushMal, line 68: Extra \n

src/source/Ice/Tls.c
```diff
@@ -0,0 +1,267 @@
...
+    INT32 sslReadRet = 0;
+    UINT32 writtenBytes = 0;
+    UINT64 sslErrorRet;
+
+    CHK(pTlsSession != NULL && pData != NULL && pDataLen != NULL, STATUS_NULL_ARG);
+    CHK(pTlsSession->state != TLS_SESSION_STATE_NEW, STATUS_SOCKET_CONNECTION_NOT_READY_TO_SEND);
+    CHK(pTlsSession->state != TLS_SESSION_STATE_CLOSED, STATUS_SOCKET_CONNECTION_CLOSED_ALREADY);
+
+    // return early if there's no data
+    CHK(*pDataLen > 0, retStatus);
```
MushMal, line 146: As this is unsigned, a better check would be CHK(*pDataLen != 0, retStatus);

src/source/Ice/Tls.c
```diff
@@ -0,0 +1,267 @@
...
+                    continueRead = FALSE;
+                    break;
+                case SSL_ERROR_WANT_READ:
+                    break;
+                default:
+                    sslErrorRet = ERR_get_error();
+                    DLOGW("SSL_read failed with %s", ERR_error_string(sslErrorRet, NULL));
+                    break;
+            }
+            break;
```
MushMal, line 166: don't break here. set continueRead = FALSE

src/source/Ice/Tls.c
```diff
@@ -0,0 +1,267 @@
...
+    PCHAR wBioBuffer = NULL;
+
+    CHK(pTlsSession != NULL, STATUS_NULL_ARG);
+
+    if (SSL_is_init_finished(pTlsSession->pSsl)) {
+        tlsSessionChangeState(pTlsSession, TLS_SESSION_STATE_CONNECTED);
+
+        /* Should have a valid buffer */
+        CHK(pData != NULL && dataLen > 0, STATUS_INVALID_ARG);
+        sslRet = SSL_write(pTlsSession->pSsl, pData, dataLen);
```
MushMal, line 199: can pData be NULL here? If not, make sure you either have validated pData in above call sites or validate here with pTlsSession not being NULL

src/source/Ice/SocketConnection.c
lherman-cs: Why is that? Because my main purpose here is to avoid name conflicts. Since this is just a callback, I thought it was appropriate to keep the scope within the file only.

src/source/Ice/SocketConnection.c
lherman-cs: makes sense.

src/source/Ice/SocketConnection.c
lherman-cs: same reason as above.

src/source/Ice/SocketConnection.c
lherman-cs: My bad, I forgot to do that. I'll update it. Thanks.

src/source/Ice/Tls.c
lherman-cs: same reason.

src/source/Ice/Tls.c
lherman-cs: Yep, makes sense. pData shouldn't be null here, since it's already checked in the above call.

src/source/Ice/SocketConnection.c
MushMal: First, static has no bearing on the name conflicts - those are compile time issues. The class modifier in this case is a linker directive to mangle the export name to make it local class. Second, many compilers will not support this as this is not C spec.

src/source/Ice/SocketConnection.c
MushMal: Ditto

src/source/Ice/SocketConnection.c
lherman-cs: Ah I see, that makes sense. I'll update it.

PR #553, opened 2020-06-15
log ice candidate protocol type. change sample to handle upcoming get…

chehefen:
…IceServerConfigChange. Other minor fixs.

getIceServerConfig will soon return urls in follow forms:
  - turn:ip:port?transport=udp
  - turns:ip:port?transport=udp
  - turns:ip:port?transport=tcp

Originally in Common.c we are manually appending the query param. It's removed in this commit. After this commit and before the getIceServerConfig change, ICE will try both TURN over UDP and TURN over TCP/TLS.

*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/src/source/Ice/SocketConnection.c
+++ b/src/source/Ice/SocketConnection.c
@@ -260,6 +260,7 @@ STATUS socketConnectionReadData(PSocketConnection pSocketConnection, PBYTE pBuf,
                     continueRead = FALSE;
                     break;
                 case SSL_ERROR_WANT_READ:
+                case SSL_ERROR_ZERO_RETURN:
                     break;
                 default:
                     sslErrorRet = ERR_get_error();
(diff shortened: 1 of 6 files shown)


src/source/Ice/SocketConnection.c
```diff
@@ -260,6 +260,7 @@ STATUS socketConnectionReadData(PSocketConnection pSocketConnection, PBYTE pBuf,
                     continueRead = FALSE;
                     break;
                 case SSL_ERROR_WANT_READ:
+                case SSL_ERROR_ZERO_RETURN:
```
lherman-cs, line 263: I think we should mark the connection as closed to let the above call knows this state update.

lherman-cs: Actually, nvm, the documentation apparently says that there might still be some pending data.

PR #557, opened 2020-06-18
Fixed compilation issue.

fernandoc1:
*Issue #, if available:*

*Description of changes:*


CMakeLists.txt
```diff
@@ -252,6 +252,7 @@ target_link_libraries(
   kvsWebrtcClient
   PRIVATE kvspicUtils
           kvspicState
+          m
```
MushMal, line 255: Can you describe why we need libm in case of the webrtc library? I wasn't sure whether this is needed

PR #560, opened 2020-06-22
Merge free rtp packet

### LINKED ISSUE #518: [BUG]Memory is not released completely after session disconnecting
**Describe the bug**
The memory usage rate is increasing during the connection.

I have customized the following params to reduce the size of the rolling buffers.
https://github.com/awslabs/amazon-kinesis-video-streams-webrtc-sdk-c/blob/master/src/source/PeerConnection/Rtp.h#L11-L12
```
#define DEFAULT_ROLLING_BUFFER_DURATION_IN_SECONDS              1
#define HIGHEST_EXPECTED_BIT_RATE                               (2 * 1024 * 1024)
```
I added the definitions in CMakeLists.txt. And reset the allocators in code.
`	add_definitions(-DINSTRUMENTED_ALLOCATORS)`
`SET_INSTRUMENTED_ALLOCATORS();`
It print the size of allocation regularly as follows:
GetAllocationSize,==-=>size[71926]!

When I connected to the device for a while, the memory usage increased to 2891417. From 71926 to 2829102. After disconnecting the memory usage decreased to 1103593 which was higer than before.

This phenomenon does not happen every time. Usually it occurs when the network is not good enough or the devices and viewers are under different networks. When I use the master and viewer in our office, it is hard to reprodure. However, preview elsewhere easily to reproduce.

Here is the full log.
[master.log](https://github.com/awslabs/amazon-kinesis-video-streams-webrtc-sdk-c/files/4729944/master.log)

chehefen:
*Issue #, if available:*
fix #518 
*Description of changes:*
RtpPacket should take ownership of its pRawPacket. pRawPacket is set in the following places:
- https://github.com/awslabs/amazon-kinesis-video-streams-webrtc-sdk-c/blob/master/src/source/PeerConnection/Rtp.c#L207
  - rawPacket is a copy and allocated at line https://github.com/awslabs/amazon-kinesis-video-streams-webrtc-sdk-c/blob/master/src/source/PeerConnection/Rtp.c#L203
- https://github.com/awslabs/amazon-kinesis-video-streams-webrtc-sdk-c/blob/master/src/source/PeerConnection/Rtp.c#L216
  - rawPacket is a copy and allocated at line https://github.com/awslabs/amazon-kinesis-video-streams-webrtc-sdk-c/blob/master/src/source/PeerConnection/Rtp.c#L203
- https://github.com/awslabs/amazon-kinesis-video-streams-webrtc-sdk-c/blob/master/src/source/Rtp/RtpPacket.c#L113
  - createRtpPacketFromBytes is called in https://github.com/awslabs/amazon-kinesis-video-streams-webrtc-sdk-c/blob/6f741f4c6830fcad7eb2ccb9f07f38367ee7fb40/src/source/Rtcp/RtpRollingBuffer.c#L66 and https://github.com/awslabs/amazon-kinesis-video-streams-webrtc-sdk-c/blob/b2d1600c62ac14f1a7c4557f914086d195caccd6/src/source/PeerConnection/PeerConnection.c#L187, in both places the rawPacket is a copy
- https://github.com/awslabs/amazon-kinesis-video-streams-webrtc-sdk-c/blob/93dbd03699b9eaaa9c173de888da7e0dd2f5daec/src/source/Rtp/RtpPacket.c#L158. pRawPacket is allocated and copied

Since pRawPacket is always a copy owned by pRtpPacket, it should be safe to free pRawPacket also when free pRtpPacket.

### DIFF
--- a/src/source/PeerConnection/JitterBuffer.c
+++ b/src/source/PeerConnection/JitterBuffer.c
@@ -98,7 +98,7 @@ STATUS jitterBufferPush(PJitterBuffer pJitterBuffer, PRtpPacket pRtpPacket)
         || pRtpPacket->header.timestamp >= pJitterBuffer->lastPushTimestamp - pJitterBuffer->maxLatency) {
         pCurPacket = pJitterBuffer->pktBuffer[pRtpPacket->header.sequenceNumber];
         if (pCurPacket != NULL) {
-            freeRtpPacketAndRawPacket(&pCurPacket);
+            freeRtpPacket(&pCurPacket);
             pJitterBuffer->pktBuffer[pRtpPacket->header.sequenceNumber] = NULL;
         }
         pJitterBuffer->pktBuffer[pRtpPacket->header.sequenceNumber] = pRtpPacket;
@@ -106,7 +106,7 @@ STATUS jitterBufferPush(PJitterBuffer pJitterBuffer, PRtpPacket pRtpPacket)
         DLOGS("jitterBufferPush get packet timestamp %lu seqNum %lu", pRtpPacket->header.timestamp, pRtpPacket->header.sequenceNumber);
     } else {
         // Free the packet if it is out of range, jitter buffer need to own the packet and do free
-        freeRtpPacketAndRawPacket(&pRtpPacket);
+        freeRtpPacket(&pRtpPacket);
     }
 
     CHK_STATUS(jitterBufferPop(pJitterBuffer, FALSE));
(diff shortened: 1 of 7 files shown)


Sean-Der:
@chehefen I am reading now, but is everything in `Retransimitter.c` safe as well?

Before it wouldn't touch the rawPacket (but now it is freeing it as well). We have tests that cover this, but I am worried that they aren't E2E.

Would you able to run a simple call using Chromium with ~3% loss and just make sure ASAN (or Valgrind) is happy?

Sean-Der:
`constructRetransmitRtpPacketFromBytes` has 0 coverage, that is concerning :/

chehefen:
> (but now it is freeing it as well). We have tests that cover this, but I am worried that they aren't E

since rawPacket is always a copy, we probably should also free the rawPacket when freeing pRtpPacket. Maybe this is what caused the leak in #518. I ran a call with chromium with 5% loss with ASAN and everything worked ok.

PR #565, opened 2020-06-23
Add mbedTLS support

lherman-cs:
Changes for mbedTLS specific implementation:
  * Add DTLS and TLS support
  * Add SRTP (media channel) and SCTP (data channel) support
  * Add common crypto defs
  * Create a new Crypto module
  * Move TLS and DTLS to Crypto module
  * Add pre-generated certificate and key support
  * Add unit tests  
  * Add end-to-end tests
  * Add a CI build for mbedTLS

Todo:
  * Create a memory/cpu footprint observation between OpenSSL vs mbedTLS
  * Reduce memory footprint by configuring mbedTLS to exclude unnecessary things
  * There are still some duplicated codes between OpenSSL and mbedTLS specific codes. I'm planning to do more refactoring in the future about this. But, for now, the goal is to avoid any regression with OpenSSL, it makes it easier to see the diff for the future PR.
  * Better pre-generated certificate library agnostic support


CMake/Dependencies/libmbedtls-CMakeLists.txt
```diff
@@ -0,0 +1,32 @@
...
+    set(CMAKE_MACOSX_RPATH 1)
+endif()
+
+set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -DMBEDTLS_SSL_DTLS_SRTP")
+ExternalProject_Add(
+  project_libmbedtls
+  " TODO: needs to replace this with official mbedTLS branch
+  "       when they add DTLS-SRTP extension support.
+  " Reference: https://github.com/ARMmbed/mbedtls/pull/3235
+  GIT_REPOSITORY 	https://github.com/lherman-cs/mbedtls.git
```
lherman-cs, line 23: I forked the PR branch for the DTLS-SRTP extension support. The main reason is so that the branch is persisted. Otherwise, there might be a new update to the branch, or the PR gets merged and the branch gets deleted.

CMakeLists.txt
```diff
@@ -210,10 +250,12 @@ target_link_libraries(
   kvsWebrtcClient
   PRIVATE kvspicUtils
           kvspicState
+          ${CMAKE_THREAD_LIBS_INIT}
```
lherman-cs, line 253: This is needed because mbedTLS doesn't link to pthread. If I didn't do this, the compilation will fail due to not able finding pthread APIs.

MushMal: What about other platforms like Windows

src/source/Crypto/Dtls.h
```diff
@@ -11,14 +11,17 @@
...
-#define GENERATED_CERTIFICATE_NAME (PUINT8) "KVS-WebRTC-Client"
-#define KEYING_EXTRACTOR_LABEL "EXTRACTOR-dtls_srtp"
+#define MAX_SRTP_MASTER_KEY_LEN             16
+#define MAX_SRTP_SALT_KEY_LEN               14
+#define MAX_DTLS_RANDOM_BYTES_LEN           32
+#define MAX_DTLS_MASTER_KEY_LEN             48
+
+#define GENERATED_CERTIFICATE_MAX_SIZE      4096
+#define GENERATED_CERTIFICATE_BITS          2048
+#define GENERATED_CERTIFICATE_SERIAL        1
```
MushMal, line 21: Not sure why this change though.

src/source/Crypto/Dtls.h
```diff
@@ -29,15 +32,16 @@ extern "C" {
...
 
-#define LOG_OPENSSL_ERROR(s)                    while ((sslErr = ERR_get_error()) != 0) { \
-                                                    if (sslErr != SSL_ERROR_WANT_WRITE && sslErr != SSL_ERROR_WANT_READ) { \
-                                                        DLOGW("%s failed with %s", (s), ERR_error_string(sslErr, NULL)); \
-                                                    } \
-                                                }
+#define HUNDREDS_OF_NANOS_IN_A_DAY          (HUNDREDS_OF_NANOS_IN_AN_HOUR * 24LL)
 
 typedef enum {
+#ifdef KVS_USE_OPENSSL
```
MushMal, line 38: Minor. These could be outsourced to Crypto.h that you've created. Alias them like KVS_SRTP_AES128_CM_SHA1_80 or something. Not a biggie

src/source/Crypto/Dtls.h
```diff
@@ -71,31 +75,66 @@ typedef struct {
...
 } DtlsKeyingMaterial, *PDtlsKeyingMaterial;
 
+#ifdef KVS_USE_OPENSSL
 typedef struct {
     BOOL created;
     X509 *pCert;
     EVP_PKEY *pKey;
 } DtlsSessionCertificateInfo, *PDtlsSessionCertificateInfo;
 
+#elif KVS_USE_MBEDTLS
```
MushMal, line 85: Minor: This should be else. If you are using elif then perhaps you should have the else clause with an error that the TLS is not defined

src/source/Crypto/Dtls_mbedtls.c
```diff
@@ -0,0 +1,739 @@
...
+    SAFE_MEMFREE(*ppDtlsSession);
+
+CleanUp:
+    LEAVES();
+    return retStatus;
+}
+
+INT32 dtlsSessionSendCallback(PVOID customData, const unsigned char *pBuf, ULONG len)
+{
+  STATUS retStatus = STATUS_SUCCESS;
```
MushMal, line 118: spacing is off - 2 instead of 4 spaces below this

src/source/Crypto/Dtls_mbedtls.c
```diff
@@ -0,0 +1,739 @@
...
+  PDtlsSession pDtlsSession = (PDtlsSession) customData;
+
+  CHK(pDtlsSession != NULL, STATUS_NULL_ARG);
+
+  pDtlsSession->dtlsSessionCallbacks.outboundPacketFn(pDtlsSession->dtlsSessionCallbacks.outBoundPacketFnCustomData, (PBYTE) pBuf, len);
+
+CleanUp:
+
+  return STATUS_FAILED(retStatus) ? -retStatus : len;
```
MushMal, line 127: Do we have any consumers aware of this? Perhaps you should document in a comment or something

src/source/Crypto/Dtls_mbedtls.c
```diff
@@ -0,0 +1,739 @@
...
+    PDtlsSession pDtlsSession = (PDtlsSession) customData;
+    BOOL locked = FALSE;
+
+    CHK(pDtlsSession != NULL, STATUS_NULL_ARG);
+
+    MUTEX_LOCK(pDtlsSession->sslLock);
+    locked = TRUE;
+
+    handshakeStatus = mbedtls_ssl_handshake(&pDtlsSession->sslCtx);
+    if (handshakeStatus == 0) {
```
MushMal, line 204: Best to use switch/case statement

src/source/Crypto/Dtls_mbedtls.c
```diff
@@ -0,0 +1,739 @@
...
+        sslRet = mbedtls_ssl_read(&pDtlsSession->sslCtx, pData + readBytes, pReadBuffer->len - pReadBuffer->off);
+        if (sslRet > 0) {
+            readBytes += sslRet;
+        } else if (sslRet == 0 || sslRet == MBEDTLS_ERR_SSL_PEER_CLOSE_NOTIFY) {
+            // if sslRet is 0, the connection is closed already.
+            // if sslRet is MBEDTLS_ERR_SSL_PEER_CLOSE_NOTIFY, the client notified us that the connection is going to be closed.
+            // In either case, we'll make sure that the state will change to CLOSED. If it's already closed, it'll be just a noop.
+            DLOGD("Detected DTLS close_notify alert");
+            CHK_STATUS(dtlsSessionShutdown(pDtlsSession));
+            break;
```
MushMal, line 348: Best to set a boolean variable that gets checked in the loops condition instead of flow modification. Something like iterate = FALSE, then in the while loop have while (iterate && blah)

src/source/Crypto/Dtls_mbedtls.c
```diff
@@ -0,0 +1,739 @@
...
+
+    MUTEX_LOCK(pDtlsSession->sslLock);
+    locked = TRUE;
+    
+    while (writtenBytes < dataLen) {
+        sslRet = mbedtls_ssl_write(&pDtlsSession->sslCtx, pData + writtenBytes, dataLen - writtenBytes);
+        if (sslRet > 0) {
+            writtenBytes += sslRet;
+        } else if (sslRet == MBEDTLS_ERR_SSL_WANT_READ || sslRet == MBEDTLS_ERR_SSL_WANT_WRITE) {
+            break;
```
MushMal, line 395: Same minor comment about break;

src/source/Crypto/Dtls_mbedtls.c
```diff
@@ -0,0 +1,739 @@
...
+    STATUS retStatus = STATUS_SUCCESS;
+    BOOL locked = FALSE;
+
+    CHK(pDtlsSession != NULL && pBuff != NULL, STATUS_NULL_ARG);
+    CHK(buffLen >= CERTIFICATE_FINGERPRINT_LENGTH, STATUS_INVALID_ARG_LEN);
+
+    MUTEX_LOCK(pDtlsSession->sslLock);
+    locked = TRUE;
+
+    // TODO: Use the 0th certificate for now
```
MushMal, line 425: GitHub issue link:?

src/source/Crypto/Dtls_mbedtls.c
```diff
@@ -0,0 +1,739 @@
...
+    INT32 sslRet;
+
+    CHK(pDtlsSession != NULL, STATUS_NULL_ARG);
+
+    MUTEX_LOCK(pDtlsSession->sslLock);
+    locked = TRUE;
+
+    CHK(!ATOMIC_LOAD_BOOL(&pDtlsSession->shutdown), retStatus);
+
+    do sslRet = mbedtls_ssl_close_notify(&pDtlsSession->sslCtx);
```
MushMal, line 536: Use curlies even with a single line statements

MushMal: How do we make sure this doesn't evaluate into a tight loop?

src/source/Crypto/IOBuffer.c
```diff
@@ -0,0 +1,101 @@
+#define LOG_CLASS "IOBuffer"
```
MushMal, line 1: CamelCase the name of the file and the names. Something like IoBuffer.c

The APIs will then be createIoBuffer

src/source/Crypto/IOBuffer.c
```diff
@@ -0,0 +1,101 @@
...
+#include "../Include_i.h"
+
+STATUS createIOBuffer(UINT32 initialCap, PIOBuffer* ppBuffer) {
+    STATUS retStatus = STATUS_SUCCESS;
+    PIOBuffer pBuffer = NULL;
+
+    pBuffer = MEMCALLOC(SIZEOF(IOBuffer), 1);
```
MushMal, line 8: Need to cast the result of MEMCALLOC to (PIoBuffer) type as compilers will warn

src/source/Crypto/Tls_mbedtls.c
```diff
@@ -0,0 +1,239 @@
+/**
+ * Kinesis Video Tcp
```
MushMal, line 2: Tcp?

src/source/Crypto/Tls_mbedtls.c
```diff
@@ -0,0 +1,239 @@
...
+#include "../Include_i.h"
+
+STATUS createTlsSession(PTlsSessionCallbacks pCallbacks, PTlsSession *ppTlsSession)
+{
+    ENTERS();
+    STATUS retStatus = STATUS_SUCCESS;
+    PTlsSession pTlsSession = NULL;
+
+    CHK(ppTlsSession != NULL && pCallbacks != NULL, STATUS_NULL_ARG);
+    CHK(pCallbacks->outboundPacketFn != NULL, STATUS_NULL_ARG);
```
MushMal, line 14: Combine?

src/source/Crypto/Tls_mbedtls.c
```diff
@@ -0,0 +1,239 @@
...
+
+    pTlsSession = MEMCALLOC(1, SIZEOF(TlsSession));
```
MushMal, line 16: Need casting

src/source/Crypto/Tls_mbedtls.c
```diff
@@ -0,0 +1,239 @@
...
+    retStatus = tlsSessionShutdown(pTlsSession);
+    SAFE_MEMFREE(*ppTlsSession);
+
+CleanUp:
+    return retStatus;
+}
+
+INT32 tlsSessionSendCallback(PVOID customData, const unsigned char *buf, ULONG len)
+{
+  STATUS retStatus = STATUS_SUCCESS;
```
MushMal, line 72: spacing

src/source/Crypto/Tls_mbedtls.c
```diff
@@ -0,0 +1,239 @@
...
+
+STATUS tlsSessionShutdown(PTlsSession pTlsSession)
+{
+    STATUS retStatus = STATUS_SUCCESS;
+    INT32 sslRet;
+
+    CHK(pTlsSession != NULL, STATUS_NULL_ARG);
+    CHK(pTlsSession->state != TLS_SESSION_STATE_CLOSED, retStatus);
+
+    do sslRet = mbedtls_ssl_close_notify(&pTlsSession->sslCtx);
```
MushMal, line 230: curlies

src/source/Crypto/Dtls_mbedtls.c
lherman-cs: Not really sure, we've been doing this for a while, I just copied this from the openssl implementation, https://github.com/awslabs/amazon-kinesis-video-streams-webrtc-sdk-c/blob/master/src/source/Dtls/Dtls.c#L723-L746.

src/source/Crypto/Dtls_mbedtls.c
lherman-cs: Done

src/source/Crypto/Dtls_mbedtls.c
lherman-cs: Done

src/source/Crypto/Dtls_mbedtls.c
lherman-cs: Done

src/source/Crypto/Dtls_mbedtls.c
lherman-cs: This is an internal status code. I had to multiply by -1 so that mbedtls thinks that it's an error. When it's an error, `mbedtls_ssl_read` will return an error and reach this block, https://github.com/awslabs/amazon-kinesis-video-streams-webrtc-sdk-c/pull/565/files#diff-c1ebb6c5ff118358fc32fd996dd11f8bR169-R172. The status that we'll return to the customer is going to be `STATUS_INTERNAL_ERROR`.

This is definitely not perfect since the error code can be ambiguous, it can be the same error code from mbedtls. But, at least, we'll still log the error code which might help us in debugging, https://github.com/awslabs/amazon-kinesis-video-streams-webrtc-sdk-c/pull/565/files#diff-c1ebb6c5ff118358fc32fd996dd11f8bR169.

src/source/Crypto/Dtls_mbedtls.c
lherman-cs: Done

src/source/Crypto/Dtls.h
lherman-cs: Ah, forgot that C has #error directive. I'll do that. Thanks.

src/source/Crypto/Dtls.h
lherman-cs: This is the problem that I mentioned a while ago that mbedtls has a bug, when it's zero, mbedtls will not append a bit of 0, https://github.com/ARMmbed/mbedtls/issues/2166.

src/source/Crypto/Dtls_mbedtls.c
lherman-cs: This is going to tell mbedtls to flush its internal buffer. If a handshake has happened, mbedtls will also send a close notify to the other peer. 

Sending a close notify might fail if sometimes the send callback fails, https://github.com/awslabs/amazon-kinesis-video-streams-webrtc-sdk-c/pull/565/files#diff-1ec7a454de24efa5d640c90c61260220R116-R128. But, since the outgoing callback that's passed to DTLS is always "successful", VOID as return type, mbedtls should be able to always flush its internal buffer, until it runs out or not `MBEDTLS_ERR_SSL_WANT_WRITE`.

Should I add the number of max retries to make sure that this loop will always end?

CMakeLists.txt
lherman-cs: @MushMal and I talked offline about this, and decided that we'll add Windows support later, not a part of this PR.

src/source/Crypto/Tls_mbedtls.c
lherman-cs: done

src/source/Crypto/Tls_mbedtls.c
lherman-cs: Done

src/source/Crypto/Tls_mbedtls.c
lherman-cs: Done

src/source/Crypto/Tls_mbedtls.c
lherman-cs: Ah, right. Done.

src/source/Crypto/Tls_mbedtls.c
lherman-cs: Oops, copy error. Done.

src/source/Crypto/IOBuffer.c
lherman-cs: Done.

src/source/Crypto/IOBuffer.c
lherman-cs: I think it's better to be "IO", input output, since they're not full words. If "Io", it sound like as if there's a word "io".

src/source/Crypto/Dtls_mbedtls.c
MushMal: I doubt it... Perhaps, if this proves to be an issue on the MBED TLS side

src/source/Crypto/IOBuffer.c
MushMal: OK

src/source/Crypto/Dtls.h
lherman-cs: Done

src/source/Crypto/Dtls.h
lherman-cs: Yep, makes sense!

.travis.yml
```diff
@@ -167,6 +167,15 @@ matrix:
         - cmake .. -DBUILD_OPENSSL=TRUE -DBUILD_OPENSSL_PLATFORM=linux-generic32 -DBUILD_LIBSRTP_HOST_PLATFORM=x86_64-unknown-linux-gnu -DBUILD_LIBSRTP_DESTINATION_PLATFORM=arm-unknown-linux-uclibcgnueabi
       script: make
 
+    # mbedTLS. TODO: we probably should run mbedTLS build after each OpenSSL build to cover all platforms?
```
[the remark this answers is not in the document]
lherman-cs: Yep, changed it to gcc instead of clang. And, the os is still Linux.

lherman-cs:
@MushMal I've updated the PR, I didn't force push the original commit but instead I created new commits for the changes to reflect the comments. This is to make it easier to review per change since it's organized per commit. 

I'll squash all the commits when it's final.

src/source/Crypto/Dtls_mbedtls.c
```diff
@@ -0,0 +1,739 @@
+#define LOG_CLASS "DTLS_mbedtls"
```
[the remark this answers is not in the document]
lherman-cs: I did differentiate this on purpose, because this is a "special case" since the source is filtered using this pattern, https://github.com/awslabs/amazon-kinesis-video-streams-webrtc-sdk-c/pull/565/files#diff-af3b638bc2a3e6c650974192a53c7291R231-R236.

This is just to let us aware that, the binary will have a mutual exclusion depending on the flag.

src/source/Crypto/Dtls_openssl.c
```diff
@@ -1,4 +1,4 @@
-#define LOG_CLASS "DTLS"
+#define LOG_CLASS "DTLS_openssl"
```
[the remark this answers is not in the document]
lherman-cs: ditto

src/source/Crypto/Tls_mbedtls.c
```diff
@@ -0,0 +1,241 @@
...
+
+STATUS tlsSessionShutdown(PTlsSession pTlsSession)
+{
+    STATUS retStatus = STATUS_SUCCESS;
+    INT32 sslRet;
+
+    CHK(pTlsSession != NULL, STATUS_NULL_ARG);
+    CHK(pTlsSession->state != TLS_SESSION_STATE_CLOSED, retStatus);
+
+    while (mbedtls_ssl_close_notify(&pTlsSession->sslCtx) == MBEDTLS_ERR_SSL_WANT_WRITE) {
```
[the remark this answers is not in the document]
lherman-cs: The same reason as https://github.com/awslabs/amazon-kinesis-video-streams-webrtc-sdk-c/pull/565#discussion_r452385656.

src/source/Crypto/Tls_mbedtls.c
```diff
@@ -0,0 +1,241 @@
...
+    LEAVES();
+    return retStatus;
+}
+
+STATUS tlsSessionPutApplicationData(PTlsSession pTlsSession, PBYTE pData, UINT32 dataLen)
+{
+    ENTERS();
+    STATUS retStatus = STATUS_SUCCESS;
+    UINT32 writtenBytes = 0;
+    BOOL iterate = TRUE;
```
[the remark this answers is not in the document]
lherman-cs: Hmm, I don't see why not since it's really just a flag marking if it should iterate, nothing else specific. I don't mind to rename it if you can come up with a better name.

src/source/Crypto/Tls_mbedtls.c
```diff
@@ -0,0 +1,241 @@
...
+}
+
+INT32 tlsSessionSendCallback(PVOID customData, const unsigned char *buf, ULONG len)
+{
+    STATUS retStatus = STATUS_SUCCESS;
+    PTlsSession pTlsSession = (PTlsSession) customData;
+
+    CHK(pTlsSession != NULL, STATUS_NULL_ARG);
+
+    pTlsSession->callbacks.outboundPacketFn(pTlsSession->callbacks.outBoundPacketFnCustomData, (PBYTE) buf, len);
```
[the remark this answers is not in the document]
lherman-cs: It's already checked in the createTlsSession, https://github.com/awslabs/amazon-kinesis-video-streams-webrtc-sdk-c/pull/565/files/23e82a59366d6ef8a3f333a2bdbc2e88188e1377#diff-c1ebb6c5ff118358fc32fd996dd11f8bR13, it's guaranteed to be not null for the whole session life cycle.

src/source/Crypto/Tls_mbedtls.c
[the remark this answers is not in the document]
disa6302: After looking a little more closely, I think it is ok.

PR #567, opened 2020-06-24
Enable deep debug mode in CI

lherman-cs:
The goal of this PR is to make it easier for us to debug build errors (for example, https://travis-ci.com/github/awslabs/amazon-kinesis-video-streams-webrtc-sdk-c/jobs/352369910), and debug issues locally.


CMake/Dependencies/libkvsCommonLws-CMakeLists.txt
```diff
@@ -6,7 +6,7 @@ include(ExternalProject)
 
 ExternalProject_Add(libkvsCommonLws-download
     GIT_REPOSITORY    https://github.com/awslabs/amazon-kinesis-video-streams-producer-c.git
-    GIT_TAG           bb5c4a7ff913fc0f557a5decaa35f45f802e10d9
+    GIT_TAG           11d0d4a5e906b26376b2cd816954d00af13503f9
```
lherman-cs, line 9: This update is necessary to forward the build mode from Producer C to PIC

.travis.yml
```diff
@@ -34,7 +34,7 @@ matrix:
       os: osx
       compiler: gcc
       before_script:
-        - mkdir build && cd build && cmake .. -DBUILD_TEST=TRUE
+        - mkdir build && cd build && cmake .. -DCMAKE_BUILD_TYPE=Debug -DBUILD_TEST=TRUE
```
MushMal, line 37: Enable debug run ONLY on one of the runs without any of the sanitizers. You could use this one for example or the Linux one. We need our CI to be running with the release mode in general.

lherman-cs: Makes sense. I'll update it.

PR #571, opened 2020-06-26
do not do socket bind for relay canddiate sockets

### LINKED ISSUE #568: [question] can you explain why knowing interface+ip ahead of time is required for turn?
https://github.com/awslabs/amazon-kinesis-video-streams-webrtc-sdk-c/blob/327a1aafd51c040d6d7661f1dc655199117bfe56/src/source/Ice/IceAgent.c#L1643-L1656

i've been thinking of replacing this with a programmatic `ip route get $MY_TURN_SERVER_IP_ADDR` which most of the time is default gateway

chehefen:
*Issue #, if available:*
fix #568 
*Description of changes:*

### DIFF
--- a/src/source/Ice/Network.c
+++ b/src/source/Ice/Network.c
@@ -148,25 +148,18 @@ CleanUp:
     return retStatus;
 }
 
-STATUS createSocket(PKvsIpAddress pHostIpAddress, PKvsIpAddress pPeerAddress, KVS_SOCKET_PROTOCOL protocol, UINT32 sendBufSize, PINT32 pSockFd)
+STATUS createSocket(KVS_IP_FAMILY_TYPE familyType, KVS_SOCKET_PROTOCOL protocol, UINT32 sendBufSize, PINT32 pOutSockFd)
 {
     STATUS retStatus = STATUS_SUCCESS;
 
-    struct sockaddr_in ipv4Addr, ipv4PeerAddr;
-    struct sockaddr_in6 ipv6Addr, ipv6PeerAddr;
-    INT32 sockfd, sockType, flags, retVal;
-    struct sockaddr *sockAddr = NULL, *peerSockAddr = NULL;
-    socklen_t addrLen;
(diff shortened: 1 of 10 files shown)


src/source/Ice/Network.c
```diff
@@ -148,25 +148,18 @@ STATUS getLocalhostIpAddresses(PKvsIpAddress destIpList, PUINT32 pDestIpListLen,
     return retStatus;
 }
 
-STATUS createSocket(PKvsIpAddress pHostIpAddress, PKvsIpAddress pPeerAddress, KVS_SOCKET_PROTOCOL protocol, UINT32 sendBufSize, PINT32 pSockFd)
+STATUS createSocket(UINT16 familyType, KVS_SOCKET_PROTOCOL protocol, UINT32 sendBufSize, PINT32 pOutSockFd)
```
MushMal, line 151: Not sure whether familyType should be UINT16 or the KVS_IP_FAMILY_TYPE enum type? Same with KvsIpAddress.family member. This is not a major issue for this change though

src/source/Ice/Network.c
```diff
@@ -182,6 +175,44 @@ STATUS createSocket(PKvsIpAddress pHostIpAddress, PKvsIpAddress pPeerAddress, KV
         CHK(FALSE, STATUS_SOCKET_SET_SEND_BUFFER_SIZE_FAILED);
     }
 
+    *pOutSockFd = (INT32) sockfd;
+
+#ifdef _WIN32
+    UINT32 nonblock = 1;
```
MushMal, line 181: This might cause a compiler warning/error. Perhaps, move this in the beginning of the function. If non-Windows compilers complain that it's not used you could simply add UNUSED_PARAM(nonblock)

src/source/Ice/Network.c
chehefen: changed to KVS_IP_FAMILY_TYPE

src/source/Ice/Network.c
chehefen: this is not causing any warnings in non-windows compiler.

PR #574, opened 2020-06-28
A standard way of install build artifacts

mateuszboryn:
Dear Developers,
here is a little enhancement that helps integrating this library with other projects. It installs resulting artifacts as long as header files.


MushMal:
The changes look OK but I would like other folks to comment as well.

PR #577, opened 2020-06-30
Signaling metrics work

MushMal:
*Issue #, if available:*

This is the first cut for the signaling metrics work. Not yet finished. 

Need to

* Add EMA (Exponential Mean Averaging) for the durations once this is exposed from PIC which is ongoing
* Look for ways to not use interlocking (for performance and complexity reasons) as the metrics operations could soft-lock with reading them/updating.

### DIFF
--- a/src/include/com/amazonaws/kinesis/video/webrtcclient/Stats.h
+++ b/src/include/com/amazonaws/kinesis/video/webrtcclient/Stats.h
@@ -296,17 +296,17 @@ typedef struct {
     UINT64 dpApiCallLatency; //!< Latency (in 100 ns) incurred per backend API call for the data plane APIs
     UINT64 signalingClientUptime; //!< Client uptime (in 100 ns). Timestamp will be recorded at every SIGNALING_CLIENT_STATE_CONNECTED
     UINT64 connectionDuration; //!< Duration of connection (in 100 ns)
-    UINT64 numberOfMessagesSent; //!< Number of messages sent by the signaling client
-    UINT64 numberOfMessagesReceived; //!< Number of messages received by the signaling client
-    UINT64 iceRefreshCount; //!< Number of times the ICE is refreshed
-    UINT64 numberOfDynamicErrors; //!< Number of indirect errors. These are errors that are not returned as part of
+    UINT32 numberOfMessagesSent; //!< Number of messages sent by the signaling client
+    UINT32 numberOfMessagesReceived; //!< Number of messages received by the signaling client
+    UINT32 iceRefreshCount; //!< Number of times the ICE is refreshed
+    UINT32 numberOfDynamicErrors; //!< Number of indirect errors. These are errors that are not returned as part of
                                   //!< public API calls but rather when an error occurs on background threads
                                   //!< for example:
                                   //!< * When received message on the background thread is "bad"
                                   //!< * When re-connect logic fails after pre-configured retries/times out
                                   //!< * When refreshing ICE server configuration fails after pre-configured retries
                                   //!< In all of these cases the error callback (if specified) will be called.
-    UINT64 numberOfReconnects; //!< Number of reconnects in the session
+    UINT32 numberOfReconnects; //!< Number of reconnects in the session
 } SignalingClientStats, PSignalingClientStats;
 
 /**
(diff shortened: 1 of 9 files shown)


src/source/Signaling/Signaling.c
```diff
@@ -1173,3 +1208,39 @@ UINT64 signalingGetCurrentTime(UINT64 customData)
...
+    pSignalingClientMetrics->signalingClientStats.iceRefreshCount = (UINT32) pSignalingClient->diagnostics.iceRefreshCount;
+    pSignalingClientMetrics->signalingClientStats.numberOfDynamicErrors = (UINT32) pSignalingClient->diagnostics.numberOfDynamicErrors;
+    pSignalingClientMetrics->signalingClientStats.numberOfReconnects = (UINT32) pSignalingClient->diagnostics.numberOfReconnects;
+    pSignalingClientMetrics->signalingClientStats.cpApiCallLatency = pSignalingClient->diagnostics.cpApiLatency;
+    pSignalingClientMetrics->signalingClientStats.dpApiCallLatency = pSignalingClient->diagnostics.dpApiLatency;
+
+    if (ATOMIC_LOAD_BOOL(&pSignalingClient->connected)) {
+        pSignalingClientMetrics->signalingClientStats.connectionDuration = curTime - pSignalingClient->diagnostics.connectTime;
+    } else {
+        pSignalingClientMetrics->signalingClientStats.connectionDuration = INVALID_TIMESTAMP_VALUE;
```
lherman-cs, line 1237: I think it's better if it's set to 0 instead. Let say we want to dump these metrics to Cloudwatch, dumping it as 0 will give more meaningful data, we can infer that the device is not connected to our signaling. However, if we use `INVALID_TIMESTAMP_VALUE`, it is ambiguous, it can be that there's a device that has run for `0xFFFFFFFFFFFFFFFF` (it's definitely unlikely) or the metrics get aggregated.

src/include/com/amazonaws/kinesis/video/webrtcclient/Stats.h
```diff
@@ -292,14 +292,21 @@ typedef struct {
...
-    UINT64 numberOfRecoverableErrors; //!< Number of recoverable errors. Will be a static count returned when requested
-    UINT64 numberOfReconnects; //!< Number of reconnects in the session
+    UINT64 cpApiCallLatency; //!< Latency (in 100 ns) incurred per backend API call for the control plane APIs
+    UINT64 dpApiCallLatency; //!< Latency (in 100 ns) incurred per backend API call for the data plane APIs
+    UINT64 signalingClientUptime; //!< Client uptime (in 100 ns). Timestamp will be recorded at every SIGNALING_CLIENT_STATE_CONNECTED
+    UINT64 connectionDuration; //!< Duration of connection (in 100 ns)
+    UINT32 numberOfMessagesSent; //!< Number of messages sent by the signaling client
+    UINT32 numberOfMessagesReceived; //!< Number of messages received by the signaling client
+    UINT32 iceRefreshCount; //!< Number of times the ICE is refreshed
+    UINT32 numberOfDynamicErrors; //!< Number of indirect errors. These are errors that are not returned as part of
```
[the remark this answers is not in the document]
MushMal: Yeah, perhaps you are right. Let me think more if I can find a better name. If not, I will use indirect infix. Perhaps runtime?

src/source/Signaling/Signaling.c
MushMal: You are absolutely correct. Will do that

disa6302:
Just throwing out a suggestion here: can we have a utility to log the signaling metrics periodically as well? Of course, this would not be done unless the user chooses to log, maybe by setting a flag in SampleConfiguration or through an environment variable. Whichever makes sense.

MushMal:
> Just throwing out a suggestion here: can we have a utility to log the signaling metrics periodically as well? Of course, this would not be done unless the user chooses to log, maybe by setting a flag in SampleConfiguration or through an environment variable. Whichever makes sense.

I think a better approach for this would be to enhance the samples to printout some of the metrics periodically. Though, I wouldn't tie the work up with this PR

src/source/Signaling/Signaling.c
```diff
@@ -328,6 +342,9 @@ STATUS signalingSendMessageSync(PSignalingClient pSignalingClient, PSignalingMes
                               pSignalingMessage->payload, pSignalingMessage->payloadLen,
                               pSignalingMessage->correlationId, 0));
 
+    // Update the internal diagnostics only after successfully sending
+    ATOMIC_INCREMENT(&pSignalingClient->diagnostics.numberOfMessagesSent);
+
 CleanUp:
 
     CHK_LOG_ERR(retStatus);
```
lherman-cs, line 350: I think we should check the error here, and increment `numberOfRuntimeErrors`

MushMal: Not necessary for now as the sending of the messages is a synchronous operation and the error will be returned in the call itself making it not a runtime error but rather an operation failure. I will be thinking a little more about the types of errors to add - something like general errors which will account for any operation failures to increment the counter. Not sure how useful that will be though. Perhaps it makes sense, let me do that

PR #594, opened 2020-07-06
configure libwebsocket to fallback to gethostbyname if getaddrinfo fails

chehefen:
*Issue #, if available:*

*Description of changes:*


lherman-cs:
@chehefen I think we should also add a fallback on our side too, https://github.com/awslabs/amazon-kinesis-video-streams-webrtc-sdk-c/blob/master/src/source/Ice/Network.c#L290-L330. Because if `getaddrinfo` fails on lws, very likely it'll fail also on our side.

PR #595, opened 2020-07-06
Fix line length to include all chars

lherman-cs:
This is only a minor fix for line length calculation. 

Let say we have this string: "m=audio\n"

Before: lineLen = 6 (missing the last char, 'o')

After: lineLen = 7

### DIFF
--- a/src/source/Sdp/Serialize.c
+++ b/src/source/Sdp/Serialize.c
@@ -10,7 +10,7 @@ STATUS parseMediaName(PSessionDescription pSessionDescription, PCHAR pch, UINT32
 
     STRNCPY(pSessionDescription->mediaDescriptions[pSessionDescription->mediaCount].mediaName,
             (pch + SDP_ATTRIBUTE_LENGTH),
-            MIN(MAX_SDP_MEDIA_NAME_LENGTH, lineLen - (SDP_ATTRIBUTE_LENGTH - 1)));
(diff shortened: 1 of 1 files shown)


src/source/Sdp/Serialize.c
```diff
@@ -107,43 +106,43 @@ STATUS serializeSessionDescription(PSessionDescription pSessionDescription, PCHA
...
-                        MIN(MAX_SDP_MEDIA_NAME_LENGTH, lineLen - (SDP_ATTRIBUTE_LENGTH - 1)));
+                        MIN(MAX_SDP_MEDIA_NAME_LENGTH, lineLen - SDP_ATTRIBUTE_LENGTH));
             }
 
             // SDP Phone number
             if (0 == STRNCMP(curr, SDP_PHONE_NUMBER_MARKER, (ARRAY_SIZE(SDP_PHONE_NUMBER_MARKER) - 1))) {
                 STRNCPY(pSessionDescription->phoneNumber,
                         (curr + SDP_ATTRIBUTE_LENGTH),
-                        MIN(MAX_SDP_MEDIA_NAME_LENGTH, lineLen - (SDP_ATTRIBUTE_LENGTH - 1)));
+                        MIN(MAX_SDP_MEDIA_NAME_LENGTH, lineLen - SDP_ATTRIBUTE_LENGTH));
```
MushMal, line 145: Not sure how this works...

Issue #587, opened 2020-07-03
[BUG] setRemoteDescription fails with STATUS_INVALID_DIGIT if using CRLF


gerwin3:
**Describe the bug**
When using CRLF line-endings rather than LF *some* SDPs will cause `setRemoteDescription` to fail with `STATUS_INVALID_DIGIT`.

**SDK version number**
58c77c45839547db5b780366b09ad18172146594

**To Reproduce**

The following code demonstrates the issue (`EXPECT_EQ(setRemoteDescription(pRtcPeerConnection, &rtcSessionDescriptionInit), STATUS_SUCCESS)` fails!).

Note that either
 1) removing the LF -> CRLF conversion **OR**
 2) removing just the last line of the SDP contents will mitigate the issue (???).

```cpp
  char remoteSessionDescription[10000] = R"(v=0
o=- 8372277215766250349 2 IN IP4 127.0.0.1
s=-
t=0 0
a=group:BUNDLE 0 1
a=msid-semantic: WMS
m=audio 9 UDP/TLS/RTP/SAVPF 111 103 104 9 0 8 106 105 13 110 112 113 126
c=IN IP4 0.0.0.0
a=rtcp:9 IN IP4 0.0.0.0
a=ice-ufrag:TKNK
a=ice-pwd:YATDTbOgdvtG/11GLGAPsa1V
a=ice-options:trickle
a=fingerprint:sha-256 88:E5:FA:D4:D0:41:8E:A6:94:EB:ED:93:C4:AC:2B:CF:B8:16:BB:48:12:4D:AD:FB:45:CF:40:1A:B3:0A:B6:EC
a=setup:actpass
a=mid:0
a=extmap:1 urn:ietf:params:rtp-hdrext:ssrc-audio-level
a=extmap:2 http://www.webrtc.org/experiments/rtp-hdrext/abs-send-time
a=extmap:3 http://www.ietf.org/id/draft-holmer-rmcat-transport-wide-cc-extensions-01
a=extmap:4 urn:ietf:params:rtp-hdrext:sdes:mid
a=extmap:5 urn:ietf:params:rtp-hdrext:sdes:rtp-stream-id
a=extmap:6 urn:ietf:params:rtp-hdrext:sdes:repaired-rtp-stream-id
a=recvonly
a=rtcp-mux
a=rtpmap:111 opus/48000/2
a=rtcp-fb:111 transport-cc
[...]

PR #600, opened 2020-07-07
Improve SDP conformation with RFC

lherman-cs:
Improve SDP conformation with RFC

Resolves https://github.com/awslabs/amazon-kinesis-video-streams-webrtc-sdk-c/issues/587

Changes:
 * SDP parser now works with CRLF and LF
* SDP line separator is now using CRLF to conform with RFC, https://tools.ietf.org/html/rfc4566#section-5
* Add unit tests for CRLF cases

### DIFF
--- a/src/source/PeerConnection/SessionDescription.c
+++ b/src/source/PeerConnection/SessionDescription.c
@@ -27,7 +27,11 @@ STATUS serializeSessionDescriptionInit(PRtcSessionDescriptionInit pSessionDescri
     while ((next = STRNCHR(curr, (UINT32) (tail - curr), '\n')) != NULL) {
         lineLen = (UINT32) (next - curr);
 
-         amountWritten = SNPRINTF(
+        if (lineLen > 0 && curr[lineLen - 1] == '\r') {
+            lineLen--;
+        }
+
+        amountWritten = SNPRINTF(
                 sessionDescriptionJSON + *sessionDescriptionJSONLen,
                 sessionDescriptionJSON == NULL ? 0 : inputSize - *sessionDescriptionJSONLen,
                 "%*.*s%s",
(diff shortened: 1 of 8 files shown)


src/source/PeerConnection/SessionDescription.c
```diff
@@ -85,23 +89,25 @@ STATUS deserializeSessionDescriptionInit(PCHAR sessionDescriptionJSON, UINT32 se
...
+                lineLen = (INT32) (next - curr);
 
-                    last = curr;
+                // there are escape characters before \n and \r, so we need to move back 1 more character
+                if (lineLen > 1 && curr[lineLen - 2] == '\\' && curr[lineLen - 1] == 'r') {
+                    lineLen -= 2;
                 }
-            }
+
+                MEMCPY((pSessionDescriptionInit->sdp) + j, curr, lineLen);
```
MushMal, line 104: You might want to use lineLen * SIZEOF(CHAR) here

src/source/PeerConnection/SessionDescription.c
```diff
@@ -85,23 +89,25 @@ STATUS deserializeSessionDescriptionInit(PCHAR sessionDescriptionJSON, UINT32 se
...
+                j += (lineLen + 2);
+                pSessionDescriptionInit->sdp[j - 2] = '\r';
+                pSessionDescriptionInit->sdp[j - 1] = '\n';
+
+                curr = next + 2;
```
MushMal, line 109: Ideally, you should add some comments describing that here you are skipping the source '\' 'n' characters. whereas in LN 105 you are adding 2 as you are adding \r and \n. Some readers might not have enough context on \r\n vs \n

src/source/PeerConnection/SessionDescription.c
lherman-cs: Yep, makes sense. Done.

src/source/PeerConnection/SessionDescription.c
lherman-cs: Ah, you're right. I'll add some comments.

lherman-cs: done

PR #602, opened 2020-07-09
free socketConnection in TURN_STATE_CLEAN_UP

### LINKED ISSUE #596: [QUESTION] How to reduce CPU usage?
I have ported Kinesis to a camera we are developing, but I found that it is using a lot of CPU. Is there some way to make Kinesis use less CPU power?
We are just reading H264 packets from our image pipeline and sending them trough Kinesis.
Our implementation is basically the same as kvsWebRTCClientMaster.c, with small changes in order to read H264 packets from a shared memory source, instead of the file sequence.

Many thanks for your attention,

chehefen:
*Issue #, if available:*
fix #596 
*Description of changes:*

### DIFF
--- a/src/source/Ice/TurnConnection.c
+++ b/src/source/Ice/TurnConnection.c
@@ -96,8 +96,10 @@ STATUS freeTurnConnection(PTurnConnection* ppTurnConnection)
     }
 
     // shutdown control channel
-    CHK_LOG_ERR(connectionListenerRemoveConnection(pTurnConnection->pConnectionListener, pTurnConnection->pControlChannel));
-    CHK_LOG_ERR(freeSocketConnection(&pTurnConnection->pControlChannel));
+    if (pTurnConnection->pControlChannel) {
+        CHK_LOG_ERR(connectionListenerRemoveConnection(pTurnConnection->pConnectionListener, pTurnConnection->pControlChannel));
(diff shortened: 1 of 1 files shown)


src/source/Ice/TurnConnection.c
```diff
@@ -96,8 +96,10 @@ STATUS freeTurnConnection(PTurnConnection* ppTurnConnection)
     }
 
     // shutdown control channel
-    CHK_LOG_ERR(connectionListenerRemoveConnection(pTurnConnection->pConnectionListener, pTurnConnection->pControlChannel));
-    CHK_LOG_ERR(freeSocketConnection(&pTurnConnection->pControlChannel));
+    if (pTurnConnection->pControlChannel) {
```
MushMal, line 99: Explicitly check against NULL - something like if (pTurnConnection->pControlChannel != NULL)

sharadrj: I am also interested to know how this change helps in reducing CPU consumption.

chehefen:
> Can you provide more info on what was the cause of a higher CPU usage? It's not immediately clear fro the change. The change looks good though with a minor comment.

After we've deallocated the unused turn connections, we were not removing the turn sockets from the connectionListener so they continued to be listened in select(). I'm not sure what the behavior would be in that case perhaps select() error out immediately but it caused the high cpu usage.

PR #603, opened 2020-07-09
rtcp sender reports

zhuker:
*Issue #, if available:*

implemented https://tools.ietf.org/html/rfc3550#section-6.4.1
as part of outbound metrics implementation

*Description of changes:*
sends rtcp sender reports
calculates some outgoing metrics

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.

PR #607, opened 2020-07-10
[WIP] Outbound stats

zhuker:
work in progress implementing 
https://www.w3.org/TR/webrtc-stats/#dom-rtcoutboundrtpstreamstats

NOTE: this PR assumes "rtcp sender reports" PR is landed https://github.com/awslabs/amazon-kinesis-video-streams-webrtc-sdk-c/pull/603


src/include/com/amazonaws/kinesis/video/webrtcclient/Include.h
```diff
@@ -1321,6 +1321,25 @@ typedef struct {
     RtcStatsObject rtcStatsObject; //!< Object that is populated by the SDK on request
 } RtcStats, *PRtcStats;
 
+typedef struct {
```
MushMal, line 1324: I have multiple issues with this approach.

1) Most importantly, why are we carrying the extra params needed for the stats in the frame itself? Do we assume the encoder will be changing the bitrate and the resolution with each frame?
2) The backward compat will only work if we modify the version of the underlying frame struct to some sentinel version that will allow the code to access the bits following it.

How else can the encoder communicate the necessary stats to the SDK? Perhaps a new API? In a real-life situation, are there any encoders that actually expose the bitrate and the frame encoder time? I haven't seen any. 

The width and height can be extracted from the codec private data - namely, from SPS for H264.

src/include/com/amazonaws/kinesis/video/webrtcclient/Stats.h
```diff
@@ -224,12 +224,38 @@ typedef struct {
     UINT32 selectedCandidatePairChanges; //!< The number of times that the selected candidate pair of this transport has changed
 } RtcTransportStats, *PRtcTransportStats;
 
+typedef char DOMString[MAX_STATS_STRING_LENGTH + 1];
```
MushMal, line 227: char -> CHAR. Also, think of changing to a more generic name perhaps as this can be used by other stats

src/source/PeerConnection/Rtcp.c
```diff
@@ -2,31 +2,125 @@
 
 #include "../Include_i.h"
 
+STATUS findTransceiverBySsrc(PKvsPeerConnection pKvsPeerConnection, PKvsRtpTransceiver* ppTransceiver)
```
MushMal, line 5: Can we assume the parameters are sanitized for this call?

src/source/PeerConnection/PeerConnection.c
```diff
@@ -474,7 +474,66 @@ STATUS generateJSONSafeString(PCHAR pDst, UINT32 len)
...
+        octetCount  = pKvsRtpTransceiver->sender.outboundRtpStreamStats.sentRtpStreamStats.bytesSent;
+        DLOGD("sender report %u %" PRIu64 " %" PRIu64 " : %u packets %u bytes", ssrc, ntpTime, rtpTime, packetCount, octetCount);
+        packetLen = RTCP_PACKET_HEADER_LEN + 24;
+        allocSize = packetLen + SRTP_AUTH_TAG_OVERHEAD;
+        CHK(NULL != (rawPacket = (PBYTE) MEMALLOC(allocSize)), STATUS_NOT_ENOUGH_MEMORY);
+        rawPacket[0]                       = RTCP_PACKET_VERSION_VAL << 6;
+        rawPacket[RTCP_PACKET_TYPE_OFFSET] = RTCP_PACKET_TYPE_SENDER_REPORT;
+        putUnalignedInt16BigEndian(rawPacket + RTCP_PACKET_LEN_OFFSET,
+                                   (packetLen / RTCP_PACKET_LEN_WORD_SIZE) - 1); // The length of this RTCP packet in 32-bit words minus one
+        putUnalignedInt32BigEndian(rawPacket + 4, ssrc);
```
MushMal, line 514: Minor: Could we simply do SIZEOF(UINT32) instead of +4, etc? Normally, for packing, I use a current pointer which gets incremented by the size of the thing that I just copy. For example

put(pCurPtr, ssrc);
pCurPtr += SIZEOF(UINT32);
put(pCurPtr, ntpTime);
pCurPtr += SIZEOF(UINT64);
...

src/source/PeerConnection/Rtcp.c
```diff
@@ -2,31 +2,125 @@
...
+                } else if (rtcpPacket.header.receptionReportCount == RTCP_PSFB_PLI) {
+                    CHK_STATUS(onRtcpPLIPacket(&rtcpPacket, pKvsPeerConnection));
+                } else {
+                    DLOGW("unhandled packet type RTCP_PACKET_TYPE_PAYLOAD_SPECIFIC_FEEDBACK");
+                }
+                break;
+            case RTCP_PACKET_TYPE_SENDER_REPORT:
+                // https://tools.ietf.org/html/rfc3550#section-6.4.1
+                if (rtcpPacket.payloadLength == RTCP_PACKET_SENDER_REPORT_MINLEN) {
+                    senderSSRC = getUnalignedInt32BigEndian(rtcpPacket.payload);
```
MushMal, line 62: same comment about the arranging the bits

src/source/PeerConnection/Rtcp.c
```diff
@@ -2,31 +2,125 @@
...
+                if (rtcpPacket.payloadLength == RTCP_PACKET_RECEIVER_REPORT_MINLEN) {
+                    senderSSRC          = getUnalignedInt32BigEndian(rtcpPacket.payload);
+                    ssrc1               = getUnalignedInt32BigEndian(rtcpPacket.payload + 4);
+
+                    CHK_STATUS(findTransceiverBySsrc(pKvsPeerConnection, &pTransceiver));
+                    CHK(pTransceiver != NULL, STATUS_SUCCESS); // TODO: change to some "not found" status
+
+                    fractionLost        = rtcpPacket.payload[8];
+                    cumulativeLost      = ((UINT32) getUnalignedInt32BigEndian(rtcpPacket.payload + 8)) & 0x00ffffffu;
+                    extHiSeqNumReceived = getUnalignedInt32BigEndian(rtcpPacket.payload + 12);
```
MushMal, line 86: same

src/include/com/amazonaws/kinesis/video/webrtcclient/Stats.h
```diff
@@ -281,11 +307,12 @@ typedef struct {
  */
 typedef struct {
     CHAR localId[MAX_STATS_STRING_LENGTH + 1]; //!< Used to look up RTCOutboundRtpStreamStats for the SSRC
-    UINT64 roundTripTime; //!< Estimated round trip time (seconds) for this SSRC based on the RTCP timestamps
-    UINT64 totalRoundTripTime; //!< The cumulative sum of all round trip time measurements in seconds since the beginning of the session
-    UINT64 fractionLost; //!< The fraction packet loss reported for this SSRC
-    UINT64 reportsReceived; //!< Total number of RTCP RR blocks received for this SSRC
-    UINT64 roundTripTimeMeasurements; //!< Total number of RTCP RR blocks received for this SSRC that contain a valid round trip time
+    volatile SIZE_T roundTripTime; //!< Estimated round trip time (milliseconds) for this SSRC based on the RTCP timestamps
```
MushMal, line 310: Important! The volatile variables will be used in atomic operations. Those should be machine word aligned. I see the localId and later fractionLost variable that might break the alignment. Perhaps, move the volatile SIZE_Ts up front to ensure proper alignment. The same goes for all of the parameters.

src/source/PeerConnection/Rtp.h
```diff
@@ -58,7 +74,7 @@ STATUS freeKvsRtpTransceiver(PKvsRtpTransceiver*);
 
 STATUS kvsRtpTransceiverSetJitterBuffer(PKvsRtpTransceiver, PJitterBuffer);
 
-UINT64 convertTimestampToRTP(UINT64, UINT64);
+#define convertTimestampToRTP(clockRate, pts) (pts * clockRate / HUNDREDS_OF_NANOS_IN_A_SECOND)
```
MushMal, line 77: Macros should be capitalized and underscored - something like CONVERT_TIMESTAMP_TO_RTP

MushMal: Also, it's a good practice to wrap the params in parenthesis 

#define CONVERT_TIMESTAMP_TO_RTP(clockRate, pts) ((pts) * (clockRate) / HUNDREDS_OF_NANOS_IN_A_SECOND)

src/include/com/amazonaws/kinesis/video/webrtcclient/Include.h
zhuker: > encoder will be changing the bitrate and the resolution with each frame?
changing target bitrate and even resolutiin is a part of the spec

src/include/com/amazonaws/kinesis/video/webrtcclient/Stats.h
zhuker: `DOMString` type is a part of webrtc spec, decided why not just use it

src/source/PeerConnection/Rtcp.c
zhuker: good question
i dont think we should assume that

src/source/PeerConnection/Rtp.h
zhuker: ok

src/include/com/amazonaws/kinesis/video/webrtcclient/Include.h
zhuker: i dont like this approach either
but couldnt come up with anything better
ideas?
separate API?

src/include/com/amazonaws/kinesis/video/webrtcclient/Stats.h
MushMal: It's a good idea in my opinion. The comment is mostly about using the internal type CHAR

src/include/com/amazonaws/kinesis/video/webrtcclient/Include.h
MushMal: Certainly, a separate API is a lot better than overloading the frame - something like onEncoderStatsChanged or similar. What I am not certain about is how this will be wired in but I assume some platforms might allow getting these stats from the encoder.

src/source/PeerConnection/PeerConnection.c
zhuker: i add explicit offsets to quickly debug what was written at which offset

src/include/com/amazonaws/kinesis/video/webrtcclient/Include.h
zhuker: `onEncoderStatsChanged` makes sense

src/source/PeerConnection/PeerConnection.c
MushMal: Really up to you. Some of these are minor comments/suggestions

src/include/com/amazonaws/kinesis/video/webrtcclient/Stats.h
zhuker: is there any pragma or attribute to force 64bit alignment?
as you can see i update stats with atomics, didnt want to add yet another lock for stats, why risk deadlock and/or slowdown when only updating stats. right?

src/source/PeerConnection/Rtcp.c
zhuker: i am pretty sure i'll be changing this part of code quite a bit soon

src/include/com/amazonaws/kinesis/video/webrtcclient/Stats.h
MushMal: I don't think there is a single nice portable way to enforce a 64 bit alignment - you might also have issues on 32 bit platforms. This said, you can always move the volatiles up on top - this is what we do with every structure we have that has members which will be used with atomics.

The other question is whether you can have generic atomics for 64 bit type and the answer is NO. Some of the compilers/platforms/libraries (most don't support) allowing 64bit atomic operations on 32bit internally actually use interlocking. In our case, due to the platform and compiler portability we only assume SIZE_T type atomics operations.

zhuker: > move the volatiles up on top
good idea
> 64bit atomic operations on 32bit internally actually use interlocking
just a thought - do you know how JVM implements it's `AtomicLong` ? might make sense to look into it....

src/source/PeerConnection/Rtcp.c
```diff
@@ -2,31 +2,125 @@
...
+
+    CHK_STATUS(doubleListGetHeadNode(pKvsPeerConnection->pTransceievers, &pCurNode));
+    while (pCurNode != NULL) {
+        CHK_STATUS(doubleListGetNodeData(pCurNode, &item));
+        pTransceiver = (PKvsRtpTransceiver) item;
+        if (pTransceiver->sender.ssrc || pTransceiver->sender.rtxSsrc) {
+            break;
+        }
+    }
+    *ppTransceiver = pTransceiver;
```
disa6302, line 20: What if `pTransceiver` is NULL? Any repurcussions while making the call to this function?

src/source/PeerConnection/Rtp.c
```diff
@@ -122,27 +122,47 @@ STATUS transceiverOnBandwidthEstimation(PRtcRtpTransceiver pRtcRtpTransceiver, U
...
+    if (MEDIA_STREAM_TRACK_KIND_VIDEO == pKvsRtpTransceiver->sender.track.kind) {
+        ATOMIC_INCREMENT(&pKvsRtpTransceiver->sender.outboundRtpStreamStats.framesEncoded);
+        if (0 != (pFrame->flags & FRAME_FLAG_KEY_FRAME)) {
+            ATOMIC_INCREMENT(&pKvsRtpTransceiver->sender.outboundRtpStreamStats.keyFramesEncoded);
+        }
+        if (pKvsRtpTransceiver->sender.lastKnownFrameCountTime == 0) {
+            pKvsRtpTransceiver->sender.lastKnownFrameCountTime = now;
+            pKvsRtpTransceiver->sender.lastKnownFrameCount     = pKvsRtpTransceiver->sender.outboundRtpStreamStats.framesEncoded;
+        } else if (now - pKvsRtpTransceiver->sender.lastKnownFrameCountTime > HUNDREDS_OF_NANOS_IN_A_SECOND) {
+            UINT64 frames = pKvsRtpTransceiver->sender.outboundRtpStreamStats.framesEncoded - pKvsRtpTransceiver->sender.lastKnownFrameCount;
```
disa6302, line 156: Btw, I have not set clang checks with aligned alignments. So, this would break.

src/source/PeerConnection/Rtp.c
```diff
@@ -122,27 +122,47 @@ STATUS transceiverOnBandwidthEstimation(PRtcRtpTransceiver pRtcRtpTransceiver, U
...
+            UINT64 time   = now - pKvsRtpTransceiver->sender.lastKnownFrameCountTime;
+            DOUBLE fps    = (DOUBLE)(frames * HUNDREDS_OF_NANOS_IN_A_SECOND) / (DOUBLE) time;
+            ATOMIC_STORE(&pKvsRtpTransceiver->sender.outboundRtpStreamStats.framesPerSecond, fps);
+            pKvsRtpTransceiver->sender.lastKnownFrameCountTime = now;
+            pKvsRtpTransceiver->sender.lastKnownFrameCount     = pKvsRtpTransceiver->sender.outboundRtpStreamStats.framesEncoded;
+        }
+        ATOMIC_STORE(&pKvsRtpTransceiver->sender.outboundRtpStreamStats.frameWidth, pRtcFrame->width);
```
disa6302, line 163: Where is this `pRtcFrame->width` getting set? 

Same with height and targetBitrate.

src/source/PeerConnection/Rtp.c
zhuker: ok

src/source/PeerConnection/Rtp.c
zhuker: externally
@MushMal offered to add separate api call for this like `onEncoderSettingsChanged`

src/source/PeerConnection/Rtcp.c
zhuker: noted

src/source/PeerConnection/PeerConnection.c
```diff
@@ -474,7 +474,66 @@ STATUS generateJSONSafeString(PCHAR pDst, UINT32 len)
...
+        putUnalignedInt64BigEndian(rawPacket + 8, ntpTime);
+        putUnalignedInt32BigEndian(rawPacket + 16, rtpTime);
+        putUnalignedInt32BigEndian(rawPacket + 20, packetCount);
+        putUnalignedInt32BigEndian(rawPacket + 24, octetCount);
+        CHK_STATUS(encryptRtcpPacket(pKvsPeerConnection->pSrtpSession, rawPacket, (PINT32) &packetLen));
+        CHK_STATUS(iceAgentSendPacket(pKvsPeerConnection->pIceAgent, rawPacket, packetLen));
+    }
+
+    delay = 100 + (RAND() % 200);
+    DLOGD("next sender report %u in %" PRIu64 " msec", ssrc, delay);
```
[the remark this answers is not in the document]
zhuker: `DLOGV` instead?

src/include/com/amazonaws/kinesis/video/webrtcclient/Stats.h
```diff
@@ -224,12 +224,38 @@ typedef struct {
     UINT32 selectedCandidatePairChanges; //!< The number of times that the selected candidate pair of this transport has changed
 } RtcTransportStats, *PRtcTransportStats;
 
+typedef char DOMString[MAX_STATS_STRING_LENGTH + 1];
+
+// https://www.w3.org/TR/webrtc-stats/#dom-rtcrtpstreamstats
+typedef struct {
+    UINT32 ssrc; //!< The 32-bit unsigned integer value per [RFC3550] used to identify the source of the stream of RTP packets that this stats object
+                 //!< concerns.
+    DOMString kind; //!< Either "audio" or "video". This MUST match the media type part of the information in the corresponding codecType member of
```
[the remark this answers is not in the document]
zhuker: i actually copied length from you, everywhere else you are using `CHAR[MAX_STATS_STRING_LENGTH + 1]`
e.g.
`CHAR trackId[MAX_STATS_STRING_LENGTH + 1]; //!< ID representing current track attached to the sender of the stream`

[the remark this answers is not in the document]
zhuker: i am not very concerned about memory consumption since its negligible compared to media being copied
these structures are allocated once

src/source/PeerConnection/PeerConnection.c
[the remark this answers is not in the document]
disa6302: DLOGV should work as well.

src/include/com/amazonaws/kinesis/video/webrtcclient/Stats.h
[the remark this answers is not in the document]
disa6302: Yes. I had set everything MAX_STATS_STRING_LENGTH at the time with the goal to change it ito constant accurate f the lengths were not variable. But, fine, I have no strong opinion here.

zhuker:
@MushMal introduced `RtcEncoderStats` and `updateEncoderStats` instead of extending frame

src/include/com/amazonaws/kinesis/video/webrtcclient/Stats.h
MushMal: I am not entirely sure but it can't be using CPU intrinsics on 32 bit. Likely on nix-es it's wrapped by a futex (similar to synchronized keyword application) and on Windows-es must be using RW locks

src/include/com/amazonaws/kinesis/video/webrtcclient/Stats.h
```diff
@@ -247,34 +265,34 @@ typedef struct {
     CHAR rid[MAX_STATS_STRING_LENGTH + 1]; //!< Exposes the rid encoding parameter of this RTP stream if it has been set, otherwise it is undefined
     CHAR encoderImplementation[MAX_STATS_STRING_LENGTH + 1]; //!< Identifies the encoder implementation used.
     UINT32 packetsDiscardedOnSend; //!< Total number of RTP packets for this SSRC that have been discarded due to socket errors
-    UINT32 framesSent; //!< Only valid for video. Represents the total number of frames sent on this RTP stream
-    UINT32 hugeFramesSent; //!< Only valid for video. Represents the total number of huge frames sent by this RTP stream
-                           //!< Huge frames have an encoded size at least 2.5 times the average size of the frames
-    UINT32 framesEncoded; //!< Only valid for video. It represents the total number of frames successfully encoded for this RTP media stream
-    UINT32 keyFramesEncoded; //!< Only valid for video. It represents the total number of key frames encoded successfully in the RTP Stream
+    volatile UINT64 framesSent; //!< Only valid for video. Represents the total number of frames sent on this RTP stream
```
MushMal, line 268: Seems that these are also used in atomics operations. These should also be moved up front and we should use SIZE_T as that's what the machine word will be.

If we must use UINT64 then we do need to use a LOCK. This is unfortunate but this is also the reason I had to introduce a lock in Signaling metrics to handle interlocking for the UINT64 types of variables.

src/include/com/amazonaws/kinesis/video/webrtcclient/Stats.h
```diff
@@ -247,34 +265,34 @@ typedef struct {
...
     UINT32 nackCount; //!< Count the total number of Negative ACKnowledgement (NACK) packets received by this sender.
     UINT32 firCount; //!< Only valid for video. Count the total number of Full Intra Request (FIR) packets received by this sender
     UINT32 pliCount; //!< Only valid for video. Count the total number of Picture Loss Indication (PLI) packets received by this sender
     UINT32 sliCount; //!< Only valid for video. Count the total number of Slice Loss Indication (SLI) packets received by this sender
     UINT32 qualityLimitationResolutionChanges; //!< Only valid for video. The number of times that the resolution has changed because we are quality limited
     INT32 fecPacketsSent; //!< Total number of RTP FEC packets sent for this SSRC. Can also be incremented while sending FEC packets in band
-    UINT64 lastPacketSentTimestamp; //!< The timestamp at which the last packet was sent for this SSRC
-    UINT64 headerBytesSent; //!< Total number of RTP header and padding bytes sent for this SSRC
+    volatile UINT64 lastPacketSentTimestamp; //!< The timestamp in milliseconds at which the last packet was sent for this SSRC
```
MushMal, line 283: Yeah, seems like we need a lock.

zhuker:
> Seems that the only outstanding issue is the lock. @disa6302 @zhuker I think we should introduce a higher-level lock in PeerConnection perhaps that will control interlocking for all of the stats read/write that can't be atomized due to their type.

i think it should be on transceiver level

MushMal:
> > Seems that the only outstanding issue is the lock. @disa6302 @zhuker I think we should introduce a higher-level lock in PeerConnection perhaps that will control interlocking for all of the stats read/write that can't be atomized due to their type.
> 
> i think it should be on transceiver level

I tend to agree. Perhaps keeping it more granular is a better idea.

disa6302:
> > Seems that the only outstanding issue is the lock. @disa6302 @zhuker I think we should introduce a higher-level lock in PeerConnection perhaps that will control interlocking for all of the stats read/write that can't be atomized due to their type.
> 
> i think it should be on transceiver level

Transceiver level makes sense.

src/include/com/amazonaws/kinesis/video/webrtcclient/Stats.h
zhuker: i cant make myself to use locks

zhuker: c++11 has `std::atomic<uint64>`

zhuker: from AtomicLong.getAndAddLong
```
    public final long getAndAddLong(Object o, long offset, long delta) {
        long v;
        do {
            v = getLongVolatile(o, offset);
        } while (!weakCompareAndSetLong(o, offset, v, v + delta));
        return v;
    }
```
and then
```
UNSAFE_ENTRY(jboolean, Unsafe_CompareAndSwapLong(JNIEnv *env, jobject unsafe, jobject obj, jlong offset, jlong e, jlong x))
  UnsafeWrapper("Unsafe_CompareAndSwapLong");
  Handle p (THREAD, JNIHandles::resolve(obj));
  jlong* addr = (jlong*)(index_oop_from_field_offset_long(p(), offset));
  if (VM_Version::supports_cx8())
    return (jlong)(Atomic::cmpxchg(x, addr, e)) == e;
  else {
    jboolean success = false;
    ObjectLocker ol(p, THREAD);
    if (*addr == e) { *addr = x; success = true; }
    return success;
  }
```

zhuker: @MushMal

zhuker: uses `cmpxchg` if available and mutex otherwise

zhuker: `#include <stdatomic.h>` for plain C

MushMal: There are no 64bit intrinsic atomic operations on 32 bit systems. Some compiler backends allow type promotion to SSE2/SIMD larger register types which on the given system allow atomicity, others, generate spinlock code that handles synchronization. Our atomics can't do that - it's a limitation we need to deal with.

Now, this piece of code is of no concern when it comes to performance so I recommend:

1) For any 64 bit type representing counts, use SIZE_T directly
2) For any duration/time, cast it down to SIZE_T by exposing not 100ns granularity but something more meaningful - say in ms and be explicit about it
3) If there are types that do need 64 bit (which I don't see yet) we need to use a MUTEX

zhuker: ok i'll try to use locks but if it looks too ugly, i'll give atomics another shot

MushMal: Thanks @zhuker . Per my comment, see if you really need the 64 bit type or SIZE_T works OK in which case we could use atomics as is

disa6302:
@zhuker , 

One more thing I missed mentioning, can you add unit tests, wherever feasible, for the stats you are adding?

src/source/PeerConnection/Rtp.c
```diff
@@ -159,25 +163,26 @@ STATUS writeFrame(PRtcRtpTransceiver pRtcRtpTransceiver, PFrame pFrame)
...
-            pKvsRtpTransceiver->sender.lastKnownFrameCount     = pKvsRtpTransceiver->sender.outboundRtpStreamStats.framesEncoded;
+            pKvsRtpTransceiver->sender.lastKnownFrameCount = pKvsRtpTransceiver->sender.outboundRtpStreamStats.framesEncoded;
         } else if (now - pKvsRtpTransceiver->sender.lastKnownFrameCountTime > HUNDREDS_OF_NANOS_IN_A_SECOND) {
-            UINT64 frames = pKvsRtpTransceiver->sender.outboundRtpStreamStats.framesEncoded - pKvsRtpTransceiver->sender.lastKnownFrameCount;
-            UINT64 time   = now - pKvsRtpTransceiver->sender.lastKnownFrameCountTime;
-            DOUBLE fps    = (DOUBLE)(frames * HUNDREDS_OF_NANOS_IN_A_SECOND) / (DOUBLE) time;
-            ATOMIC_STORE(&pKvsRtpTransceiver->sender.outboundRtpStreamStats.framesPerSecond, fps);
-            pKvsRtpTransceiver->sender.lastKnownFrameCountTime = now;
-            pKvsRtpTransceiver->sender.lastKnownFrameCount     = pKvsRtpTransceiver->sender.outboundRtpStreamStats.framesEncoded;
+            UINT64 _frames = pKvsRtpTransceiver->sender.outboundRtpStreamStats.framesEncoded - pKvsRtpTransceiver->sender.lastKnownFrameCount;
```
MushMal, line 183: Perhaps name this something other than _frames and do the definition up at the beginning of the function with other params.

src/source/PeerConnection/Rtp.c
```diff
@@ -159,25 +163,26 @@ STATUS writeFrame(PRtcRtpTransceiver pRtcRtpTransceiver, PFrame pFrame)
...
+            UINT64 time = now - pKvsRtpTransceiver->sender.lastKnownFrameCountTime;
```
MushMal, line 184: Important! Aside from the previous comment, this needs to be sanitized against being 0 for the call below

src/source/PeerConnection/Rtp.c
```diff
@@ -295,6 +292,28 @@ STATUS writeFrame(PRtcRtpTransceiver pRtcRtpTransceiver, PFrame pFrame)
...
+    pKvsRtpTransceiver->sender.lastKnownFrameCountTime = now;
+    pKvsRtpTransceiver->sender.lastKnownFrameCount = pKvsRtpTransceiver->sender.outboundRtpStreamStats.framesEncoded;
+    pKvsRtpTransceiver->sender.outboundRtpStreamStats.sentRtpStreamStats.bytesSent += bytesSent;
+    pKvsRtpTransceiver->sender.outboundRtpStreamStats.sentRtpStreamStats.packetsSent++;
+    if (lastPacketSentTimestamp > 0)
+        pKvsRtpTransceiver->sender.outboundRtpStreamStats.lastPacketSentTimestamp = lastPacketSentTimestamp;
+    pKvsRtpTransceiver->sender.outboundRtpStreamStats.headerBytesSent += headerBytesSent;
+    pKvsRtpTransceiver->sender.outboundRtpStreamStats.framesSent += framesSent;
+    if (pKvsRtpTransceiver->sender.outboundRtpStreamStats.framesPerSecond > 0.0) {
+        DOUBLE avgFrameSize =
```
MushMal, line 310: Define up at the beginning

src/source/PeerConnection/Rtp.h
```diff
@@ -34,6 +34,7 @@ typedef struct {
     PRtpRollingBuffer packetBuffer;
     PRetransmitter retransmitter;
 
+    MUTEX statsLock;
```
MushMal, line 37: This needs to be freed at some stage. Take a loot at the API that frees the transceiver bits

src/source/PeerConnection/Rtp.c
zhuker: already sanitized in if `if (now - pKvsRtpTransceiver->sender.lastKnownFrameCountTime > HUNDREDS_OF_NANOS_IN_A_SECOND)`

MushMal: D'oh!

src/source/PeerConnection/Rtcp.c
```diff
@@ -73,6 +102,15 @@ STATUS onRtcpPacket(PKvsPeerConnection pKvsPeerConnection, PBYTE pBuff, UINT32 b
...
                         DLOGD("RTCP_PACKET_TYPE_RECEIVER_REPORT rttPropDelay %u msec", rttPropDelayMsec);
                     }
+
+                    MUTEX_LOCK(pTransceiver->sender.statsLock);
+                    pTransceiver->sender.remoteInboundRtpStreamStats.reportsReceived++;
+                    pTransceiver->sender.remoteInboundRtpStreamStats.fractionLost = (DOUBLE) fractionLost / 255.0;
+                    pTransceiver->sender.remoteInboundRtpStreamStats.roundTripTimeMeasurements++;
+                    pTransceiver->sender.remoteInboundRtpStreamStats.totalRoundTripTime += rttPropDelayMsec;
+                    pTransceiver->sender.remoteInboundRtpStreamStats.roundTripTime = rttPropDelayMsec;
+                    MUTEX_UNLOCK(pTransceiver->sender.statsLock);
```
zhuker, line 112: @MushMal 
locking code turned out to be not bad at all

MushMal: Yeah, indeed, it's not as bad as one would expect - your approach was fantastic with evaluating local vars which later get set in one go.

I think the only thing that's missing is freeing the mutex

MushMal: Just approved the changes. THANK YOU!!!

zhuker:
@MushMal waiting for @disa6302 to land #611 on master
i'll re PR this one

zhuker:
> @zhuker ,
> 
> One more thing I missed mentioning, can you add unit tests, wherever feasible, for the stats you are adding?

@disa6302 still working on it
are you merging the #611 ?

disa6302:
> > @zhuker ,
> > One more thing I missed mentioning, can you add unit tests, wherever feasible, for the stats you are adding?
> 
> @disa6302 still working on it
> are you merging the #611 ?

Waiting for the build to complete. Will merge once it is done.

disa6302:
Not sure why the PR closed. 
I have merged #611 into master @zhuker

disa6302:
Ah! It is because the branch got deleted in the other PR automatically. I restored the branch.

zhuker:
@disa6302 i am closing this PR, see https://github.com/awslabs/amazon-kinesis-video-streams-webrtc-sdk-c/pull/615 same thing but targets master

PR #612, opened 2020-07-14
Fix wrong ICE candidate priority calculation

lherman-cs:
The ice candidate calculation is wrong, the spec uses `^` symbol which represents power. But, in C, this symbol means XOR.


lherman-cs:
> Oh wow...

Exactly what I was thinking...

PR #613, opened 2020-07-14
fix edge case when deallocate turn allocation gives error response. d…

chehefen:
…ata race during ice restart when turn is involved.
- also still create permission and channel bind if needed in turnConnection ready state.
*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/src/source/Ice/ConnectionListener.c
+++ b/src/source/Ice/ConnectionListener.c
@@ -277,7 +277,7 @@ PVOID connectionListenerReceiveDataRoutine(PVOID arg)
             locked = FALSE;
 
             ATOMIC_STORE_BOOL(&pConnectionListener->connectionListChanged, FALSE);
-            CVAR_SIGNAL(pConnectionListener->removeConnectionComplete);
+            CVAR_BROADCAST(pConnectionListener->removeConnectionComplete);
         }
 
         for (i = 0; i < socketCount; ++i) {
--- a/src/source/Ice/TurnConnection.c
+++ b/src/source/Ice/TurnConnection.c
@@ -57,7 +57,6 @@ STATUS createTurnConnection(PIceServer pTurnServer, TIMER_QUEUE_HANDLE timerQueu
     pTurnConnection->allocationExpirationTime = INVALID_TIMESTAMP_VALUE;
     pTurnConnection->nextAllocationRefreshTime = 0;
     pTurnConnection->currentTimerCallingPeriod = DEFAULT_TURN_TIMER_INTERVAL_BEFORE_READY;
-    CHK_STATUS(doubleListCreate(&pTurnConnection->turnPeerList));
 
 CleanUp:
 
@@ -80,10 +79,9 @@ STATUS freeTurnConnection(PTurnConnection* ppTurnConnection)
     ENTERS();
     STATUS retStatus = STATUS_SUCCESS;
     PTurnConnection pTurnConnection = NULL;
-    PDoubleListNode pCurNode = NULL;
-    UINT64 data;
     PTurnPeer pTurnPeer = NULL;
     SIZE_T timerCallbackId;
+    UINT32 i;
 
     CHK(ppTurnConnection != NULL, STATUS_NULL_ARG);
     // free is idempotent
@@ -103,17 +101,10 @@ STATUS freeTurnConnection(PTurnConnection* ppTurnConnection)
     }
 
     // free transactionId store for each turn peer
-    CHK_LOG_ERR(doubleListGetHeadNode(pTurnConnection->turnPeerList, &pCurNode));
-    while (pCurNode != NULL) {
-        CHK_STATUS(doubleListGetNodeData(pCurNode, &data));
-        pCurNode = pCurNode->pNext;
-
-        pTurnPeer = (PTurnPeer) data;
+    for (i = 0; i < pTurnConnection->turnPeerCount; ++i) {
+        pTurnPeer = &pTurnConnection->turnPeerList[i];
         freeTransactionIdStore(&pTurnPeer->pTransactionIdStore);
     }
-    // free turn peers
-    CHK_LOG_ERR(doubleListClear(pTurnConnection->turnPeerList, TRUE));
-    CHK_LOG_ERR(doubleListFree(pTurnConnection->turnPeerList));
 
     if (IS_VALID_MUTEX_VALUE(pTurnConnection->lock)) {
         MUTEX_FREE(pTurnConnection->lock);
@@ -203,11 +194,9 @@ STATUS turnConnectionHandleStun(PTurnConnection pTurnConnection, PBYTE pBuffer, 
     CHAR ipAddrStr[KVS_IP_ADDRESS_STRING_BUFFER_LEN];
     BOOL locked = FALSE;
     ATOMIC_BOOL hasAllocation = FALSE;
-
-    PDoubleListNode pCurNode = NULL;
-    UINT64 data;
     PTurnPeer pTurnPeer = NULL;
     UINT64 currentTime;
+    UINT32 i;
 
     CHK(pTurnConnection != NULL, STATUS_NULL_ARG);
     CHK(pBuffer != NULL && bufferLen > 0, STATUS_INVALID_ARG);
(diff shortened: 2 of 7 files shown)


src/source/Ice/ConnectionListener.c
```diff
@@ -133,12 +133,12 @@ STATUS connectionListenerRemoveConnection(PConnectionListener pConnectionListene
      * the change. */
     while (ATOMIC_LOAD_BOOL(&pConnectionListener->listenerRoutineStarted) && !ATOMIC_LOAD_BOOL(&pConnectionListener->terminate) &&
            ATOMIC_LOAD_BOOL(&pConnectionListener->connectionListChanged) && STATUS_SUCCEEDED(cvarWaitStatus)) {
-        cvarWaitStatus =
-            CVAR_WAIT(pConnectionListener->removeConnectionComplete, pConnectionListener->lock, CONNECTION_AWAIT_CONNECTION_REMOVAL_TIMEOUT);
         /* CVAR_WAIT should never time out */
         if (STATUS_FAILED(cvarWaitStatus)) {
```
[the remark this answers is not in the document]
MushMal: Something is amiss here. We have a check in the while loop and then immediately check if it failed. This is an invariant now. The intent is likely to handle timeout, etc in a more graceful manner. Seems like the inner check if (STATUS_FAILED(cvarWaitStatus) is redundant

src/source/Ice/ConnectionListener.c
```diff
@@ -177,12 +177,13 @@ STATUS connectionListenerRemoveAllConnection(PConnectionListener pConnectionList
...
```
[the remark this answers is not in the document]
MushMal: same

src/source/Ice/TurnConnection.c
```diff
@@ -418,14 +408,11 @@ STATUS turnConnectionHandleStunError(PTurnConnection pTurnConnection, PBYTE pBuf
...
-            while (pCurNode != NULL) {
-                pTurnPeer = (PTurnPeer) pCurNode->data;
-                pCurNode = pCurNode->pNext;
+            for (i = 0; i < pTurnConnection->turnPeerCount; ++i) {
+                pTurnPeer = &pTurnConnection->turnPeerList[i];
                 if (transactionIdStoreHasId(pTurnPeer->pTransactionIdStore, pBuffer + STUN_PACKET_TRANSACTION_ID_OFFSET)) {
                     pTurnPeer->connectionState = TURN_PEER_CONN_STATE_FAILED;
-                    /* break the loop */
-                    pCurNode = NULL;
+                    break;
```
MushMal, line 415: I strongly recommend not using condition flow modification keywords inside switch statements.

src/source/Ice/TurnConnection.c
```diff
@@ -1000,22 +974,17 @@ STATUS turnConnectionStepState(PTurnConnection pTurnConnection)
...
+                if (pTurnConnection->turnPeerList[i].connectionState == TURN_PEER_CONN_STATE_BIND_CHANNEL ||
+                    pTurnConnection->turnPeerList[i].connectionState == TURN_PEER_CONN_STATE_READY) {
                     channelWithPermissionCount++;
                 }
-                totalPeerCount++;
             }
 
             // push back timeout if no peer is available yet
-            if (totalPeerCount == 0) {
+            if (pTurnConnection->turnPeerCount == 0) {
```
MushMal, line 987: I assume this is running under a lock

src/source/Ice/TurnConnection.c
```diff
@@ -1031,19 +1000,13 @@ STATUS turnConnectionStepState(PTurnConnection pTurnConnection)
...
         case TURN_STATE_BIND_CHANNEL:
 
-            CHK_STATUS(doubleListGetHeadNode(pTurnConnection->turnPeerList, &pCurNode));
-            while (pCurNode != NULL) {
-                CHK_STATUS(doubleListGetNodeData(pCurNode, &data));
-                pCurNode = pCurNode->pNext;
-
-                pTurnPeer = (PTurnPeer) data;
-                if (pTurnPeer->connectionState == TURN_PEER_CONN_STATE_READY) {
+            for (i = 0; i < pTurnConnection->turnPeerCount; ++i) {
```
MushMal, line 1003: this is locked right?

chehefen: yes

src/source/Ice/TurnConnection.c
chehefen: yes

src/source/Ice/ConnectionListener.c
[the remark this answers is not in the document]
chehefen: fixed

src/source/Ice/ConnectionListener.c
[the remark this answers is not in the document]
chehefen: fixed

src/source/Ice/TurnConnection.c
```diff
@@ -57,7 +57,6 @@ STATUS createTurnConnection(PIceServer pTurnServer, TIMER_QUEUE_HANDLE timerQueu
     pTurnConnection->allocationExpirationTime = INVALID_TIMESTAMP_VALUE;
     pTurnConnection->nextAllocationRefreshTime = 0;
     pTurnConnection->currentTimerCallingPeriod = DEFAULT_TURN_TIMER_INTERVAL_BEFORE_READY;
```
MushMal, line 59: Don't you need to initialize the entries in the TurnPeer array now? If MEMCALLOC takes care of it, could you please comment it? Though, from what I understand, you might need to set the status of the turn peer right?

src/source/Ice/TurnConnection.c
```diff
@@ -629,32 +615,28 @@ STATUS turnConnectionAddPeer(PTurnConnection pTurnConnection, PKvsIpAddress pPee
 
     /* check for duplicate */
     CHK(turnConnectionGetPeerWithIp(pTurnConnection, pPeerAddress) == NULL, retStatus);
-    CHK_STATUS(doubleListGetNodeCount(pTurnConnection->turnPeerList, &peerCount));
-    CHK_WARN(peerCount < DEFAULT_TURN_MAX_PEER_COUNT, STATUS_INVALID_OPERATION, "Add peer failed. Max peer count reached");
+    CHK_WARN(pTurnConnection->turnPeerCount < DEFAULT_TURN_MAX_PEER_COUNT, STATUS_INVALID_OPERATION, "Add peer failed. Max peer count reached");
```
MushMal, line 618: Perhaps I am not understanding this. So you have an array of turn peers. It should get initialized to an initial value at first with status being set to something like not_existent. Then, you add a new peer. If we never "delete" a peer then adding it at the turnPeerCount is good but we might delete some turn peers making "holes" in the array right?

src/source/Ice/TurnConnection.c
```diff
@@ -629,32 +615,28 @@ STATUS turnConnectionAddPeer(PTurnConnection pTurnConnection, PKvsIpAddress pPee
...
-
-    CHK_STATUS(doubleListInsertItemTail(pTurnConnection->turnPeerList, (UINT64) pTurnPeer));
     pTurnPeer = NULL;
 
 CleanUp:
 
     if (STATUS_FAILED(retStatus) && pTurnPeer != NULL) {
         freeTransactionIdStore(&pTurnPeer->pTransactionIdStore);
-        MEMFREE(pTurnPeer);
+        pTurnConnection->turnPeerCount--;
```
MushMal, line 639: So it does seem that we always add/remove to the tail but that won't account for dynamic adding/freeing peers.

src/source/Ice/TurnConnection.h
```diff
@@ -123,7 +123,8 @@ struct __TurnConnection {
 
     PSocketConnection pControlChannel;
 
-    PDoubleList turnPeerList;
+    TurnPeer turnPeerList[DEFAULT_TURN_MAX_PEER_COUNT];
```
MushMal, line 126: I really like that we are using an array instead of dynamic list.

src/source/Ice/TurnConnection.c
chehefen: initially there are zero peers so the value isnt used anyways. Once a peer get added, its status is set properly

src/source/Ice/TurnConnection.c
chehefen: we dont delete turn peers other than here when it fails to create transaction store due to OOM.

src/source/Ice/TurnConnection.c
chehefen: We dont delete peers because there is no explicit signal that a peer has gone bad.

PR #623, opened 2020-07-16
Remove unnecessary allocation by updating in-place

lherman-cs:
It's unnecessary to create a new table to simply update each value in the table. 


MushMal (dismissed):
Very cool

disa6302 (dismissed):

Issue #618, opened 2020-07-15
[QUESTION] Sendto() failed with errno Network is unreachable


Lion-blade:
In a wireless network. The live view on the AWS console via kvsWebRtcMaster.  After 10 -15 seconds the preview terminates and returns the following exception logs.

Log:
11:45:50  2020-07-15 03:49:10 DEBUG   lwsWssCallbackRoutine(): Client is writable
11:46:00  2020-07-15 03:49:20 DEBUG   lwsWssCallbackRoutine(): Client is writable
11:46:01  2020-07-15 03:49:21 DEBUG   socketSendDataWithRetry(): sendto() failed with errno Network is unreachable
11:46:01  2020-07-15 03:49:21 DEBUG   socketSendDataWithRetry(): Close socket 51
11:46:01  2020-07-15 03:49:21 DEBUG   socketSendDataWithRetry(): Failed to send data. Bytes sent 0. Data len 188. Retry count 0
11:46:01  2020-07-15 03:49:21 ERROR   socketSendDataWithRetry(): operation returned status code: 0x5800001a
11:46:01  2020-07-15 03:49:21 ERROR   iceUtilsSendData(): operation returned status code: 0x5800001a
11:46:01  2020-07-15 03:49:21 WARN    turnConnectionSendData(): iceUtilsSendData failed with 0x5800001a
11:46:01  2020-07-15 03:49:21 WARN    socketConnectionSendData(): Failed to send data. Socket closed already
11:46:01  2020-07-15 03:49:21 ERROR   iceUtilsSendData(): operation returned status code: 0x58000022
11:46:01  2020-07-15 03:49:21 WARN    turnConnectionSendData(): iceUtilsSendData failed with 0x58000022
11:46:01  2020-07-15 03:49:21 WARN    socketConnectionSendData(): Failed to send data. Socket closed already
11:46:01  2020-07-15 03:49:21 ERROR   iceUtilsSendData(): operation returned status code: 0x58000022
11:46:01  2020-07-15 03:49:21 WARN    turnConnectionSendData(): iceUtilsSendData failed with 0x58000022
11:46:01  2020-07-15 03:49:21 WARN    socketConnectionSendData(): Failed to send data. Socket closed already
11:46:01  2020-07-15 03:49:21 ERROR   iceUtilsSendData(): operation returned status code: 0x58000022
[...]

PR #625, opened 2020-07-17
Fix the interruption of the sending process due to a short network fa…

Nomidia:
…ilure

1. Network is unreachable  occasionally but the situation immediately resumes.
2. Save errno to prevent modification such as after calling strerror.

*Issue #, if available:*
#618 

*Description of changes:*


src/source/Ice/SocketConnection.c
```diff
@@ -329,7 +330,8 @@ STATUS socketSendDataWithRetry(PSocketConnection pSocketConnection, PBYTE buf, U
     while (socketWriteAttempt < MAX_SOCKET_WRITE_RETRY && bytesWritten < bufLen) {
         result = sendto(pSocketConnection->localSocket, buf, bufLen, NO_SIGNAL, destAddr, addrLen);
         if (result < 0) {
-            if (errno == EAGAIN || errno == EWOULDBLOCK) {
+            errorNum = errno;
```
MushMal, line 333: I believe most implementations would store this value in TLS which is pretty fast but I like this change

disa6302:
@Nomidia , 

Changes look good! Clang format check seems to be failing. Can you run `./clang-format.sh -f <filename>` and then run `./check-clang.sh` to confirming the formatting is right?

disa6302:
@Nomidia , 

Something really funky happened when I tried to clang format and push the change to your branch. Can you see if you can reopen this PR?

sharadrj:
Re-opening this pull request.

PR #634, opened 2020-07-20
Ice server and signaling stats usage example

disa6302:
*Issue #, if available:*

*Description of changes:*
 Stats usage example.

### DIFF
--- a/samples/Common.c
+++ b/samples/Common.c
@@ -322,6 +322,7 @@ STATUS initializePeerConnection(PSampleConfiguration pSampleConfiguration, PRtcP
     UINT32 i, j, iceConfigCount, uriCount;
     PIceConfigInfo pIceConfigInfo;
     const UINT32 maxTurnServer = 1;
+    uriCount = 0;
 
     CHK(pSampleConfiguration != NULL && ppRtcPeerConnection != NULL, STATUS_NULL_ARG);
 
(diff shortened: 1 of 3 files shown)


samples/Common.c
```diff
@@ -361,14 +361,31 @@ STATUS initializePeerConnection(PSampleConfiguration pSampleConfiguration, PRtcP
...
 
+// Return ICE server stats for a specific streaming session
+STATUS getIceServerStats(PSampleStreamingSession pSampleStreamingSession)
+{
+    ENTERS();
+    STATUS retStatus = STATUS_SUCCESS;
+    RtcStats rtcmetrics;
+    UINT32 j = 0;
+    rtcmetrics.requestedTypeOfStats = RTC_STATS_TYPE_ICE_SERVER;
+    for (; j < pSampleStreamingSession->pSampleConfiguration->iceUriCount; j++) {
```
lherman-cs, line 380: I might've missed something, but how are the metrics returned? I saw that rtcmetrics is defined locally.

samples/Common.c
```diff
@@ -698,8 +736,11 @@ STATUS freeSampleConfiguration(PSampleConfiguration* ppSampleConfiguration)
     pSampleConfiguration = *ppSampleConfiguration;
 
     CHK(pSampleConfiguration != NULL, retStatus);
-
     for (i = 0; i < pSampleConfiguration->streamingSessionCount; ++i) {
+        // Run export LOG_STATS=TRUE to see logs
```
lherman-cs, line 740: I'm a little bit confused here. Why do we need `LOG_STATS=TRUE`? If users are not interested, can't they just not call `getIceServerStats`?

samples/Common.c
disa6302: The metrics get logged as and when the call is made. The rtcmetrics is populated when call to `rtcPeerConnectionGetMetrics` is made.

samples/Common.c
disa6302: I have added it to keep it consistent with the way I set up in Metrics.c. I use an export flag to control logging of all stats to avoid any unnecessary spew. And yes, if users are not interested they can avoid calling. In this example, I am just logging the stats. But, users can consume the information any way they want. So logging and extracting metrics can be two different things.

samples/Common.c
lherman-cs: right, but rtcmetrics is not being used anywhere. As soon as the function ends, rtcmetrics will be removed from the stack.

samples/Common.c
lherman-cs: I see what you mean. That makes sense.  Since it's a part of `rtcPeerConnectionGetMetrics`, maybe move the comment to be above `rtcPeerConnectionGetMetrics` call?

samples/Common.c
disa6302: Yes. In this case, we are not using it anywhere. It could be used right after the call. 
Now that I think about it, I probably can log in the application itself and not depend on internal logging. Let me do that.

lherman-cs: I like this approach better! Less reliant on environment variables, makes it much more explicit to the user.

samples/Common.c
```diff
@@ -361,14 +362,37 @@ STATUS initializePeerConnection(PSampleConfiguration pSampleConfiguration, PRtcP
             }
         }
     }
-
+    pSampleConfiguration->iceUriCount = uriCount + 1;
```
MushMal, line 365: Not sure why we have this - uriCount is an invariant. Is the intent to change this later?

samples/Common.c
```diff
@@ -698,8 +745,10 @@ STATUS freeSampleConfiguration(PSampleConfiguration* ppSampleConfiguration)
     pSampleConfiguration = *ppSampleConfiguration;
 
     CHK(pSampleConfiguration != NULL, retStatus);
-
     for (i = 0; i < pSampleConfiguration->streamingSessionCount; ++i) {
+        if ((retStatus = getIceServerStats(pSampleConfiguration->sampleStreamingSessionList[i])) != STATUS_SUCCESS) {
```
MushMal, line 749: Use STATUS_FAILED(retStatus = blah())

samples/Common.c
disa6302: I am using pSampleConfiguration->iceUriCount while pulling stats for all the servers. That is the reason I am storing uriCount in pSampleConfiguration.

MushMal: I guess I am not understanding the reason of setting uriCount to 0 then assigning it to pSampleConfiguration->iceUriCount + 1; Perhaps, we will be mutating this value later? If not, why not simply assign 1?

samples/Common.c
disa6302: Done

samples/Common.c
disa6302: > Ah! Ok. I assigned it to 0 to be consistent with the rest of the logic in the function. Because we do a uriCount + 1 within the for loops. If I set it to 1 outside, the value inside the for loops would be incorrect. And we do mutate it inside the loops

PR #670, opened 2020-07-27
Release notes

disa6302:
*Issue #, if available:*

*Description of changes:*

Release 1.3.0 This PR will not be merged. It is a placeholder for review

### DIFF
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -1,0 +1,35 @@
+Release 1.2.0 of WebRTC C SDK
+
+Release tagged at: <Commit:  <to-be-filled>
+
+Whats’s new:
+- Support on Windows
+- Fixed broken data channel message receive
+- Added capability to handle more transceivers 
+- Allowing ICE username/password unto 256 bytes
+- Added PRtcDataChannel to data channel callback arguments
+- Added new data type to allow certain nullable types specified in RTC spec
+- Provision to support trickle ICE based on Remote description
+- File logging in samples
+
+Improvements:
+- Unit tested data channel send message after DTLS handshake
+- Fix leak when ice restart fails
+
+PIC Changes: (Commit:  <to-be-filled>)
+- Build system enhancements
+- File logger functionality moved to PIC from Producer C
+- Fixed re-alloc bug
+
+Producer C Changes: (Commit:  <to-be-filled>)
+- Build system enhancement - navigation away from use of submodules
+- Using only JSMN header
+- Added file logging in samples
+- New status codes for file credential provider
+
+Known Issues:
+* Incorrect transceiver direction breaks SDP exchange in Firefox and iOS
+* Partial reliability does not work as expected. Need to add library level tests to validate
+* Reduce memory allocations in sending audio/video packets
+
+Documentation: Information about the WebRTC public APIs can be found here: https://awslabs.github.io/amazon-kinesis-video-streams-webrtc-sdk-c/
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -1,35 +1,28 @@
-Release 1.2.0 of WebRTC C SDK
+Release 1.3.0 of WebRTC C SDK
 
 Release tagged at: <Commit:  <to-be-filled>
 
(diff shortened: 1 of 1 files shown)


RELEASE.md
```diff
@@ -0,0 +1,29 @@
...
+
+Release tagged at:
<Commit:  <to-be-filled>
+
+Whats’s new:
+- ICE server stats added
+- RTP outbound and inbound stats added (open source contribution)
+- Signaling stats available
+- Clang check added to Travis
+- MBedTLS support added. Choice between OpenSSL and mbedTLS available
```
lherman-cs, line 10: I think we should postpone this to the next release since I still need to verify some key metrics.

RELEASE.md
```diff
@@ -0,0 +1,29 @@
...
+
+Improvements:
+- ICE candidate priority calculation bug fixed improving the candidate nomination process
+- Improved SDP conformance with RFC to work with CRLF and LF
+- Fixed interruption of sending process due to momentary network unreachable errors (open source contribution)
```
sharadrj, line 15: We should add a line about fix for high CPU usage issues

RELEASE.md
```diff
@@ -0,0 +1,29 @@
...
+Producer C Changes: (Commit: e7d4868)
+- Build system improvements
+- README improvements on SDK versioning, development and release
+- Added non-BSD strnstr implementation
+- Fixed static libcurl build failure
+
+Known Issues:
+* Incorrect transceiver direction breaks SDP exchange in Firefox and iOS
+* Reduce memory allocations in sending audio/video packets
+* Get information about the selected candidates such as IP address, server source, type of candidate chosen and protocol used
```
sharadrj, line 26: Not sure what this line means. Are you saying we currently don't have a to surface these information. If so, please rephrase.

RELEASE.md
```diff
@@ -0,0 +1,29 @@
...
+
+Producer C Changes: (Commit: e7d4868)
+- Build system improvements
+- README improvements on SDK versioning, development and release
+- Added non-BSD strnstr implementation
+- Fixed static libcurl build failure
+
+Known Issues:
+* Incorrect transceiver direction breaks SDP exchange in Firefox and iOS
+* Reduce memory allocations in sending audio/video packets
```
sharadrj, line 25: Lets re-phrase it from customer POV.

RELEASE.md
```diff
@@ -0,0 +1,29 @@
+Release 1.2.1 of WebRTC C SDK
+
+Release tagged at: <Commit:  <to-be-filled>
+
+Whats’s new:
+- ICE server stats added
+- RTP outbound and inbound stats added (open source contribution)
+- Signaling stats available
+- Clang check added to Travis
```
lherman-cs, line 9: Since this doesn't benefit users, I think we can omit it.

RELEASE.md
lherman-cs: Probably reword this to `High memory allocations in sending audio/video packets`

RELEASE.md
lherman-cs: Probably reword this to `Missing logging information about the selected candidates, e.g. IP address, server source, type of candidate chosen and protocol used`.

RELEASE.md
sharadrj: Agreed.

RELEASE.md
MushMal: I would phrase as "optimized memory allocations during RTP transmission"

RELEASE.md
disa6302: Ok. But, the changes would still be part of the release. Do we want to add a note, maybe something like MBedTLS (beta version). ?

RELEASE.md
disa6302: Will add once PR #671 is in.

RELEASE.md
lherman-cs: @disa6302 That's a good idea. I like it.

PR #678, opened 2020-07-30
Add an optional flag to not include samples

lherman-cs:
When users want to use our SDK as a library, it's not necessary to also compile samples. So, this PR will give them an ability to not compile the samples.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -12,6 +12,7 @@ option(BUILD_STATIC_LIBS "Build all libraries statically. (This includes third-p
 option(BUILD_OPENSSL_PLATFORM "If buildng OpenSSL what is the target platform" OFF)
 option(BUILD_LIBSRTP_HOST_PLATFORM "If buildng LibSRTP what is the current platform" OFF)
 option(BUILD_LIBSRTP_DESTINATION_PLATFORM "If buildng LibSRTP what is the destination platform" OFF)
+option(BUILD_SAMPLE "Build available samples" ON)
 
 # Developer Flags
 option(BUILD_TEST "Build the testing tree." OFF)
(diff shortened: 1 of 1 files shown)


lherman-cs:
* There's no particular ask, but I found this to be the common pattern for having this optional. And, I think it's a good practice to separate them, samples might have dependencies that the libraries don't depend.
* I'm making BUILD_SAMPLE to be on by default because customers might've been used to this behavior.

MushMal:
> * There's no particular ask, but I found this to be the common pattern for having this optional. And, I think it's a good practice to separate them, samples might have dependencies that the libraries don't depend.
> * I'm making BUILD_SAMPLE to be on by default because customers might've been used to this behavior.

Sounds great then

ycyang1229:
> When users want to use our SDK as a library, it's not necessary to also compile samples. So, this PR will give them an ability to not compile the samples.
> 
> By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.

I like this idea, but do you mind adding three more options for kvsWebrtcClientMaster, kvsWebrtcClientViewer, and kvsWebrtcClientMasterGstSample separately below this option BUILD_SAMPLE? Sometimes, customers only need one part of kvs samples.

MushMal:
> > When users want to use our SDK as a library, it's not necessary to also compile samples. So, this PR will give them an ability to not compile the samples.
> > By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.
> 
> I like this idea, but do you mind adding three more options for kvsWebrtcClientMaster, kvsWebrtcClientViewer, and kvsWebrtcClientMasterGstSample separately below this option BUILD_SAMPLE? Sometimes, customers only need one part of kvs samples.

I would recommend avoiding the complexity in the build scripts as those are very hard to debug and get right for various platforms. Could you perhaps give some examples when compiling the samples is difficult?

lherman-cs:
> > When users want to use our SDK as a library, it's not necessary to also compile samples. So, this PR will give them an ability to not compile the samples.
> > By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.
> 
> I like this idea, but do you mind adding three more options for kvsWebrtcClientMaster, kvsWebrtcClientViewer, and kvsWebrtcClientMasterGstSample separately below this option BUILD_SAMPLE? Sometimes, customers only need one part of kvs samples.

@ycyang1229 I think for this use case, the customer should not compile with just `make` but instead,  `make <specific example>`.

PR #691, opened 2020-08-07
Measure start up latency from offer to first frame

disa6302:
*Issue #, if available:*

*Description of changes:*

- Measures start up latency from offer to first frame for every streaming session
- Changed status code from `STATUS_SUCCESS` to `STATUS_SRTP_NOT_READY_YET` to clearly indicate packets are discarded because SRTP is not set up.

### DIFF
--- a/samples/kvsWebRTCClientMaster.c
+++ b/samples/kvsWebRTCClientMaster.c
@@ -109,7 +109,6 @@ CleanUp:
     }
 
     printf("[KVS Master] Cleaning up....\n");
-
     if (pSampleConfiguration != NULL) {
         // Kick of the termination sequence
         ATOMIC_STORE_BOOL(&pSampleConfiguration->appTerminateFlag, TRUE);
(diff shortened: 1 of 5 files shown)


src/include/com/amazonaws/kinesis/video/webrtcclient/Include.h
```diff
@@ -904,6 +904,8 @@ typedef struct {
 typedef struct {
     RTC_RTP_TRANSCEIVER_DIRECTION direction; //!< Transceiver direction
     RtcRtpReceiver receiver;                 //!< RtcRtpReceiver that has track specific information
+    BOOL firstFrame;
```
lherman-cs, line 907: Do we really need these to be a part of our SDK? Why can't the users just store these values themselves?

disa6302: Because I am calculating it within writeFrame() call. And the packetization happens inside writeFrame. Though, currently, it is one packet per frame usually

lherman-cs: I would calculate this outside of the SDK, I would add the same logic that you have here but after calling `writeFrame` in the video/audio threads.

disa6302: But what about the case when we have multiple RtpPackets per frame?

lherman-cs: I think that's okay. Since we're measuring "first frame" delay, fully processing all of the RtpPackets in a frame should be the one that's measured.

disa6302: Hmm. I had measured first RTP packet based on a request. However, since one frame corresponds to one packet mostly, I am measuring first frame delay in application. Since writeFrame could get called before SRTP is set up, I added a new status code to indicate that instead of just returning STATUS_SUCCESS so that the latency is measured appropriately.

samples/kvsWebRTCClientMaster.c
```diff
@@ -240,10 +239,16 @@ PVOID sendVideoPackets(PVOID args)
             status = writeFrame(pSampleConfiguration->sampleStreamingSessionList[i]->pVideoRtcRtpTransceiver, &frame);
             encoderStats.encodeTimeMsec = 4; // update encode time to an arbitrary number to demonstrate stats update
             updateEncoderStats(pSampleConfiguration->sampleStreamingSessionList[i]->pVideoRtcRtpTransceiver, &encoderStats);
-            if (status != STATUS_SUCCESS) {
+            if (status != STATUS_SUCCESS && status != STATUS_SRTP_NOT_READY_YET) {
```
MushMal, line 242: This looks wrong. How about

if (status != STATUS_SRTP_NOT_READY_YET) {
  if (STATUS_FAILED(status)) { blah} else if (firstFrame) {print}
}

samples/kvsWebRTCClientMaster.c
```diff
@@ -313,8 +318,14 @@ PVOID sendAudioPackets(PVOID args)
         MUTEX_LOCK(pSampleConfiguration->sampleConfigurationObjLock);
         for (i = 0; i < pSampleConfiguration->streamingSessionCount; ++i) {
             status = writeFrame(pSampleConfiguration->sampleStreamingSessionList[i]->pAudioRtcRtpTransceiver, &frame);
-            if (status != STATUS_SUCCESS) {
+            if (status != STATUS_SUCCESS && status != STATUS_SRTP_NOT_READY_YET) {
```
MushMal, line 321: same question here

samples/kvsWebRTCClientMaster.c
```diff
@@ -313,8 +318,14 @@ PVOID sendAudioPackets(PVOID args)
...
                 printf("writeFrame failed with 0x%08x", status);
```
MushMal, line 322: BTW, we might also use the iffef whether we print or not

samples/kvsWebRTCClientMaster.c
disa6302: Reason for not doing that is to avoid multiple if levels unnecessarily. That is all. But, I have made the change

samples/kvsWebRTCClientMaster.c
disa6302: Done

PR #701, opened 2020-08-12
Change log level to DLOGS for rtcp receiver reports

### LINKED ISSUE #699: [QUESTION] Too many onRtcpReceiverReport messages
I'm receiving a lot of these messages, when running Kinesis on my hardware:
onRtcpReceiverReport(): RTCP_PACKET_TYPE_RECEIVER_REPORT rttPropDelay

What does these messages mean? How can I fix this?

disa6302:
*Issue #, if available:*
#699 
*Description of changes:*

Resolved #699 


disa6302:
> Looks good. At some stage we should think whether this really needs to be DLOGS - I don't think this will be important even for DLOGV if it's too verbose even for that

Actually, that makes sense. It does get generated very frequently.

PR #712, opened 2020-08-18
Add null termination to string generated by generateJSONSafeString

disa6302:
*Issue #, if available:*

*Description of changes:*
`generateJSONSafeString` generates string without NULL termination which causes an abort trap when being copied. This PR adds the NULL termination and also updates the length of the different strings that are generated using this API to give wiggle room for the NULL character

### DIFF
--- a/src/source/PeerConnection/PeerConnection.c
+++ b/src/source/PeerConnection/PeerConnection.c
@@ -530,10 +530,10 @@ STATUS generateJSONSafeString(PCHAR pDst, UINT32 len)
 
     CHK(pDst != NULL, STATUS_NULL_ARG);
 
-    for (i = 0; i < len; i++) {
+    for (i = 0; i < len - 1; i++) {
         pDst[i] = VALID_CHAR_SET_FOR_JSON[RAND() % (ARRAY_SIZE(VALID_CHAR_SET_FOR_JSON) - 1)];
     }
-
+    pDst[i] = '\0'; // String must be NULL terminated
 CleanUp:
 
     LEAVES();
--- a/src/source/Ice/IceAgent.h
+++ b/src/source/Ice/IceAgent.h
@@ -52,7 +52,7 @@ extern "C" {
 #define ICE_HASH_TABLE_BUCKET_COUNT  50
 #define ICE_HASH_TABLE_BUCKET_LENGTH 2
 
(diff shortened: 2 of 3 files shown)


src/source/PeerConnection/PeerConnection.c
```diff
@@ -530,10 +530,10 @@ STATUS generateJSONSafeString(PCHAR pDst, UINT32 len)
 
     CHK(pDst != NULL, STATUS_NULL_ARG);
 
-    for (i = 0; i < len; i++) {
+    for (i = 0; i < len - 1; i++) {
```
MushMal, line 533: While this is an internal API, still I would like to standardize on the naming and the buffer ownership.

Normally, a string length indicates the length of the string without the null terminator. If we keep the naming - aka - the "len" then the caller should assume that there will be an extra space for the null terminator. Alternatively, if you want to use this way, perhaps you could rename the len to something like "size"?

Most importantly, the usage sites for the API do assume the extra space for the NULL terminator that gets zero-ed initially. For example, https://github.com/awslabs/amazon-kinesis-video-streams-webrtc-sdk-c/blob/148e0db354f43e2049531293f72c525a4e5e790e/src/source/Ice/IceAgent.h#L131

Could you perhaps add some info on what's actually broken right now?

disa6302: IMO, the application should not have to worry about NULL terminating the string. I cannot think of an application that would not want a string to be null terminated. Nothing is broken in the current stack, but, when the string with same length is copied into another array (using STRNCPY), it abort traps. Yes, adding an extra byte to accommodate a NULL character does fix the issue, but, I did not see a problem in NULL terminating the string at the source, hence the PR.

MushMal: The main usage case for the non-null termination for a low-level routine is to avoid alloc/copy when mangling the names. Say, I want to generate a string in a form of "Blah_XXXXX_blah" where the 5 X-es are the placeholder for a generated name.

I am OK either way but perhaps others will find a usage case that I just described.

disa6302: That makes sense! My only thought was SNPRINTF copies LENGTH - 1 anyways, so the NULL character would not make a difference. But, I see your point. Let me close this PR and we can revisit in the future if needed.

Issue #578, opened 2020-06-30
[FEATURE] WebRTC Metrics


disa6302:
**Is your feature request related to a problem? Please describe.**
An application using the WebRTC SDK expects to monitor the performance of the underlying SDK performance and to provide the flexibility for the applications to take decisions based on a metric

**Describe the solution you'd like**
The application should be able to pull the metrics out periodically based on a type. The following is a broad list of metrics that would be tracked:

- [x] ICE Candidate - #638 (@disa6302 )
- [x] ICE Candidate Pair (@disa6302 )
- [x] ICE Server - #598 (@disa6302 )
- [x] Remote Inbound RTP Stats - #615 (@zhuker )
- [x] Outbound RTP Stats - #615 (@zhuker )
- [x] Inbound stats - #673  (@zhuker )
- [x] Signaling - #577 (@MushMal )
- [x]  Data channel - #704 (@disa6302 )

Related to #576

MushMal:
https://github.com/awslabs/amazon-kinesis-video-streams-webrtc-sdk-c/pull/577 for Signaling metrics

zhuker:
@disa6302
>    I will merge this into awslabs:set-up-ice-metrics, but, I am splitting the different ICE metrics work into different branches to avoid a huge PR with lot of changes. Just FYI.

do you mind sharing them as separate PRs?

 how do we avoid duplicate work?

zhuker:
working on
- Inbound RTP Stats
- Outbound RTP Stats

specifically focused on round trip times

disa6302:
@zhuker ,

Thank you for your contribution once again. I am so thrilled to see your active participation! I should have added more details around the work distribution. RTP stats sounds good. I will add more details to the ticket.

disa6302:
@zhuker , 

Outbound stats are merged :)
Great work!

zhuker:
Thanks

zhuker:
@disa6302 inbound stats https://github.com/awslabs/amazon-kinesis-video-streams-webrtc-sdk-c/pull/673

zhuker:
@disa6302 i will need data channel stats as well
https://github.com/awslabs/amazon-kinesis-video-streams-webrtc-sdk-c/pull/697

disa6302:
@zhuker , 

Sure! I can work on it. Should not take long to implement.

zhuker:
> [ ] Data channel (@disa6302 )

[...]

PR #715, opened 2020-08-18
[Issue-578] Ice candidate pair stats

disa6302:
*Issue #, if available: #578 *

*Description of changes:*
- Ice candidate pair stats (https://www.w3.org/TR/webrtc-stats/#candidatepair-dict*)
- Added usage pattern and average computation based on stats in samples.

Todo:
- Unit test to simulate values such as bytes sent, packets received, etc and validate the values returned by the public API call. This would more likely be an evolved version of a unit test. Currently, the only test for the ice candidate pair stats is Api Test.

### DIFF
--- a/samples/Common.c
+++ b/samples/Common.c
@@ -279,6 +279,14 @@ STATUS handleOffer(PSampleConfiguration pSampleConfiguration, PSampleStreamingSe
         if (pSampleConfiguration->audioSource != NULL) {
             THREAD_CREATE(&pSampleConfiguration->audioSenderTid, pSampleConfiguration->audioSource, (PVOID) pSampleConfiguration);
         }
+
+        if ((retStatus = timerQueueAddTimer(pSampleConfiguration->timerQueueHandle, SAMPLE_STATS_DURATION, SAMPLE_STATS_DURATION,
+                                            getIceCandidatePairStatsCallback, (UINT64) pSampleConfiguration,
+                                            &pSampleConfiguration->iceCandidatePairStatsTimerId)) != STATUS_SUCCESS) {
+            DLOGW("Failed to add getIceCandidatePairStatsCallback to add to timer queue (code 0x%08x). Cannot pull ice candidate pair metrics "
+                  "periodically",
+                  retStatus);
+        }
     }
 
     // The audio video receive routine should be per streaming session
@@ -476,6 +484,7 @@ STATUS createSampleStreamingSession(PSampleConfiguration pSampleConfiguration, P
     ATOMIC_STORE_BOOL(&pSampleStreamingSession->peerIdReceived, TRUE);
 
     pSampleStreamingSession->pSampleConfiguration = pSampleConfiguration;
+    pSampleStreamingSession->rtcMetricsHistory.prevTs = GETTIME();
 
     ATOMIC_STORE_BOOL(&pSampleStreamingSession->terminateFlag, FALSE);
     ATOMIC_STORE_BOOL(&pSampleStreamingSession->candidateGatheringDone, FALSE);
@@ -725,11 +734,14 @@ STATUS createSampleConfiguration(PCHAR channelName, SIGNALING_CHANNEL_ROLE_TYPE 
 
     pSampleConfiguration->clientInfo.version = SIGNALING_CLIENT_INFO_CURRENT_VERSION;
     pSampleConfiguration->clientInfo.loggingLevel = logLevel;
+    pSampleConfiguration->iceCandidatePairStatsTimerId = MAX_UINT32;
 
     ATOMIC_STORE_BOOL(&pSampleConfiguration->interrupted, FALSE);
     ATOMIC_STORE_BOOL(&pSampleConfiguration->mediaThreadStarted, FALSE);
     ATOMIC_STORE_BOOL(&pSampleConfiguration->appTerminateFlag, FALSE);
     ATOMIC_STORE_BOOL(&pSampleConfiguration->recreateSignalingClient, FALSE);
+
+    CHK_STATUS(timerQueueCreate(&pSampleConfiguration->timerQueueHandle));
 
     pSampleConfiguration->iceUriCount = 0;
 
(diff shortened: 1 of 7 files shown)


samples/kvsWebRTCClientMaster.c
```diff
@@ -197,6 +197,7 @@ PVOID sendVideoPackets(PVOID args)
     frame.presentationTs = 0;
     startTime = GETTIME();
     lastFrameTime = startTime;
+    periodicIceTime = GETTIME();
```
chehefen, line 200: periodicIceTime doesnt seem to be used anywhere.

disa6302: Yeah. I am modifying the samples to poll the stats periodically. So the current changes in samples will go.

src/include/com/amazonaws/kinesis/video/webrtcclient/Stats.h
```diff
@@ -159,23 +158,19 @@ typedef struct {
...
+    UINT64 totalRoundTripTime;       //!< The sum of all round trip time (seconds) since the beginning of the session, based
                                      //!< on STUN connectivity check responses (responsesReceived), including those that reply to requests
                                      //!< that are sent in order to verify consent. The average round trip time can be computed from
                                      //!< totalRoundTripTime by dividing it by responsesReceived.
-    DOUBLE currentRoundTripTime;     //!< Latest round trip time (seconds)
-    DOUBLE availableOutgoingBitrate; //!< Total available bit rate for all the outgoing RTP streams on this candidate pair. Calculated by underlying
-                                     //!< congestion control
-    DOUBLE availableIncomingBitrate; //!< Total available bit rate for all the outgoing RTP streams on this candidate pair. Calculated by underlying
-                                     //!< congestion control
+    UINT64 currentRoundTripTime;     //!< Latest round trip time (seconds)
```
chehefen, line 165: round trip time is in seconds and it's an uint64. wouldnt it be 0 most of the time?

src/source/Ice/IceAgent.c
```diff
@@ -982,7 +998,19 @@ STATUS createIceCandidatePairs(PIceAgent pIceAgent, PIceCandidate pIceCandidate,
...
+            STRNCPY(pIceCandidatePair->rtcIceCandidatePairDiagnostics.localCandidateId, pIceCandidatePair->local->id,
+                    ARRAY_SIZE(pIceCandidatePair->rtcIceCandidatePairDiagnostics.localCandidateId));
+            STRNCPY(pIceCandidatePair->rtcIceCandidatePairDiagnostics.remoteCandidateId, pIceCandidatePair->remote->id,
+                    ARRAY_SIZE(pIceCandidatePair->rtcIceCandidatePairDiagnostics.remoteCandidateId));
+            pIceCandidatePair->rtcIceCandidatePairDiagnostics.state = pIceCandidatePair->state;
+            pIceCandidatePair->rtcIceCandidatePairDiagnostics.nominated = pIceCandidatePair->nominated;
+            pIceCandidatePair->rtcIceCandidatePairDiagnostics.lastPacketSentTimestamp = pIceCandidatePair->lastDataSentTime;
+            pIceCandidatePair->firstStunRequest = TRUE;
             pIceCandidatePair->priority = computeCandidatePairPriority(pIceCandidatePair, pIceAgent->isControlling);
+            pIceCandidatePair->rtcIceCandidatePairDiagnostics.totalRoundTripTime = 0.0;
```
chehefen, line 1010: in this uint or float?

src/source/Ice/IceAgent.c
```diff
@@ -2151,9 +2190,18 @@ STATUS incomingDataHandler(UINT64 customData, PSocketConnection pSocketConnectio
     // for stun packets, first 8 bytes are 4 byte type and length, then 4 byte magic byte
     if ((bufferLen < 8 || !IS_STUN_PACKET(pBuffer)) && pIceAgent->iceAgentCallbacks.inboundPacketFn != NULL) {
         // release lock early
+        retStatus = findIceCandidatePairWithLocalSocketConnectionAndRemoteAddr(pIceAgent, pSocketConnection, pSrc, TRUE, &pIceCandidatePair);
         MUTEX_UNLOCK(pIceAgent->lock);
         locked = FALSE;
         pIceAgent->iceAgentCallbacks.inboundPacketFn(pIceAgent->iceAgentCallbacks.customData, pBuffer, bufferLen);
+        if (pIceCandidatePair != NULL) {
+            if (pIceCandidatePair == pIceAgent->pDataSendingIceCandidatePair) {
```
chehefen, line 2198: why not just compare pDataSendingIceCandidatePair with pSocketConnection and remote address?

src/source/Ice/IceAgent.c
```diff
@@ -2348,9 +2409,16 @@ STATUS handleStunPacket(PIceAgent pIceAgent, PBYTE pBuffer, UINT32 bufferLen, PS
...
+                pIceCandidatePair->rtcIceCandidatePairDiagnostics.currentRoundTripTime = pIceCandidatePair->roundTripTime / HUNDREDS_OF_NANOS_IN_A_MILLISECOND;
+                pIceCandidatePair->rtcIceCandidatePairDiagnostics.totalRoundTripTime += pIceCandidatePair->roundTripTime / HUNDREDS_OF_NANOS_IN_A_MILLISECOND;
                 pIceCandidatePair->state = ICE_CANDIDATE_PAIR_STATE_SUCCEEDED;
             }
-
+            if (pIceCandidatePair == pIceAgent->pDataSendingIceCandidatePair && pIceCandidatePair != NULL) {
+                pIceAgent->pDataSendingIceCandidatePair->rtcIceCandidatePairDiagnostics.responsesReceived += connectivityCheckResponsesReceived;
+                pIceAgent->pDataSendingIceCandidatePair->rtcIceCandidatePairDiagnostics.lastResponseTimestamp = GETTIME();
+            }
+            pIceCandidatePair->rtcIceCandidatePairDiagnostics.responsesReceived += connectivityCheckResponsesReceived;
```
chehefen, line 2420: are you double counting here when pIceCandidatePair is pDataSendingIceCandidatePair?

disa6302: My bad. It is redundant. I removed the updating of DataSendingCandidatePair because it would get updated in `iceAgentConnectedStateSetup` once state changes to `ICE_CANDIDATE_PAIR_STATE_SUCCEEDED`.

src/source/Ice/IceAgent.c
disa6302: Agreed. Fixed.

src/include/com/amazonaws/kinesis/video/webrtcclient/Stats.h
disa6302: Earlier, I had set it to DOUBLE to align with spec. Hence the confusion. I have changed it back to DOUBLE and value computation to seconds.

samples/kvsWebRTCClientMaster.c
disa6302: Samples modified. Fixed.

src/source/Ice/IceAgent.c
disa6302: DOUBLE. Fixed.

samples/Common.c
```diff
@@ -766,6 +771,89 @@ STATUS logSignalingClientStats(PSignalingClientMetrics pSignalingClientMetrics)
...
+                    pSampleConfiguration->sampleStreamingSessionList[i]->rtcMetricsHistory.prevNumberOfBytesSent =
+                        rtcMetrics.rtcStatsObject.iceCandidatePairStats.bytesSent;
+                    pSampleConfiguration->sampleStreamingSessionList[i]->rtcMetricsHistory.prevNumberOfBytesReceived =
+                        rtcMetrics.rtcStatsObject.iceCandidatePairStats.bytesReceived;
+                    pSampleConfiguration->sampleStreamingSessionList[i]->rtcMetricsHistory.prevPacketsDiscardedOnSend =
+                        rtcMetrics.rtcStatsObject.iceCandidatePairStats.packetsDiscardedOnSend;
+                }
+            }
+        }
+        THREAD_SLEEP(SAMPLE_STATS_DURATION);
```
MushMal, line 851: This will mean that on termination we have to block for up-to 60 seconds in order to clean it properly.

Instead, I recommend using TimerQueue construct and schedule periodic tasks. Once we have the timer queue, we can use the timer queue to schedule the audio and video and avoid having separate threads.

src/source/Ice/IceAgent.c
```diff
@@ -586,17 +590,30 @@ STATUS iceAgentSendPacket(PIceAgent pIceAgent, PBYTE pBuffer, UINT32 bufferLen)
...
         }
         retStatus = STATUS_SUCCESS;
+    } else {
+        bytesSent = bufferLen;
+        packetsSent++;
     }
 
 CleanUp:
 
+    if (pIceAgent->pDataSendingIceCandidatePair != NULL) {
```
MushMal, line 608: You can jump here if pIceAgent is NULL. Recommend having a higher-level check for retStatus being success before proceeding

src/source/Ice/IceAgent.c
```diff
@@ -1193,6 +1224,15 @@ STATUS iceAgentSendStunPacket(PStunPacket pStunPacket, PBYTE password, UINT32 pa
             DLOGD("mark candidate pair %s_%s as failed", pIceCandidatePair->local->id, pIceCandidatePair->remote->id);
             pIceCandidatePair->state = ICE_CANDIDATE_PAIR_STATE_FAILED;
         }
+    } else {
+        CHK_STATUS(findIceCandidatePairWithLocalSocketConnectionAndRemoteAddr(pIceAgent, pLocalCandidate->pSocketConnection, pDestAddr, TRUE,
+                                                                              &pIceCandidatePair));
+        if (pIceCandidatePair == pIceAgent->pDataSendingIceCandidatePair && pIceCandidatePair != NULL) {
+            if (pIceAgent->pDataSendingIceCandidatePair->firstStunRequest) {
```
MushMal, line 1231: Combine the two ifs together In this case the equivalent boolean would be

if (pIceCandidatePair == pIceAgent->pDataSendingIceCandidatePair && pIceCandidatePair != NULL && pIceAgent->pDataSendingIceCandidatePair->firstStunRequest) {

src/source/Ice/IceAgent.c
```diff
@@ -2151,9 +2193,21 @@ STATUS incomingDataHandler(UINT64 customData, PSocketConnection pSocketConnectio
...
         MUTEX_UNLOCK(pIceAgent->lock);
         locked = FALSE;
         pIceAgent->iceAgentCallbacks.inboundPacketFn(pIceAgent->iceAgentCallbacks.customData, pBuffer, bufferLen);
+        if (candidatePairFound) {
+            pIceAgent->pDataSendingIceCandidatePair->rtcIceCandidatePairDiagnostics.lastPacketReceivedTimestamp = GETTIME();
+            bytesReceived += bufferLen;
+            packetsReceived++; // Since every byte buffer translates to a single RTP packet
+        }
     } else {
         if (ATOMIC_LOAD_BOOL(&pIceAgent->processStun)) {
```
MushMal, line 2212: Is this  } else if (ATOMIC_LOAD_BOOL(&pIceAgent->processStun) {...

src/source/Metrics/Metrics.c
```diff
@@ -7,11 +7,48 @@
 STATUS getIceCandidatePairStats(PRtcPeerConnection pRtcPeerConnection, PRtcIceCandidatePairStats pRtcIceCandidatePairStats)
 {
     STATUS retStatus = STATUS_SUCCESS;
-    PKvsPeerConnection pKvsPeerConnection = (PKvsPeerConnection) pRtcPeerConnection;
-    UNUSED_PARAM(pKvsPeerConnection);
+    BOOL locked = FALSE;
+    PIceAgent pIceAgent = ((PKvsPeerConnection) pRtcPeerConnection)->pIceAgent;
```
MushMal, line 11: pRtcPeerConnection might be NULL. You can define the pIceAgent = NULL and set it after the checks

src/source/Metrics/Metrics.c
```diff
@@ -7,11 +7,48 @@
...
     CHK((pRtcPeerConnection != NULL || pRtcIceCandidatePairStats != NULL), STATUS_NULL_ARG);
-    CHK(FALSE, STATUS_NOT_IMPLEMENTED);
+    MUTEX_LOCK(pIceAgent->lock);
+    locked = TRUE;
+    if (pIceAgent->pDataSendingIceCandidatePair == NULL) {
+        CHK(FALSE, STATUS_SUCCESS);
+    }
```
MushMal, line 17: instead, use CHK(pIceAgent->pDataSendingIceCandidatePair != NULL, STATUS_SUCCESS) if you want an early success return

samples/Common.c
disa6302: Great suggestion! I have modified the code to use Timer queue.

src/source/Ice/IceAgent.c
disa6302: Agreed. Fixed.

src/source/Ice/IceAgent.c
disa6302: Done

src/source/Ice/IceAgent.c
disa6302: Could be. But, do not want to change anything here because it does not affect any functionality

src/source/Metrics/Metrics.c
disa6302: Done

src/source/Metrics/Metrics.c
disa6302: Done.

src/source/Ice/IceAgent.c
disa6302: Done

src/source/Metrics/Metrics.c
disa6302: Done

src/source/Metrics/Metrics.c
disa6302: Done

src/source/Ice/IceAgent.c
disa6302: Could be. But do not want to change anything here because it does not affect the functionality

src/source/Ice/IceAgent.c
disa6302: Good point. Fixed.

samples/Common.c
disa6302: Great suggestion! I have modified the code to use timer queues. It is useful because I can use it to expand on other metrics as well in the future.

samples/Common.c
```diff
@@ -807,6 +904,11 @@ STATUS freeSampleConfiguration(PSampleConfiguration* ppSampleConfiguration)
 
     freeStaticCredentialProvider(&pSampleConfiguration->pCredentialProvider);
 
+    if (pSampleConfiguration->iceCandidatePairStatsTimerId != UINT32_MAX) {
+        CHK_STATUS(timerQueueCancelTimer(pSampleConfiguration->timerQueueHandle, pSampleConfiguration->iceCandidatePairStatsTimerId,
+                                         (UINT64) pSampleConfiguration));
+        pSampleConfiguration->iceCandidatePairStatsTimerId = UINT32_MAX;
+    }
```
MushMal, line 911: You might also need to free the timer queue itself

PR #719, opened 2020-08-19
Sample refactor

chehefen:
*Issue #, if available:*

*Description of changes:*
- Organize signaling message handling by first putting them into a queue, then use a dedicated thread to process them one by one. 
- use a separate lock for media thread to access sampleStreamingSessionList to avoid media flow blocked by some potentially time consuming operations such as getIceServerConfig.

### DIFF
--- a/samples/Common.c
+++ b/samples/Common.c
@@ -143,6 +143,9 @@ STATUS handleOffer(PSampleConfiguration pSampleConfiguration, PSampleStreamingSe
     CHK_STATUS(createAnswer(pSampleStreamingSession->pPeerConnection, &pSampleStreamingSession->answerSessionDescriptionInit));
     CHK_STATUS(setLocalDescription(pSampleStreamingSession->pPeerConnection, &pSampleStreamingSession->answerSessionDescriptionInit));
 
+    /*
+     * If remote support trickle ice, send answer now. Otherwise answer will be sent once ice candidate gathering is complete.
+     */
     if (pSampleStreamingSession->remoteCanTrickleIce) {
         CHK_STATUS(respondWithAnswer(pSampleStreamingSession));
         DLOGD("time taken to send answer %" PRIu64 " ms",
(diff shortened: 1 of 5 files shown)


chehefen:
> This is an involved change and I am unsure why you want to cache the signaling messages and not process them directly in the caller thread? Messages are delivered in a separate thread.

I've removed the caching part and handles signaling messages directly.

samples/Common.c
```diff
@@ -999,3 +915,132 @@ STATUS sessionCleanupWait(PSampleConfiguration pSampleConfiguration)
...
+
+    switch (pReceivedSignalingMessage->signalingMessage.messageType) {
+        case SIGNALING_MESSAGE_TYPE_OFFER:
+            /*
+             * Create new streaming session for each offer, then insert the client id and streaming session into
+             * pRtcPeerConnectionForRemoteClient for subsequent ice candidate messages. Lastly check if there is
+             * any ice candidate messages queued in pPendingSignalingMessageForRemoteClient. If so then submit
+             * all of them.
+             */
+            CHK_STATUS(createSampleStreamingSession(pSampleConfiguration, pReceivedSignalingMessage->signalingMessage.peerClientId, TRUE,
```
MushMal, line 974: Since we are refactoring the sample, we should think on how to make this operation prompt. The main issue here is that it can block on the awaiting for ice config which then blocks the callback thread that executes this API. However, this is OK as it's a separate thread just for the message delivery and it in fact assumes non-prompt operations could happen so we should be OK.

samples/kvsWebRTCClientMaster.c
```diff
@@ -343,16 +331,16 @@ PVOID sendAudioPackets(PVOID args)
     return (PVOID)(ULONG_PTR) retStatus;
 }
 
-PVOID sampleReceiveAudioFrame(PVOID args)
+PVOID sampleReceiveVideoFrame(PVOID args)
```
MushMal, line 334: This looks to be generic enough for any type. Should be make it agnostic of type?

chehefen: the routine we assign it to is pSampleConfiguration->receiveAudioVideoSource. User can assign any routine to receiveAudioVideoSource to setup which transceivers to receive.

MushMal: OK

Issue #614, opened 2020-07-14
[FEATURE] Travis CI for samples


disa6302:
**Is your feature request related to a problem? Please describe.**
It is important to check for regression in samples.
**Describe the solution you'd like**
Add samples to travis CI

PR #721, opened 2020-08-20
Add a CI job to check samples

lherman-cs:
Add a CI job to check samples.

Resolves https://github.com/awslabs/amazon-kinesis-video-streams-webrtc-sdk-c/issues/614

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -200,6 +200,21 @@ matrix:
         - .github/build_windows.bat
         - cd build/tst && ./webrtc_client_test.exe --gtest_filter="-DataChannelFunctionalityTest.*:IceApiTest.*:IceFunctionalityTest.*:PeerConnectionFunctionalityTest.*:SignalingApiFunctionalityTest.*:TurnConnectionFunctionalityTest.*:RtpFunctionalityTest.marshallUnmarshallH264Data:RtpFunctionalityTest.packingUnpackingVerifySameH264Frame"
             
+    - name: "Sample check"
+      os: linux
+      compiler: gcc
+      before_install:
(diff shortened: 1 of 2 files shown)


.travis.yml
```diff
@@ -200,6 +200,19 @@ matrix:
...
             
+    - name: "Sample check"
+      os: linux
+      compiler: gcc
+      before_install:
+        # TODO: Remove the following line. This is only a workaround for enabling IPv6, https://github.com/travis-ci/travis-ci/issues/8891.
+        - sudo sh -c 'echo 0 > /proc/sys/net/ipv6/conf/all/disable_ipv6'
+        - sudo add-apt-repository -y ppa:ubuntu-toolchain-r/test
+        - sudo apt-get -q update
+        - sudo apt-get -y install gcc-4.4
```
MushMal, line 211: I like that we are running this on older compiler

check-sample.sh
```diff
@@ -0,0 +1,18 @@
...
+pids=""
+cd build
+./kvsWebrtcClientMaster SampleChannel &
+pids+=" $!"
+
+sleep 2
+./kvsWebrtcClientViewer SampleChannel &
+pids+=" $!"
+
+bash -c "sleep 10 && kill -s INT $(jobs -p)" &
```
MushMal, line 12: This is a good start. Let's make sure we add some configuration to the samples to run for a specified amount of time and exit gracefully. We also need to ensure that the samples actually communicate and exchange data. We can enhance that together with the samples refactor work.

check-sample.sh
```diff
@@ -0,0 +1,22 @@
+#!/bin/bash
+
+set -xv
```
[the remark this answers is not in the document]
lherman-cs: After second thought, I found this to be useful anyway. So, I added a comment to describe what it does.

Issue #556, opened 2020-06-18
[BUG] Unable to build libwebsockets


fernandoc1:
I'm trying to compile the code for Amazon Kinesis with gcc version 4.8.3 20131202 (Hisilicon_v300) for an ARM board.

My compilation command is:

`CC=arm-hisiv300-linux-gcc CXX=arm-hisiv300-linux-g++ AR=arm-hisiv300-linux-ar cmake .. -DBUILD_OPENSSL=TRUE -DBUILD_OPENSSL_PLATFORM=linux-generic32 -DBUILD_LIBSRTP_HOST_PLATFORM=x86_64-unknown-linux-gnu -DBUILD_LIBSRTP_DESTINATION_PLATFORM=arm-unknown-linux-uclibcgnueabi`

I'm facing the following error:

```
amazon-kinesis-video-streams-webrtc-sdk-c/open-source/libwebsockets/build/src/project_libwebsockets/lib/roles/pipe/ops-pipe.c:30:2: error: implicit declaration of function ‘eventfd_read’ [-Werror=implicit-function-declaration]
  if (eventfd_read(wsi->desc.sockfd, &value) < 0)

```

fernandoc1:
I fixed my compilation issues with the following changes:
[](https://gist.github.com/fernandoc1/e2b6821bf7431caec8bac7d8dd47921c)

MushMal:
@fernandoc1 this is a good change I believe. Could you please submit a PR? All you need to do is rebase on the latest master branch, git checkout -b <your_branch_name> and git push. This will produce a link for the PR. Please submit it so we can integrate.

fernandoc1:
Done:
[557](https://github.com/awslabs/amazon-kinesis-video-streams-webrtc-sdk-c/pull/557)

ycyang1229:
@fernandoc1 Hey, the platform is interesting. Does linux run on your platform or other non-linux-based RTOS?

fernandoc1:
My platform is Linux based.
I can share my compiler toolchain for kinesis testing it's compatibility.

MushMal:
@fernandoc1 after some thinking it might be possible that the fix you've submitted could break some customers. In order to not regress, the customers who are affected by this we recommend modifying the appropriate lib...Cmake directly in their environment. We will try to think how we can improve this custom flags passing to third party libs better.

Resolving this for now, please don't hesitate to re-open

[...]

PR #723, opened 2020-08-20
Removing the ldexp from the codebase

MushMal:
Issue #556 

No need for ldexp which will also remove the need to link against libm library on some platforms


src/source/Rtcp/RtcpPacket.c
```diff
@@ -122,7 +122,7 @@ STATUS rembValueGet(PBYTE pPayload, UINT32 payloadLen, PDOUBLE pMaximumBitRate,
     mantissa &= RTCP_PACKET_REMB_MANTISSA_BITMASK;
 
     exponent = pPayload[RTCP_PACKET_REMB_IDENTIFIER_OFFSET + SIZEOF(UINT32) + SIZEOF(BYTE)] >> 2;
-    maximumBitRate = ldexp(mantissa, exponent);
+    maximumBitRate = mantissa * (1 << exponent);
```
hassanctech, line 125: Minor comment, this could also be: 
`maximumBitRate = mantissa << exponent`

MushMal: That won’t be equivalent right? Here, we are trying to multiply mantissa with the 2 in the power of the exponent. Doesn’t it?

PR #735, opened 2020-08-24
Fix sample bugs

### LINKED ISSUE #709: [BUG] race condition in freeSampleConfiguration
see https://github.com/awslabs/amazon-kinesis-video-streams-webrtc-sdk-c/blob/master/samples/Common.c#L780
before accessing `sampleStreamingSessionList`, should acquire `sampleConfigurationObjLock`?

### LINKED ISSUE #730: [QUESTION] Should streamingSessionCount be atomic?
Should streamingSessionCount be atomic in the sample apps?

disa6302:
*Issue #, if available:*

- #709 
- #730 

*Description of changes:*

- Fixes potential race condition introduced while accessing streamingSessionList
- streamingSessionCount is made atomic.
- Fixing bug in start up latency calculation to record offer time instead of using `firstSdpMessageReceiveTime`

Resolves #709 
Resolves #730 


samples/Common.c
```diff
@@ -133,6 +134,7 @@ STATUS masterMessageReceived(UINT64 customData, PReceivedSignalingMessage pRecei
     locked = TRUE;
     // ice candidate message and offer message can come at any order. Therefore, if we see a new peerId, then create
     // a new SampleStreamingSession, which in turn creates a new peerConnection
+    ATOMIC_LOAD(&pSampleConfiguration->streamingSessionCount);
```
MushMal, line 137: same

samples/Common.c
```diff
@@ -148,13 +150,15 @@ STATUS masterMessageReceived(UINT64 customData, PReceivedSignalingMessage pRecei
         CHK_STATUS(createSampleStreamingSession(pSampleConfiguration, pReceivedSignalingMessage->signalingMessage.peerClientId, TRUE,
                                                 &pSampleStreamingSession));
         pSampleStreamingSession->firstSdpMsgReceiveTime = GETTIME();
-        pSampleConfiguration->sampleStreamingSessionList[pSampleConfiguration->streamingSessionCount++] = pSampleStreamingSession;
+        pSampleConfiguration->sampleStreamingSessionList[pSampleConfiguration->streamingSessionCount] = pSampleStreamingSession;
+        ATOMIC_INCREMENT(&pSampleConfiguration->streamingSessionCount);
```
MushMal, line 154: This one is not atomic. Could you check with Felix whether this is already protected by the object lock and no need to provide for atomicity.

samples/Common.c
```diff
@@ -798,7 +802,8 @@ STATUS getIceCandidatePairStatsCallback(UINT32 timerId, UINT64 currentTime, UINT
         DLOGW("[KVS Master] getPeriodicStats(): operation returned status code: 0x%08x \n", STATUS_NULL_ARG);
         goto CleanUp;
     }
-    for (i = 0; i < pSampleConfiguration->streamingSessionCount; ++i) {
+    ATOMIC_LOAD(&pSampleConfiguration->streamingSessionCount);
```
MushMal, line 805: same issue

samples/Common.c
```diff
@@ -880,14 +885,16 @@ STATUS freeSampleConfiguration(PSampleConfiguration* ppSampleConfiguration)
     pSampleConfiguration = *ppSampleConfiguration;
 
     CHK(pSampleConfiguration != NULL, retStatus);
-    for (i = 0; i < pSampleConfiguration->streamingSessionCount; ++i) {
+    ATOMIC_LOAD(&pSampleConfiguration->streamingSessionCount);
```
MushMal, line 888: Usage of atomics is incorrect

samples/Common.c
```diff
@@ -880,14 +885,16 @@ STATUS freeSampleConfiguration(PSampleConfiguration* ppSampleConfiguration)
...
+    MUTEX_LOCK(pSampleConfiguration->sampleConfigurationObjLock);
```
MushMal, line 889: I actually think if we got here then the mutex is valid. However, think if you want to move the locks outside of the loop and in that case you will need to check whether the mutex is valid before executing the entire loop

samples/Common.c
```diff
@@ -944,12 +951,13 @@ STATUS sessionCleanupWait(PSampleConfiguration pSampleConfiguration)
 
     while (!ATOMIC_LOAD_BOOL(&pSampleConfiguration->interrupted)) {
         // scan and cleanup terminated streaming session
-        for (i = 0; i < pSampleConfiguration->streamingSessionCount; ++i) {
+        ATOMIC_LOAD(&pSampleConfiguration->streamingSessionCount);
```
MushMal, line 954: same issue with atomics

samples/Samples.h
```diff
@@ -80,7 +80,7 @@ typedef struct {
     BOOL enableFileLogging;
     UINT64 customData;
     PSampleStreamingSession sampleStreamingSessionList[DEFAULT_MAX_CONCURRENT_STREAMING_SESSION];
-    UINT32 streamingSessionCount;
+    volatile SIZE_T streamingSessionCount;
```
MushMal, line 83: See if you really need an atomic here or all of the operations requiring the streamingSessionCount are actually interlocked and you don't need an atomic.
In case you decide that you still need atomic, move the volatile SIZE_T to be the first first member of the stuct declaration to be machine word aligned even if it gets tight packed - we always declare the atomics first in order to assure they are machine word aligned, otherwise most hardware would generate bus faults

Issue #741, opened 2020-08-25
[BUG?] REMB for multiple video tracks handling


zhuker:
according to REMB spec 
> This feedback message is used to notify a sender of multiple media streams over the same RTP session of the *total* estimated available bit rate on the path to the receiving side of this RTP session.

meaning a single bitrate value is sent for multiple tracks, while we are handling it as if bandwidth hint is per track see `Rtcp.c:199`

zhuker:
@Sean-Der what you say?

MushMal:
Perhaps we should add some handling for this case. Most scenarios are video only so it would be OK. In cases where we have audio and video, the data is a little skewed in this case. True, the audio density is multiple times smaller than video but perhas we should handle it. I am not yet sure how to properly handle this and what the net result should be from the applications perspective.

The SDK currently just passes the information along to the customers application but it can't control the actual ABR or frame thinning.

zhuker:
if you get a REMB like `[2mbps, audio_ssrc, video_ssrc]`
both audio and video track will receive 2mbps bandwidth hint separately so user app will decide that both audio AND video can be encoded at 2mbps for a total of 4mbps which is not what REMB intended to say

MushMal:
Indeed, however, in practice I have not seen any ABR that can apply neatly to audio. So, likely, the applications will simply ignore it for audio.

This said, I would like to tease apart what would be needed to handle this from the SDK perspective. Perhaps, the applications should have the understanding that the reported bandwidth is for all the tracks summarily as the SDK can't subdivide the bandwidth proportionally

zhuker:
i also found a bug in remb handling https://github.com/awslabs/amazon-kinesis-video-streams-webrtc-sdk-c/pull/743

MushMal:
NIIIIIICE! Commented on it.

PR #743, opened 2020-08-25
remb mutliple tracks

zhuker:
REMB for multiple tracks
*Description of changes:*

before: if transceiver was not found last known good transceiver was used
after: call `onBandwidthEstimation` for all transceivers specified in REMB packet but ONLY specified transceivers

related to but does NOT fix https://github.com/awslabs/amazon-kinesis-video-streams-webrtc-sdk-c/issues/741

### DIFF
--- a/src/source/PeerConnection/Rtcp.c
+++ b/src/source/PeerConnection/Rtcp.c
@@ -203,30 +203,18 @@ STATUS onRtcpRembPacket(PRtcpPacket pRtcpPacket, PKvsPeerConnection pKvsPeerConn
     DOUBLE maximumBitRate = 0;
     UINT8 ssrcListLen;
     UINT32 i;
-    PDoubleListNode pCurNode = NULL;
     PKvsRtpTransceiver pTransceiver = NULL;
-    UINT64 item;
 
     CHK(pKvsPeerConnection != NULL && pRtcpPacket != NULL, STATUS_NULL_ARG);
 
     CHK_STATUS(rembValueGet(pRtcpPacket->payload, pRtcpPacket->payloadLength, &maximumBitRate, (PUINT32) &ssrcList, &ssrcListLen));
 
     for (i = 0; i < ssrcListLen; i++) {
-        CHK_STATUS(doubleListGetHeadNode(pKvsPeerConnection->pTransceivers, &pCurNode));
-        while (pCurNode != NULL && pTransceiver == NULL) {
-            CHK_STATUS(doubleListGetNodeData(pCurNode, &item));
-            CHK(item != 0, STATUS_INTERNAL_ERROR);
-
-            pTransceiver = (PKvsRtpTransceiver) item;
-            if (pTransceiver->sender.ssrc != ssrcList[i] && pTransceiver->sender.rtxSsrc != ssrcList[i]) {
-                pTransceiver = NULL;
-            }
-
-            pCurNode = pCurNode->pNext;
+        pTransceiver = NULL;
+        if (STATUS_FAILED(findTransceiverBySsrc(pKvsPeerConnection, &pTransceiver, ssrcList[i]))) {
+            DLOGW("Received REMB for non existing ssrcs: ssrc %lu", ssrcList[i]);
         }
-
-        CHK_ERR(pTransceiver != NULL, STATUS_RTCP_INPUT_SSRC_INVALID, "Received REMB for non existing ssrcs: ssrc %lu", ssrcList[i]);
-        if (pTransceiver->onBandwidthEstimation != NULL) {
+        if (pTransceiver != NULL && pTransceiver->onBandwidthEstimation != NULL) {
             pTransceiver->onBandwidthEstimation(pTransceiver->onBandwidthEstimationCustomData, maximumBitRate);
         }
     }
--- a/src/source/PeerConnection/Rtp.c
+++ b/src/source/PeerConnection/Rtp.c
@@ -421,6 +421,7 @@ STATUS findTransceiverBySsrc(PKvsPeerConnection pKvsPeerConnection, PKvsRtpTrans
         if (pTransceiver->sender.ssrc == ssrc || pTransceiver->sender.rtxSsrc == ssrc || pTransceiver->jitterBufferSsrc == ssrc) {
             break;
         }
+        pTransceiver = NULL;
         pCurNode = pCurNode->pNext;
     }
     CHK(pTransceiver != NULL, STATUS_NOT_FOUND);
--- a/tst/RtcpFunctionalityTest.cpp
+++ b/tst/RtcpFunctionalityTest.cpp
@@ -11,13 +11,15 @@ class RtcpFunctionalityTest : public WebRtcClientTestBase {
     PKvsRtpTransceiver pKvsRtpTransceiver = nullptr;
     PKvsPeerConnection pKvsPeerConnection = nullptr;
     PRtcPeerConnection pRtcPeerConnection = nullptr;
+    PRtcRtpTransceiver pRtcRtpTransceiver = nullptr;
 
     STATUS initTransceiver(UINT32 ssrc)
     {
         RtcConfiguration config{};
         EXPECT_EQ(STATUS_SUCCESS, createPeerConnection(&config, &pRtcPeerConnection));
         pKvsPeerConnection = reinterpret_cast<PKvsPeerConnection>(pRtcPeerConnection);
-        pKvsRtpTransceiver = reinterpret_cast<PKvsRtpTransceiver>(addTransceiver(ssrc));
+        pRtcRtpTransceiver = addTransceiver(ssrc);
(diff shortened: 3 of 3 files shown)


src/source/PeerConnection/Rtcp.c
```diff
@@ -199,34 +199,28 @@ STATUS onRtcpPacket(PKvsPeerConnection pKvsPeerConnection, PBYTE pBuff, UINT32 b
...
-            CHK_STATUS(doubleListGetNodeData(pCurNode, &item));
-            CHK(item != 0, STATUS_INTERNAL_ERROR);
-
-            pTransceiver = (PKvsRtpTransceiver) item;
-            if (pTransceiver->sender.ssrc != ssrcList[i] && pTransceiver->sender.rtxSsrc != ssrcList[i]) {
-                pTransceiver = NULL;
+        pTransceiver = NULL;
+        tmpRetStatus = findTransceiverBySsrc(pKvsPeerConnection, &pTransceiver, ssrcList[i]);
+        if (STATUS_FAILED(tmpRetStatus)) {
+            DLOGE("Received REMB for non existing ssrcs: ssrc %lu", ssrcList[i]);
```
MushMal, line 217: Not sure why we are returning an error here? Perhaps, just DLOGW printout is enough and skipping the transceiver? I am not sure if this adds much value to return an error, though I can see it both ways. The core SDK won't process the error and it will be dumped to the ground anyway.

If you decide to go with the DLOGW then you can simply skip over the transceiver and continue.

zhuker: makes sense

zhuker:
@MushMal is there a way to extract core dump from the failing CI build https://travis-ci.com/github/awslabs/amazon-kinesis-video-streams-webrtc-sdk-c/jobs/378017200 
cant reproduce it here

MushMal:
> @MushMal is there a way to extract core dump from the failing CI build https://travis-ci.com/github/awslabs/amazon-kinesis-video-streams-webrtc-sdk-c/jobs/378017200
> cant reproduce it here

This is odd... Just run the test locally without auth info and it passes just fine. We did have some instability in TravisCI before runs related to the open source components such as usrsctp. I have re-started that particular run. Let's see how that works

zhuker:
http://jsteemann.github.io/blog/2014/10/30/getting-core-dumps-of-failed-travisci-builds/

MushMal:
> http://jsteemann.github.io/blog/2014/10/30/getting-core-dumps-of-failed-travisci-builds/

Hmm, the build crashes again. Enabling core dumps is a must in my opinion. Do you mind taking a stab at it perhaps?

Also, have you rebased on the latest? @lherman-cs added a different commit for one of the dependencies. @lherman-cs any ideas?

lherman-cs:
@zhuker, it looks like your PR is behind the master. Could you try to rebase your PR to the current master?

zhuker:
@lherman-cs rebased

MushMal:
> @lherman-cs rebased

@zhuker I didn't see perhaps but have you pushed a new commit on top of the commits @lherman-cs mentioned that could fix the crash? I think the failure is from the previous run right?

lherman-cs:
If it's Linux, we can do `ulimit -c unlimited` to get the stack trace. But this approach won't work for Mac. I'm still looking a way to get the stack trace.

zhuker:
@MushMal @lherman-cs as you already know i am not a git guru at all, i seem to have rebased on top of master and it _looks_ like the branch is no longer behind
but CI still fails :man_shrugging:

zhuker:
when i am running tests on my mac i get this in crash reporter even if all tests pass 
Screen Shot 2020-08-27 at 9 34 01 AM
```
Thread 0 Crashed:: Dispatch queue: com.apple.main-thread
0   libsystem_kernel.dylib        	0x00007fff701ee33a __pthread_kill + 10
1   libsystem_pthread.dylib       	0x00007fff702aae60 pthread_kill + 430
2   libsystem_c.dylib             	0x00007fff70175808 abort + 120
3   libsystem_c.dylib             	0x00007fff70174ac6 __assert_rtn + 314
4   libkvsWebrtcClient.dylib      	0x000000010dcbcff2 transactionIdStoreClear + 82
5   webrtc_client_test            	0x000000010d97b282 com::amazonaws::kinesis::video::webrtcclient::IceApiTest_IceUtilApiTest_Test::TestBody() + 4866
6   webrtc_client_test            	0x000000010dacfc14 void testing::internal::HandleSehExceptionsInMethodIfSupported<testing::Test, void>(testing::Test*, void (testing::Test::*)(), char const*) + 132
7   webrtc_client_test            	0x000000010da9c3eb void testing::internal::HandleExceptionsInMethodIfSupported<testing::Test, void>(testing::Test*, void (testing::Test::*)(), char const*) + 123
8   webrtc_client_test            	0x000000010da9c326 testing::Test::Run() + 182
9   webrtc_client_test            	0x000000010da9d70a testing::TestInfo::Run() + 202
10  webrtc_client_test            	0x000000010da9e674 testing::TestCase::Run() + 228
11  webrtc_client_test            	0x000000010daaaae8 testing::internal::UnitTestImpl::RunAllTests() + 680
12  webrtc_client_test            	0x000000010dad4da4 bool testing::internal::HandleSehExceptionsInMethodIfSupported<testing::internal::UnitTestImpl, bool>(testing::internal::UnitTestImpl*, bool (testing::internal::UnitTestImpl::*)(), char const*) + 132
13  webrtc_client_test            	0x000000010daaa5bb bool testing::internal::HandleExceptionsInMethodIfSupported<testing::internal::UnitTestImpl, bool>(testing::internal::UnitTestImpl*, bool (testing::internal::UnitTestImpl::*)(), char const*) + 123
14  webrtc_client_test            	0x000000010daaa470 testing::UnitTest::Run() + 208
15  webrtc_client_test            	0x000000010da87821 RUN_ALL_TESTS() + 17
16  webrtc_client_test            	0x000000010da87783 main + 179
17  libdyld.dylib                 	0x00007fff700a6cc9 start + 1
```

MushMal:
This is an odd one. Let me whether this is a consequence of failed test or an overzealous assert. @chehefen any ideas?

chehefen:
> This is an odd one. Let me whether this is a consequence of failed test or an overzealous assert. @chehefen any ideas?

it's not an overzealous assert, if assert isn't there it would crash with null pointer instead.

MushMal:
> > This is an odd one. Let me whether this is a consequence of failed test or an overzealous assert. @chehefen any ideas?
> 
> it's not an overzealous assert, if assert isn't there it would crash with null pointer instead.

If this is a legitimate negative case then we need to return an error and not tear down the process. What is the actual case here we are dealing with? I need to look into the code Tik to understand the negative case in this test

lherman-cs:
@zhuker could you try to apply this patch, https://github.com/awslabs/amazon-kinesis-video-streams-webrtc-sdk-c/pull/761? It's a fix for a segmentation fault issue, this might be related to the issue that we're seeing here.

MushMal:
I think the crash is very specific to clang and google test. In this test we are expecting a crash in the test. Something might not be capturing the crash that’s expected

zhuker:
@lherman-cs just applied the patch

lherman-cs:
@zhuker the Travis CI seems to be happy after the patch, except the sample test. But, this OK, since this is a fork build, Travis will not have the AWS credentials that we use for unit tests, so I made this https://github.com/awslabs/amazon-kinesis-video-streams-webrtc-sdk-c/pull/763, to skip the sample check for fork builds.

zhuker:
@lherman-cs do you have the same tsan issue? https://travis-ci.com/github/awslabs/amazon-kinesis-video-streams-webrtc-sdk-c/jobs/379617327

lherman-cs:
@zhuker Nope I don't. But, it seems to be an actual problem. I probably didn't hit the code path. Let me take a look.

MushMal:
Folks I thought I updated this but seems not. The issue reproduces only on clang and it seems that the test expects a crash and it doesn’t. @lherman-cs could you take a a look?

lherman-cs:
@MushMal The patch did solve the segmentation fault issue. This data race issue is another issue, which I have another patch for fixing this, https://github.com/awslabs/amazon-kinesis-video-streams-webrtc-sdk-c/pull/764.

lherman-cs:
@zhuker all the fixes have been merged to master, https://github.com/awslabs/amazon-kinesis-video-streams-webrtc-sdk-c/commit/66ef16fb53a978e42b167d471263ea5807ea05c4, https://github.com/awslabs/amazon-kinesis-video-streams-webrtc-sdk-c/commit/a66f462c40d0b32654df2ec2275b75fa79434f31, and https://github.com/awslabs/amazon-kinesis-video-streams-webrtc-sdk-c/commit/80241f383c7fa1b2d3a303022b3bb844b4e12336.

Your change should pass the tests now.

lherman-cs:
@zhuker Also, would you mind to squash all of your commits into a single commit?

zhuker:
@lherman-cs thanks for your help getting this to work!

Issue #688, opened 2020-08-05
[BUG]Windows: Assertion during runtime


codingspirit:
**Describe the bug**
Run Windows built ClientMaster sample will have assertion like this:
![image](https://user-images.githubusercontent.com/20569533/89390891-0169b000-d73a-11ea-8139-de61c6528789.png)
After that sample will be terminated.

Maybe we should use `closesocket` under Windows instead of `close` in https://github.com/awslabs/amazon-kinesis-video-streams-webrtc-sdk-c/blob/af187a586ccc6f785d856c6fcfe39872dc0fe0eb/src/source/Ice/SocketConnection.c#L77 

I tried repleace it with `closesocket`, no assertion came out but sample still can't work:

```
2020-08-05 08:04:51 ERROR   signalingCacheLoadFromFile(): operation returned status code: 0x0000000a
2020-08-05 08:04:51 WARN    createSignalingSync(): Failed to load signaling cache from file
2020-08-05 08:04:51 ERROR   signalingCacheSaveToFile(): operation returned status code: 0x0000000a
2020-08-05 08:04:51 WARN    getChannelEndpoint(): Failed to save signaling cache to file
[KVS Viewer] Signaling client created successfully
[KVS Viewer] Signaling client connection to socket established
2020-08-05 08:04:56 DEBUG   iceAgentValidateKvsRtcConfig():
        iceLocalCandidateGatheringTimeout: 10000 ms
        iceConnectionCheckTimeout: 10000 ms
        iceCandidateNominationTimeout: 10000 ms
        iceConnectionCheckPollingInterval: 50 ms
[KVS Viewer] Creating streaming session...completed
[KVS Viewer] Offer creation successful
2020-08-05 08:04:56 DEBUG   createSocket(): setsockopt() failed with errno No error
2020-08-05 08:04:56 DEBUG   createSocket(): setsockopt() failed with errno No error
2020-08-05 08:04:56 DEBUG   createSocket(): setsockopt() failed with errno No error
2020-08-05 08:04:56 DEBUG   createSocket(): setsockopt() failed with errno No error
2020-08-05 08:04:56 DEBUG   createSocket(): setsockopt() failed with errno No error
2020-08-05 08:04:56 DEBUG   createSocket(): setsockopt() failed with errno No error
2020-08-05 08:04:56 DEBUG   createSocket(): setsockopt() failed with errno No error
[...]

PR #747, opened 2020-08-26
Windows: fix assertion when close socket

codingspirit:
Windows: fix assertion when close socket

*Issue #, if available:*
#688 
*Description of changes:*
socket() on windows do not use *nix-style file descriptors but returns a
handle to a kernel object instead, it must be closed via closesocket()

### DIFF
--- a/src/source/Ice/SocketConnection.c
+++ b/src/source/Ice/SocketConnection.c
@@ -74,7 +74,7 @@ STATUS freeSocketConnection(PSocketConnection* ppSocketConnection)
         freeTlsSession(&pSocketConnection->pTlsSession);
     }
 
-    close(pSocketConnection->localSocket);
+    CHK_STATUS(closeSocket(pSocketConnection->localSocket));
 
     MEMFREE(pSocketConnection);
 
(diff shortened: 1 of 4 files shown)


src/source/Ice/SocketConnection.c
```diff
@@ -74,7 +74,11 @@ STATUS freeSocketConnection(PSocketConnection* ppSocketConnection)
         freeTlsSession(&pSocketConnection->pTlsSession);
     }
 
+#ifdef _WIN32
```
MushMal, line 77: Can I ask what's asserting here and why? 

If there is a strong reason to platform differentiate then I would recommend outsourcing the functionality to closeSocket() API which should be defined in Network.c/h where we already do conditional compilation based on platform/os

codingspirit: Sorry if commit message is not clear enough.
For *nix system, `socket()` will create socket FD, so we can close it by `close()`.
For Windows, what `socket()` created is a not a FD, we can't use `close()` to a `SOCKET` object.

In [Win32 API Doc](https://docs.microsoft.com/en-us/windows/win32/api/winsock2/nf-winsock2-socket)
> When a session has been completed, a closesocket must be performed.

MushMal: Thanks for clarification. We will take a note to hit this path with the tests. 

The changes themselves are good but I would recommend adding them to Network.c and have a common closeSocket similar to createSocket?

codingspirit: I have updated PR with a common `closeSocket` in Network.c/.h.

lherman-cs:
@codingspirit I'm sorry for the inconvenience. Do you mind to rebase from master? There's a minor fix in the test script for handling fork builds, https://github.com/awslabs/amazon-kinesis-video-streams-webrtc-sdk-c/commit/66ef16fb53a978e42b167d471263ea5807ea05c4.

codingspirit:
@lherman-cs Sure, done

PR #761, opened 2020-08-28
Fix NULL pointer dereference

lherman-cs:
Resolves https://github.com/awslabs/amazon-kinesis-video-streams-webrtc-sdk-c/issues/760

A NULL pointer dereference at `pDataSendingIceCandidatePair` can occur when the ICE agent receives data before it's finished setup. Since ICE gathering starts as soon as the user calls `setLocalDescription` and the `pDataSendingIceCandidatePair` setup is done after calling `setRemoteDescription` within the `iceAgentStartAgent`, it's possible that the ICE agent receives data in between `setLocalDescription` and `setRemoteDescription`, which will lead to a NULL pointer dereference.

### DIFF
--- a/src/source/Ice/IceAgent.c
+++ b/src/source/Ice/IceAgent.c
@@ -2193,7 +2193,8 @@ STATUS incomingDataHandler(UINT64 customData, PSocketConnection pSocketConnectio
     if ((bufferLen < 8 || !IS_STUN_PACKET(pBuffer)) && pIceAgent->iceAgentCallbacks.inboundPacketFn != NULL) {
         // release lock early
         addrLen = IS_IPV4_ADDR(pSrc) ? IPV4_ADDRESS_LENGTH : IPV6_ADDRESS_LENGTH;
-        if (pIceAgent->pDataSendingIceCandidatePair->local->pSocketConnection == pSocketConnection &&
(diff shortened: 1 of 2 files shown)


MushMal:
Any way we could validate this in a targeted test?

hassanctech:
Do we also need to check `pIceAgent->pDataSendingIceCandidatePair->local` for NULL?

lherman-cs:
> Do we also need to check `pIceAgent->pDataSendingIceCandidatePair->local` for NULL?

I don't think it's possible. The local candidate is set here, https://github.com/awslabs/amazon-kinesis-video-streams-webrtc-sdk-c/blob/bc5306dc3be8f1b9a6b047818b8903828f7d9719/src/source/Ice/IceAgent.c#L986-L992

If it’s remote, it’ll use localCandidates linked list which is guaranteed that the list item is not NULL. The local candidates list are appended through `iceAgentInitRelayCandidate`, `iceAgentInitHostCandidate`, and `iceAgentInitSrflxCandidate`. These must be not NULL because that’s how we keep track local socket connections to receive incoming data.

lherman-cs:
> Any way we could validate this in a targeted test?

@MushMal Added a unit test. Removing the fix, the unit test could hit 100% of the times in my tests.

PR #763, opened 2020-08-31
Ignore fork builds for check-sample

lherman-cs:
Currently, `check-sample` always fails when there's no AWS credentials provided. This is great and expected but we should not run the check when it's used for checking fork builds as they won't have AWS credentials in Travis.

### DIFF
--- a/check-sample.sh
+++ b/check-sample.sh
@@ -1,6 +1,13 @@
 #!/bin/bash
 
+if [[ -z "$AWS_ACCESS_KEY_ID" || -z "$AWS_SECRET_ACCESS_KEY" ]]
+then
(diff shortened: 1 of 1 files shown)


disa6302:
Only one concern: What if the PR from fork involves a change in the samples?

lherman-cs:
@disa6302 Unfortunately, we need to pay attention more closely manually. This is similar to any other unit tests, we'll skip them when it's a fork build. e.g.

https://github.com/awslabs/amazon-kinesis-video-streams-webrtc-sdk-c/blob/f25dc67d80a906d32de0fcdc530f7c92baca2639/tst/SignalingApiTest.cpp#L62-L64

Issue #801, opened 2020-09-16
[BUG] Race condition of `signalingClientHandle` in the sample code


staowyze:
Hi experts,

In the sample code, `signalingClientHandle` is freed and recreated at https://github.com/awslabs/amazon-kinesis-video-streams-webrtc-sdk-c/blob/master/samples/Common.c#L904
`recreateSignalingClient` is atomic, however is it possible  that `signalingClientHandle` is accessed by multiple threads?

Not very familiar with the implementation of signaling state machine, couldn't get confident answer by myself, any help is appreciated!

Thank you.

staowyze:
@MushMal what do you think?

MushMal:
@staowyze there is indeed a chance of a race condition as we are not interlocking. The race could happen when we receive a message (either trickle ICE candidate or an offer for example) - this is handled on a separate thread that's created within the signaling client. The race could happen if we are still within the thread handling the message and an error gets triggered by the signaling (either due to ICE refresh failure or reconnect failure) and we execute the free call before the message thread terminates.

While this is not possible to hit in practice due to the retry logic in the signaling client, it's still possible to hit this in an odd timing case when the signaling client is configured with no retry or no reconnect.

This should be properly addressed in the samples.

I am changing the title and the label to BUG. Thanks for reporting.

staowyze:
Thanks for confirming @MushMal ! Could you please elaborate why it is not possible to hit in practice due to the retry logic in the signaling client?
My metrics show there are crashes on our devices, especially when they are in trouble to reconnect signaling channel, i am wondering is it possible that, ICE refreshing could hit the race condition?

MushMal:
It is possible to hit race condition even with the retries in refreshing ICE config. This said, I would recommend capturing the crash stack to ensure it's not something else that's crashing.

We will prioritize the fix for this

staowyze:
thank you! this is the top issue I have seen on my devices.

staowyze:
@MushMal any updates on this?

MushMal:
Sorry, nothing yet. We will update as soon as we have something

MushMal:
Just had a chance to look into this. I believe the crashes you were seeing are related to the NULL deref. There is also the race condition that's been fixed in this PR: #810 - please take a look. We will get this through and push this into master branch ASAP.

staowyze:
[...]

PR #810, opened 2020-09-20
Fixing NULL deref and a race condition

MushMal:
We had a regression with NULL deref as the remote peer connection was not being removed from the hash table. This is what's been causing the crashes on some reported cases.

There is also a potential for a race condition that's also been fixed with these changes.

Issue #801 

*Description of changes:*


samples/Common.c
```diff
@@ -868,16 +864,18 @@ STATUS sessionCleanupWait(PSampleConfiguration pSampleConfiguration)
     ENTERS();
     STATUS retStatus = STATUS_SUCCESS;
     PSampleStreamingSession pSampleStreamingSession = NULL;
-    UINT32 i;
-    BOOL locked = FALSE;
+    UINT32 i, clientIdHash;
+    ;
```
hassanctech, line 868: Stray `;`

samples/Common.c
```diff
@@ -890,15 +888,19 @@ STATUS sessionCleanupWait(PSampleConfiguration pSampleConfiguration)
                 pSampleConfiguration->sampleStreamingSessionList[i] =
                     pSampleConfiguration->sampleStreamingSessionList[pSampleConfiguration->streamingSessionCount];
 
+                // Remove from the hash table
+                clientIdHash = COMPUTE_CRC32((PBYTE) pSampleStreamingSession->peerId, (UINT32) STRLEN(pSampleStreamingSession->peerId));
+                CHK_STATUS(hashTableContains(pSampleConfiguration->pRtcPeerConnectionForRemoteClient, clientIdHash, &peerConnectionFound));
+                if (peerConnectionFound) {
+                    CHK_STATUS(hashTableRemove(pSampleConfiguration->pRtcPeerConnectionForRemoteClient, clientIdHash));
```
disa6302, line 895: Do we want to to jump to CleanUp if `hashTableRemove` fails? We would not be freeing the streaming session in this case.

samples/Common.c
MushMal: D'oh!. Will fix. Apparently, the clang formatter does smart things :)

samples/Common.c
MushMal: This is the case when we have something unexpected. I think the best would be to fail indeed.

hassanctech (approved):
LGTM

disa6302 (approved):

Issue #812, opened 2020-09-20
[BUG] Gst Sample Starting Writing Frames to RTP Transceiver before DTLS session is established


suggestedfixes:
**Describe the bug**
Would be nice to include a DTLS callback, or a way of detecting the session is ready before writing to transceiver.

chehefen:
can you give us some details what issue did you run it?

suggestedfixes:
When a second session joins, we start seeing a bunch of writeFrame message failed. Would that cause any issues?

disa6302:
@suggestedfixes , 

What is the failure? Do you have logs?

suggestedfixes:
Everything leading up to DTLS init looks fine. writeFrame might have been non-critical?

```
2020-09-21 23:05:16 DEBUG   on_new_sample(): writeFrame failed with 0x5b000006
2020-09-21 23:05:16 DEBUG   on_new_sample(): writeFrame failed with 0x5b000006
2020-09-21 23:05:16 DEBUG   turnConnectionHandleStun(): create permission succeeded for peer 3.16.13.80
2020-09-21 23:05:16 DEBUG   turnConnectionHandleStun(): create permission succeeded for peer 3.16.13.80
2020-09-21 23:05:16 DEBUG   on_new_sample(): writeFrame failed with 0x5b000006
2020-09-21 23:05:16 DEBUG   turnConnectionHandleStun(): Channel bind succeeded with peer 3.16.13.80, port: 57069, channel number 16385
2020-09-21 23:05:16 DEBUG   turnConnectionHandleStun(): Channel bind succeeded with peer 3.16.13.80, port: 63391, channel number 16386
2020-09-21 23:05:16 DEBUG   on_new_sample(): writeFrame failed with 0x5b000006
2020-09-21 23:05:16 DEBUG   turnConnectionHandleStun(): TURN Allocation freed.
2020-09-21 23:05:16 DEBUG   socketConnectionClosed(): Close socket 32
2020-09-21 23:05:16 DEBUG   turnConnectionStepState(): TurnConnection state changed from TURN_STATE_CLEAN_UP to TURN_STATE_NEW
2020-09-21 23:05:16 DEBUG   on_new_sample(): writeFrame failed with 0x5b000006
2020-09-21 23:05:16 DEBUG   on_new_sample(): writeFrame failed with 0x5b000006
[...]

PR #814, opened 2020-09-21
Check for SRTP session ready in samples

disa6302:
*Issue #, if available:*
#812 
*Description of changes:*
Add check for STATUS_SRTP_NOT_READY_YET status code in gst sample to avoid log spew of writeFrame() status code.

### DIFF
--- a/samples/kvsWebRTCClientMasterGstreamerSample.c
+++ b/samples/kvsWebRTCClientMasterGstreamerSample.c
@@ -77,9 +77,8 @@ GstFlowReturn on_new_sample(GstElement* sink, gpointer data, UINT64 trackid)
                 frame.decodingTs = frame.presentationTs;
                 pSampleStreamingSession->videoTimestamp += SAMPLE_VIDEO_FRAME_DURATION; // assume video fps is 30
             }
-
             status = writeFrame(pRtcRtpTransceiver, &frame);
-            if (status != STATUS_SUCCESS) {
+            if (status != STATUS_SRTP_NOT_READY_YET && status != STATUS_SUCCESS) {
 #ifdef VERBOSE
                 printf("writeFrame() failed with 0x%08x", status);
 #endif


samples/kvsWebRTCClientMasterGstreamerSample.c
```diff
@@ -78,11 +78,12 @@ GstFlowReturn on_new_sample(GstElement* sink, gpointer data, UINT64 trackid)
                 pSampleStreamingSession->videoTimestamp += SAMPLE_VIDEO_FRAME_DURATION; // assume video fps is 30
             }
 
-            status = writeFrame(pRtcRtpTransceiver, &frame);
-            if (status != STATUS_SUCCESS) {
+            if (status != STATUS_SRTP_NOT_READY_YET) {
```
MushMal, line 81: I don't understand any of this code. Who sets the status in the first place?

disa6302: The writeFrame call sets it. Earlier, we were returning `STATUS_SUCCESS` until SRTP was not set up. But, there was an issue while trying to calculate start up latency since the frames would not be sent out till SRTP was ready and application had no way of knowing it.

MushMal: Oh - it's actually right on LN 80. My bad, I was staring at it and didn't see it being set.

Suggest to combine it

if (STATUS_FAILED(status) && status != STATUS_SRTP_NOT_READY_YET) {
blah;
}

disa6302: Did not use `STATUS_FAILED` since the sample does not use any of our usual `CHK_x` or "SDK" macros. So keeping it consistent with that. But i can certainly combine the if's

MushMal (approved):

PR #816, opened 2020-09-22
Remove log level override in createPeerConnection.

andyhoffy:
*Issue #, if available:*

N/A 

*Description of changes:*

This pull request reverts a change that causes the log level to be overridden (set to DEBUG if no env var is set) inside createPeerConnection. This results in programmatically set values being overridden.

Before this change: Any log level set via ```loggerSetLogLevel``` and/or ```SignalingClientInfo::logLevel``` would be overridden by a call to createPeerConnection.

After this change: ```loggerSetLogLevel``` and ```SignalingClientInfo::logLevel``` (in conjunction) successfully define/maintain a desired log output level.

Original commit, if helpful: https://github.com/awslabs/amazon-kinesis-video-streams-webrtc-sdk-c/commit/6fbedcb380f42263ff0bd8e5638912e8ad19c5f8

### DIFF
--- a/src/source/PeerConnection/PeerConnection.c
+++ b/src/source/PeerConnection/PeerConnection.c
@@ -636,8 +636,6 @@ STATUS createPeerConnection(PRtcConfiguration pConfiguration, PRtcPeerConnection
     PKvsPeerConnection pKvsPeerConnection = NULL;
     IceAgentCallbacks iceAgentCallbacks;
     DtlsSessionCallbacks dtlsSessionCallbacks;
-    UINT32 logLevel = LOG_LEVEL_DEBUG;
-    PCHAR logLevelStr = NULL;
     PConnectionListener pConnectionListener = NULL;
 
     CHK(pConfiguration != NULL && ppPeerConnection != NULL, STATUS_NULL_ARG);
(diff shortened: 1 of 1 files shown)


MushMal (commented):
Not sure if this is the right approach. Think of the WebRTC object graph routed to PeerConnection object as a separate set of objects than that of the signaling client. Each will need to have it's own mechanism of setting and clearing it. Perhaps, adding a logic similar to the signaling client is the right approach and make sure it's properly coordinated at the higher-level object such as the sample. Think of the applications that simply won't use the signaling client and will use the PeerConnection standalone.

andyhoffy:
Hmm, if you'd like a side effect of configuring logging at various entry points, how about making automatic log configuration a one-shot deal? Something like:
```
// Global variable that lives with other logging globals... ensures that logging is auto-config'd at most once:
extern volatile bool gLogInitComplete; // default to false

// New method - safe to call many times at critical entry-points in the lib, but
// does not override settings if logging has been config'd already.
void loggerInit() {
  if (!gLogInitComplete) {
    // Logic here to read from env, if not already config'd.
    loggerSetLogLevel(env_level);
  }
}
...
void loggerSetLogLevel(...) {
  // Acquire lock if not locked.
  gLoggerInitComplete = true;
  // Configure logging (existing method).
}
```

andyhoffy:
Addressed by https://github.com/awslabs/amazon-kinesis-video-streams-webrtc-sdk-c/pull/819 - thank you!

PR #819, opened 2020-09-23
Removing log level setting from the peer connection

MushMal:
*Issue #, if available:*

Removing the log level setting from the peer connection. The problem is that the customers application will not be able to control the log level as the peer connection will keep resetting it every time a new one is created. Moreover, the dynamic setting of the env var will mess things up. The application should have a full control on what and when the log level should be set.

Adding extra check in the Common.c to validate the range.

Not touching the signaling - it's a longer living object. The application can control it via the ClientInfo.logLevel.

### DIFF
--- a/samples/Common.c
+++ b/samples/Common.c
@@ -570,7 +570,7 @@ STATUS createSampleConfiguration(PCHAR channelName, SIGNALING_CHANNEL_ROLE_TYPE 
     STATUS retStatus = STATUS_SUCCESS;
     PCHAR pAccessKey, pSecretKey, pSessionToken, pLogLevel;
     PSampleConfiguration pSampleConfiguration = NULL;
-    UINT32 logLevel;
+    UINT32 logLevel = LOG_LEVEL_DEBUG;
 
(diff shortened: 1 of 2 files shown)


lherman-cs:
@MushMal should we also remove/ignore ClientInfo.logLevel? It's definitely an API change, but I think it's the right thing. Let say users have 2 signaling clients (although uncommon). The last signaling client creation will override the logging level.

Since we use global logging in PIC, I think we should let users to set the logging level globally as well instead of per instance, with `SET_LOGGER_LOG_LEVEL`.

MushMal:
> @MushMal should we also remove/ignore ClientInfo.logLevel? It's definitely an API change, but I think it's the right thing. Let say users have 2 signaling clients (although uncommon). The last signaling client creation will override the logging level.
> 
> Since we use global logging in PIC, I think we should let users to set the logging level globally as well instead of per instance, with `SET_LOGGER_LOG_LEVEL`.

This is a little different. The application still is in the control of the log level. The application can ensure the order of creation and which values to specify. I think we should be OK for the signaling client.

lherman-cs (approved):
This is an ok solution for now. Let's not break any API.

PR #830, opened 2020-09-30
Change lwsapicalls buffer allocation

ycyang1229:
*Issue #, if available:*

*Description of changes:*
Change lwsapicalls buffer allocation


MushMal (commented):
Could you please format the changes as your spacing is off.

My main concern is about this bulk migration from stack allocation to heap. We need to get these changes properly justified. What's the actual stack size on the OS you are running and is it possible instead to adjust the stack size rather than going over the entire codebase and making these changes. The concern is not only the churn that could introduce bugs which are hard to debug but also performance hit on some platforms.

ycyang1229:
> Could you please format the changes as your spacing is off.
> 
> My main concern is about this bulk migration from stack allocation to heap. We need to get these changes properly justified. What's the actual stack size on the OS you are running and is it possible instead to adjust the stack size rather than going over the entire codebase and making these changes. The concern is not only the churn that could introduce bugs which are hard to debug but also performance hit on some platforms.

 Could you please format the changes as your spacing is off. ==> Would you mind explaining this to me? I did not get it.

Indeed, it may introduce the memory leak bugs if it is not freed. I think this can be back to one basic question, how much stack size does this sdk need? No matter what platform is, the stack size should be almost the same. Would you mind profiling the maximum of stack size in this sdk based on your platform? I think this kind of measurement result of memory profiling is quite important for c sdk and embedded devices.

For now, I used ESP32 platform(which has 520KB internal ram and 4MB external ram) to run this sdk, and I statically allocate each tasks with the same stack size as much as I can(20KB). Of course, I can dynamically allocate the different stack size and recollect the unused stack back, but it needs to modify much code of this sdk to run it. Will try to run this sdk on another vendor's board which has similar ram design in the next weeks. 

Btw, what kind of performance issue have you hit? Would you mind giving me more detail? In generally speaking, I do not think it will hit performance issue since those code only run once or the period is pretty long.

MushMal:
Your suggestion of profiling is a really good one. We have a work item to add profiling that among other things would baseline the stack, dynamic memory, CPU usage, etc.. so we can rely that consequent changes do not degrade the perf.

However, profiling overall stack usage is not a trivial task as it would heavily depend on the end-to-end scenario, and, in our case it would also depend on the networking and retries. Our format state machine uses a stack for the retries.

The performance is one of the big reasons we prefer the stack allocations way over the heap allocations. This is especially important for the functionality that's called frequently. Stack allocation is two orders of magnitude more performant than heap allocation. This would be impacting low-footprint devices as well as super scalable high-end scenarios such as backend large machine instances with multiple cores processing thousands of streams.

As such, it's a difficult decision but one that we would need to be super careful with

ycyang1229:
> Your suggestion of profiling is a really good one. We have a work item to add profiling that among other things would baseline the stack, dynamic memory, CPU usage, etc.. so we can rely that consequent changes do not degrade the perf.
> 
> However, profiling overall stack usage is not a trivial task as it would heavily depend on the end-to-end scenario, and, in our case it would also depend on the networking and retries. Our format state machine uses a stack for the retries.
> 
> The performance is one of the big reasons we prefer the stack allocations way over the heap allocations. This is especially important for the functionality that's called frequently. Stack allocation is two orders of magnitude more performant than heap allocation. This would be impacting low-footprint devices as well as super scalable high-end scenarios such as backend large machine instances with multiple cores processing thousands of streams.
> 
> As such, it's a difficult decision but one that we would need to be super careful with

Understood. But likes you said, in our case, networking is the major factor. I think networking dominates performance first. The difference between stack and heap may not take huge part in the process of signaling, and there may be other places can be improved for this process. 

How about starting reduce the length of this buffer first? I am not sure why this takes 10KB? Is it possible to set it as smaller size? I set this as 1024Bytes, and it still works. 
https://github.com/awslabs/amazon-kinesis-video-streams-pic/blob/master/src/client/include/com/amazonaws/kinesis/video/client/Include.h#L253

MushMal:
> > Your suggestion of profiling is a really good one. We have a work item to add profiling that among other things would baseline the stack, dynamic memory, CPU usage, etc.. so we can rely that consequent changes do not degrade the perf.
> > However, profiling overall stack usage is not a trivial task as it would heavily depend on the end-to-end scenario, and, in our case it would also depend on the networking and retries. Our format state machine uses a stack for the retries.
> > The performance is one of the big reasons we prefer the stack allocations way over the heap allocations. This is especially important for the functionality that's called frequently. Stack allocation is two orders of magnitude more performant than heap allocation. This would be impacting low-footprint devices as well as super scalable high-end scenarios such as backend large machine instances with multiple cores processing thousands of streams.
> > As such, it's a difficult decision but one that we would need to be super careful with
> 
> Understood. But likes you said, in our case, networking is the major factor. I think networking dominates performance first. The difference between stack and heap may not take huge part in the process of signaling, and there may be other places can be improved for this process.
> 
> How about starting reduce the length of this buffer first? I am not sure why this takes 10KB? Is it possible to set it as smaller size? I set this as 1024Bytes, and it still works.
> https://github.com/awslabs/amazon-kinesis-video-streams-pic/blob/master/src/client/include/com/amazonaws/kinesis/video/client/Include.h#L253

If you are referring to MAX_URI_CHAR_LEN then this is a generic URI max len which is what seem to be the lowest common denominator on various platforms/systems/use cases. Most URIs are a lot shorter indeed. This said, the usage in the structures can change to be more specific. For example, in the WebRTC Signaling, we don't use this but instead a stricter/shorter version

ycyang1229:
> > > Your suggestion of profiling is a really good one. We have a work item to add profiling that among other things would baseline the stack, dynamic memory, CPU usage, etc.. so we can rely that consequent changes do not degrade the perf.
> > > However, profiling overall stack usage is not a trivial task as it would heavily depend on the end-to-end scenario, and, in our case it would also depend on the networking and retries. Our format state machine uses a stack for the retries.
> > > The performance is one of the big reasons we prefer the stack allocations way over the heap allocations. This is especially important for the functionality that's called frequently. Stack allocation is two orders of magnitude more performant than heap allocation. This would be impacting low-footprint devices as well as super scalable high-end scenarios such as backend large machine instances with multiple cores processing thousands of streams.
> > > As such, it's a difficult decision but one that we would need to be super careful with
> > 
> > 
> > Understood. But likes you said, in our case, networking is the major factor. I think networking dominates performance first. The difference between stack and heap may not take huge part in the process of signaling, and there may be other places can be improved for this process.
> > How about starting reduce the length of this buffer first? I am not sure why this takes 10KB? Is it possible to set it as smaller size? I set this as 1024Bytes, and it still works.
> > https://github.com/awslabs/amazon-kinesis-video-streams-pic/blob/master/src/client/include/com/amazonaws/kinesis/video/client/Include.h#L253
> 
> If you are referring to MAX_URI_CHAR_LEN then this is a generic URI max len which is what seem to be the lowest common denominator on various platforms/systems/use cases. Most URIs are a lot shorter indeed. This said, the usage in the structures can change to be more specific. For example, in the WebRTC Signaling, we don't use this but instead a stricter/shorter version

Thanks for the information. Is it possible that you can provide the maximum length for webrtc since webrtc does not need 10Kbytes?

MushMal:
> > > > Your suggestion of profiling is a really good one. We have a work item to add profiling that among other things would baseline the stack, dynamic memory, CPU usage, etc.. so we can rely that consequent changes do not degrade the perf.
> > > > However, profiling overall stack usage is not a trivial task as it would heavily depend on the end-to-end scenario, and, in our case it would also depend on the networking and retries. Our format state machine uses a stack for the retries.
> > > > The performance is one of the big reasons we prefer the stack allocations way over the heap allocations. This is especially important for the functionality that's called frequently. Stack allocation is two orders of magnitude more performant than heap allocation. This would be impacting low-footprint devices as well as super scalable high-end scenarios such as backend large machine instances with multiple cores processing thousands of streams.
> > > > As such, it's a difficult decision but one that we would need to be super careful with
> > > 
> > > 
> > > Understood. But likes you said, in our case, networking is the major factor. I think networking dominates performance first. The difference between stack and heap may not take huge part in the process of signaling, and there may be other places can be improved for this process.
> > > How about starting reduce the length of this buffer first? I am not sure why this takes 10KB? Is it possible to set it as smaller size? I set this as 1024Bytes, and it still works.
> > > https://github.com/awslabs/amazon-kinesis-video-streams-pic/blob/master/src/client/include/com/amazonaws/kinesis/video/client/Include.h#L253
> > 
> > 
> > If you are referring to MAX_URI_CHAR_LEN then this is a generic URI max len which is what seem to be the lowest common denominator on various platforms/systems/use cases. Most URIs are a lot shorter indeed. This said, the usage in the structures can change to be more specific. For example, in the WebRTC Signaling, we don't use this but instead a stricter/shorter version
> 
> Thanks for the information. Is it possible that you can provide the maximum length for webrtc since webrtc does not need 10Kbytes?

I am not sure which one you refer to. We were trying to not use MAX_URI_CHAR_LEN but directly specify the lengths that are required in the WebRTC object graph

Issue #756, opened 2020-08-27
kvsWebrtcClientMaster not working on Windows


Ben8823:
I'm trying to get the kvsWebrtcClientMaster program to run under windows. It compiles fine, if i run it it creates the signaling channels and prints out the following.
```
[KVS Master] Signaling client connection to socket established
[KVS Master] Channel test set up done
```
Unfortunately no video is showing up on the KVS WebRTC Test Page. If I try the same from ubuntu I can watch the video.
Did anyone get it to run under windows? If yes are there any changes requried?

MushMal:
No changes should be required. Could you provide full debug logs for both the master and the viewer. Include detailed information about the network topology, your application, scenario, etc

Ben8823:
Ok that's strange. I noticed that in the handleOffer function in Common.c the call to setLocalDescription fails.
The callstack looks like this:
```
iceAgenstStartGathering->iceAgentInitRelayCandidates->iceAgentInitRelayCandidate->createSocketConnection->socketConnect
```
the connect in socketConnect returns -1 and only errno is checked which is not set on windows. on windows WSAGetLastError has to be called.
but this would mean that the socket codebase is not ready for windows yet?

I'm testing it on LAN behind NAT, both on the same machine.
Attached you can find the log of master and viewer.

[viewer.log](https://github.com/awslabs/amazon-kinesis-video-streams-webrtc-sdk-c/files/5137742/viewer.log)
[master.log](https://github.com/awslabs/amazon-kinesis-video-streams-webrtc-sdk-c/files/5137743/master.log)

disa6302:
@Ben8823 , 

It does look like we do not support `WSAGetLastError` in our stack. We will look into it. Meanwhile, can you use WSAGetLastError and FormatMessage in [createSocket()](https://github.com/awslabs/amazon-kinesis-video-streams-webrtc-sdk-c/blob/a7da1e06166ca2931c626b11e24f4bb2c05d4a0f/src/source/Ice/Network.c#L146) to get the error string so that we can understand what the error is?

Ben8823:
@disa6302 

ok but as far as i know there is not really an alternative to `WSAGetLastError()` if compiling on windows. 
[...]

PR #834, opened 2020-09-30
Add internal getErrorCode() and getErrorString()

grahn:
*Issue #, if available:*
#756 references `WSAGetLastError()`. So this pull request might relate to that issue.

*Description of changes:*
These functions allow the addition of error handling support for
Windows Sockets Error Codes. On Windows the functions use `WSAGetLastError()` and
`FormatMessage()`, and on non-Windows platforms they use `errno` and
`strerror()`.

In addition, these functions provide mapping from Windows Sockets
Error Codes to BSD error codes for the error codes that are explicitly
checked in code.

### DIFF
--- a/src/source/Ice/Network.c
+++ b/src/source/Ice/Network.c
@@ -156,17 +156,17 @@ STATUS createSocket(KVS_IP_FAMILY_TYPE familyType, KVS_SOCKET_PROTOCOL protocol,
 
     sockfd = socket(familyType == KVS_IP_FAMILY_TYPE_IPV4 ? AF_INET : AF_INET6, sockType, 0);
     if (sockfd == -1) {
-        DLOGW("socket() failed to create socket with errno %s", strerror(errno));
+        DLOGW("socket() failed to create socket with errno %s", getErrorString(getErrorCode()));
         CHK(FALSE, STATUS_CREATE_UDP_SOCKET_FAILED);
     }
 
     optionValue = 1;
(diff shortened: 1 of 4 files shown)


src/source/Ice/Network.c
```diff
@@ -382,3 +383,68 @@ BOOL isSameIpAddress(PKvsIpAddress pAddr1, PKvsIpAddress pAddr2, BOOL checkPort)
...
+            break;
+        case EINTR:
+            error = WSAEINTR;
+            break;
+        default:
+            /* leave unchanged */
+            break;
+    }
+    if (FormatMessage((FORMAT_MESSAGE_FROM_SYSTEM | FORMAT_MESSAGE_IGNORE_INSERTS), NULL, error, MAKELANGID(LANG_NEUTRAL, SUBLANG_DEFAULT), buffer,
+                      sizeof(buffer), NULL) == 0) {
```
MushMal, line 439: Could you please change sizeof() -> SIZEOF() and sprintf_s -> SNPRINTF

MushMal (commented):
Niiiice! Thanks a lot for taking care of it. 

Just have a minor comment. Could you please address this? We would like to keep the consistency in the codebase that will allow system-wide single place changes by redefining the macros

src/source/Ice/Network.c
grahn: Yes, I pushed another commit that makes this change.

I can squash the additional commit if you want me too.

MushMal (approved):

MushMal:
Thanks. Merged!

PR #841, opened 2020-10-05
some toolchains may not have the interface of getifaddrs.

ycyang1229:
*Issue #, if available:*

*Description of changes:*


src/source/Ice/Network.c
```diff
@@ -9,14 +9,17 @@ STATUS getLocalhostIpAddresses(PKvsIpAddress destIpList, PUINT32 pDestIpListLen,
...
     UINT32 ipCount = 0, destIpListLen;
+
     BOOL filterSet = TRUE;
 
 #ifdef _WIN32
     DWORD retWinStatus, sizeAAPointer;
     PIP_ADAPTER_ADDRESSES adapterAddresses, aa = NULL;
     PIP_ADAPTER_UNICAST_ADDRESS ua;
 #else
+#ifdef KVCWEBRTC_HAVE_IFADDRS_H
```
MushMal, line 20: These could be handled (aka - an error can be generated) in the CMakeList.txt file instead of setting this param and erring here.

ycyang1229: Indeed, actually it is a favor for people are porting on different platforms. Before building everything, people can see everything needed by this sdk, and have a whole picture on these platform-related stuff. And also hope developers are aware of platform-related stuffs. When these similar stuffs becomes more, we should move those stuffs into dedicated folders or files to make a wrapper for them. 
Should I remove this error part? What do you suggest?

MushMal: Developers working on very specific custom platforms should know the platforms well - there is little we can help. For now, as we don't have a support for those platforms, I think we should fail in the CMake itself. What your changes do is that they detect the condition and set a variable but then they fail the compiler in the code. Instead, I am suggesting to fail the compilation in the CMakeFile.txt instead of setting KVSWEBRTC_HAVE_IFADDRES_H

ycyang1229: Sure. Should I close this pr or,  remove the KVSWEBRTC_HAVE_IFADDRES_H  and keep KVSWEBRTC_HAVE_GETIFADDRS? thanks.

MushMal: Just upload another commit that only has changes in CMakeList.txt. The main body of determining the support. Then fail if the var is not defined. I am not certain of the syntax but it should be something like this

if(NOT KVSWEBRTC_HAVE_GETIFADDRS)
    message(FATAL_ERROR "Platform should support getIfAddr API")
endif()

We should also ensure Windows works fine too.

ycyang1229: Done. force push into one commit. thanks.

MushMal (approved):

PR #843, opened 2020-10-05
Modify sdp ice candidate parser

ycyang1229:
*Issue #, if available:*

*Description of changes:*
terminate the parsing of sdp candidate early when the ip of ice candidate is not valid such as "candidate": "candidate:2977641484 1 udp 2113937151 b3c423be-e111-420a-9b06-755a59cf42d1.local 47036 typ host generation 0 ufrag wwMY network-cost 999".


src/source/Ice/IceAgent.c
```diff
@@ -376,6 +376,9 @@ STATUS iceAgentAddRemoteCandidate(PIceAgent pIceAgent, PCHAR pIceCandidateString
         curr = next + 1;
     }
 
+    CHK(foundPort, STATUS_ICE_CANDIDATE_STRING_MISSING_PORT);
```
MushMal: These are good checks but we might need some unit tests around them.

Most importantly, there is a bad logic in LN 368-369. I am not sure why we set foundPort to TRUE and immediately check what we set. This needs to be properly fixed - from the look of it it seems that we need to remove LN 369.

Could you please think about how to add unit tests for these?

ycyang1229: Indeed, fixed it. Forgot to remove it  after modifying the code for the unit test. Speaking of the unit test, I can do it after creating those pr. For now, add the example I mention may be enough. What do you think?

MushMal (approved):

PR #846, opened 2020-10-06
disable data-channel related stuffs when disabling it in the cmakelist.

ycyang1229:
*Issue #, if available:*

*Description of changes:*


ycyang1229:
I have another commit for disabling the streaming part and keeping the data channel part only, but I am not sure you will take it or not. If yes, I can prepare it.

MushMal (commented):
What's prompting these changes? 

In general, we can't take these types of changes lightly. First of all, the implementation will proliferate conditional compilation through the codebase - this is one of the things that we are very careful of - even though the entire codebase is very platform independent, unlike other codebases, it's extremely easy to read due to the severe emphasis of not using conditional compilation at all throughout the code but ONLY in a very few specific files which are designed to abstract the platform capability away.

If, for some reason, we decide to enable/disable the channels, it would be implemented by abstracting the functionality behind a common interface which would then be defined in a particular file to map to a type of functionality that's needed.

Again, let's address the data channel removal motivation first - this is an inherent part of the webrtc spec and there is no intent to remove those.

ycyang1229:
> What's prompting these changes?
> 
> In general, we can't take these types of changes lightly. First of all, the implementation will proliferate conditional compilation through the codebase - this is one of the things that we are very careful of - even though the entire codebase is very platform independent, unlike other codebases, it's extremely easy to read due to the severe emphasis of not using conditional compilation at all throughout the code but ONLY in a very few specific files which are designed to abstract the platform capability away.
> 
> If, for some reason, we decide to enable/disable the channels, it would be implemented by abstracting the functionality behind a common interface which would then be defined in a particular file to map to a type of functionality that's needed.
> 
> Again, let's address the data channel removal motivation first - this is an inherent part of the webrtc spec and there is no intent to remove those.

Sure, agree with what you said but the idea of this commit comes from one customer asking is it possible to only have the functionality of data channel before.  When I ported the this sdk on the esp32 and other rtos, I realized the porting of usrsctp may not be a easy thing, and it depends on tcp/ip stack heavily. So if one embedded ipc device only wants to have the functionality of video streaming and its tcp/ip stack is simple, customers may want to drop the functionality of data channel. That's why I separated the data part first, and then tried to disable media part(enable data part only) later.

For me, I will say webrtc provides two independent functionalities, one is media part, and another is data part. If the embedded  devices only wants one of them, it should be doable. Not sure my understanding is right, so please correct me if I am wrong. I also hope the relationship between webrtc and other third party can be weaker, although libwebsockets, libsrtp and usrsctp are the best choices right now. I understand this commit is not a good one. So if you want to make them independently, I can change this into one abstract implementation. If you think this is not a good idea, I will close this pr. thanks.

MushMal:
> > What's prompting these changes?
> > In general, we can't take these types of changes lightly. First of all, the implementation will proliferate conditional compilation through the codebase - this is one of the things that we are very careful of - even though the entire codebase is very platform independent, unlike other codebases, it's extremely easy to read due to the severe emphasis of not using conditional compilation at all throughout the code but ONLY in a very few specific files which are designed to abstract the platform capability away.
> > If, for some reason, we decide to enable/disable the channels, it would be implemented by abstracting the functionality behind a common interface which would then be defined in a particular file to map to a type of functionality that's needed.
> > Again, let's address the data channel removal motivation first - this is an inherent part of the webrtc spec and there is no intent to remove those.
> 
> Sure, agree with what you said but the idea of this commit comes from one customer asking is it possible to only have the functionality of data channel before. When I ported the this sdk on the esp32 and other rtos, I realized the porting of usrsctp may not be a easy thing, and it depends on tcp/ip stack heavily. So if one embedded ipc device only wants to have the functionality of video streaming and its tcp/ip stack is simple, customers may want to drop the functionality of data channel. That's why I separated the data part first, and then tried to disable media part(enable data part only) later.
> 
> For me, I will say webrtc provides two independent functionalities, one is media part, and another is data part. If the embedded devices only wants one of them, it should be doable. Not sure my understanding is right, so please correct me if I am wrong. I also hope the relationship between webrtc and other third party can be weaker, although libwebsockets, libsrtp and usrsctp are the best choices right now. I understand this commit is not a good one. So if you want to make them independently, I can change this into one abstract implementation. If you think this is not a good idea, I will close this pr. thanks.

Understood. WebRTC is a loosely coupled set of RFCs and one of those is the data channel. While I understand the desired outcome on the small platforms, most of the customers would like to have more compliance and a complete feature-set. Let us get back to the drawing board and think how we can isolate this better if needed instead of peppering the code with the conditional compilation. Those make any codebase impossible to read and reason about and we have tried to manage this in the entire codebase. If this really is a big issue and one that has a set of potential customers then could I ask you to email the list of the customers to kinesis-video-support@amazon.com so we can prioritize this work appropriately. As is, I would rather close this PR.

ycyang1229:
> > > What's prompting these changes?
> > > In general, we can't take these types of changes lightly. First of all, the implementation will proliferate conditional compilation through the codebase - this is one of the things that we are very careful of - even though the entire codebase is very platform independent, unlike other codebases, it's extremely easy to read due to the severe emphasis of not using conditional compilation at all throughout the code but ONLY in a very few specific files which are designed to abstract the platform capability away.
> > > If, for some reason, we decide to enable/disable the channels, it would be implemented by abstracting the functionality behind a common interface which would then be defined in a particular file to map to a type of functionality that's needed.
> > > Again, let's address the data channel removal motivation first - this is an inherent part of the webrtc spec and there is no intent to remove those.
> > 
> > 
> > Sure, agree with what you said but the idea of this commit comes from one customer asking is it possible to only have the functionality of data channel before. When I ported the this sdk on the esp32 and other rtos, I realized the porting of usrsctp may not be a easy thing, and it depends on tcp/ip stack heavily. So if one embedded ipc device only wants to have the functionality of video streaming and its tcp/ip stack is simple, customers may want to drop the functionality of data channel. That's why I separated the data part first, and then tried to disable media part(enable data part only) later.
> > For me, I will say webrtc provides two independent functionalities, one is media part, and another is data part. If the embedded devices only wants one of them, it should be doable. Not sure my understanding is right, so please correct me if I am wrong. I also hope the relationship between webrtc and other third party can be weaker, although libwebsockets, libsrtp and usrsctp are the best choices right now. I understand this commit is not a good one. So if you want to make them independently, I can change this into one abstract implementation. If you think this is not a good idea, I will close this pr. thanks.
> 
> Understood. WebRTC is a loosely coupled set of RFCs and one of those is the data channel. While I understand the desired outcome on the small platforms, most of the customers would like to have more compliance and a complete feature-set. Let us get back to the drawing board and think how we can isolate this better if needed instead of peppering the code with the conditional compilation. Those make any codebase impossible to read and reason about and we have tried to manage this in the entire codebase. If this really is a big issue and one that has a set of potential customers then could I ask you to email the list of the customers to [kinesis-video-support@amazon.com](mailto:kinesis-video-support@amazon.com) so we can prioritize this work appropriately. As is, I would rather close this PR.

Sure, let me close this one, and ask customers send the mail if they need this. thanks.

PR #856, opened 2020-10-08
Fix RTX parsing logic to not require fmtp apt line to come immediately after rtpmap rtx line

hassanctech:
*Issue #, if available:* 811

*Description of changes:*

In Chrome and Safari the SDP Offer with RTX the relevant attributes appear like this:

```
a=rtpmap:102 H264/90000
a=rtcp-fb:102 goog-remb
a=rtcp-fb:102 transport-cc
a=rtcp-fb:102 ccm fir
a=rtcp-fb:102 nack
a=rtcp-fb:102 nack pli
a=fmtp:102 level-asymmetry-allowed=1;packetization-mode=1;profile-level-id=42001f
a=rtpmap:121 rtx/90000
a=fmtp:121 apt=102
```

In Firefox > v80, the appear like this:

```
a=fmtp:126 profile-level-id=42e01f;level-asymmetry-allowed=1;packetization-mode=1
a=fmtp:97 profile-level-id=42e01f;level-asymmetry-allowed=1
a=fmtp:127 apt=126
a=fmtp:98 apt=97
...
a=rtpmap:126 H264/90000
a=rtpmap:127 rtx/90000
a=rtpmap:97 H264/90000
a=rtpmap:98 rtx/90000
```

Until and including Firefox v79 there was no support for RTX but in v80 it was added and Firefox orders the SDP attributes in alphabetical order, so the fmtp lines actually all occur *before* the rtpmap rtx lines.  Therefore it is necessary to first construct a map between encountered apt values (parsedPayloadType) and the fmtp value which is the rtx payload type.  Then at the end we iterate over these values and if the stream matches a supported codec we insert them into the RTX table.

I've removed the logic for actually checking for the RTX value in the rtpmap line, according to the spec here:  https://tools.ietf.org/html/rfc4588#section-8.1, apt only occurs in the context of RTX so this was no longer necessary. However if desired one could maintain a set of rtpmap values specified with rtx and prior to adding anything to the rtx table, make sure the parsed payload type is present in the set of rtpmap values.

I've verified the master sample works on Firefox v80/81 (which were previously broken), as well as Safari, Chrome, Firefox v79.


hassanctech:
It might be useful to note that the fmtp and apt values are going to be typically in the range 96-127 but certainly less than 8 bits. In the linked list each item is 64 bits, I store both of them, the apt at bits 0-7 and the fmtp at bits 8-15.
```
63 62 61 ....... 16 15 14 13 12 10 9 8 7 6 5 4 3 2 1 0
|      unused      |     |       fmtp          |  |         apt        |
```

src/source/PeerConnection/SessionDescription.c
```diff
@@ -155,12 +155,14 @@ STATUS setPayloadTypesFromOffer(PHashTable codecTable, PHashTable rtxTable, PSes
...
-    UINT64 parsedPayloadType, rtxPayloadType, hashmapPayloadType;
+    UINT64 parsedPayloadType, rtxPayloadType, hashmapPayloadType, fmtpVal, aptVal, aptFmtVal;
+    PSingleList pListAptFmtpVals = NULL;
     BOOL supportCodec;
-    UINT32 tokenLen;
+    UINT32 tokenLen, nodeCount, i;
 
     for (currentMedia = 0; currentMedia < pSessionDescription->mediaCount; currentMedia++) {
         pMediaDescription = &(pSessionDescription->mediaDescriptions[currentMedia]);
+        singleListCreate(&pListAptFmtpVals);
```
MushMal, line 165: There is a handful of these values. You could define a fixed array of MAX_FMTP_VALUES and a count and set the values in the array. You could have two arrays to store the the fmtp value and the type in each for example. It will be way faster than the list and will be almost the same size if there are only a few of these values expected.

src/source/PeerConnection/SessionDescription.c
```diff
@@ -214,32 +217,42 @@ STATUS setPayloadTypesFromOffer(PHashTable codecTable, PHashTable rtxTable, PSes
...
-                    if (supportCodec) {
-                        CHK_STATUS(hashTableGet(codecTable, RTC_CODEC_VP8, &hashmapPayloadType));
-                        if (parsedPayloadType == hashmapPayloadType) {
-                            CHK_STATUS(hashTableUpsert(rtxTable, RTC_RTX_CODEC_VP8, rtxPayloadType));
-                        }
-                    }
+        singleListGetNodeCount(pListAptFmtpVals, &nodeCount);
+
+        for (i = 0; i < nodeCount; i++) {
+            singleListGetNodeDataAt(pListAptFmtpVals, i, &aptFmtVal);
```
MushMal, line 232: Just a note that this will make your algorithm quadratic as getting node at N is a linear op. If you decide to go with the list solution then you should perhaps use an iterator here to linearly scan the list.

node = singleListGetHead();
while (node != NULL) {
...
node = singleListGetNextNode();
}

something like this - don't remember the syntax.

MushMal (commented):
Kudos to get to the bottom of this. The logic is convincing. I just made a comment about the implementation

src/source/PeerConnection/SessionDescription.c
hassanctech: Yeah I wasn't concerned with performance here since as you mentioned there aren't many of these, I would just pick from the head, delete and move on -- however I'm going to go with the fixed size array approach so this wouldn't be an issue.

src/source/PeerConnection/SessionDescription.c
hassanctech: I really wanted to do this but it was taking too long to find what MAX_FMTP_VALUES should be without using too much memory.  Now that you mention it as as well I'm going to use this, currently the max SDP attributes is 128 sdp attributes and each rtx type requires 2 attributes so 64 is a safe value for MAX_FMTP_VALUES, I'm going to go with that.

MushMal (approved):

PR #866, opened 2020-10-13
Handle separate error codes from connect()

grahn:
*Issue #, if available:*
#756 
*Description of changes:*
On Windows when calling connect() with a non-blocking socket, the
error code WSAEWOULDBLOCK is used to indicate the connection attempt
cannot be completed immediately. However the POSIX error code for this
case is EINPROGRESS.

This commit updates socketConnect() to handle both the WIN32 and POSIX
error codes mentioned above.

### DIFF
--- a/src/source/Ice/Network.c
+++ b/src/source/Ice/Network.c
@@ -291,7 +291,7 @@ STATUS socketConnect(PKvsIpAddress pPeerAddress, INT32 sockfd)
     }
 
     retVal = connect(sockfd, peerSockAddr, addrLen);
-    CHK_ERR(retVal >= 0 || getErrorCode() == EINPROGRESS, STATUS_SOCKET_CONNECT_FAILED, "connect() failed with errno %s",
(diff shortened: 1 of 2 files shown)


src/source/Ice/Network.c
```diff
@@ -291,7 +291,14 @@ STATUS socketConnect(PKvsIpAddress pPeerAddress, INT32 sockfd)
     }
 
     retVal = connect(sockfd, peerSockAddr, addrLen);
-    CHK_ERR(retVal >= 0 || getErrorCode() == EINPROGRESS, STATUS_SOCKET_CONNECT_FAILED, "connect() failed with errno %s",
+    /* Windows uses EWOULDBLOCK (WSAEWOULDBLOCK) to indicate connection attempt
+     * cannot be completed immediately, whereas POSIX uses EINPROGRESS. */
+#ifdef _WIN32
```
MushMal, line 296: Consider defining a KVS specific error code - say KVS_SOCKET_IN_PROGRESS in the header file which would then define at compile time to the specific platform status code. Here, it would simply check for the value without conditional compilation. This, making the main code more readable.

MushMal (commented):
Thanks a lot for taking care of this. I have a more of a readability comment than anything else

MushMal (approved):
Niiiice!

MushMal (approved):

Issue #861, opened 2020-10-09
core dump due to invalid mutex


staowyze:
Hi experts,

I got a couple of crash dumps on my devices, would like to have your eyes on it, here is the call stack
```
#0  0x76ff4f74 in pthread_mutex_lock () from /opt/nfs/lib/libpthread-0.9.33.2.so
#1  0x771c08e0 in defaultLockMutex (mutex=0)
    at /home/admin/workspace/amazon-kinesis-video-streams-webrtc-sdk-c-mirror/open-source/libkvsCommonLws/build/src/libkvsCommonLws-download/dependency/libkvspic/kvspic-src/src/utils/src/Mutex.c:202
#2  0x77149b40 in stepSignalingStateMachine (pSignalingClient=0x56bac8, status=0)
    at /home/admin/workspace/amazon-kinesis-video-streams-webrtc-sdk-c-mirror/src/source/Signaling/StateMachine.c:64
#3  0x7714285c in reconnectHandler (warning: GDB can't find the start of the function at 0x76ff36de.
```

Thank you!

staowyze:
@MushMal for visibility

staowyze:
a private tool is required to look at the full core dump, so i didn't upload it here

MushMal:
Do you have any more info? Debug logs leading to the crash? Is this the stock master application? Any scenario specifics?

staowyze:
Here are the log
```
[10-07 18:26:51][2020-09-07 10:26:50] [KVS]connectSignalingChannelLws(): operation returned status code: 0x0000000f
[10-07 18:27:00][2020-09-07 10:27:00] [KVS]refreshIceConfigurationCallback(): operation returned status code: 0x0000000f
[10-07 18:37:18][2020-09-07 10:37:18] [KVS]connectSignalingChannelLws(): operation returned status code: 0x0000000f
[10-07 18:37:34][2020-09-07 10:37:34] [KVS]connectSignalingChannelLws(): operation returned status code: 0x0000000f
[10-07 18:42:42][2020-09-07 10:42:42] [KVS]createSignalingSync(): operation returned status code: 0x0000000f
[10-07 18:47:20][2020-09-07 10:47:20] [KVS]connectSignalingChannelLws(): operation returned status code: 0x0000000f
[10-07 18:52:26][2020-09-07 10:52:26] [KVS]connectSignalingChannelLws(): operation returned status code: 0x0000000f
[...]

PR #871, opened 2020-10-15
Increment the shutdown timeout to be greater than LWS connection time…

MushMal:
…out and add a check for the shutdown flag

Issue #861 

*Description of changes:*


src/source/Signaling/LwsApiCalls.c
```diff
@@ -1292,12 +1292,17 @@ PVOID reconnectHandler(PVOID args)
...
     // Await for the listener to clear
     MUTEX_LOCK(pSignalingClient->listenerTracker.lock);
     MUTEX_UNLOCK(pSignalingClient->listenerTracker.lock);
 
-    // Indicate that we started
-    ATOMIC_STORE_BOOL(&pSignalingClient->reconnecterTracker.terminated, FALSE);
+    // Exit immediately if we are shutting down in case we are getting terminated while we were waiting for the
+    // listener thread to terminate. The shutdown flag would have been checked prior kicking off the reconnect
+    // thread but there is a slight chance of a race condition.
+    CHK(!ATOMIC_LOAD_BOOL(&pSignalingClient->shutdown), retStatus);
```
staowyze, line 1305: I am not confident this would fix the race condition mentioned in #861, at least for the usage in the sample code. Assuming create reconnectHandler thread takes super long time, the pSignalingClient is about to be freed due to refresh ICE error but not yet, so the pSignalingClient might be in a weird state, e.g. pSignalingClient is not NULL, but pSignalingClient->listenerTracker.lock is freed?

MushMal: You are correct. We have two options here

1) Try to set the terminated before creating the reconnect thread. This is actually pretty clean but need proper cleanup sequence in case the thread creation fails.
2) Use the thread trackers lock and the conditional variable to sequence the startup for the thread. This is actually more code than the first approach.

Will code up the first.

src/source/Signaling/LwsApiCalls.c
```diff
@@ -294,8 +294,15 @@ INT32 lwsWssCallbackRoutine(struct lws* wsi, enum lws_callback_reasons reason, P
             ATOMIC_STORE(&pSignalingClient->result, (SIZE_T) SERVICE_CALL_UNKNOWN);
 
             if (connected && !ATOMIC_LOAD_BOOL(&pSignalingClient->shutdown)) {
-                // Handle re-connection in a reconnect handler thread
-                CHK_STATUS(THREAD_CREATE(&pSignalingClient->reconnecterTracker.threadId, reconnectHandler, (PVOID) pSignalingClient));
+                // Handle re-connection in a reconnect handler thread. Set the terminated indicator before the thread
+                // creation and the thread itself will reset it. NOTE: Need to check for a failure and reset.
+                ATOMIC_STORE_BOOL(&pSignalingClient->reconnecterTracker.terminated, FALSE);
```
staowyze, line 299: this would definitely improve, but I am still concerned that the race could happen with little chance, as we cannot batch loading `shutdown` and clearing `terminated`in an atomic operation, so for magic reason, we loaded the `shutdown` flag, but have not cleared `terminated` yet, then freeSignaling starts and finishes.

MushMal: They can't be indeed be processed atomically. The main difficulty and the complexity of the code comes from the fact that there is no platform independent way to await on the thread handle like in Windows so we have to do this insane interlocking/atomics operations to guarantee ordering.

In this case, there is a chance indeed that the shutdown could be set but the terminated for the reconnect tracker is not. However. there will be the listener tracker thats currently running the event loop which would then block. Let me check if there is a chance for that to misbehave.

Regardless of that, these changes should go in and we should in parallel track if there is still an issue with the shutdown sequence on edge cases like reconnect

MushMal: Yeah, the freeSignaling attempts to tear down the listener first (which is driving the reconnect scenario that spawns the reconnect handler).

lherman-cs: In that case, how about if we remove "terminated" flag in the thread tracker, and have an atomic integer in the signaling client that contains the three states: signaling shutdown, reconnect thread terminated, and listener thread terminated? Since these are just booleans, we only need 3 bits to represent all states. 

This way, we'll be able to atomically load and store the three states in batch.

staowyze: what would happen to the ongoing reconnect handler when the listener tracker is teared down?
sorry for too many questions, just like to learn more about the multi-threading in the SDK.

MushMal: The atomicity is not really about two reads. It's more about multi-read-check-write. In either case, there is a chance for the race if you just consider this local example. 

A better solution perhaps is a semaphore similar to what we have in PIC/C Producer transitions. In this case, every time there is a source of liveness, we increment a semaphore (listener, ICE refresher, reconnecter). The shutdown would then flag and await the semaphore with a timeout (instead of the awaitForThreadTermination API sequence).

The semaphore would not remove the need for the existing primitives like the tracker as we need to do certain synchronization for listener/reconnect threading.

staowyze: oh, you are saying the listenerTracker would be blocked if the reconnect handler is about to start?

MushMal: > what would happen to the ongoing reconnect handler when the listener tracker is teared down?
> sorry for too many questions, just like to learn more about the multi-threading in the SDK.

The reconnect is getting kicked off from the event loop. The event loop is powered by the listener thread. The reconnect gets kicked off on error and on disconnect callback handling to ensure we have a way to reconnect. We can't do the reconnection from within the existing event loop/websocket so we need a new source of liveness.

The reconnect thread would then launch and block until the listener thread frees the lock on exit, meaning that the reconnect thread could then go and drive the state machinery and exit. Part of the state machinery state is to launch a new listener thread.

So, the net result is old listener -> reconnect if an error -> reconnect handler wait for old listener to exit and free the lock -> old listener exits and unlocks the lock -> reconnect thread drives the state machinery -> kicks off a new listener -> reconnect thread exits.

lherman-cs: > The atomicity is not really about two reads. It's more about multi-read-check-write. In either case, there is a chance for the race if you just consider this local example.

Right, I was thinking to use CAS for this, which allows us to do multi-read-check-write.

> A better solution perhaps is a semaphore similar to what we have in PIC/C Producer transitions. In this case, every time there is a source of liveness, we increment a semaphore (listener, ICE refresher, reconnecter). The shutdown would then flag and await the semaphore with a timeout (instead of the awaitForThreadTermination API sequence).

I was thinking about using a semaphore as well, but wouldn't it face the same problem as there's no way to update the shutdown flag and increment the semaphore at the same time?

MushMal: > oh, you are saying the listenerTracker would be blocked if the reconnect handler is about to start?

The listener thread is the one that is powering the libWebSocket threading and the event loop. The callback is primarily (though, not always) is happening from the thread that is actually powering the event loop in lwsCompleteSync

MushMal: > > The atomicity is not really about two reads. It's more about multi-read-check-write. In either case, there is a chance for the race if you just consider this local example.
> 
> Right, I was thinking to use CAS for this, which allows us to do multi-read-check-write.
> 
> > A better solution perhaps is a semaphore similar to what we have in PIC/C Producer transitions. In this case, every time there is a source of liveness, we increment a semaphore (listener, ICE refresher, reconnecter). The shutdown would then flag and await the semaphore with a timeout (instead of the awaitForThreadTermination API sequence).
> 
> I was thinking about using a semaphore as well, but wouldn't it face the same problem as there's no way to update the shutdown flag and increment the semaphore at the same time?

There are no atomic primitives for this type of CAS. You will need interlocking.

The blocking rw semaphore we have is the right primitive. Readers would increment the lock count, the writer (shutdown sequence) would block until the semaphore value is 0 and continue. All this time no new readers will be permitted. 

Still, there is chance for a timeout (due to the complexity in some edge cases when the socket dies) and we might end up with timing out to continue the tear down which in some cases could lead to a crash. This is still better than a hang

MushMal: I would still prefer to go with these changes and cut an enhancement work item for the semaphore solution

staowyze: alright, good to learn all the details, i think we all on the same page regarding the race we need to fix.
what would be the perfect solution? is this PR good enough? if so, i would encourage to get it merged asap, then i could try it out too.

MushMal: This PR should fix the issue with the teardown race for the issue we had identified earlier and the need to increase the shutdown timeout (most important part of the fix actually).

As mentioned, there is a theoretical chance of a race but that particular issue is handled at the higher level logic by which the shutdown tears down the listener first and by that time the reconnect would have a chance to react on the shutdown and exit. In this case, there is no case with the race as the shutdown thread would go and block on the termination flag for the reconnect.

However, this is still brittle and handles just this case and perhaps we have other cases there. I would prefer a better solution for the shutdown sequence based not on a shutdown flag (atomic) but rather on a semaphore. That should be a separate work item though

lherman-cs (approved):
The PR itself looks good overall. As discussed, the theoretical race condition can occur and will be a separate work item

MushMal:
https://github.com/awslabs/amazon-kinesis-video-streams-webrtc-sdk-c/issues/873 added a feature item to track work for serializing the shutdown sequence using a semaphore

Issue #870, opened 2020-10-14
[QUESTION] core dump in refreshIceConfigurationCallback


staowyze:
Hi experts,

Got a core dump as follows, v1.30 tag, any comments are appreciated!
```
Program terminated with signal SIGSEGV, Segmentation fault.
#0  0x77b79db8 in ?? ()
(gdb) set solib-search-path /opt/nfs/lib/:/opt/nfs/webrtc/
(gdb) bt
#0  0x77b79db8 in __lws_close_free_wsi (wsi=0xc2e570, reason=LWS_CLOSE_STATUS_NOSTATUS, caller=<optimized out>)
    at /home/admin/workspace/amazon-kinesis-video-streams-webrtc-sdk-c-mirror/open-source/libwebsockets/build/src/project_libwebsockets/lib/core-net/close.c:342
#1  0x77b8e04c in lws_client_connect_2 (wsi=0xc2e570)
    at /home/admin/workspace/amazon-kinesis-video-streams-webrtc-sdk-c-mirror/open-source/libwebsockets/build/src/project_libwebsockets/lib/roles/http/client/client-handshake.c:768
#2  0x77b83cfc in lws_header_table_attach (wsi=0xc2e570, autoservice=0)
    at /home/admin/workspace/amazon-kinesis-video-streams-webrtc-sdk-c-mirror/open-source/libwebsockets/build/src/project_libwebsockets/lib/roles/http/server/parsers.c:322
#3  0x77b85920 in rops_client_bind_h1 (wsi=0xc2e570, i=<optimized out>)
    at /home/admin/workspace/amazon-kinesis-video-streams-webrtc-sdk-c-mirror/open-source/libwebsockets/build/src/project_libwebsockets/lib/roles/h1/ops-h1.c:965
#4  rops_client_bind_h1 (wsi=0xc2e570, i=<optimized out>)
    at /home/admin/workspace/amazon-kinesis-video-streams-webrtc-sdk-c-mirror/open-source/libwebsockets/build/src/project_libwebsockets/lib/roles/h1/ops-h1.c:927
#5  0x77b8bd80 in lws_client_connect_via_info (i=0x74e34e38)
    at /home/admin/workspace/amazon-kinesis-video-streams-webrtc-sdk-c-mirror/open-source/libwebsockets/build/src/project_libwebsockets/lib/core-net/connect.c:257
[...]

PR #878, opened 2020-10-19
lws_client_connect_via_info API is not multi-thread reentrant. Simply…

MushMal:
… interlocking it will lead to live locks.

LibWebSockets assumes a single threaded execution. While we are connected to websocket we need to run ICE server info refresh which requires interlacing a few of the API calls with the callbacks.
This leaves us with two choices - either disconnect from the connected websocket when we do ICE refresh and reconnect it once that's done or to spin up two separate contexts. The first solution is more straight forward but at the expense of potentially losing a message (if it falls out of TTL) or having high latency when viewers are connecting. The dual context solution is bad because it doesn't scale down to small footprint devices. So we decided to use the same context to drive synchronous in-flight requests.
The interlocking is not a trivial thing to have as it leads to live-locks in case where the service thread is blocked on a socket and has the lock and there is no way to wake it up as the wake up API itself would require a lock. The problem is more exacerbated by the fact that we could have multiple entities in-flight - connection listener, ICE refresh and the reconnect thread and they need to be carefully coordinated and interlocked during the negative cases and teardown.

The solution here is to store the WSI when connecting in an interlocked fashion which uses a separate serialization lock and not the service lock to avoid live locking. While acquiring the lock, we try to ackquire it and if we fail, that means another thread is locked (for example the listener thread servicing a websocket while we are executing ICE refresh). As the holder can be blocked on the socket, we need to "wake it up" without having the lock held. Apparently, there are only TWO APIs that can do that. From https://libwebsockets.org/lws-api-doc-master/html/md_READMEs_README_8coding.html

<
Therefore the two apis mentioned above that may be used from another thread are

For LWS using the default poll() event loop, lws_callback_on_writable()
For LWS using libuv/libev/libevent event loop, lws_cancel_service()
>

Issue #870 

*Description of changes:*

### DIFF
--- a/CMake/Dependencies/libkvsCommonLws-CMakeLists.txt
+++ b/CMake/Dependencies/libkvsCommonLws-CMakeLists.txt
@@ -6,7 +6,7 @@ include(ExternalProject)
 
 ExternalProject_Add(libkvsCommonLws-download
     GIT_REPOSITORY    https://github.com/awslabs/amazon-kinesis-video-streams-producer-c.git
-    GIT_TAG           3f437b936adf1899533a8f82c9f374e7041004eb
+    GIT_TAG           15f124e5a5561679cf001bb36c2c758073740681
     PREFIX            ${CMAKE_CURRENT_BINARY_DIR}/build
     CMAKE_ARGS        
       -DCMAKE_INSTALL_PREFIX=${OPEN_SRC_INSTALL_PREFIX}
--- a/src/source/Signaling/LwsApiCalls.c
+++ b/src/source/Signaling/LwsApiCalls.c
@@ -550,8 +550,30 @@ STATUS lwsCompleteSync(PLwsCallInfo pCallInfo)
     serializerLocked = TRUE;
 
     // Ensure we are not running another https protocol
+    // The WSIs for all of the protocols are set and cleared in this function only.
+    // The HTTPS is serialized via the state machine lock and we should not encounter
+    // another https protocol in flight. The only case is when we have an http request
+    // and a wss is in progress. This is the case when we have a current websocket listener
+    // and need to perform an https call due to ICE server config refresh for example.
+    // If we have an ongoing wss operations, we can't call lws_client_connect_via_info API
+    // due to threading model of LWS. WHat we need to do is to wake up the potentially blocked
+    // ongoing wss handler for it to release the service lock which it holds while calling lws_service()
+    // API so we can grab the lock in order to perform the lws_client_connect_via_info API call.
+    // The need to wake up the wss handler (if any) to compete for the lock is the reason for this
+    // loop. In order to avoid pegging of the CPU while the contention for the lock happes,
+    // we are setting an atomic and releasing it to trigger a timed wait when the lws_service call
+    // awakes to make sure we are not going to starve this thread.
+
+    // NOTE: The THREAD_SLEEP calls in this routine are not designed to adjust
+    // the execution timing/race conditions but to eliminate a busy wait in a spin-lock
+    // type scenario for resource contention.
+
+    // We should have HTTPS protocol serialized at the state machine level
     CHK_ERR(pCallInfo->pSignalingClient->currentWsi[PROTOCOL_INDEX_HTTPS] == NULL, STATUS_INVALID_OPERATION,
             "HTTPS requests should be processed sequentially.");
+
+    // Indicate that we are trying to acquire the lock
+    ATOMIC_STORE_BOOL(&pCallInfo->pSignalingClient->serviceLockContention, TRUE);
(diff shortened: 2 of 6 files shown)


src/source/Signaling/LwsApiCalls.c
```diff
@@ -544,24 +544,56 @@ STATUS lwsCompleteSync(PLwsCallInfo pCallInfo)
...
-    locked = FALSE;
+    // Attempt to iterate and acquire the locks
+    // NOTE: The https protocol should be called sequentially only
+    MUTEX_LOCK(pCallInfo->pSignalingClient->lwsSerializerLock);
+    serializerLocked = TRUE;
+
+    // Ensure we are not running another https protocol
+    CHK_ERR(pCallInfo->pSignalingClient->currentWsi[PROTOCOL_INDEX_HTTPS] == NULL, STATUS_INVALID_OPERATION,
+            "HTTPS requests should be processed sequentially.");
+    while (iterate && pCallInfo->pSignalingClient->currentWsi[PROTOCOL_INDEX_WSS] != NULL) {
```
lherman-cs, line 555: Would you mind to add some comments when the currentWsi for WSS is nullified? From my understanding, when we failed to acquire the service lock, we would wake up the thread that will eventually nullify the currentWsi for WSS. Is this right?

How do we make sure that the awaken thread will finish in timely manner?

src/source/Signaling/LwsApiCalls.c
```diff
@@ -1718,8 +1751,11 @@ STATUS terminateConnectionWithStatus(PSignalingClient pSignalingClient, SERVICE_
         ATOMIC_STORE_BOOL(&pSignalingClient->pOngoingCallInfo->cancelService, TRUE);
     }
 
-    // Wake up the service event loop
-    CHK_STATUS(wakeLwsServiceEventLoop(pSignalingClient));
+    // Wake up the service event loop for all of the protocols
+    for (i = 0; i < LWS_PROTOCOL_COUNT; i++) {
```
lherman-cs, line 1755: Since we removed the locks that are in `wakeLwsServiceEventLoop`, shouldn't we need to acquire a lock here?

src/source/Signaling/LwsApiCalls.c
MushMal: I added some wording in the comment. Basically, there is a lock contention. If there is an ongoing WSS operation indicated by a non-null wss pointer, and only then, we need to wake up the service thread handler which would release the lock that we could acquire. I realized that this will lead to a temporary busy loops while the contention of the lock is happening so I added an atomic to indicate to the awakening service loop to wait.

If the wss pointer is null then there is no need to acquire the lock.

src/source/Signaling/LwsApiCalls.c
MushMal: Good question. The wake up calls lws_callback_on_writable which is pretty much the only API that is re-entrant. The question then is whether the earlier check for if (pSignalingClient->currentWsi[protocolIndex] != NULL) stays valid. I traced the calls and it seems that the higher level state machine locks prevent this case.

staowyze:
is this diff ready for landing?

lherman-cs:
The CI says DataChannelFunctionalityTest.createDataChannel_Disconnected failed with a segfault

MushMal:
CI is clear. Need CR to go through so we can push it

CMake/Dependencies/libkvsCommonLws-CMakeLists.txt
```diff
@@ -6,7 +6,7 @@ include(ExternalProject)
 
 ExternalProject_Add(libkvsCommonLws-download
     GIT_REPOSITORY    https://github.com/awslabs/amazon-kinesis-video-streams-producer-c.git
-    GIT_TAG           3f437b936adf1899533a8f82c9f374e7041004eb
+    GIT_TAG           15f124e5a5561679cf001bb36c2c758073740681
```
staowyze, line 9: anything in kvsCommonLws could affect the bug?

MushMal: We implemented blocking shutdown in the timer queue in PIC which was pushed into common and now in here. This is needed for proper termination sequence for the PeerConnection. None directly involved in the presumed fix for the crash you are seeing

src/source/Signaling/LwsApiCalls.c
MushMal: There could be two threads at one time driving the service loop - one that's exiting and the other that's either running reconnect or the https APIs. The changes I made will remove the potentially busy loop in case of thread starvation by adding THREAD_SLEEP. 

THREAD_SLEEP is a non-blocking sleep implemented by the OS scheduled. What this means, is that the thread is relinquishing it's quanta, allowing another competing thread to grab the lock.

hassanctech (approved):
lgtm!

lherman-cs (approved):
LGTM

PR #881, opened 2020-10-19
Enable compiler warning on linux

disa6302:
*Issue #, if available:*

*Description of changes:*
Enable compiler warning on Linux. This was removed at some point of time, but, we should have this enabled and reported as errors to catch any serious warnings. 


disa6302:
The CI fails because of warnings in PIC. Will get back to this PR after a few days.

PR #882, opened 2020-10-20
Generate pc file for libsrtp

### LINKED ISSUE #538: Missing libsrtp2 pc file [BUG]
**Describe the bug**
Latest master does not create a package config file for libsrtp2 but includes it as a library

**SDK version number**
commit: acc94940a715ac5dec8f20cf5e51b65607e09e43

**Open source building**

I am building the repository and creating a deb package for installation in our build system. When I install the libraries, dependencies, and created pc files, I am missing the srtp2 pc file. When I manually add in the pc file from previous older versions of webrtc, it works fine. This pc file used to be included in older versions of webrtc.

**To Reproduce**
Build the repository, look in `open-source/lib/pkgconfig`. There are config files for all the other created libraries except `libsrtp2`

**Expected behavior**
Need a `libsrtp2.pc` in `open-source/lib/pkgconfig`:

```
prefix=/usr
exec_prefix=${prefix}
libdir=${exec_prefix}/lib
includedir=${prefix}/include

Name: libsrtp2
Version: 2.2.0
Description: Library for SRTP (Secure Realtime Transport Protocol)

Libs: -L${libdir} -lsrtp2 -lcrypto -ldl -pthread 
Cflags: -I${includedir}
```

disa6302:
*Issue #, if available:*
 #538 

*Description of changes:*

This PR generates a PC file for unix and apple platforms by running `configure`.
There is a known issue in libsrtp2.3.0 which leads to `./configure` to fail when cross compiling for ARM platforms with the error: 

```checking if OPENSSL_cleanse is broken... configure: error: in .....
configure: error: cannot run test program while cross compiling
```

SRTP Issue reference: `https://github.com/cisco/libsrtp/issues/479`

As a workaround, this PR checks if `BUILD_LIBSRTP_HOST_PLATFORM == BUILD_LIBSRTP_DESTINATION_PLATFORM` to ensure `./configure` is run only in non cross compile mode. This check can be removed once we can update libsrtp with the fix in the libsrtp repo. There is no timeline on that yet. 

We can also move to 2.2.0 but there might be some fixes in 2.3.0 which we might lose, which is not ideal.

Resolves  #538 

### DIFF
--- a/CMake/Dependencies/libsrtp-CMakeLists.txt
+++ b/CMake/Dependencies/libsrtp-CMakeLists.txt
@@ -2,17 +2,47 @@ cmake_minimum_required(VERSION 2.8)
 
 project(libsrtp-download NONE)
 
+
+SET(CONFIGURE_COMMAND "")
+
+# There is known bug in libsrtp where cross compiling using configure on ARM fails. Do not
+# enable this option if cross compilng on ARM
+# Check https://github.com/cisco/libsrtp/pull/496
+if(BUILD_LIBSRTP_DESTINATION_PLATFORM STREQUAL BUILD_LIBSRTP_HOST_PLATFORM)
+  if(UNIX OR APPLE)
+    if(USE_OPENSSL)
+      SET(CONFIGURE_COMMAND ${CMAKE_CURRENT_BINARY_DIR}/build/src/project_libsrtp/configure "CFLAGS=${CMAKE_C_FLAGS}" --prefix=${OPEN_SRC_INSTALL_PREFIX}  --enable-openssl --with-openssl-dir=${OPENSSL_DIR})
+    else()
+      SET(CONFIGURE_COMMAND ${CMAKE_CURRENT_BINARY_DIR}/build/src/project_libsrtp/configure "CFLAGS=${CMAKE_C_FLAGS}" --prefix=${OPEN_SRC_INSTALL_PREFIX})
+    endif()
+
+    if (DEFINED BUILD_LIBSRTP_DESTINATION_PLATFORM AND NOT BUILD_LIBSRTP_DESTINATION_PLATFORM STREQUAL OFF)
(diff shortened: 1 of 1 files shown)


disa6302:
Having this PR open to track windows build. Will push from draft to ready to review once windows build passes.

.travis.yml
```diff
@@ -177,7 +177,7 @@ matrix:
       before_script:
         - export CC=arm-linux-gnueabi-gcc CXX=arm-linux-gnueabi-g++
         - mkdir build && cd build
-        - cmake .. -DBUILD_OPENSSL=TRUE -DBUILD_OPENSSL_PLATFORM=linux-generic32 -DBUILD_LIBSRTP_HOST_PLATFORM=x86_64-unknown-linux-gnu -DBUILD_LIBSRTP_DESTINATION_PLATFORM=arm-unknown-linux-uclibcgnueabi
+        - cmake .. -DGENERATE_LIBSRTP_PC=OFF -DBUILD_OPENSSL=TRUE -DBUILD_OPENSSL_PLATFORM=linux-generic32 -DBUILD_LIBSRTP_HOST_PLATFORM=x86_64-unknown-linux-gnu -DBUILD_LIBSRTP_DESTINATION_PLATFORM=arm-unknown-linux-uclibcgnueabi
```
lherman-cs, line 180: Not sure if we want to add yet another flag to our CMake for this use case. Should we infer from `BUILD_LIBSRTP_HOST_PLATFORM` != `BUILD_LIBSRTP_DESTINATION_PLATFORM`. I think this means the same thing, what do you think?

lherman-cs (commented):
Minor comment

.travis.yml
disa6302: Oh! That is a good idea actually!
I have made the change. Lets see if the CI passes.

disa6302: It worked!

lherman-cs (approved):

Issue #890, opened 2020-10-22
[BUG] Handle trickle ice attribute appropriately


disa6302:
**Describe the bug**
On receiving `ice-options` in the offer, we perform an exact match to `trickle` to enable the trickle ice. However, the `ice-options` can contain multiple attribute values and we should not perform an exact match but check if `trickle` value exists in the attribute value list.

**SDK version number**
Top of master

**To Reproduce**
1. Generate an offer with values: `ice-option`: `trickle renomination`

**Expected behavior**
Trickle gets enabled. But, trickle does not get enabled with current SDK

Created from #889

disa6302:
@yuma-m , 

The change is in. Let me know if this fixes the issue for you. Closing this

yuma-m:
Hi @disa6302 , thank you for fixing it quickly.
I confirmed `trickle renomination` is properly handled as `canTrickleIce = TRUE`.

PR #891, opened 2020-10-22
Fix parsing of ice options in SDP

disa6302:
*Issue #, if available:*
#890 

*Description of changes:*
`ice-options` can contain multiple attribute values along with trickle. The SDK parsed the option with the assumption that `ice-options` would contain only a single value leading to performing an exact match to enable trickle ice based on the value. This PR fixes the assumption by looking for `trickle` in the list of values instead of an exact match. An unit test is also added to test this case.

### DIFF
--- a/tst/SdpApiTest.cpp
+++ b/tst/SdpApiTest.cpp
@@ -848,6 +848,54 @@ a=fmtp:102 strange
     });
 }
 
+TEST_F(SdpApiTest, populateSingleMediaSection_TestMultipleIceOptions)
+{
+    CHAR remoteSessionDescription[] = R"(v=0
+o=- 7732334361409071710 2 IN IP4 127.0.0.1
+s=-
+t=0 0
+a=group:BUNDLE 0
+a=msid-semantic: WMS
+m=video 16485 UDP/TLS/RTP/SAVPF 96 102
+c=IN IP4 205.251.233.176
+a=rtcp:9 IN IP4 0.0.0.0
+a=ice-ufrag:9YRc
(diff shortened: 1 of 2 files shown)


tst/SdpApiTest.cpp
```diff
@@ -848,6 +848,56 @@ a=fmtp:102 strange
...
+a=recvonly
+a=rtcp-mux
+a=rtcp-rsize
+a=rtpmap:96 VP8/90000
+a=rtpmap:102 H264/90000
+)";
+
+    assertLFAndCRLF(remoteSessionDescription, ARRAY_SIZE(remoteSessionDescription) - 1, [](PCHAR sdp) {
+        PRtcPeerConnection pRtcPeerConnection = NULL;
+        PRtcRtpTransceiver pRtcRtpTransceiver = NULL;
```
hassanctech, line 877: This is not used anywhere, can be deleted.

tst/SdpApiTest.cpp
```diff
@@ -848,6 +848,56 @@ a=fmtp:102 strange
...
+        RtcConfiguration rtcConfiguration;
+        RtcMediaStreamTrack rtcMediaStreamTrack;
+        RtcRtpTransceiverInit rtcRtpTransceiverInit;
```
hassanctech, line 880: This is not used anywhere, can be deleted.

MushMal (approved):

tst/SdpApiTest.cpp
disa6302: Ouch! Fixed.

PR #893, opened 2020-10-23
Update mbedTLS to official pre-released version (discarded)

lherman-cs:
The DTLS-SRTP extension PR from upstream has been revisioned. Its revision has received an approval and gone through more unit tests. As of now, the revision has been merged to master.

Changes:
  * Update mbedTLS version to the official pre-release
  * Update CMakeLists to enable custom mbedTLS config
  * Optimize mbedTLS config to be bare minimum
  * Update codebase to reflect the new changes from mbedTLS


configs/config_mbedtls.h
```diff
@@ -0,0 +1,166 @@
...
+
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+/**
+ * Minimal Suite-B configuration
+ *
+ * Inherited from https://github.com/ARMmbed/mbedtls/blob/1e14827bebc8480786498f007304bd852ef953c6/configs/config-suite-b.h.
+ * Some of the configurations are disabled due to incompatibility with what we need. The configs that are disabled are still
```
MushMal, line 18: So the only change from the "standard" config is the commented lines with notes?

configs/config_mbedtls.h
```diff
@@ -0,0 +1,166 @@
...
+#define MBEDTLS_BASE64_C
+#define MBEDTLS_CERTS_C
+#define MBEDTLS_PEM_PARSE_C
+
+/* Save RAM at the expense of ROM */
+#define MBEDTLS_AES_ROM_TABLES
+
+/* Save RAM by adjusting to our exact needs
+ *
+ * Note: disabled because we want RSA as well, which is much bigger that ECP.
```
MushMal, line 69: Not sure I understood the comment. We are disabling what and why?

src/source/Crypto/Dtls_mbedtls.c
```diff
@@ -3,8 +3,9 @@
 
 /**  https://tools.ietf.org/html/rfc5764#section-4.1.2 */
 mbedtls_ssl_srtp_profile DTLS_SRTP_SUPPORTED_PROFILES[] = {
-    MBEDTLS_SRTP_AES128_CM_HMAC_SHA1_80,
-    MBEDTLS_SRTP_AES128_CM_HMAC_SHA1_32,
+    MBEDTLS_TLS_SRTP_AES128_CM_HMAC_SHA1_80,
+    MBEDTLS_TLS_SRTP_AES128_CM_HMAC_SHA1_32,
+    MBEDTLS_TLS_SRTP_UNSET,
```
MushMal, line 8: for my info - what exactly the unset profile means here?

MushMal: oh, I see, they must have changed the API to have a terminator value?

MushMal (commented):
mostly questions

src/source/Crypto/Dtls_mbedtls.c
lherman-cs: Yes, they updated the API, it now uses a terminator value

configs/config_mbedtls.h
lherman-cs: The `MBEDTLS_ECP_MAX_BITS` and `MBEDTLS_MPI_MAX_SIZE` flags are set in the original suite-b config, https://github.com/ARMmbed/mbedtls/blob/1e14827bebc8480786498f007304bd852ef953c6/configs/config-suite-b.h. This is an optimization to reduce the big number max size, but we can't reduce the max size since we're using RSA which requires a much bigger amount of bits. So, we're essentially disabling these optimizations.

configs/config_mbedtls.h
lherman-cs: Only the ones that are commented out and have notes. Apart from that, there's another section called "mbedTLS configuration extensions for KVS", which contains extra flags for our needs.

MushMal (approved):

MushMal (approved):

lherman-cs:
https://github.com/awslabs/amazon-kinesis-video-streams-webrtc-sdk-c/pull/929 obsoletes this PR

Issue #914, opened 2020-11-02
[BUG] Possible memory leak when viewer has no videos


Nomidia:
**Logging**
[11-02-11-1080P-COM4.log](https://github.com/awslabs/amazon-kinesis-video-streams-webrtc-sdk-c/files/5472601/11-02-11-1080P-COM4.log)

**Describe the bug**
In my chrome brower, I have to close and reopen the page every time I start the viewer. If I don't close the page and start viewer, there is no videos, ss shown in Figure 1. And when the program exits it will print a memory leak. The memory can be 12060, 24120, 36180, integer multiples of 12060, like a size of a structure.

**Screenshots**
Figure 1
![image](https://user-images.githubusercontent.com/20633988/97826772-79c21c00-1cfd-11eb-9680-5877f6db32c4.png)

hassanctech:
This is the same issue as:  https://github.com/awslabs/amazon-kinesis-video-streams-webrtc-sdk-c/issues/904

We are investigating this, in the mean time if you use the JS sample page:  https://awslabs.github.io/amazon-kinesis-video-streams-webrtc-sdk-js/examples/index.html

You should be able to play/stop as many times as you want without issue.

MushMal:
I am not sure I understand the scenario. Here is what I am trying.

* Using the latest master sample.
* Using Google Chrome in the browser hosting https://awslabs.github.io/amazon-kinesis-video-streams-webrtc-sdk-js/examples/index.html
* Start the master app (under the debugger or without)
* Start the viewer (works fine)
* Stop the viewer by clicking stop button and start it again clicking start (works fine)
* Refresh the page and click start (works fine)

How can I reproduce this issue?

The logs you've attached have an odd STATUS_NULL_ARG in the signaling message callback. This shouldn't be the case with the stock master sample.

hassanctech:
Right there's no issue in the JS page, it's in the KVS console, if you go to the signaling channel and try to play it'll work the first time, if you pause and try to play again you'll see this behavior, specially I see it with the gstreamer master sample, not with the non-gstreamer.

Nomidia:
Correct, I am using kvsWebRTCClientMaster and view from kvs.

![image](https://user-images.githubusercontent.com/20633988/97944864-cc184100-1dc0-11eb-9854-680d8ad46278.png)

Nomidia:
Reproduce steps
- Using kvsWebRTCClientMaster 
[...]

PR #916, opened 2020-11-03
Race condition fixes in the sample on teardown and on dtls initializa…

MushMal:
…tion in the socket connection.

Issue #914 

While investigating #914 I realized that the code had race conditions. What's really bad is that one of the issues is not even in the sample but in the SocketConnection.c and what's worse is that it's been suppressed erroneously. We really ought to be careful with the TSUN suppressions.

Running the scenario described in #914 I can't reproduce the memory leak. Perhaps not doing the right scenario?


src/source/Ice/SocketConnection.c
```diff
@@ -132,8 +132,13 @@ STATUS socketConnectionInitSecureConnection(PSocketConnection pSocketConnection,
     ENTERS();
     TlsSessionCallbacks callbacks;
     STATUS retStatus = STATUS_SUCCESS;
+    BOOL locked = FALSE;
 
     CHK(pSocketConnection != NULL, STATUS_NULL_ARG);
+
+    MUTEX_LOCK(pSocketConnection->lock);
```
lherman-cs, line 139: From my understanding, the race occurs when we're accessing `pSocketConnection->localSocket`. I think the actual race happens in the outbound packet callback:

https://github.com/awslabs/amazon-kinesis-video-streams-webrtc-sdk-c/blob/73a6ec9ce5add1aab02bbda4cc31750739cffe7f/src/source/Ice/SocketConnection.c#L89-L100

`socketSendDataWithRetry` expects the caller to hold the socket connection lock.

MushMal: The particular TSAN issue was flagged detecting a race in connect and recvfrom API calls. We need to look closely on how we can better protect the socketConnectionIsConnected API against the ongoing socket operations. It's not trivial with the existing code form what I see.

lherman-cs (approved):

disa6302 (approved):

Issue #569, opened 2020-06-26
[BUG] Data race in generateSignatureDateTime 


lherman-cs:
**Describe the bug**
After https://github.com/awslabs/amazon-kinesis-video-streams-webrtc-sdk-c/pull/559, `generateJSONSafeString` became faster and introduced a data race in `generateSignatureDateTime`, very likely due to a timing change. Currently, the data race is being suppressed here, https://github.com/awslabs/amazon-kinesis-video-streams-webrtc-sdk-c/commit/58c77c45839547db5b780366b09ad18172146594.

**SDK version number**
bdb39e22f04572cd9ad4bc1696a017968dfa6e51

**Open source building**
N/A

**To Reproduce**
The data race seems to be intermittent. From my observation, slow network connection speed increases the chance to get the data race.

Here's the exact Travis job that caught this data race: https://travis-ci.com/github/awslabs/amazon-kinesis-video-streams-webrtc-sdk-c/jobs/352369910.

**Additional context**
Stacktrace:

```
[ RUN      ] SignalingApiFunctionalityTest.invalidChannelInfoInput

[2020/06/24 15:45:19:7855] E: Unable to load SSL Client certs file from /usr/test -- client ssl isn't going to work

2020-06-24 15:45:19 INFO    SetUp(): 

Setting up test: SignalingApiFunctionalityTest

2020-06-24 15:45:19 INFO    createSignalingClientSync(): Creating Signaling Client Sync

2020-06-24 15:45:19 ERROR   signAwsRequestInfo(): operation returned status code: 0x00000002

2020-06-24 15:45:19 ERROR   signAwsRequestInfo(): operation returned status code: 0x00000002

2020-06-24 15:45:19 ERROR   signAwsRequestInfo(): operation returned status code: 0x00000002

2020-06-24 15:45:20 ERROR   signAwsRequestInfo(): operation returned status code: 0x00000002

2020-06-24 15:45:22 ERROR   signAwsRequestInfo(): operation returned status code: 0x00000002

2020-06-24 15:45:25 ERROR   signAwsRequestInfo(): operation returned status code: 0x00000002

2020-06-24 15:45:31 ERROR   signAwsRequestInfo(): operation returned status code: 0x00000002

2020-06-24 15:45:31 ERROR   createSignalingSync(): operation returned status code: 0x0000000f

2020-06-24 15:45:31 INFO    freeSignalingClient(): Freeing Signaling Client

[...]

PR #930, opened 2020-11-08
Removing TSAN suppression as it's not needed.

MushMal:
Issue #569 

I don't think we should suppress this issue. Local run can't reproduce this issue. Trying to get a Travis CI run clean and if it fails, then fix it.


lherman-cs (approved):
Oh interesting, it used to cause race conditions. Since this PR passed the CI, LGTM!

Issue #931, opened 2020-11-10
[BUG] Memory leak


Nomidia:
**Logging**
[normal.log](https://github.com/awslabs/amazon-kinesis-video-streams-webrtc-sdk-c/files/5514297/normal.log)
[abnomal.log](https://github.com/awslabs/amazon-kinesis-video-streams-webrtc-sdk-c/files/5514298/abnomal.log)

**Describe the bug**
Quit the program after connectig several times, it shows memory leak. The value is integer multiple of 12060. I add printf in **instrumentedAllocatorsMemCalloc** and **instrumentedAllocatorsMemFree**. In abnormal.txt, it allocated 12060 twice but released only once.

**Screenshots**
![image](https://user-images.githubusercontent.com/20633988/98618252-b6b09300-233b-11eb-9ac9-30475b154317.png)

Nomidia:
[11-10-10-1080P-COM4.log](https://github.com/awslabs/amazon-kinesis-video-streams-webrtc-sdk-c/files/5514492/11-10-10-1080P-COM4.log)

![image](https://user-images.githubusercontent.com/20633988/98621182-ab606600-2341-11eb-9649-863d8002f1a9.png)


![image](https://user-images.githubusercontent.com/20633988/98621192-aef3ed00-2341-11eb-9224-e73ba32f8123.png)

disa6302:
@Nomidia , 

You are right. That is the location where memory leak occurs. Can you try the PR to see if the issue gets fixed?

MushMal:
Fixed. Closing

PR #933, opened 2020-11-10
Fix memory leak

disa6302:
*Issue #, if available:*
#931 
*Description of changes:*
We fail to free pReceivedSignalingMessageCopy after allocating. This PR fixes the problem.

### DIFF
--- a/samples/Common.c
+++ b/samples/Common.c
@@ -1102,7 +1102,7 @@ STATUS signalingMessageReceived(UINT64 customData, PReceivedSignalingMessage pRe
                 }
 
                 pReceivedSignalingMessageCopy = MEMCALLOC(1, SIZEOF(ReceivedSignalingMessage));
-                *pReceivedSignalingMessageCopy = *pReceivedSignalingMessage;
+                MEMCPY(pReceivedSignalingMessageCopy, pReceivedSignalingMessage, SIZEOF(ReceivedSignalingMessage));
 
                 CHK_STATUS(stackQueueEnqueue(pPendingMessageQueue, (UINT64) pReceivedSignalingMessageCopy));
             } else {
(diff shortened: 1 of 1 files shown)


samples/Common.c
```diff
@@ -1102,7 +1102,7 @@ STATUS signalingMessageReceived(UINT64 customData, PReceivedSignalingMessage pRe
                 }
 
                 pReceivedSignalingMessageCopy = MEMCALLOC(1, SIZEOF(ReceivedSignalingMessage));
-                *pReceivedSignalingMessageCopy = *pReceivedSignalingMessage;
+                MEMCPY(pReceivedSignalingMessageCopy, pReceivedSignalingMessage, SIZEOF(ReceivedSignalingMessage));
```
MushMal, line 1105: This is the same but a little slower.

samples/Common.c
```diff
@@ -1116,7 +1116,7 @@ STATUS signalingMessageReceived(UINT64 customData, PReceivedSignalingMessage pRe
     }
 
 CleanUp:
-
+	SAFE_MEMFREE(pReceivedSignalingMessageCopy);
```
MushMal, line 1119: This will cause an access violation as you are freeing a piece of memory that you've added to the queue to be processed later. The fix should be somewhat different.

Nomidia: https://github.com/awslabs/amazon-kinesis-video-streams-webrtc-sdk-c/blob/27b95a928c05810f901949fa02364acdb17b16e3/samples/Common.c#L1096-L1102

In fact, it ignores the else condition, the _pPendingMessageQueue_ will be nill, so the _pReceivedSignalingMessageCopy_ will not be put into the queue.

MushMal: Will be submitting a fix asap

PR #934, opened 2020-11-10
Samples memory leak fix

MushMal:
Issue #931 

*Description of changes:*

### DIFF
--- a/samples/Common.c
+++ b/samples/Common.c
@@ -996,16 +996,18 @@ STATUS submitPendingIceCandidate(PStackQueue pPendingMessageQueue, PSampleStream
             hashValue = 0;
             CHK_STATUS(stackQueueDequeue(pPendingMessageQueue, &hashValue));
             pReceivedSignalingMessage = (PReceivedSignalingMessage) hashValue;
+            CHK(pReceivedSignalingMessage != NULL, STATUS_INTERNAL_ERROR);
             if (pReceivedSignalingMessage->signalingMessage.messageType == SIGNALING_MESSAGE_TYPE_ICE_CANDIDATE) {
                 CHK_STATUS(handleRemoteCandidate(pSampleStreamingSession, &pReceivedSignalingMessage->signalingMessage));
             }
-            MEMFREE(pReceivedSignalingMessage);
+            SAFE_MEMFREE(pReceivedSignalingMessage);
         }
     } while (!noPendingSignalingMessageForClient);
     CHK_STATUS(stackQueueFree(pPendingMessageQueue));
 
 CleanUp:
 
+    SAFE_MEMFREE(pReceivedSignalingMessage);
     CHK_LOG_ERR(retStatus);
     return retStatus;
 }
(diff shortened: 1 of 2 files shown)


disa6302:
Now the patch makes sense!

samples/Common.c
```diff
@@ -996,16 +996,18 @@ STATUS submitPendingIceCandidate(PStackQueue pPendingMessageQueue, PSampleStream
             hashValue = 0;
             CHK_STATUS(stackQueueDequeue(pPendingMessageQueue, &hashValue));
             pReceivedSignalingMessage = (PReceivedSignalingMessage) hashValue;
+            CHK(pReceivedSignalingMessage != NULL, STATUS_INTERNAL_ERROR);
             if (pReceivedSignalingMessage->signalingMessage.messageType == SIGNALING_MESSAGE_TYPE_ICE_CANDIDATE) {
                 CHK_STATUS(handleRemoteCandidate(pSampleStreamingSession, &pReceivedSignalingMessage->signalingMessage));
             }
-            MEMFREE(pReceivedSignalingMessage);
+            SAFE_MEMFREE(pReceivedSignalingMessage);
```
disa6302, line 1003: Question: Since `pReceivedSignalingMessage` is not allocated on heap, is is not enough to just set this to `NULL`? 

Besides, using `free()` on a memory not allocated using malloc() or calloc() might generate errors in some OSs.

tst/SignalingApiFunctionalityTest.cpp
```diff
@@ -1445,8 +1445,8 @@ TEST_F(SignalingApiFunctionalityTest, iceRefreshEmulationWithFaultInjectionError
     BYTE firstByte = pSignalingClient->pAwsCredentials->secretKey[0];
     pSignalingClient->pAwsCredentials->secretKey[0] = 'A';
 
-    // Wait for ICE refresh while connected
-    THREAD_SLEEP(15 * HUNDREDS_OF_NANOS_IN_A_SECOND);
+    // Wait for ICE refresh while connected and terminate with an error that will be called via error callback
+    THREAD_SLEEP(30 * HUNDREDS_OF_NANOS_IN_A_SECOND);
```
disa6302, line 1449: Any reason for this change?

samples/Common.c
hassanctech: Right hashValue is on the stack, we should not be calling free on this pointer.  That is left from when I fixed a bug earlier when we were not using a stack value.

Also

`pReceivedSignalingMessage = (PReceivedSignalingMessage) hashValue;`

this will be a null ptr if hashValue is 0.  Is that the desired behavior, it seems like hashValue is the value of the data item stored in start of the queue.  Is 0 a valid value here?

MushMal: The pReceivedSignalingMessage is allocated dynamically in the receive message handler. We shouldn't really have 0 in the hashValue and it will evaluate to NULL. If, however, it does, we bail out.

tst/SignalingApiFunctionalityTest.cpp
MushMal: It seems that in some of our runs the timing is not enough to kick off the ice refresh (in 10 seconds) and also ensure it fails after it retries numerous times.

disa6302 (approved):

samples/Common.c
hassanctech: I see ok it's allocated here:

https://github.com/awslabs/amazon-kinesis-video-streams-webrtc-sdk-c/blob/master/samples/Common.c#L1104-L1107

If pReceivedSignalingMessageCopy was NULL after the alloc then the next line would deref a null ptr, perhaps right after the alloc we should add a check that it wasn't null.  That way we can be sure there will never be a 0 value in there and we don't need to do the null check later (i.e. put it at item insertion to fail fast instead of deletion).

hassanctech (approved):

samples/Common.c
MushMal: Thought about it too but in this case, perhaps the right thing is to let it crash. I generally recommend having proper checks for the large or primary structure allocation/initialization at the start of the application. Auxiliary allocations are OK to validate within the SDK but in the application itself there is not much you one can do. Letting it fail might be the right thing to do

Issue #902, opened 2020-10-29
[BUG] SDP setPayloadTypesForOffer may select an incompatible H264 protocol from the SDP offer


andyhoffy:
**Logging**

Sample offer/answer from testing that reproduces the issue:

[offer.log](https://github.com/awslabs/amazon-kinesis-video-streams-webrtc-sdk-c/files/5459733/offer.log)
[answer.log](https://github.com/awslabs/amazon-kinesis-video-streams-webrtc-sdk-c/files/5459737/answer.log)

**Describe the bug**

It appears that the SDP answer formulation process does not discriminate between different fmtp options when selecting an H264 fmtp value to put into the answer.

Specifically, the sole supported H264 protocol in the SDK is:
`RTC_CODEC_H264_PROFILE_42E01F_LEVEL_ASYMMETRY_ALLOWED_PACKETIZATION_MODE`

In Google Chrome, the correct advertised FMTP value to select looks something like:
```a=fmtp:108 level-asymmetry-allowed=1;packetization-mode=0;profile-level-id=42e01f```

In firefox, the default packetization-mode value is assumed (0) so the offer looks like:
```a=fmtp:97 profile-level-id=42e01f;level-asymmetry-allowed=1```

In my testing, I found that the SDK is instead selecting whatever H264 fmtp line is last in the offer, which at times is incompatible with the stream being sent (e.g. packetization-mode=1, for example) which results in a breakdown of streaming.

Without a fix in place, I found that Google Chrome on Mac, Firefox on Linux and Mac, and Chromium on Linux would not process streams correctly, and communication would break down shortly after streaming begins, resulting often in a NACK shower.

With a fix in place, streaming seems to occur normally/as expected.

**SDK version number**

The issue was initially diagnosed and patched off of this commit point: https://github.com/awslabs/amazon-kinesis-video-streams-webrtc-sdk-c/commit/4e21b6e7e3029db94d3e260352ef1f07c261c50a

**Open source building**

N/A

**To Reproduce**
1. Configure a master that is offering an H264 video stream (this may work using the gstreamer sample, but we used our own custom library.) Enable logging of the SDP offer/answer.
2. Attempt to connect to the master via a client that advertises a multitude of H264 options (e.g. Chrome on Mac).
[...]

PR #952, opened 2020-11-14
Fix issue with H264 fmtp offer selection.

andyhoffy:
*Issue #, if available:*
Issue #902 

*Description of changes:*

This PR make the payloadType selection aware of the fmtp payload description, and attempts to select the one that is most compatible with the current implementation of H264 streaming in this library.

Note that with the loose interpretation in this proposed PR, there may be an additional issue (defect) with the way that the answer is produced in `populateSingleMediaSection` when formulating an SDP answer, as it mirrors the fmtp line from the SDP offer, even if there isn't a perfect protocol match. (Based on the naming of the constant `RTC_RTX_CODEC_H264_PROFILE_42E01F_LEVEL_ASYMMETRY_ALLOWED_PACKETIZATION_MODE` one may assume that the library will advertise `42E01F` back even in cases where `42E01F` isn't offered but level asymmetry is allowed.)


tst/SdpApiTest.cpp
```diff
@@ -625,13 +662,13 @@ a=group:BUNDLE 0
 
         track1.kind = MEDIA_STREAM_TRACK_KIND_VIDEO;
         track1.codec = RTC_CODEC_VP8;
-        STRNCPY(track1.streamId, "videoTrack1", MAX_MEDIA_STREAM_ID_LEN);
-        STRNCPY(track1.trackId, "videoTrack1", MAX_MEDIA_STREAM_ID_LEN);
+        STRNCPY(track1.streamId, "videoStream", MAX_MEDIA_STREAM_ID_LEN);
+        STRNCPY(track1.trackId, "videoTrack1", MAX_MEDIA_STREAM_TRACK_ID_LEN);
```
andyhoffy, line 666: Just some naming consistency cleanup... (no impact)

tst/SdpApiTest.cpp
```diff
@@ -811,7 +848,9 @@ a=rtcp-mux
 a=rtcp-rsize
 a=rtpmap:96 VP8/90000
 a=rtpmap:102 H264/90000
-a=fmtp:102 strange
```
andyhoffy, line 814: Change: the logic no longer accepts an H264 fmtp value without exact required characteristics.

tst/WebRTCClientTestFixture.h
```diff
@@ -1,6 +1,7 @@
 #include "gtest/gtest.h"
 #include "../src/source/Include_i.h"
 #include <memory>
+#include <string>
```
andyhoffy, line 4: I noticed that some tests were using std::string, but the header was not included here or in the tests.

src/source/PeerConnection/SessionDescription.c
```diff
@@ -316,6 +327,64 @@ PCHAR fmtpForPayloadType(UINT64 payloadType, PSessionDescription pSessionDescrip
...
+  if (fmtp == NULL) {
+      return 0;
+  }
+
+  // Currently, the packetization mode must be 1, as the packetization logic
+  // is currently not configurable, and sends both NALU and FU-A packets.
+  // https://tools.ietf.org/html/rfc7742#section-6.2
+  if (readHexValue(fmtp, "packetization-mode=", &packetizationMode) == FALSE || packetizationMode != 1) {
+      return 0;
+  }
```
andyhoffy, line 367: Note that this may be TOO stringent in cases where we are in rcvonly mode, as technically, the depayloader will handle packetization modes 0 and 1. However, in sndonly and sndrecv cases, there isn't any logic to protect the incoming packets from exceeding the MTU, triggering the creation of an FU-A type packet, which is only supported in packetization-mode=1. 

(Based on the packet types I saw, packetization-mode=2 does not appear to be fully implemented/supported, nor is it likely desirable in cases where you are real-time streaming?)

I'm a bit torn on this implementation, and could see making `packetization-mode` matching softly scored, but ideally that would only happen if the transceiver was only configure to receive and not send.

src/source/PeerConnection/SessionDescription.c
```diff
@@ -316,6 +326,64 @@ PCHAR fmtpForPayloadType(UINT64 payloadType, PSessionDescription pSessionDescrip
...
 }
 
+/*
+ * Extracts a (hex) value after the provided prefix string. Returns true if
+ * successful.
+ */
+static BOOL readHexValue(const PCHAR input, const PCHAR prefix, PINT32 value) {
+  const PCHAR substr = STRSTR(input, prefix);
+  if (substr != NULL) {
+    if (sscanf(substr + STRLEN(prefix), "%x", value) == 1) {
```
MushMal, line 336: Combine the two ifs like if (a && b)

Most importantly, we currently don't have a macro for sscanf. Not recommending to use sscanf either. In this case, recommend using 

CHAR valueStr[16];

if (substr != NULL && STATUS_SUCCEEDED(ULTOSTR(value,  valueStr, ARRAY_SIZE(valueStr), 16, NULL)) {
    return TRUE;
} else {
    return FALSE;
}

src/source/PeerConnection/SessionDescription.c
```diff
@@ -160,10 +160,13 @@ STATUS setPayloadTypesFromOffer(PHashTable codecTable, PHashTable rtxTable, PSes
     UINT16 aptFmtVal;
     BOOL supportCodec;
     UINT32 tokenLen, i, aptFmtpValCount;
+    PCHAR fmtp;
+    INT32 fmtpScore, bestFmtpScore;
```
MushMal, line 164: Always use unsigned integers unless you really need signed. In this case, there is no need for signed so use UINT32

src/source/PeerConnection/SessionDescription.c
```diff
@@ -316,6 +326,64 @@ PCHAR fmtpForPayloadType(UINT64 payloadType, PSessionDescription pSessionDescrip
...
+
+/*
+ * Scores the provided fmtp string based on this library's ability to
+ * process various types of H264 streams. A score of 0 indicates an
+ * incompatible fmtp line. Beyond this, a higher score indicates more
+ * compatibility with the desired characteristics, packetization-mode=1,
+ * level-asymmetry-allowed=1, and inbound match with our preferred
+ * profile-level-id.
+ */
+INT32 getH264FmtpScore(PCHAR fmtp) {
```
MushMal, line 351: UINT32

src/source/PeerConnection/SessionDescription.c
```diff
@@ -316,6 +326,64 @@ PCHAR fmtpForPayloadType(UINT64 payloadType, PSessionDescription pSessionDescrip
...
+  INT32 profileId = 0, packetizationMode = 0, levelAsymmetry = 0;
+  BOOL isAsymmetryAllowed = FALSE;
+  BOOL isProfileMatch = FALSE;
+  INT32 score = 0;
```
MushMal, line 355: UINT32

src/source/PeerConnection/SessionDescription.c
```diff
@@ -316,6 +326,64 @@ PCHAR fmtpForPayloadType(UINT64 payloadType, PSessionDescription pSessionDescrip
...
+/*
+ * Scores the provided fmtp string based on this library's ability to
+ * process various types of H264 streams. A score of 0 indicates an
+ * incompatible fmtp line. Beyond this, a higher score indicates more
+ * compatibility with the desired characteristics, packetization-mode=1,
+ * level-asymmetry-allowed=1, and inbound match with our preferred
+ * profile-level-id.
+ */
+INT32 getH264FmtpScore(PCHAR fmtp) {
+  INT32 profileId = 0, packetizationMode = 0, levelAsymmetry = 0;
```
MushMal, line 352: UINT32

src/source/PeerConnection/SessionDescription.c
```diff
@@ -316,6 +326,64 @@ PCHAR fmtpForPayloadType(UINT64 payloadType, PSessionDescription pSessionDescrip
...
+
+  // No ftmp match found.
+  if (fmtp == NULL) {
+      return 0;
+  }
+
+  // Currently, the packetization mode must be 1, as the packetization logic
+  // is currently not configurable, and sends both NALU and FU-A packets.
+  // https://tools.ietf.org/html/rfc7742#section-6.2
+  if (readHexValue(fmtp, "packetization-mode=", &packetizationMode) == FALSE || packetizationMode != 1) {
```
MushMal, line 365: if (!readHexValue(...) || packetizationMode != 1)

MushMal: Also think about combining with if (fmtp == NULL || readHex() || packetizationMode  1)

src/source/PeerConnection/SessionDescription.c
MushMal: I agree. We really need to look deeper at whether we need to add support for interleaved mode with MTAP or we don't need to. Based on how widespread the support is

src/source/PeerConnection/SessionDescription.c
```diff
@@ -316,6 +326,64 @@ PCHAR fmtpForPayloadType(UINT64 payloadType, PSessionDescription pSessionDescrip
...
+      return 0;
+  }
+
+  if (readHexValue(fmtp, "profile-level-id=", &profileId) == TRUE) {
+      isProfileMatch = profileId == 0x42e01f;
```
MushMal, line 370: use parenthesis - more readable

isProfileMatch = (profileId == blah);

src/source/PeerConnection/SessionDescription.c
```diff
@@ -316,6 +326,64 @@ PCHAR fmtpForPayloadType(UINT64 payloadType, PSessionDescription pSessionDescrip
...
+  }
+
+  if (readHexValue(fmtp, "level-asymmetry-allowed=", &levelAsymmetry) == TRUE) {
+      isAsymmetryAllowed = (levelAsymmetry == 1);
+  }
+
+  if (isProfileMatch) {
```
MushMal, line 377: Seems that you could use uint32s instead of these booleans and simply return the addition

src/source/PeerConnection/SessionDescription.h
```diff
@@ -64,6 +64,7 @@ STATUS populateSessionDescription(PKvsPeerConnection, PSessionDescription, PSess
 STATUS reorderTransceiverByRemoteDescription(PKvsPeerConnection, PSessionDescription);
 STATUS setReceiversSsrc(PSessionDescription, PDoubleList);
 PCHAR fmtpForPayloadType(UINT64, PSessionDescription);
+INT32 getH264FmtpScore(PCHAR);
```
MushMal, line 67: UINT32

MushMal (commented):
Fantastic work! Thank you. Some comments.

lherman-cs:
Adding some documentations for other readers:

Taken from https://tools.ietf.org/html/rfc6184#section-5.4

      Table 3.  Summary of allowed NAL unit types for each packetization
                mode (yes = allowed, no = disallowed, ig = ignore)

      Payload Packet    Single NAL    Non-Interleaved    Interleaved
      Type    Type      Unit Mode           Mode             Mode
      -------------------------------------------------------------
      0      reserved      ig               ig               ig
      1-23   NAL unit     yes              yes               no
      24     STAP-A        no              yes               no
      25     STAP-B        no               no              yes
      26     MTAP16        no               no              yes
      27     MTAP24        no               no              yes
      28     FU-A          no              yes              yes
      29     FU-B          no               no              yes
      30-31  reserved      ig               ig               ig

As of 0348f5724268f5539f8b95e09d3f5a798147f719, our SDK only supports single NALU https://tools.ietf.org/html/rfc6184#section-5.6 and FU-A https://tools.ietf.org/html/rfc6184#section-5.8 (only when the packet is larger than the MTU) in H264 RTP packetization. Therefore, the sender MUST offer `packetization-mode=1` when sending since there is a chance that H264 frames will exceed the MTU size (suggested by @andyhoffy, https://github.com/awslabs/amazon-kinesis-video-streams-webrtc-sdk-c/issues/902)

src/source/PeerConnection/SessionDescription.c
```diff
@@ -189,7 +192,13 @@ STATUS setPayloadTypesFromOffer(PHashTable codecTable, PHashTable rtxTable, PSes
             CHK_STATUS(hashTableContains(codecTable, RTC_CODEC_H264_PROFILE_42E01F_LEVEL_ASYMMETRY_ALLOWED_PACKETIZATION_MODE, &supportCodec));
             if (supportCodec && (end = STRSTR(attributeValue, H264_VALUE)) != NULL) {
                 CHK_STATUS(STRTOUI64(attributeValue, end - 1, 10, &parsedPayloadType));
-                CHK_STATUS(hashTableUpsert(codecTable, RTC_CODEC_H264_PROFILE_42E01F_LEVEL_ASYMMETRY_ALLOWED_PACKETIZATION_MODE, parsedPayloadType));
+                fmtp = fmtpForPayloadType(parsedPayloadType, pSessionDescription);
+                fmtpScore = getH264FmtpScore(fmtp);
+                if (fmtpScore > bestFmtpScore) {
+                    DLOGV("Found H264 payload type %" PRId64 " with score %d: %s", parsedPayloadType, fmtpScore, fmtp);
+                    CHK_STATUS(hashTableUpsert(codecTable, RTC_CODEC_H264_PROFILE_42E01F_LEVEL_ASYMMETRY_ALLOWED_PACKETIZATION_MODE, parsedPayloadType));
```
lherman-cs, line 199: As of now, this codec selection will work fine since `getH264FmtpScore` returns fitness distance solely based on the packetization mode. But, this algorithm can be used in the future to find the best profile.

src/source/PeerConnection/SessionDescription.c
lherman-cs: I think `0x42e01f` should be a constant, and we should add a `FIXME` as we may add more profiles in the future.

tst/SdpApiTest.cpp
```diff
@@ -895,6 +964,74 @@ a=rtpmap:102 H264/90000
...
+
+    EXPECT_PRED_FORMAT2(testing::IsSubstring, match(),
+                        rtcSessionDescriptionInit.sdp) << "Offer:\n" << sdp << "\nAnswer:\n" << rtcSessionDescriptionInit.sdp;
+    closePeerConnection(pRtcPeerConnection);
+    freePeerConnection(&pRtcPeerConnection);
+}
+
+INSTANTIATE_TEST_CASE_P(SdpApiTest_SdpMatch_Chrome, SdpApiTest_SdpMatch,
+                        ::testing::Values(SdpMatch{
+                                              "../samples/SDP/offers/1v1a1d-chrome-linux.txt",
```
lherman-cs, line 1006: knit: I think we should put these sample SDPs under `tst/SDP/offers/` as this belongs to unit tests only.

lherman-cs (commented):
Thank you for the contribution! This is an awesome work!! Dropped some minor comments.

staowyze:
We encountered the same issue, thank you for working on the fix, hope to get the fix in ASAP.

samples/SDP/offers/1v1a1d-chrome-linux.txt
```diff
@@ -0,0 +1,144 @@
+v=0
```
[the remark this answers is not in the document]
andyhoffy: Yeah, I'll move them into a new test folder. Was a bit timid about adding a folder for test artifacts.

src/source/PeerConnection/SessionDescription.c
```diff
@@ -316,6 +326,64 @@ PCHAR fmtpForPayloadType(UINT64 payloadType, PSessionDescription pSessionDescrip
     return NULL;
 }
 
+/*
+ * Extracts a (hex) value after the provided prefix string. Returns true if
+ * successful.
+ */
+static BOOL readHexValue(const PCHAR input, const PCHAR prefix, PINT32 value) {
```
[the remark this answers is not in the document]
andyhoffy: Ok, will pull the const usage out - looks like that's the preferred pattern in this library.

Re static, this makes the function private to this file, which is what I was going for (hide from linker...) - this was to prevent it from appearing in the global symbol space. Would you prefer if I made this function linkable outside this compilation unit (and declare in SessionDescription.h?)

src/source/PeerConnection/SessionDescription.c
```diff
@@ -316,6 +326,64 @@ PCHAR fmtpForPayloadType(UINT64 payloadType, PSessionDescription pSessionDescrip
...
+  const PCHAR substr = STRSTR(input, prefix);
```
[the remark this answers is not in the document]
andyhoffy: Ok, will do!

src/source/PeerConnection/SessionDescription.c
andyhoffy: Hmm, this is a bit tricky... my preference would be to use `sscanf` from the standard library, as it can read arbitrary length case-agnostic hex values, which is what we want here, I think? If this sounds okay, I'll put up a PR to get a macro wrapper into the pic code. :-)

src/source/PeerConnection/SessionDescription.c
andyhoffy: Ok, will combine with other feedback and switch to DOUBLE, so as to allow scoring between [0, 1]

src/source/PeerConnection/SessionDescription.c
andyhoffy: Interesting! Will try changing this to DOUBLE, let me know what you think. Can revert if it's too much.

src/source/PeerConnection/SessionDescription.c
andyhoffy: Note: If using sscanf(str, "%x", ...), I think these have to be INT32s... see "type of argument" for x, X: https://www.tutorialspoint.com/c_standard_library/c_function_sscanf.htm

src/source/PeerConnection/SessionDescription.c
andyhoffy: Ooh, like the parens as well - that's cleaner.

Will add a #define for this profile ID, lmk if you agree with it. Didn't want to cast to an INT32/UINT32 as I could see it being used in both signed and unsigned contexts. Good news is that it's equivalent since it has leading zeros... (0x4...)

src/source/PeerConnection/SessionDescription.c
andyhoffy: How strongly do you feel about this? I just tried it out with that approach, and it felt a bit less intuitive/readable to someone trying to intuit what the scoring meant...

src/source/PeerConnection/SessionDescription.h
andyhoffy: :+1: Changing to DOUBLE per earlier discussion.

samples/SDP/offers/1v1a1d-chrome-linux.txt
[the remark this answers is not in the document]
andyhoffy: FYI, I continued an assumption that the build dir is one directory under the project root, and that tests are run from that directory. That may be a bad assumption, but I noticed that RtpFunctionalityTest.cpp is already assuming this.

We could change this at some point (at least the build directory location assumption) by adding a cmake target to copy over the files into the build directory, so it could at least be agnostic to where the build dir lives (e.g. some editors put their build dirs at the same level as the project directory, instead of under it...)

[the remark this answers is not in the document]
MushMal: The other alternative is to simply define these as static const strings in test cpp

src/source/PeerConnection/SessionDescription.c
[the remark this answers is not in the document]
MushMal: We use visibility to expose or not to from the compile unit. All our public APIs are marked as PUBLIC_API

src/source/PeerConnection/SessionDescription.c
MushMal: sscanf will not do much better in terms of handling hex as in your case you are not even trying uint64 but rather int. Sscanf is multiple orders of magnitude in terms of runtime costs. However, this usage is not in a main "hot" path so I am OK perf-wise. As a rule of thumb, the usage of these APIs should be super limited to not proliferate across the SDK to keep the runtime footprint low.

src/source/PeerConnection/SessionDescription.c
MushMal: DOUBLE sounds good as well. Though, it would be funny defining the range. Using UINT32 would allow to simply add the "features" and return the value as the score, allowing later to expand

src/source/PeerConnection/SessionDescription.c
MushMal: I disagree. Using float will mean that you need to come up with forward compat scheme. Whereas an integral type will allow to simply add features to get higher score and be better future proof

src/source/PeerConnection/SessionDescription.c
MushMal: Good documentation is the key. This approach is better future proof in terms of getting even closer match

src/source/PeerConnection/SessionDescription.c
```diff
@@ -316,6 +326,65 @@ PCHAR fmtpForPayloadType(UINT64 payloadType, PSessionDescription pSessionDescrip
...
+
+  // Currently, the packetization mode must be 1, as the packetization logic
+  // is currently not configurable, and sends both NALU and FU-A packets.
+  // https://tools.ietf.org/html/rfc7742#section-6.2
+  if (!readHexValue(fmtp, "packetization-mode=", &packetizationMode) || packetizationMode != 1) {
+      return 0.;
+  }
+
+  if (readHexValue(fmtp, "profile-level-id=", &profileId)) {
+      isProfileMatch = (profileId == H264_PROFILE_42E01F);
```
docee, line 371: Some browsers do not have the `profile-level-id` of `42E01F`. The `profile-level-id` consists of three parts, which are `profile_idc (8 bits)`, `profile_iop (8 bits)`, and `level_idc (8 bits)`.

So I think it should be more DETAILED when calculating the fmtp score. The score of `profile_idc` of `0x42` must be higher than `0x64`.

MushMal: This is a good suggestion, however, for the context of this PR I would leave it out and follow-up with another change to determine the profile-level-id. This is a terrible RFC all together and mangling different meanings in a free-handed text is a bad solution. At any rate, we should take each one of these individually as supporting the entire RFC makes no sense - just the parts that are real-life use cases which these actually are.

MushMal:
@andyhoffy Any update on this PR? I would like to get this in but I don't feel comfortable with the DOUBLE for the score instead of an additive UINT. The issue with the float type is that it works only when you have an upper bound.

src/source/PeerConnection/SessionDescription.c
andyhoffy: @docee Thanks! It's a good point. Have you come across any SDP offers that don't support `42E01F`? If so, could you send me a sample? I'm trying to build a small library of real-world SDP offers in this PR so that the SDK doesn't unknowingly exclude a previously supported platform with a change. 

I'd like to try and get this in ASAP (I've already let it linger too long) as the current code config'd to use H264 seems to be broken "as-is" on several major browsers/platforms. But I see what you're saying and we probably ought to follow up by somehow generalizing H264 support beyond `RTC_CODEC_H264_PROFILE_42E01F_LEVEL_ASYMMETRY_ALLOWED_PACKETIZATION_MODE`. 

As @MushMal pointed out, the RFC is both broad and convoluted, so it seems like achieving broad compatibility is much more dependent on finding and supporting the most commonly implemented encoding schemes than it is about satisfying the full breadth of what's offered in the spec. I'm definitely nervous that this change "misses" a sweet spot among other devices.

The support for transceiver configuration in this SDK is (as you noted) very constrained, as is support for any sort of asymmetrical encoding configuration. For example, it doesn't even expose what's in the offer to the API user, so there's no opportunity (for example) to reconfigure encoding based on available options. That's certainly an area for growth!

docee: This is Chrome SDP on android:
```
"v=0
o=- 2383219383355379692 2 IN IP4 127.0.0.1
s=-
t=0 0
a=group:BUNDLE 0 1 2
a=msid-semantic: WMS 3CXV4snScv28Bl5Ltn7V4StSDzTGKOnaaAdf
m=audio 9 UDP/TLS/RTP/SAVPF 111 103 9 0 8 105 13 110 113 126
c=IN IP4 0.0.0.0
a=rtcp:9 IN IP4 0.0.0.0
a=ice-ufrag:Wafl
a=ice-pwd:IS933a+6pookL48tYXASXnwW
a=ice-options:trickle
a=fingerprint:sha-256 49:3C:77:6E:6B:2B:90:69:00:AE:47:4A:87:A7:F4:F6:F0:B3:6D:D6:FA:7F:84:DD:A5:6E:6E:2E:1F:64:D7:7F
a=setup:actpass
a=mid:0
a=extmap:1 urn:ietf:params:rtp-hdrext:ssrc-audio-level
a=extmap:2 http://www.webrtc.org/experiments/rtp-hdrext/abs-send-time
a=extmap:3 http://www.ietf.org/id/draft-holmer-rmcat-transport-wide-cc-extensions-01
a=extmap:4 urn:ietf:params:rtp-hdrext:sdes:mid
a=extmap:5 urn:ietf:params:rtp-hdrext:sdes:rtp-stream-id
a=extmap:6 urn:ietf:params:rtp-hdrext:sdes:repaired-rtp-stream-id
a=sendrecv
a=msid:3CXV4snScv28Bl5Ltn7V4StSDzTGKOnaaAdf f94c5ff6-26b9-4315-815d-40b4dd2efdef
a=rtcp-mux
a=rtpmap:111 opus/48000/2
a=rtcp-fb:111 transport-cc
a=fmtp:111 minptime=10;useinbandfec=1
a=rtpmap:103 ISAC/16000
a=rtpmap:9 G722/8000
a=rtpmap:0 PCMU/8000
a=rtpmap:8 PCMA/8000
a=rtpmap:105 CN/16000
a=rtpmap:13 CN/8000
a=rtpmap:110 telephone-event/48000
a=rtpmap:113 telephone-event/16000
a=rtpmap:126 telephone-event/8000
a=ssrc:2261433415 cname:5L7rDchCDtUnZEok
a=ssrc:2261433415 msid:3CXV4snScv28Bl5Ltn7V4StSDzTGKOnaaAdf f94c5ff6-26b9-4315-815d-40b4dd2efdef
a=ssrc:2261433415 mslabel:3CXV4snScv28Bl5Ltn7V4StSDzTGKOnaaAdf
a=ssrc:2261433415 label:f94c5ff6-26b9-4315-815d-40b4dd2efdef
m=video 9 UDP/TLS/RTP/SAVPF 96 97 98 99 127 125 104 122 106 107 108 109 124 121 100 101 102
c=IN IP4 0.0.0.0
a=rtcp:9 IN IP4 0.0.0.0
a=ice-ufrag:Wafl
a=ice-pwd:IS933a+6pookL48tYXASXnwW
a=ice-options:trickle
a=fingerprint:sha-256 49:3C:77:6E:6B:2B:90:69:00:AE:47:4A:87:A7:F4:F6:F0:B3:6D:D6:FA:7F:84:DD:A5:6E:6E:2E:1F:64:D7:7F
a=setup:actpass
a=mid:1
a=extmap:14 urn:ietf:params:rtp-hdrext:toffset
a=extmap:2 http://www.webrtc.org/experiments/rtp-hdrext/abs-send-time
a=extmap:13 urn:3gpp:video-orientation
a=extmap:3 http://www.ietf.org/id/draft-holmer-rmcat-transport-wide-cc-extensions-01
a=extmap:12 http://www.webrtc.org/experiments/rtp-hdrext/playout-delay
a=extmap:11 http://www.webrtc.org/experiments/rtp-hdrext/video-content-type
a=extmap:7 http://www.webrtc.org/experiments/rtp-hdrext/video-timing
a=extmap:8 http://www.webrtc.org/experiments/rtp-hdrext/color-space
a=extmap:4 urn:ietf:params:rtp-hdrext:sdes:mid
a=extmap:5 urn:ietf:params:rtp-hdrext:sdes:rtp-stream-id
a=extmap:6 urn:ietf:params:rtp-hdrext:sdes:repaired-rtp-stream-id
a=recvonly
a=rtcp-mux
a=rtcp-rsize
a=rtpmap:96 VP8/90000
a=rtcp-fb:96 goog-remb
a=rtcp-fb:96 transport-cc
a=rtcp-fb:96 ccm fir
a=rtcp-fb:96 nack
a=rtcp-fb:96 nack pli
a=rtpmap:97 rtx/90000
a=fmtp:97 apt=96
a=rtpmap:98 VP9/90000
a=rtcp-fb:98 goog-remb
a=rtcp-fb:98 transport-cc
a=rtcp-fb:98 ccm fir
a=rtcp-fb:98 nack
a=rtcp-fb:98 nack pli
a=fmtp:98 profile-id=0
a=rtpmap:99 rtx/90000
a=fmtp:99 apt=98
a=rtpmap:127 VP9/90000
a=rtcp-fb:127 goog-remb
a=rtcp-fb:127 transport-cc
a=rtcp-fb:127 ccm fir
a=rtcp-fb:127 nack
a=rtcp-fb:127 nack pli
a=fmtp:127 profile-id=1
a=rtpmap:125 VP9/90000
a=rtcp-fb:125 goog-remb
a=rtcp-fb:125 transport-cc
a=rtcp-fb:125 ccm fir
a=rtcp-fb:125 nack
a=rtcp-fb:125 nack pli
a=fmtp:125 profile-id=2
a=rtpmap:104 H264/90000
a=rtcp-fb:104 goog-remb
a=rtcp-fb:104 transport-cc
a=rtcp-fb:104 ccm fir
a=rtcp-fb:104 nack
a=rtcp-fb:104 nack pli
a=fmtp:104 level-asymmetry-allowed=1;packetization-mode=1;profile-level-id=420015
a=rtpmap:122 rtx/90000
a=fmtp:122 apt=104
a=rtpmap:106 H264/90000
a=rtcp-fb:106 goog-remb
a=rtcp-fb:106 transport-cc
a=rtcp-fb:106 ccm fir
a=rtcp-fb:106 nack
a=rtcp-fb:106 nack pli
a=fmtp:106 level-asymmetry-allowed=1;packetization-mode=1;profile-level-id=4d0015
a=rtpmap:107 rtx/90000
a=fmtp:107 apt=106
a=rtpmap:108 H264/90000
a=rtcp-fb:108 goog-remb
a=rtcp-fb:108 transport-cc
a=rtcp-fb:108 ccm fir
a=rtcp-fb:108 nack
a=rtcp-fb:108 nack pli
a=fmtp:108 level-asymmetry-allowed=1;packetization-mode=1;profile-level-id=640015
a=rtpmap:109 rtx/90000
a=fmtp:109 apt=108
a=rtpmap:124 H264/90000
a=rtcp-fb:124 goog-remb
a=rtcp-fb:124 transport-cc
a=rtcp-fb:124 ccm fir
a=rtcp-fb:124 nack
a=rtcp-fb:124 nack pli
a=fmtp:124 level-asymmetry-allowed=1;packetization-mode=1;profile-level-id=42e015
a=rtpmap:121 rtx/90000
a=fmtp:121 apt=124
a=rtpmap:100 red/90000
a=rtpmap:101 rtx/90000
a=fmtp:101 apt=100
a=rtpmap:102 ulpfec/90000
m=application 9 UDP/DTLS/SCTP webrtc-datachannel
c=IN IP4 0.0.0.0
a=ice-ufrag:Wafl
a=ice-pwd:IS933a+6pookL48tYXASXnwW
a=ice-options:trickle
a=fingerprint:sha-256 49:3C:77:6E:6B:2B:90:69:00:AE:47:4A:87:A7:F4:F6:F0:B3:6D:D6:FA:7F:84:DD:A5:6E:6E:2E:1F:64:D7:7F
a=setup:actpass
a=mid:2
a=sctp-port:5000
a=max-message-size:262144
"
```

lherman-cs:
Closing this in favor of https://github.com/awslabs/amazon-kinesis-video-streams-webrtc-sdk-c/pull/966

src/source/PeerConnection/SessionDescription.c
andyhoffy: Thanks @docee! I think @lherman-cs and @MushMal moved over to another PR (they were going to push some final changes) - I'll make sure that I follow up and get your SDP capture in the test set!!!

lherman-cs: Thanks @docee! The SDP has been included in the unit test, https://github.com/awslabs/amazon-kinesis-video-streams-webrtc-sdk-c/pull/966/files#diff-560b56b36b72eba42e1ed20d819402995117a0cdab25953080198ca7c13e8613R1011-R1014

Issue #908, opened 2020-10-30
[FEATURE] add API to set cache file apth


zwei333:
We are using File Cache feature. (code implementation in src/source/Signaling/FileCache.c)
However, there is no API to set DEFAULT_CACHE_FILE_PATH.  It is hard coded to be the same directory as webrtc binary.
In our system, app binary is in read only folder so the cache file "./.SignalingCache_v0" doesn't work.

Please provide API to override the default cache file path.

**Is your feature request related to a problem? Please describe.**

**Describe the solution you'd like**

**Describe alternatives you've considered**

**Additional context**

zwei333:
Yes, being able to specify the path will be good enough.

disa6302:
Change merged. Closing.

PR #964, opened 2020-11-18
Adding ability to specify the cache file path

MushMal:
Issue #908 

Revving the version of ClientInfo struct to include the optional path to the cache file. Default value is used by specifying NULL or EMPTY_STRING. Adding test variations.

NOTE: This is an important change to allow cache file usage for many embedded devices where the partition for the binaries is locked/read-only.

### DIFF
--- a/src/source/Signaling/Signaling.c
+++ b/src/source/Signaling/Signaling.c
@@ -44,7 +44,9 @@ STATUS createSignalingSync(PSignalingClientInfoInternal pClientInfo, PChannelInf
 
     if (pSignalingClient->pChannelInfo->cachingPolicy == SIGNALING_API_CALL_CACHE_TYPE_FILE) {
         // Signaling channel name can be NULL in case of pre-created channels in which case we use ARN as the name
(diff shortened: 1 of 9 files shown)


src/source/Signaling/Signaling.c
```diff
@@ -527,6 +528,32 @@ STATUS validateSignalingClientInfo(PSignalingClient pSignalingClient, PSignaling
...
+                    STATUS_SIGNALING_INVALID_CLIENT_INFO_CACHE_FILE_PATH_LEN);
+                STRCPY(pSignalingClient->clientInfo.cacheFilePath, pSignalingClient->clientInfo.signalingClientInfo.cacheFilePath);
+            } else {
+                // Set the default path
+                STRCPY(pSignalingClient->clientInfo.cacheFilePath, DEFAULT_CACHE_FILE_PATH);
+            }
+
+            break;
+
+        default:
```
disa6302, line 553: Just wondering, do we want to fail here? Maybe we can fallback on v0 instead of failing?

disa6302 (approved):
Just one small comment. The rest looks good to me.

src/source/Signaling/Signaling.c
MushMal: This really is an internal code failure case - more like an assert. We should never reach this state as we validate the versions but in case we add another version at some stage and forget to add the case for it. I would, in fact, use CHECK to crash the execution but this is similar to just returning a failure

Issue #950, opened 2020-11-13
[BUG] h264 main profile is picked in SDK answer


staowyze:
The Offer
```
Offer -----v=0
o=- 7371364611992188261 2 IN IP4 127.0.0.1
s=-
t=0 0
a=group:BUNDLE 0 1
a=msid-semantic: WMS
m=audio 63441 UDP/TLS/RTP/SAVPF 111 103 104 9 0 8 106 105 13 110 112 113 126
c=IN IP4 54.186.171.255
a=rtcp:9 IN IP4 0.0.0.0
a=candidate:98406528 1 udp 2122260223 192.168.31.76 34387 typ host generation 0 network-id 1 network-cost 10
a=candidate:1264479344 1 tcp 1518280447 192.168.31.76 9 typ host tcptype active generation 0 network-id 1 network-cost 10
a=candidate:2224395828 1 udp 1686052607 67.171.13.99 34387 typ srflx raddr 192.168.31.76 rport 34387 generation 0 network-id 1 network-cost 10
a=candidate:765018076 1 udp 41885439 54.186.171.255 63441 typ relay raddr 67.171.13.99 rport 34387 generation 0 network-id 1 network-cost 10
a=ice-ufrag:4qhB
a=ice-pwd:QloaLW7Rkt1PDBUbL7ktv8ul
a=fingerprint:sha-256 96:0D:C2:C5:75:71:05:44:90:CC:A1:B8:54:98:E4:14:07:25:F3:71:EB:D7:D4:66:3E:0E:DD:E7:3F:1F:2E:86
a=setup:actpass
a=mid:0
a=extmap:1 urn:ietf:params:rtp-hdrext:ssrc-audio-level
a=extmap:2 http://www.webrtc.org/experiments/rtp-hdrext/abs-send-time
a=extmap:3 http://www.ietf.org/id/draft-holmer-rmcat-transport-wide-cc-extensions-01
a=extmap:4 urn:ietf:params:rtp-hdrext:sdes:mid
a=extmap:5 urn:ietf:params:rtp-hdrext:sdes:rtp-stream-id
a=extmap:6 urn:ietf:params:rtp-hdrext:sdes:repaired-rtp-stream-id
a=sendrecv
a=msid:- f73809d1-50c5-44dd-9963-cef325d1d00a
a=rtcp-mux
a=rtpmap:111 opus/48000/2
a=rtcp-fb:111 transport-cc
[...]

Issue #940, opened 2020-11-11
[QUESTION] Incorrect profile-level-id may cause NullVideoDecoder in Chrome browsers of some systems


docee:
Incorrect profile-level-id may cause NullVideoDecoder in Chrome browsers of some systems. This issue is similar to #879 .

The SDP sent by Chrome contain multiple ftmp messages. The SDP Offer sample is as follows:

```
sdp: "v=0
↵o=- 537342113758176669 2 IN IP4 127.0.0.1
↵s=-
↵t=0 0
↵a=group:BUNDLE 0 1 2
↵a=msid-semantic: WMS jWS9zQ9ZLJ96v17bJ4Gkyif0Kr4qJZvs96SZ
↵m=video 9 UDP/TLS/RTP/SAVPF 96 97 98 99 100 101 102 121 127 120 125 107 108 109 124 119 123 118 114 115 116
↵c=IN IP4 0.0.0.0
↵a=rtcp:9 IN IP4 0.0.0.0
↵a=ice-ufrag:iCJm
↵a=ice-pwd:0zJvVkq9UV+txsFxIuTQeElO
↵a=ice-options:trickle
↵a=fingerprint:sha-256 26:B5:79:41:34:28:3F:C8:1D:F5:C8:4F:AC:EE:E8:21:FA:05:80:0C:DE:79:A4:71:73:7E:47:64:C4:78:52:7A
↵a=setup:actpass
↵a=mid:0
↵a=extmap:1 urn:ietf:params:rtp-hdrext:toffset
↵a=extmap:2 http://www.webrtc.org/experiments/rtp-hdrext/abs-send-time
↵a=extmap:3 urn:3gpp:video-orientation
↵a=extmap:4 http://www.ietf.org/id/draft-holmer-rmcat-transport-wide-cc-extensions-01
↵a=extmap:5 http://www.webrtc.org/experiments/rtp-hdrext/playout-delay
↵a=extmap:6 http://www.webrtc.org/experiments/rtp-hdrext/video-content-type
↵a=extmap:7 http://www.webrtc.org/experiments/rtp-hdrext/video-timing
↵a=extmap:8 http://www.webrtc.org/experiments/rtp-hdrext/color-space
↵a=extmap:9 urn:ietf:params:rtp-hdrext:sdes:mid
↵a=extmap:10 urn:ietf:params:rtp-hdrext:sdes:rtp-stream-id
[...]

PR #966, opened 2020-11-19
Fix/sdp h264 mismatch

lherman-cs:
**Related Issues**

* https://github.com/awslabs/amazon-kinesis-video-streams-webrtc-sdk-c/issues/902
* https://github.com/awslabs/amazon-kinesis-video-streams-webrtc-sdk-c/issues/940
* https://github.com/awslabs/amazon-kinesis-video-streams-webrtc-sdk-c/issues/950

**Description of changes**

This PR make the payloadType selection aware of the fmtp payload description, and attempts to select the one that is most compatible with the current implementation of H264 streaming in this library.

Note that with the loose interpretation in this proposed PR, there may be an additional issue (defect) with the way that the answer is produced in `populateSingleMediaSection` when formulating an SDP answer, as it mirrors the fmtp line from the SDP offer, even if there isn't a perfect protocol match. (Based on the naming of the constant `RTC_RTX_CODEC_H264_PROFILE_42E01F_LEVEL_ASYMMETRY_ALLOWED_PACKETIZATION_MODE` one may assume that the library will advertise `42E01F` back even in cases where `42E01F` isn't offered but level asymmetry is allowed.)

This PR obsoletes https://github.com/awslabs/amazon-kinesis-video-streams-webrtc-sdk-c/pull/952


src/source/PeerConnection/SessionDescription.c
```diff
@@ -160,10 +160,13 @@ STATUS setPayloadTypesFromOffer(PHashTable codecTable, PHashTable rtxTable, PSes
     UINT16 aptFmtVal;
     BOOL supportCodec;
     UINT32 tokenLen, i, aptFmtpValCount;
+    PCHAR fmtp;
+    UINT32 fmtpScore, bestFmtpScore;
 
     for (currentMedia = 0; currentMedia < pSessionDescription->mediaCount; currentMedia++) {
         pMediaDescription = &(pSessionDescription->mediaDescriptions[currentMedia]);
         aptFmtpValCount = 0;
+        bestFmtpScore = 0.;
```
MushMal, line 169: Remove the period after 0

src/source/PeerConnection/SessionDescription.c
```diff
@@ -189,7 +192,14 @@ STATUS setPayloadTypesFromOffer(PHashTable codecTable, PHashTable rtxTable, PSes
             CHK_STATUS(hashTableContains(codecTable, RTC_CODEC_H264_PROFILE_42E01F_LEVEL_ASYMMETRY_ALLOWED_PACKETIZATION_MODE, &supportCodec));
             if (supportCodec && (end = STRSTR(attributeValue, H264_VALUE)) != NULL) {
                 CHK_STATUS(STRTOUI64(attributeValue, end - 1, 10, &parsedPayloadType));
-                CHK_STATUS(hashTableUpsert(codecTable, RTC_CODEC_H264_PROFILE_42E01F_LEVEL_ASYMMETRY_ALLOWED_PACKETIZATION_MODE, parsedPayloadType));
+                fmtp = fmtpForPayloadType(parsedPayloadType, pSessionDescription);
+                fmtpScore = getH264FmtpScore(fmtp);
+                if (fmtpScore > bestFmtpScore) {
+                    DLOGV("Found H264 payload type %" PRId64 " with score %f.2: %s", parsedPayloadType, fmtpScore, fmtp);
```
MushMal, line 198: Change the %f.2 to %u

src/source/PeerConnection/SessionDescription.c
```diff
@@ -316,6 +327,74 @@ PCHAR fmtpForPayloadType(UINT64 payloadType, PSessionDescription pSessionDescrip
...
+ * Extracts a (hex) value after the provided prefix string. Returns true if
+ * successful.
+ */
+BOOL readHexValue(PCHAR input, PCHAR prefix, PINT32 value)
+{
+    PCHAR substr = STRSTR(input, prefix);
+    if (substr != NULL) {
+        // TODO: This should be changed to SSCANF once the PIC library macro
+        // change is propagated and available here.
+        if (sscanf(substr + STRLEN(prefix), "%x", value) == 1) {
```
MushMal, line 340: If you decided to go with SSCANF then at least combine the if statements via &&

src/source/PeerConnection/SessionDescription.c
```diff
@@ -316,6 +327,74 @@ PCHAR fmtpForPayloadType(UINT64 payloadType, PSessionDescription pSessionDescrip
...
+}
+
+/*
+ * Scores the provided fmtp string based on this library's ability to
+ * process various types of H264 streams. A score of 0 indicates an
+ * incompatible fmtp line. Beyond this, a higher score indicates more
+ * compatibility with the desired characteristics, packetization-mode=1,
+ * level-asymmetry-allowed=1, and inbound match with our preferred
+ * profile-level-id. Scoring is on a scale from [0,1], with 1 deemed as
+ * a "perfect fit".
```
MushMal, line 354: This comment does not correspond to reality

src/source/PeerConnection/SessionDescription.c
```diff
@@ -316,6 +327,74 @@ PCHAR fmtpForPayloadType(UINT64 payloadType, PSessionDescription pSessionDescrip
...
+ *
+ * At some future time, it may be worth expressing this as a true distance
+ * function as defined here, although dealing with infinite floating point
+ * values can get tricky:
+ * https://www.w3.org/TR/mediacapture-streams/#dfn-fitness-distance
```
MushMal, line 359: I really don't understand why this is any better than simply having a cumulative sum like we are doing

src/source/PeerConnection/SessionDescription.c
```diff
@@ -316,6 +327,74 @@ PCHAR fmtpForPayloadType(UINT64 payloadType, PSessionDescription pSessionDescrip
...
+ */
+UINT32 getH264FmtpScore(PCHAR fmtp)
+{
+    INT32 profileId = 0, packetizationMode = 0, levelAsymmetry = 0;
```
MushMal, line 363: Use UINT32

src/source/PeerConnection/SessionDescription.c
```diff
@@ -316,6 +327,74 @@ PCHAR fmtpForPayloadType(UINT64 payloadType, PSessionDescription pSessionDescrip
...
+
+    // Currently, the packetization mode must be 1, as the packetization logic
+    // is currently not configurable, and sends both NALU and FU-A packets.
+    // https://tools.ietf.org/html/rfc7742#section-6.2
+    if (!readHexValue(fmtp, "packetization-mode=", &packetizationMode) || packetizationMode != 1) {
+        return 0;
+    }
+
+    if (readHexValue(fmtp, "profile-level-id=", &profileId)) {
+        isProfileMatch = (profileId == H264_PROFILE_42E01F);
```
MushMal, line 381: You don't need any of the booleans. 

if (readHexValue(fmtp, "profile-level-id=", &profileId) && profileId == H264_PROFILE_42E01F) {
    score++;
}

src/source/PeerConnection/SessionDescription.c
```diff
@@ -316,6 +327,74 @@ PCHAR fmtpForPayloadType(UINT64 payloadType, PSessionDescription pSessionDescrip
...
+    }
+
+    if (readHexValue(fmtp, "level-asymmetry-allowed=", &levelAsymmetry)) {
```
MushMal, line 384: same here

src/source/PeerConnection/SessionDescription.c
```diff
@@ -316,6 +327,74 @@ PCHAR fmtpForPayloadType(UINT64 payloadType, PSessionDescription pSessionDescrip
...
+        isAsymmetryAllowed = (levelAsymmetry == 1);
+    }
+
+    if (!isProfileMatch && !isAsymmetryAllowed) {
+        return 0;
+    }
+
+    score += isAsymmetryAllowed ? 1 : 0;
+    score += isProfileMatch ? 1 : 0;
```
MushMal, line 393: remove all of this

src/source/PeerConnection/SessionDescription.c
lherman-cs: The linked document uses a cumulative sum to add "features".

src/source/PeerConnection/SessionDescription.c
lherman-cs: We need to store these booleans because we want to return 0 when the profile doesn't match and it requires a symmetric profile.

MushMal: You don't have to return 0. Remember we talked about weighted normalized multi-dimentional discrete vector? In this case, you add a lot more weight to what's important (after normalizing the units). So, say if we have a profile match then we add 100. When we have a level match we add just 1. At the end of the day, the caller can make a choice of using the score or not given it's absolute vector value.

lherman-cs: Oh right, that makes sense. I didn't realize that using more weight on the profile match can be used for this case. I'll update it

lherman-cs: I found that I don't need to add weight to profile or asymmetry dimensions as they're as important as packetization mode.

MushMal (approved):
Indeed. We can make the scoring to be far more intelligent and take into account scaling and weighting and calculate the absolute vector values but for now, you are doing something similar as all of the scaling and the weights = 1 and there is no need to calculate the vector length but simply sum the score up. Just so people are on the board. The idea is to the square of the absolute value of the vector in the discrete normalized weighted multi-dimensional space.

e.g. 

UINT64 score = 0;

// feature 1
if (feature 1) {
featureValue = value * scalingFactor * weightingFactor;
score += featueValue ^ 2;
}
... for other features which are dimensions.

return score;

tst/SdpApiTest.cpp
```diff
@@ -962,8 +948,13 @@ a=rtpmap:102 H264/90000
 
 TEST_P(SdpApiTest_SdpMatch, populateSingleMediaSection_TestH264Fmtp)
 {
-    const std::string offer = sdpOffer();
-    PCHAR sdp = const_cast<PCHAR>(offer.c_str());
+    UINT64 sdpSize;
+    PCHAR pSdp;
+
+    ASSERT_EQ(STATUS_SUCCESS, readFile((PCHAR) filename(), TRUE, NULL, &sdpSize));
```
MushMal, line 954: It's fine but for more or less knowns sizes you could have used a static buffer instead of getting the size, alloc-ing and freeing. For the test it's more than fine.

MushMal (approved):
Let's see if this fixes the issue

MushMal (approved):

Issue #957, opened 2020-11-17
[Question] Decode rate of 0 fps every 5 - 10 minutes


suggestedfixes:
**Describe the bug**
We've been observing 30-60 seconds of 0 decode rate every 5 to 10 minutes on armv7. To allow AWS to help us as much as possible, we've tried to reproduce the problem on a common platform.
We've reproduced this on both TURN and STUN for armv7 and TURN for Mac. We are still trying to come up with better reproducible steps for STUN for Mac.

**SDK version number**
armv7: v1.3.0
mac: v1.4.0

**To Reproduce**
1. build kvs gst sample on mac.
2. use the stock webrtc viewer webpage.
3. force turn, if we can use a remote viewer, it is more reproducible.
- we can reproduce this with STUN with armv7.
- we are still trying to reproduce this with STUN with mac.
4. monitor inbound video stats graph of chrome://webrtc-internals
5. after 5 to 10 minutes, we observe a gap of 30 to 60 seconds of 0 frame decoded per second, after which we see the browser decoder recovers.

**Expected behavior**
The browser can decode.

**Screenshots**
Here are screenshots of when this is happening
The negotiated profile id is level-asymmetry-allowed=1;packetization-mode=0;profile-level-id=42e01f for Windows 7
![0decodeMac](https://user-images.githubusercontent.com/61390265/99445097-f5ad8e00-28d1-11eb-8edb-a42819b96eb2.PNG)
![0decode3](https://user-images.githubusercontent.com/61390265/99445103-f8a87e80-28d1-11eb-8f27-2fa850b30d68.PNG)
![0decode4](https://user-images.githubusercontent.com/61390265/99445108-fb0ad880-28d1-11eb-93a7-4791721a10e0.PNG)

**Desktop (Viewer):**
 - OS: Win 7, (also reproduced with win 10 with armv7)
 - Browser: Chrome
 - Version: 86.0.4240.111 (Official Build) (64-bit)

**Additional context**
For armv7:
- Currently, for armv7, H264, we can only test with main profile and high profile. So there is a possibility where there is a profile mismatch between negotiated browser decoder profile and device encoder profile. Currently, on the device (master) side, we prioritize selecting 42e01f decoder with packetization mode of 0. We might have to ask our device vendor to enable baseline-profile. With packetization mode of 1 selected in the SDP answer, some browsers cannot decode properly.
- We've tried PLI handling, which did not help.
[...]

PR #968, opened 2020-11-19
Change Sample App timestamps to UINT64 to prevent overflow

suggestedfixes:
*Issue #957 *

*Description of changes:*
Current Samples have UINT32 timestamps, this means that they will easily overflow in a few minutes given 10,000,000 incremented per second. When this overflow occurs, the resulting RTP frame timestamp will decrement in value, and some decoder will fail to decode RTP frames for at least a minute. By changing them to UINT64, we delay this overflow to at least a few hours.


MushMal (approved):
Thanks!

PR #969, opened 2020-11-20
Fix gperftools

ycyang1229:
*Issue #, if available:*
Fix the abnormal behavior of gperftools.
*Description of changes:*
I can not run this sdk with gperftools correctly, since this sdk did not add the correct link option for gcc.
The similar issue is as below. 
https://stackoverflow.com/questions/26636570/using-pprof-with-gperftools-resulting-in-curl-error

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -27,7 +27,7 @@ option(ADDRESS_SANITIZER "Build with AddressSanitizer." OFF)
 option(MEMORY_SANITIZER "Build with MemorySanitizer." OFF)
 option(THREAD_SANITIZER "Build with ThreadSanitizer." OFF)
 option(UNDEFINED_BEHAVIOR_SANITIZER "Build with UndefinedBehaviorSanitizer." OFF)
-option(LINK_PROFILER "Link gperftools profiler" OFF) 
+option(LINK_PROFILER "Link gperftools profiler" OFF)
 
 if(NOT WIN32)
 CHECK_INCLUDE_FILES(ifaddrs.h       KVSWEBRTC_HAVE_IFADDRS_H)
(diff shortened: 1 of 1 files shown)


MushMal:
Do we have more info on the issue itself? How are you hitting this and why does this not (?) repro in the Travis runs we have.

Most importantly, you are adding the option to ALL of the platforms without removing this which will likely result in linkage of a lot of other libraries.

ycyang1229:
> Do we have more info on the issue itself? How are you hitting this and why does this not (?) repro in the Travis runs we have.
> 
> Most importantly, you are adding the option to ALL of the platforms without removing this which will likely result in linkage of a lot of other libraries.

I did not do anything special, and only enabled the option(LINK_PROFILER) in the cmakelist. https://github.com/awslabs/amazon-kinesis-video-streams-webrtc-sdk-c/commit/2a37ad95019221bcf379946725cf27112ac3fce6. And the error log is similar as the stack overflow. There are two option using gperftools, and this sdk does not change code for gperftools. So gcc may ignore the unused libraries. 
Yes, I do understand the link option is applies to every platforms including cmake targets in the cmakelist, but I did not see any constraint in the cmakelist. And I also did not see anything in https://github.com/awslabs/amazon-kinesis-video-streams-webrtc-sdk-c/blob/master/.travis.yml. Please let me know if I misunderstand it. thanks.

ycyang1229:
The command for gperftools I used on my ubuntu is as below .
$ CPUPROFILE=webrtc.prof ./kvsWebrtcClientMaster channel-name
$ pprof --web ./kvsWebrtcClientMaster webrtc.prof 
If you think it is relating to the environment issue, please let me know and I will close this one. thanks.

lherman-cs (approved):
LGTM

MushMal:
Sounds good. Merging.

PR #979, opened 2020-11-26
Check NULL for pFrame in writeFrame

codingspirit:

*Issue #, if available:*

*Description of changes:*

Passing NULL `pFrame` to `writeFrame` might cause segfault


MushMal (approved):
Wow. I am surprised we have no API level negative test coverage. 

The change looks good and is absolutely appropriate but I would like to have a test coverage for this if possible. @lherman-cs can you please look into adding tests to this and all of our public APIs if they have no coverage. A high level sanity checks are mandatory.

PR #984, opened 2020-11-30
Optimize startup latency from ~9s to ~80ms

lherman-cs:
The startup latency is measured with the following procedures:
  * Modified the JS example, https://awslabs.github.io/amazon-kinesis-video-streams-webrtc-sdk-js/examples/index.html,
    which basically listen to "loadeddata" event which basically it lets us
    know when the video starts flowing on the UI. When this event happens,
    it send a data channel message to the Webrtc C SDK
  * Modified the Webrtc C sample to measure the duration between connected
    to the time when it receives the data channel message from the UI

In my test, I got ~9 s delay. After the fix, the delay became ~80 ms.

The issue was that we started sending frames when the connection was still
not ready. Consequently, the first I-frame was lost, and the receiver needs
to wait until the next I-frame. So, for the delay itself, it'll be varied
depending on the configured keyframe interval. With our stock sample, I
calculated the delay to be ~8 s, which is matched to the observed delay.

The details about the sample video:
  * Keyframe interval is 250
  * Frame lost: first 16 frames
  * Frame per second: 30
  * Initial delay: (Keyframe interval - frame lost) / frame per second = 7.8 s

### DIFF
--- a/samples/Common.c
+++ b/samples/Common.c
@@ -31,22 +31,39 @@ VOID onDataChannel(UINT64 customData, PRtcDataChannel pRtcDataChannel)
 
 VOID onConnectionStateChange(UINT64 customData, RTC_PEER_CONNECTION_STATE newState)
 {
-    PSampleStreamingSession pSampleStreamingSession = (PSampleStreamingSession) customData;
-    PSampleConfiguration pSampleConfiguration = pSampleStreamingSession->pSampleConfiguration;
     STATUS retStatus = STATUS_SUCCESS;
+    PSampleStreamingSession pSampleStreamingSession = (PSampleStreamingSession) customData;
+
+    CHK(pSampleStreamingSession != NULL && pSampleStreamingSession->pSampleConfiguration != NULL, STATUS_INTERNAL_ERROR);
+
+    PSampleConfiguration pSampleConfiguration = pSampleStreamingSession->pSampleConfiguration;
     DLOGI("New connection state %u", newState);
 
-    if (newState == RTC_PEER_CONNECTION_STATE_FAILED || newState == RTC_PEER_CONNECTION_STATE_CLOSED ||
-        newState == RTC_PEER_CONNECTION_STATE_DISCONNECTED) {
-        ATOMIC_STORE_BOOL(&pSampleStreamingSession->terminateFlag, TRUE);
-        CVAR_BROADCAST(pSampleConfiguration->cvar);
-    } else if (newState == RTC_PEER_CONNECTION_STATE_CONNECTED) {
-        ATOMIC_STORE_BOOL(&pSampleConfiguration->connected, TRUE);
-        CVAR_BROADCAST(pSampleConfiguration->cvar);
-        if (STATUS_FAILED(retStatus = logSelectedIceCandidatesInformation(pSampleStreamingSession))) {
-            DLOGW("Failed to get information about selected Ice candidates: 0x%08x", retStatus);
-        }
+    switch (newState) {
+        case RTC_PEER_CONNECTION_STATE_CONNECTED:
+            ATOMIC_STORE_BOOL(&pSampleConfiguration->connected, TRUE);
+            CVAR_BROADCAST(pSampleConfiguration->cvar);
+            if (STATUS_FAILED(retStatus = logSelectedIceCandidatesInformation(pSampleStreamingSession))) {
+                DLOGW("Failed to get information about selected Ice candidates: 0x%08x", retStatus);
+            }
+            break;
+        case RTC_PEER_CONNECTION_STATE_FAILED:
+            // explicit fallthrough
+        case RTC_PEER_CONNECTION_STATE_CLOSED:
+            // explicit fallthrough
+        case RTC_PEER_CONNECTION_STATE_DISCONNECTED:
+            ATOMIC_STORE_BOOL(&pSampleStreamingSession->terminateFlag, TRUE);
+            CVAR_BROADCAST(pSampleConfiguration->cvar);
+            // explicit fallthrough
+        default:
+            ATOMIC_STORE_BOOL(&pSampleConfiguration->connected, FALSE);
+            CVAR_BROADCAST(pSampleConfiguration->cvar);
+            break;
     }
+
+CleanUp:
+
+    CHK_LOG_ERR(retStatus);
 }
 
 STATUS signalingClientStateChanged(UINT64 customData, SIGNALING_CLIENT_STATE state)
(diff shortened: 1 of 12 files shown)


tst/PeerConnectionFunctionalityTest.cpp
```diff
@@ -599,6 +599,115 @@ TEST_F(PeerConnectionFunctionalityTest, connectTwoPeersExpectFailureBecauseNoCan
...
+
+    auto onFrameHandler = [](UINT64 customData, PFrame pFrame) -> void {
+        UNUSED_PARAM(pFrame);
+        if (pFrame->frameData[0] == 1) {
+            ATOMIC_STORE_BOOL((PSIZE_T) customData, 1);
+        }
+    };
+    EXPECT_EQ(transceiverOnFrame(answerVideoTransceiver, (UINT64) &seenFirstFrame, onFrameHandler), STATUS_SUCCESS);
+
+    auto onICECandidateHdlr = [](UINT64 customData, PCHAR candidateStr) -> void {
```
lherman-cs, line 646: I can't use `connectTwoPeers` because the test is trying to exercise the `CONNECTED` state without knowing the other peer's state.

samples/kvsWebRTCClientMaster.c
```diff
@@ -199,6 +199,12 @@ PVOID sendVideoPackets(PVOID args)
         goto CleanUp;
     }
 
+    MUTEX_LOCK(pSampleConfiguration->sampleConfigurationObjLock);
```
lherman-cs, line 202: This fix doesn't reduce the startup latencies for viewers other than the first. To make this work for other viewers, we need to buffer all of the frames from the latest keyframe to the current frame. This will add complexities to the sample, and, thus, it's explicitly ignored for the sake of an example.

MushMal: This would be a separate feature by itself. Similar to pre-buffering work for the producer SDK. We can add a work item for the samples to pre-buffer from key-frame

samples/Common.c
```diff
@@ -32,14 +32,17 @@ VOID onDataChannel(UINT64 customData, PRtcDataChannel pRtcDataChannel)
 VOID onConnectionStateChange(UINT64 customData, RTC_PEER_CONNECTION_STATE newState)
 {
     PSampleStreamingSession pSampleStreamingSession = (PSampleStreamingSession) customData;
```
MushMal, line 34: Would be nice to do some checks here like

CHK(pSampleStreamingSession != NULL && pSampleStreamingSession->pSampleConfiguration != NULL, STATUS_INTERNAL_ERROR);

samples/Common.c
```diff
@@ -32,14 +32,17 @@ VOID onDataChannel(UINT64 customData, PRtcDataChannel pRtcDataChannel)
...
     STATUS retStatus = STATUS_SUCCESS;
     DLOGI("New connection state %u", newState);
 
     if (newState == RTC_PEER_CONNECTION_STATE_FAILED || newState == RTC_PEER_CONNECTION_STATE_CLOSED ||
         newState == RTC_PEER_CONNECTION_STATE_DISCONNECTED) {
         ATOMIC_STORE_BOOL(&pSampleStreamingSession->terminateFlag, TRUE);
-        CVAR_BROADCAST(pSampleStreamingSession->pSampleConfiguration->cvar);
+        CVAR_BROADCAST(pSampleConfiguration->cvar);
     } else if (newState == RTC_PEER_CONNECTION_STATE_CONNECTED) {
+        ATOMIC_STORE_BOOL(&pSampleConfiguration->connected, TRUE);
```
MushMal, line 44: Consider re-setting this on other state transitions. This also gets implicitly initialized which is bad. It's safer to initialize variables (especially BOOLs) explicitly instead of relying on calloc for zeroing the struct allocation out.

samples/kvsWebRTCClientMaster.c
MushMal, line 202: There are couple of issues with this approach.

1) This block is replicated for audio and video.
2) This only works for the stock master and not GStreamer based master.

Recommend the following.

The sample applications spin up two threads per peer connection to mimmic the media pipeline. 

1) The thread entry routine which is set in the sample configuration which is set by the appropriate sample should be wrapped by a common routine that handles things like awaiting for the connected state (this block) before calling directly the actual routine.
2) Longer term - spin a single thread that will service all of audio and video for all peer connections.

src/source/PeerConnection/PeerConnection.c
```diff
@@ -546,8 +557,16 @@ VOID onDtlsStateChange(UINT64 customData, RTC_DTLS_TRANSPORT_STATE newDtlsState)
 
     pKvsPeerConnection = (PKvsPeerConnection) customData;
 
-    if (newDtlsState == CLOSED) {
-        changePeerConnectionState(pKvsPeerConnection, RTC_PEER_CONNECTION_STATE_CLOSED);
+    switch (newDtlsState) {
+        case CONNECTED:
```
MushMal, line 561: BTW, can we change these enums PLEASE. We need to prefix these with RTC_DTLS_TRANSPORT_STATE like

RTC_DTLS_TRANSPORT_STATE_CONNECTED

I am not sure how we pushed this in. There are real chances for naming collisions. This is an internal API and we can change those just fine.

samples/Common.c
lherman-cs: Addressed

samples/Common.c
lherman-cs: Addressed.

samples/kvsWebRTCClientMaster.c
lherman-cs: I've created another routine to spin up the video and audio thread. I agree that we should have a single thread to serve all audio and video. But, this will be addressed in a different PR as it is outside the scope.

samples/Common.c
```diff
@@ -122,6 +142,36 @@ STATUS handleAnswer(PSampleConfiguration pSampleConfiguration, PSampleStreamingS
...
+
+    if (pSampleConfiguration->videoSource != NULL) {
+        THREAD_CREATE(&videoSenderTid, pSampleConfiguration->videoSource, (PVOID) pSampleConfiguration);
+    }
+
+    if (pSampleConfiguration->audioSource != NULL) {
+        THREAD_CREATE(&audioSenderTid, pSampleConfiguration->audioSource, (PVOID) pSampleConfiguration);
+    }
+
+    if (videoSenderTid != INVALID_TID_VALUE) {
```
lherman-cs, line 164: Having the sample to handle only `mediaSenderTid` will allow us to expand this in the future to use 1 thread to serve all video and audio to all peer connections.

src/source/PeerConnection/PeerConnection.c
lherman-cs: Addressed

MushMal (commented):
A comment about not blocking forever for the connected state as we can have a shutdown prior the connection established.

Another thought is that now we end up with 3 threads per peer connection.  We should follow-up with a fix to remove the audio and video threads soon.

samples/Common.c
```diff
@@ -122,6 +142,36 @@ STATUS handleAnswer(PSampleConfiguration pSampleConfiguration, PSampleStreamingS
     return retStatus;
 }
 
+PVOID mediaSenderRoutine(PVOID customData)
+{
+    PSampleConfiguration pSampleConfiguration = (PSampleConfiguration) customData;
+    TID videoSenderTid, audioSenderTid;
+
+    MUTEX_LOCK(pSampleConfiguration->sampleConfigurationObjLock);
+    while (!ATOMIC_LOAD_BOOL(&pSampleConfiguration->connected)) {
```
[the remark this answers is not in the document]
lherman-cs: That's a good point, I'll add that. There are only 3 threads for ALL peer connections: media, video, and audio. This check will guarantee this behavior, https://github.com/awslabs/amazon-kinesis-video-streams-webrtc-sdk-c/pull/984/files#diff-62548a53c1d33f3447edd43f2bfaa355c21911834766f96b97f57d02af7132ebR204-R205.

MushMal (approved):

MushMal (approved):

suggestedfixes:
@lherman-cs Have you considered the case where the 2nd to 10th peer joins? On average, they get around 4s latency correct?

lherman-cs:
@suggestedfixes yes, it was mentioned in this comment, https://github.com/awslabs/amazon-kinesis-video-streams-webrtc-sdk-c/pull/984#discussion_r532985087. To make this work, the application level needs to keep a buffer from the last keyframe to the current frame so that when there's a new viewer, it'll need to push all of the frames starting from the last keyframe, NOT from the current frame. From the SDK, we can guarantee that the application data can start flowing after PeerConnection's state changes to CONNECTED (this is the fix that I made).

I didn't implement multiviewers optimization in the sample, but rather put it as an explanatory comment since this change will overcomplicate the sample.

PR #990, opened 2020-12-04
ICE configuration refresh functionality changes

MushMal:
* Removing async ICE functionality to avoid issues with low footprint devices and parallel execution. This also has no benefits for "master" scenario. Embedded WebRTC based viewers might incur some latency hits as for the "viewer" the ICE configuration is on the "hot" path.
* Removing periodic ICE refresh. Now, it's done on access if the ICE hasn't been retrieved or whenever the ICE credentials have expired.
* Removing timer queue from the signaling client as it's no longer needed.
* The existing state machinery is unchanged and the change is very targeted and limited.
* Extra test coverage

*Issue #, if available:*

*Description of changes:*


hassanctech (approved):
lgtm

disa6302 (approved):

Issue #995, opened 2020-12-08
Crash occurs in closePeerConnection


liutuzhao:
**Logging**
[crash-in-closePeerConnection.zip](https://github.com/awslabs/amazon-kinesis-video-streams-webrtc-sdk-c/files/5657686/crash-in-closePeerConnection.zip)

**Describe the bug**
Doing two livestreming session at the same time .  When 1 session detected  broken in callbak "onConnectionStateChange" and the "terminateFlag" set as true. Another thread will check each session status and then free this broken session, found crash in  SDK  function "closePeerConnection ". Back trace as follows:

(gdb) where
#0  0x00504fa4 in pthread_mutex_lock ()
#1  0x00319500 in socketConnectionClosed ()
#2  0x0030f92c in connectionListenerRemoveAllConnection ()
#3  0x003108a0 in iceAgentShutdown ()
#4  0x002d163c in closePeerConnection ()
#5  0x0005a1ac in freeSampleStreamingSession ()
#6  0x00048a10 in CWebRTCClientMaster::SessionCleanupCheck(CQVMessageT*, unsigned int, unsigned int&) ()
#7  0x002ab82c in CQVThreadWorker::OnPolling(unsigned int&) ()
#8  0x002ac954 in CQVThreadWorker::OnThread() ()
#9  0x002aba58 in CQVThread::ThreadProc(void*) ()
#10 0x00503904 in start_thread ()
#11 0x0051cd20 in clone ()

**SDK version number**
V1.4.0

**Open source building**
default config in SDK

MushMal:
@liutuzhao this is not the stock application and the issue is not actionable without debug symbols and local variables. We will only look at crashes with stock samples. 

Please debug this further on your own. Please pull us in if you can pinpoint the actual crash in the SDK or the stock sample applications. 

As the stack trace does not correspond to the sample application, I am not sure what's causing the crash.

Removing "bug" tags.

MushMal:
Any updates? Have you been able to reproduce this with stock samples?

liutuzhao:
> Any updates? Have you been able to reproduce this with stock samples?

We're trying the  Alexa's pull request #996 and your pull request #1001 on our camera.
We dis not find crash  at the moment. We will keep testing for several days and if no relate crashe occurs again, we can close this issue.

MushMal:
[...]

PR #996, opened 2020-12-08
socketConnectionClosed: Check mutex before lock/unlock

codingspirit:

*Issue #, if available:*
#995 
*Description of changes:*
socketConnectionClosed: Check mutex before lock/unlock

### DIFF
--- a/src/source/Ice/SocketConnection.c
+++ b/src/source/Ice/SocketConnection.c
@@ -236,13 +236,17 @@ STATUS socketConnectionClosed(PSocketConnection pSocketConnection)
 
     CHK(pSocketConnection != NULL, STATUS_NULL_ARG);
     CHK(!ATOMIC_LOAD_BOOL(&pSocketConnection->connectionClosed), retStatus);
-    MUTEX_LOCK(pSocketConnection->lock);
+    if (IS_VALID_MUTEX_VALUE(pSocketConnection->lock)) {
+        MUTEX_LOCK(pSocketConnection->lock);
+    }
     DLOGD("Close socket %d", pSocketConnection->localSocket);
     ATOMIC_STORE_BOOL(&pSocketConnection->connectionClosed, TRUE);
     if (pSocketConnection->pTlsSession != NULL) {
         tlsSessionShutdown(pSocketConnection->pTlsSession);
     }
-    MUTEX_UNLOCK(pSocketConnection->lock);
+    if (IS_VALID_MUTEX_VALUE(pSocketConnection->lock)) {
+        MUTEX_UNLOCK(pSocketConnection->lock);
+    }
 
 CleanUp:
 


src/source/Ice/SocketConnection.c
```diff
@@ -236,13 +236,17 @@ STATUS socketConnectionClosed(PSocketConnection pSocketConnection)
 
     CHK(pSocketConnection != NULL, STATUS_NULL_ARG);
     CHK(!ATOMIC_LOAD_BOOL(&pSocketConnection->connectionClosed), retStatus);
-    MUTEX_LOCK(pSocketConnection->lock);
+    if (IS_VALID_MUTEX_VALUE(pSocketConnection->lock)) {
```
MushMal, line 239: There are two cases when this could happen.

1) If the construction of the object is incomplete - this is not the case as we check for the mutex successful creation in LN21 so we teardown on error. 
2) If your code accesses an already collected/freed SocketConnection object

You should identify the root cause of the second issue in your application and understand why something is attempting to access a freed object.

MushMal (commented):
I would recommend to reproduce this issue on the stock master for the referenced issue. There could be a number of root causes starting from buffer overflow and memory corruption to a bad teardown sequence to double free-ing. Try to follow the exact same steps as we have in the sample application.

src/source/Ice/SocketConnection.c
codingspirit: If it's case 2, it will goto CleanUp when do `CHK(pSocketConnection != NULL, STATUS_NULL_ARG);` at LN237. I was thinking there might be one scenario:
1. At https://github.com/awslabs/amazon-kinesis-video-streams-webrtc-sdk-c/blob/ec0d7844aaa81f062218c63d829d9fe1d64074be/src/source/Ice/SocketConnection.c#L70, we will free lock
2. If https://github.com/awslabs/amazon-kinesis-video-streams-webrtc-sdk-c/blob/ec0d7844aaa81f062218c63d829d9fe1d64074be/src/source/Ice/SocketConnection.c#L77 failed, we will skip `MEMFREE(pSocketConnection);` and `*ppSocketConnection = NULL;`

This might cause `lock` is invalid but `pSocketConnection != NULL` ?

MushMal: Oh wow.. Well, on the teardown path we won't be calling this over again to hit this case. However, this is a serious issue you've spotted. None of our "destructors" should do these jumps. The destruction is a best effort approach as the destructor might be working on a partially assembled object. LN 77 should be fixed ASAP. Will send a fix immediately

MushMal: Please check this out: https://github.com/awslabs/amazon-kinesis-video-streams-webrtc-sdk-c/pull/1001

PR #999, opened 2020-12-08
ICE server configuration sent through an Offer

MushMal:
*Issue #, if available:*

*Description of changes:*


src/source/Signaling/LwsApiCalls.c
```diff
@@ -1449,14 +1468,55 @@ STATUS sendLwsMessage(PSignalingClient pSignalingClient, PCHAR pMessageType, PCH
...
+            encodedUris[0] = '\0';
+            for (uriCount = 0; uriCount < pSignalingClient->iceConfigs[iceCount].uriCount; uriCount++) {
+                STRCAT(encodedUris, "\"");
+                STRCAT(encodedUris, pSignalingClient->iceConfigs[iceCount].uris[uriCount]);
+                STRCAT(encodedUris, "\",");
+            }
+
+            // remove the last comma
+            urisLen = STRLEN(encodedUris);
+            encodedUris[--urisLen] = '\0';
```
hassanctech, line 1492: If `pSignalingClient->iceConfigs[iceCount].uriCount` is `0` then we will try to access `encodedUris[-1]`

Is it possible for `pSignalingClient->iceConfigs[iceCount].uriCount` to be `0`, should we add validation for this and/or add if/else block around LN1491/1492 to only enter if `uriCount > 0`

MushMal: This would not be the case as this is an internal structure that gets set only after being validated. This said, it's a good idea to validate. Will add a validation earlier in the if clause.

src/source/Signaling/LwsApiCalls.c
```diff
@@ -1627,45 +1693,85 @@ STATUS receiveLwsMessage(PSignalingClient pSignalingClient, PCHAR pMessage, UINT
...
-                if (compareJsonString(pMessage, &tokens[i], JSMN_STRING, (PCHAR) "correlationId")) {
-                    strLen = (UINT32)(tokens[i + 1].end - tokens[i + 1].start);
-                    CHK(strLen <= MAX_CORRELATION_ID_LEN, STATUS_INVALID_API_CALL_RETURN_JSON);
-                    STRNCPY(pSignalingMessageWrapper->receivedSignalingMessage.signalingMessage.correlationId, pMessage + tokens[i + 1].start,
-                            strLen);
-                    pSignalingMessageWrapper->receivedSignalingMessage.signalingMessage.correlationId[MAX_CORRELATION_ID_LEN] = '\0';
+        } else if (!parsedStatusResponse && compareJsonString(pMessage, &tokens[i], JSMN_STRING, (PCHAR) "statusResponse")) {
+            parsedStatusResponse = TRUE;
+            i++;
+        } else if (parsedStatusResponse && compareJsonString(pMessage, &tokens[i], JSMN_STRING, (PCHAR) "correlationId")) {
```
disa6302, line 1699: Can we add the subsequent blocks this way? 

```
if (parsedStatusResponse) {
   if(compareJsonString....) {
   }
   else if (compareJsonString...) {
   }
   else if (compareJsonString...) {
   }
}
```

MushMal: won't work really as the JSON could contain response first and we will enter this clause for all of the consequent parsing for ICE which is wrong. I really need to think on how to simplify these JSON parsing routines further but not for this commit though

hassanctech (approved):
Lgtm

disa6302 (approved):

PR #1009, opened 2020-12-17
README: fix syntax error in IAM policy

codingspirit:

*Issue #, if available:*

*Description of changes:*


MushMal (approved):
We have had a similar issue with cpp documentation that got fixed. https://docs.aws.amazon.com/kinesisvideostreams/latest/dg/how-iot.html

Not sure how this was not.

thanks for taking care of it

PR #1010, opened 2020-12-17
Signaling: Fix possible null access in freeSignaling

codingspirit:
*Issue #, if available:*

*Description of changes:*

Possible null access when:

1. https://github.com/awslabs/amazon-kinesis-video-streams-webrtc-sdk-c/blob/b9e41d1c114acd50da3e72b05d486ebb1a0d2ff7/src/source/Signaling/Signaling.c#L132 failed, `pSignalingClient->lwsServiceLock` is invalid, goto CleanUp and call `freeSignaling`
2. https://github.com/awslabs/amazon-kinesis-video-streams-webrtc-sdk-c/blob/b9e41d1c114acd50da3e72b05d486ebb1a0d2ff7/src/source/Signaling/Signaling.c#L198

### DIFF
--- a/src/source/Signaling/Signaling.c
+++ b/src/source/Signaling/Signaling.c
@@ -97,8 +97,6 @@ STATUS createSignalingSync(PSignalingClientInfoInternal pClientInfo, PChannelInf
     creationInfo.ka_interval = SIGNALING_SERVICE_TCP_KEEPALIVE_PROBE_INTERVAL_IN_SECONDS;
     creationInfo.ws_ping_pong_interval = SIGNALING_SERVICE_WSS_PING_PONG_INTERVAL_IN_SECONDS;
 
-    CHK(NULL != (pSignalingClient->pLwsContext = lws_create_context(&creationInfo)), STATUS_SIGNALING_LWS_CREATE_CONTEXT_FAILED);
-
     ATOMIC_STORE_BOOL(&pSignalingClient->clientReady, FALSE);
     ATOMIC_STORE_BOOL(&pSignalingClient->shutdown, FALSE);
     ATOMIC_STORE_BOOL(&pSignalingClient->connected, FALSE);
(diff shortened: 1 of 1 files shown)


src/source/Signaling/Signaling.c
```diff
@@ -203,10 +203,14 @@ STATUS freeSignaling(PSignalingClient* ppSignalingClient)
     terminateOngoingOperations(pSignalingClient, TRUE);
 
     if (pSignalingClient->pLwsContext != NULL) {
-        MUTEX_LOCK(pSignalingClient->lwsServiceLock);
+        if(IS_VALID_MUTEX_VALUE(pSignalingClient->lwsServiceLock)) {
```
MushMal, line 206: If the pLwsContext != NULL then the mutex is not null

codingspirit: But https://github.com/awslabs/amazon-kinesis-video-streams-webrtc-sdk-c/blob/b9e41d1c114acd50da3e72b05d486ebb1a0d2ff7/src/source/Signaling/Signaling.c#L100 `pLwsContext` is created before `lwsServiceLock`, may I know how can we guarantee mutex is not null when `(pSignalingClient->pLwsContext != NULL)`

MushMal: You are absolutely correct and I am wrong. I confused with the lws call object we create perform the actual les calls. I think a cleaner solution perhaps is to move the lws context creation below the mutex and other object creations. Just above the stepping of the signaling state machine. What do you think?

codingspirit: Move the lws context creation sounds more reasonable than this commit, I will update it, thanks

codingspirit: ready to merge, please review again.

MushMal (approved):

MushMal:
Do you mind rebasing on top of the latest master branch? There is a merge conflict.

codingspirit:
Conflict resolved

PR #1032, opened 2021-01-07
Move to linux from osx

disa6302:
*Issue #, if available:*

*Description of changes:*
- It looks like there are some issues with clang-format install on MacOS Sierra hosts that travisCI uses. Since, all we need to do is check formatting, the type of host does not matter. Linux is more stable and hence this PR moves clang-format check from OSX to Linux.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -31,10 +31,11 @@ matrix:
   include:
     #clang check
     - name: "clang-format Check"
-      os: osx
+      os: linux
       compiler: clang
       before_script:
-        - brew install clang-format
+        - sudo apt-get -q update
+        - sudo apt-get -y install clang-format
         - mkdir build && cd build && cmake .. -DCMAKE_BUILD_TYPE=Debug -DBUILD_TEST=TRUE
       script:
         - cd ..


.travis.yml
```diff
@@ -31,14 +31,15 @@ matrix:
...
       compiler: clang
       before_script:
-        - brew install clang-format
+        - sudo apt-get -q update
+        - sudo apt-get -y install clang-format
         - mkdir build && cd build && cmake .. -DCMAKE_BUILD_TYPE=Debug -DBUILD_TEST=TRUE
       script:
         - cd ..
-        - bash scripts/check-clang.sh
+        - ./scripts/check-clang.sh
```
hassanctech, line 42: Any reason for this change, there is a difference here in terms of permissions that need to be set on the file. (Some details here: https://unix.stackexchange.com/questions/136547/what-is-the-difference-between-running-bash-script-sh-and-script-sh). In any case not a blocking change just wanted to understand.

hassanctech (approved):

.travis.yml
disa6302: No. Will change it back to bash once CI passes. I misunderstood the reason for this travis job failure and changed to ./. Also, travis check is stuck at "waiting for status to be reported..." 😔

disa6302: Moved to bash. Merging.

Issue #1041, opened 2021-01-13
[BUG] iceAgentLogNewCandidate logging TCP when UDP


suggestedfixes:
https://github.com/awslabs/amazon-kinesis-video-streams-webrtc-sdk-c/blob/master/src/source/Ice/IceAgent.c#L2615

MushMal:
Seems like a bug

PR #1051, opened 2021-01-20
fix log message for new ice candidate protocol

hassanctech:
*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/src/source/Ice/IceAgent.c
+++ b/src/source/Ice/IceAgent.c
@@ -2604,17 +2604,30 @@ CleanUp:
 VOID iceAgentLogNewCandidate(PIceCandidate pIceCandidate)
 {
     CHAR ipAddr[KVS_IP_ADDRESS_STRING_BUFFER_LEN];
-    PCHAR protocol = "UDP";
+    PCHAR protocol = "UNKNOWN";
 
     if (pIceCandidate != NULL) {
         getIpAddrStr(&pIceCandidate->ipAddress, ipAddr, ARRAY_SIZE(ipAddr));
         if (pIceCandidate->iceCandidateType == ICE_CANDIDATE_TYPE_RELAYED) {
             if (pIceCandidate->pTurnConnection == NULL) {
                 protocol = "NA";
-            } else if (pIceCandidate->pTurnConnection->protocol == KVS_SOCKET_PROTOCOL_UDP) {
-                protocol = "TCP";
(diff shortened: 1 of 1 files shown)


hassanctech:
RE: https://github.com/awslabs/amazon-kinesis-video-streams-webrtc-sdk-c/issues/1041

src/source/Ice/IceAgent.c
```diff
@@ -2604,17 +2604,30 @@ STATUS iceAgentFatalError(PIceAgent pIceAgent, STATUS errorStatus)
 VOID iceAgentLogNewCandidate(PIceCandidate pIceCandidate)
 {
     CHAR ipAddr[KVS_IP_ADDRESS_STRING_BUFFER_LEN];
-    PCHAR protocol = "UDP";
+    PCHAR protocol = "UNKNOWN";
```
disa6302, line 2607: Should we maintain this to be NONE instead of UNKNOWN to align with the KVS_SOCKET_PROTOCOL_* enum?

hassanctech: Yeah I do that below in the switch.  This is for the case that it was none of those values, suppose we add a new enum or pIceCandidate was NULL or pIceCandidate->iceCandidateType ==!= ICE_CANDIDATE_TYPE_RELAYED

MushMal: We should be consistent. There is a usage of 3 terms in this function - NA, NONE and UNKNOWN. Please chose one. I would gravitate towards UNKNOWN

MushMal (approved):

disa6302 (approved):

src/source/Ice/IceAgent.c
hassanctech: They all have different meaning.

* NA means the turn connection was NULL (which makes sense that if turn connection is none, protocol is really NA)
* NONE is just propagating the already set protocol filed in the turn connection
* UNKNOWN specifically means that it was none of those other cases so we really do not know.

hassanctech: The whole point of debugging messages is to be as specific as possible, with the 3 distinct values we'll know more about the state of the variables as opposed to mashing them all together into one.

MushMal: Oh OK. I had approved it

PR #1056, opened 2021-01-27
Fixing interlocking issues and bugs with transaction store for srflx …

MushMal:
…candidates

*Issue #, if available:*

*Description of changes:*


hassanctech (commented):
Can we please get a test for this fix?  It might be easier to understand the change when I can see a test that would have previously have failed.

MushMal:
> Can we please get a test for this fix? It might be easier to understand the change when I can see a test that would have previously have failed.

It's not trivial and will require full integ test. Some of the changes are for the negative path. I am not sure what the root cause is for the case where srflx candidate was not being selected.

disa6302 (approved):
LGTM!

hassanctech (approved):

MushMal:
Just want to confirm that after multiple experiments, srflx candidate does get selected and used in case when only TURN and SRFLX are available.

Issue #1060, opened 2021-02-02
[BUG] Possible memory leak during MasterSample runtime


Nomidia:
If the max simultaneous streaming session count reached, the new connection will be discarded. However, the new ice candidates which will not be used during runtime will be stored in hash table.

```
2021-02-02 01:36:11 WARN    signalingMessageReceived(): Max simultaneous streaming session count reached.
2021-02-02 01:36:11 DEBUG   lwsWssCallbackRoutine(): Client is writable
2021-02-02 01:36:11 DEBUG   lwsWssCallbackRoutine(): Client receive {"messagePayload":"eyJjYW5kaWRhdGUiOiJjYW5kaWRhdGU6MjcwMjIzOTY3MCAxIHVkcCAyMTEzOTM3MTUxIDAwZGFlMjEzLTQ3NmQtNGJkYi05NmI2LTU2MDY5ZDJkMTc5My5sb2NhbCA1MDM3MiB0eXAgaG9zdCBnZW5lcmF0aW9uIDAgdWZyYWcgVGpLNiBuZXR3b3JrLWNvc3QgOTk5Iiwic2RwTWlkIjoiMSIsInNkcE1MaW5lSW5kZXgiOjF9","messageType":"ICE_CANDIDATE","senderClientId":"oph0j9oihlxssmzyummb"}
2021-02-02 01:36:12 DEBUG   lwsWssCallbackRoutine(): Client is writable
2021-02-02 01:36:12 DEBUG   lwsWssCallbackRoutine(): Client receive {"messagePayload":"eyJjYW5kaWRhdGUiOiJjYW5kaWRhdGU6ODQyMTYzMDQ5IDEgdWRwIDE2Nzc3Mjk1MzUgMTE2LjIyOC45LjY2IDUwMzcyIHR5cCBzcmZseCByYWRkciAwLjAuMC4wIHJwb3J0IDAgZ2VuZXJhdGlvbiAwIHVmcmFnIFRqSzYgbmV0d29yay1jb3N0IDk5OSIsInNkcE1pZCI6IjEiLCJzZHBNTGluZUluZGV4IjoxfQ==","messageType":"ICE_CANDIDATE","senderClientId":"oph0j9oihlxssmzyummb"}
2021-02-02 01:36:12 DEBUG   lwsWssCallbackRoutine(): Client is writable
[...]

PR #1061, opened 2021-02-02
Removing pending queue if session creation fails

MushMal:
Issue #1060 1060

*Description of changes:*


Nomidia:
Usually, ice candidates are received after sdp offer. This may not take effect.

MushMal:
> Usually, ice candidates are received after sdp offer. This may not take effect.

It is indeed possible to still receive ICE candidates after. We need to do some cleaning up in the overall cleanup loop

hassanctech (approved):

PR #1063, opened 2021-02-05
Add option to build aws dependencies while using system 3rd party deps

atyshka:
On Ubuntu 20.04, all necessary 3rd party dependencies are available from the package manager, while AWS common libs are not. This PR adds a separate option to build the AWS dependencies even if system dependencies are not built from source

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -9,6 +9,7 @@ project(KinesisVideoWebRTCClient LANGUAGES C)
 # User Flags
 option(ADD_MUCLIBC "Add -muclibc c flag" OFF)
 option(BUILD_DEPENDENCIES "Whether or not to build depending libraries from source" ON)
+option(BUILD_AWS_DEPENDENCIES "Whether or not to build depending AWS libraries from source" ON)
 option(USE_OPENSSL "Use openssl as crypto library" ON)
 option(USE_MBEDTLS "Use mbedtls as crypto library" OFF)
 option(BUILD_STATIC_LIBS "Build all libraries statically. (This includes third-party libraries.)" OFF)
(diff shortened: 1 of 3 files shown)


CMakeLists.txt
```diff
@@ -140,18 +141,27 @@ if(BUILD_DEPENDENCIES)
   if (LINK_PROFILER)
     build_dependency(gperftools)
   endif()
+endif()
+
+if(BUILD_AWS_DEPENDENCIES OR BUILD_DEPENDENCIES)
```
MushMal, line 146: Perhaps I am not understanding the change but as is it makes no sense to me. In case of specifying both of the flags, the same codepath will be executed.

The SDK cmake file is capable to simply using the pre-existing libraries by not specifying BUILD_DEPENDENCIES and supplying OPEN_SRC_INSTALL_PREFIX

atyshka: I understand I could do things that way, but if I have BUILD_DEPENDENCIES off and use OPEN_SRC_INSTALL_PREFIX, that requires building open source parts ahead of time. I would like the whole setup to be automated, so that I can use FetchContent to include this in a larger project and automatically build. Adding a variable helped accomplish this in my use case, but I understand if it's niche enough that you don't want to merge

MushMal: Understood. As you mentioned, this might not be generic enough to be included in the main repository. Some of our customers simply run two-stage builds specifying different parameters while others do indeed build the dependencies or fetch them from platform specific repositories ahead of the time to bake into the image.

src/include/com/amazonaws/kinesis/video/webrtcclient/Include.h
```diff
@@ -453,7 +453,7 @@ extern "C" {
 /**
  * Maximum length of signaling message
  */
-#define MAX_SIGNALING_MESSAGE_LEN (10 * 1024)
+#define MAX_SIGNALING_MESSAGE_LEN (10 * 4096)
```
MushMal, line 456: This is a hard limit in a sense that it is defined in our backend.

atyshka:
@MushMal My apologies I did not intend to put that commit on this branch. Also for the record it doesn't seem to be a hard backend limit. My purpose in increasing was due to requiring many transceivers, which was too large for the message size and threw errors. These errors are fixed by changing the message length and messages are successfully delivered.

MushMal:
> @MushMal My apologies I did not intend to put that commit on this branch. Also for the record it doesn't seem to be a hard backend limit. My purpose in increasing was due to requiring many transceivers, which was too large for the message size and threw errors. These errors are fixed by changing the message length and messages are successfully delivered.

Understood. Perhaps the limit has been bumped recently that I am unaware. I would suggest pushing a separate commit with this change only. Someone from Amazon should reply to it whether the limits have been bumped up and point to the actual limits document

PR #1069, opened 2021-02-12
Introducing certificate pre-generation in the samples to speed up pro…

MushMal:
…cessing on low-footprint devices

Fixed the stats timer logic
Fixed the mbedTLS cert copy logic
Other minor fixes

NOTE: Tested with:

*  both OpenSSL and mbedTLS
* With and without pre-gen certs
* different timings

On my local (very powerful) machine I get a speedup around 12 - 15ms.

*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/samples/kvsWebRTCClientMasterGstreamerSample.c
+++ b/samples/kvsWebRTCClientMasterGstreamerSample.c
@@ -338,6 +338,8 @@ INT32 main(INT32 argc, CHAR* argv[])
     STATUS retStatus = STATUS_SUCCESS;
     PSampleConfiguration pSampleConfiguration = NULL;
 
+    SET_INSTRUMENTED_ALLOCATORS();
+
     signal(SIGINT, sigintHandler);
 
     // do trickle-ice by default
@@ -472,6 +474,8 @@ CleanUp:
         }
     }
     printf("[KVS Gstreamer Master] Cleanup done\n");
+
+    RESET_INSTRUMENTED_ALLOCATORS();
 
     // https://www.gnu.org/software/libc/manual/html_node/Exit-Status.html
     // We can only return with 0 - 127. Some platforms treat exit code >= 128
--- a/src/source/Crypto/Dtls_mbedtls.c
+++ b/src/source/Crypto/Dtls_mbedtls.c
@@ -573,7 +573,7 @@ STATUS copyCertificateAndKey(mbedtls_x509_crt* pCert, mbedtls_pk_context* pKey, 
         case MBEDTLS_PK_ECDSA:
             pSrcECP = mbedtls_pk_ec(*pKey);
             pDstECP = mbedtls_pk_ec(pDst->privateKey);
-            CHK(mbedtls_ecp_group_copy(&pDstECP->grp, &pSrcECP->grp) && mbedtls_ecp_copy(&pDstECP->Q, &pSrcECP->Q) == 0 &&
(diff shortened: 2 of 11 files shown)


samples/kvsWebRTCClientMasterGstreamerSample.c
```diff
@@ -338,6 +338,8 @@ INT32 main(INT32 argc, CHAR* argv[])
     STATUS retStatus = STATUS_SUCCESS;
     PSampleConfiguration pSampleConfiguration = NULL;
 
+    SET_INSTRUMENTED_ALLOCATORS();
```
disa6302, line 341: Do we need to set these allocators since we set `-DINSTRUMENTED_ALLOCATORS` to OFF in `CMake`? There is no harm in adding it to the samples, but, just want to confirm.

disa6302 (approved):

samples/kvsWebRTCClientMasterGstreamerSample.c
MushMal: The macros evaluate to no-op if the compile define is not set. The idea here is to control the behavior by not modifying the code but rather providing compile time definition

MushMal: BTW, this is very useful to have. Our unit tests do have it and we caught mem leaks on negative paths whereas the address sanitizer reported a clean run.

MushMal: That's one thing I forgot to mention in the description. I tested with and without the instrumented mem allocators.

src/source/Ice/IceUtils.c
```diff
@@ -99,7 +99,7 @@ VOID transactionIdStoreRemove(PTransactionIdStore pTransactionIdStore, PBYTE tra
 
     for (i = pTransactionIdStore->earliestTransactionIdIndex, j = 0; j < pTransactionIdStore->maxTransactionIdsCount; ++j) {
         if (MEMCMP(transactionId, pTransactionIdStore->transactionIds + i * STUN_TRANSACTION_ID_LEN, STUN_TRANSACTION_ID_LEN) == 0) {
-            MEMSET(pTransactionIdStore->transactionIds + i * STUN_TRANSACTION_ID_LEN, STUN_TRANSACTION_ID_LEN, 0x00);
+            MEMSET(pTransactionIdStore->transactionIds + i * STUN_TRANSACTION_ID_LEN, 0x00, STUN_TRANSACTION_ID_LEN);
```
hassanctech, line 102: Yikes!!

src/source/Crypto/Dtls_mbedtls.c
```diff
@@ -573,7 +573,7 @@ STATUS copyCertificateAndKey(mbedtls_x509_crt* pCert, mbedtls_pk_context* pKey,
         case MBEDTLS_PK_ECDSA:
             pSrcECP = mbedtls_pk_ec(*pKey);
             pDstECP = mbedtls_pk_ec(pDst->privateKey);
-            CHK(mbedtls_ecp_group_copy(&pDstECP->grp, &pSrcECP->grp) && mbedtls_ecp_copy(&pDstECP->Q, &pSrcECP->Q) == 0 &&
+            CHK(mbedtls_ecp_group_copy(&pDstECP->grp, &pSrcECP->grp) == 0 && mbedtls_ecp_copy(&pDstECP->Q, &pSrcECP->Q) == 0 &&
```
hassanctech, line 576: Previously if the group copy was successful we would actually fail... nice catch!!

hassanctech (approved):

PR #1077, opened 2021-02-19
Display only messsage type and recipient client ID

disa6302:
*Issue #, if available:*

*Description of changes:*
While logging the payload that is sent from the embedded SDK, in case of sending an offer, we end up logging the credentials for the ICE Servers which is part of the OFFER payload. This PR fixes it by just logging necessary information (type of message being sent and recipient ID) to keep the information to a minimum.

### DIFF
--- a/src/source/Signaling/LwsApiCalls.c
+++ b/src/source/Signaling/LwsApiCalls.c
@@ -108,10 +108,7 @@ INT32 lwsHttpCallbackRoutine(struct lws* wsi, enum lws_callback_reasons reason, 
 
         case LWS_CALLBACK_RECEIVE_CLIENT_HTTP_READ:
             DLOGD("Received client http read: %d bytes", (INT32) dataSize);
-
-            if (logLevel > LOG_LEVEL_INFO && logLevel < LOG_LEVEL_SILENT) {
-                lwsl_hexdump_notice(pDataIn, dataSize);
-            }
+            lwsl_hexdump_debug(pDataIn, dataSize);
 
             if (dataSize != 0) {
                 CHK(NULL != (pLwsCallInfo->callInfo.responseData = (PCHAR) MEMALLOC(dataSize)), STATUS_NOT_ENOUGH_MEMORY);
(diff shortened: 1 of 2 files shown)


MushMal (approved):
Very cool. Are you going to handle the receive HTTP protocol in the receive message function?

disa6302:
> Very cool. Are you going to handle the receive HTTP protocol in the receive message function?

Do you mean `receiveLwsMessage`? If yes, the change is on the way.

src/source/Signaling/LwsApiCalls.c
```diff
@@ -1943,6 +1943,28 @@ STATUS getMessageTypeFromString(PCHAR typeStr, UINT32 typeLen, SIGNALING_MESSAGE
     return retStatus;
 }
 
+PCHAR getMessageTypeInString(SIGNALING_MESSAGE_TYPE messageType)
+{
+    ENTERS();
```
MushMal, line 1948: Remove

MushMal (approved):
Minor comment

MushMal (approved):

src/source/Signaling/LwsApiCalls.h
```diff
@@ -249,6 +250,7 @@ STATUS writeLwsData(PSignalingClient, BOOL);
 STATUS terminateLwsListenerLoop(PSignalingClient);
 STATUS receiveLwsMessage(PSignalingClient, PCHAR, UINT32);
 STATUS getMessageTypeFromString(PCHAR, UINT32, SIGNALING_MESSAGE_TYPE*);
+PCHAR getMessageTypeInString(SIGNALING_MESSAGE_TYPE);
```
hassanctech, line 253: It seems like the other APIs fill a user allocated buffer and return a length maybe we should do that for consistency?  Not really a big deal it looks like the use case for this method is only for printing so definitely a lot more "convenient" to return PCHAR as you have done.

hassanctech (approved):

src/source/Signaling/LwsApiCalls.h
disa6302: Yes. You are correct, since it is purely for logging purposes, I added a very simple function without any user allocated buffer.

Issue #1085, opened 2021-03-02
[BUG] JitterBuffer.c incorrect handling of late packet from previous frame


wreuven:
Line 2 and 3 in the following if statement makes no sense (e.g. comparing timestamp to sequence number) and potentially causes trouble.

```
 if (!pJitterBuffer->started ||
        (pJitterBuffer->lastPopTimestamp == pRtpPacket->header.sequenceNumber &&
         pJitterBuffer->lastRemovedSequenceNumber >= pRtpPacket->header.sequenceNumber)) {
        // Set to started and initialize the sequence number
        pJitterBuffer->started = TRUE;
        pJitterBuffer->lastRemovedSequenceNumber = UINT16_DEC(pRtpPacket->header.sequenceNumber);
    }
```

I think the intention is perhaps that a late packet arriving after the previous frame has already been popped which has a greater sequence number than the last removed sequence number should become the last removed sequence number. In which case the expression should read (see 2 changes):

```
 if (!pJitterBuffer->started ||
        (pJitterBuffer->lastPopTimestamp == pRtpPacket->header.timestamp &&
         pJitterBuffer->lastRemovedSequenceNumber < pRtpPacket->header.sequenceNumber)) {
        // Set to started and initialize the sequence number
        pJitterBuffer->started = TRUE;
        pJitterBuffer->lastRemovedSequenceNumber = UINT16_DEC(pRtpPacket->header.sequenceNumber);
    }
```

hassanctech:
Yikes! Looks like a bug, taking a look...

hassanctech:
I put up a PR for the timestamp comparison.  I'm not yet convinced of the second change, I need to figure out what the intention there was as well, @wreuven feel free to comment on the PR.

wreuven:
By the way, there seems to be other problems with this module.
Seems to assume that timestamps always start at zero and that is not the case by RTP standard.
Seems to assume that timestamps always increase when in fact they wrap around. 
There are a couple scary sounding TODO comments in the module.

hassanctech:
Fixed: https://github.com/awslabs/amazon-kinesis-video-streams-pic/pull/84/files#diff-d181ddf3c70af1265b8b311f3249b35e9e535d4441aa83d019c4bf2d433fd08cR619

PR #1088, opened 2021-03-03
fix bug comparing timestamp with sequence number

hassanctech:
*Issue #, if available:*

Issue #1085

*Description of changes:*

### DIFF
--- a/src/source/PeerConnection/JitterBuffer.c
+++ b/src/source/PeerConnection/JitterBuffer.c
@@ -85,7 +85,7 @@ STATUS jitterBufferPush(PJitterBuffer pJitterBuffer, PRtpPacket pRtpPacket, PBOO
     CHK(pJitterBuffer != NULL && pRtpPacket != NULL, STATUS_NULL_ARG);
 
     if (!pJitterBuffer->started ||
-        (pJitterBuffer->lastPopTimestamp == pRtpPacket->header.sequenceNumber &&
+        (pJitterBuffer->lastPopTimestamp == pRtpPacket->header.timestamp &&
          pJitterBuffer->lastRemovedSequenceNumber >= pRtpPacket->header.sequenceNumber)) {
         // Set to started and initialize the sequence number
         pJitterBuffer->started = TRUE;
--- a/src/source/PeerConnection/JitterBuffer.c
+++ b/src/source/PeerConnection/JitterBuffer.c
@@ -2,6 +2,8 @@
 
 #include "../Include_i.h"
 
+//forward declaration of internal functions
+
 STATUS createJitterBuffer(FrameReadyFunc onFrameReadyFunc, FrameDroppedFunc onFrameDroppedFunc, DepayRtpPayloadFunc depayRtpPayloadFunc,
                           UINT32 maxLatency, UINT32 clockRate, UINT64 customData, PJitterBuffer* ppJitterBuffer)
 {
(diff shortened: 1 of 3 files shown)


disa6302:
Oops! Is there a way to add unit test for this?

MushMal (commented):
Oh wow. This needs in-depth look at what those tests are doing and why this is not covered. I don't understand why and how this worked. Someone needs a much closer look at this piece (not currently at my desk)

hassanctech:
> Oh wow. This needs in-depth look at what those tests are doing and why this is not covered. I don't understand why and how this worked. Someone needs a much closer look at this piece (not currently at my desk)

Wow indeed... All the *JitterBuffer* tests FAIL after I make this single change but they PASS if we're comparing the timestamp to the sequence number.  I'm going to dig into this and see what's up here...

wreuven:
> > Oh wow. This needs in-depth look at what those tests are doing and why this is not covered. I don't understand why and how this worked. Someone needs a much closer look at this piece (not currently at my desk)
> 
> Wow indeed... All the _JitterBuffer_ tests FAIL after I make this single change but they PASS if we're comparing the timestamp to the sequence number. I'm going to dig into this and see what's up here...

I think you need both changes I suggested. Doing only this change means that a late packet will "rewind" the latest sequence to a point way earlier.

src/source/PeerConnection/JitterBuffer.c
```diff
@@ -238,6 +181,87 @@ STATUS jitterBufferPop(PJitterBuffer pJitterBuffer, BOOL bufferClosed)
...
+    // freeJitterBuffer is idempotent
+    CHK(*ppJitterBuffer != NULL, retStatus);
+
+    pJitterBuffer = *ppJitterBuffer;
+
+    jitterBufferInternalParse(pJitterBuffer, TRUE);
+    jitterBufferDropBufferData(pJitterBuffer, 0, MAX_SEQUENCE_NUM, 0);
+    hashTableFree(pJitterBuffer->pPkgBufferHashTable);
+
+    MEMFREE(*ppJitterBuffer);
```
hassanctech, line 202: I see this was there before but for this to be idempotent we need to do:

`SAFE_MEMFREE(*ppJitterBuffer);`

Which will set `*ppJitterBuffer` to NULL

src/source/PeerConnection/JitterBuffer.c
```diff
@@ -146,18 +72,36 @@ STATUS jitterBufferPop(PJitterBuffer pJitterBuffer, BOOL bufferClosed)
...
 
     if (pJitterBuffer->lastPushTimestamp > pJitterBuffer->maxLatency) {
-        earliestTimestamp = pJitterBuffer->lastPushTimestamp - pJitterBuffer->maxLatency;
+        earliestAllowedTimestamp = pJitterBuffer->lastPushTimestamp - pJitterBuffer->maxLatency;
     }
 
-    lastIndex = pJitterBuffer->lastRemovedSequenceNumber;
-    index = pJitterBuffer->lastRemovedSequenceNumber + 1;
+    lastIndex = pJitterBuffer->headSequenceNumber - 1;
+    index = pJitterBuffer->headSequenceNumber;
```
hassanctech, line 79: What happens if `headSequenceNumber` is `0`.

jdelapla: The startIndex will be 0, the lastIndex will be 65535.
It's an abuse of underflow that was also used in the previous iteration.

However I could add an if() clause using MAX_UINT16 if we want to avoid this underflow usage.

src/source/PeerConnection/JitterBuffer.c
jdelapla: Ah! I wasn't aware of that API change. I'll fix that.

hassanctech:
> > > Oh wow. This needs in-depth look at what those tests are doing and why this is not covered. I don't understand why and how this worked. Someone needs a much closer look at this piece (not currently at my desk)
> > 
> > 
> > Wow indeed... All the _JitterBuffer_ tests FAIL after I make this single change but they PASS if we're comparing the timestamp to the sequence number. I'm going to dig into this and see what's up here...
> 
> I think you need both changes I suggested. Doing only this change means that a late packet will "rewind" the latest sequence to a point way earlier.

@wreuven Not sure if you're following but we did some rework to all of this and the change should be going in very soon.

src/source/PeerConnection/JitterBuffer.c
hassanctech: I'm fine with this.

hassanctech (commented):
LGTM.  I can't approve since I opened the PR initially :(

disa6302 (approved):
LGTM

PR #1103, opened 2021-03-17
Remove unnecessary logging

disa6302:
*Issue #, if available:*

*Description of changes:*
Since we do not support handling multiple receiver/sender reports, we do not need to log that we do not support this since it causes lot of noise.

### DIFF
--- a/src/source/PeerConnection/Rtcp.c
+++ b/src/source/PeerConnection/Rtcp.c
@@ -60,7 +60,6 @@ static STATUS onRtcpSenderReport(PRtcpPacket pRtcpPacket, PKvsPeerConnection pKv
 
     if (pRtcpPacket->payloadLength != RTCP_PACKET_SENDER_REPORT_MINLEN) {
         // TODO: handle sender report containing receiver report blocks
-        DLOGW("unhandled packet type RTCP_PACKET_SENDER_REPORT size %d", pRtcpPacket->payloadLength);
         return STATUS_SUCCESS;
     }
 
@@ -92,7 +91,6 @@ static STATUS onRtcpReceiverReport(PRtcpPacket pRtcpPacket, PKvsPeerConnection p
     // https://tools.ietf.org/html/rfc3550#section-6.4.2
     if (pRtcpPacket->payloadLength != RTCP_PACKET_RECEIVER_REPORT_MINLEN) {
         // TODO: handle multiple receiver report blocks
-        DLOGS("unhandled packet type RTCP_PACKET_TYPE_RECEIVER_REPORT size %d", pRtcpPacket->payloadLength);
         return STATUS_SUCCESS;
     }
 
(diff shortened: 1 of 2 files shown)


src/source/PeerConnection/Rtcp.c
```diff
@@ -60,7 +60,6 @@ static STATUS onRtcpSenderReport(PRtcpPacket pRtcpPacket, PKvsPeerConnection pKv
 
     if (pRtcpPacket->payloadLength != RTCP_PACKET_SENDER_REPORT_MINLEN) {
         // TODO: handle sender report containing receiver report blocks
-        DLOGW("unhandled packet type RTCP_PACKET_SENDER_REPORT size %d", pRtcpPacket->payloadLength);
```
MushMal, line 63: Recommend changing this to DLOGS. These won't compile unless LOG_STREAMING compiler define is used.

src/source/PeerConnection/Rtcp.c
```diff
@@ -92,7 +91,6 @@ static STATUS onRtcpReceiverReport(PRtcpPacket pRtcpPacket, PKvsPeerConnection p
     // https://tools.ietf.org/html/rfc3550#section-6.4.2
     if (pRtcpPacket->payloadLength != RTCP_PACKET_RECEIVER_REPORT_MINLEN) {
         // TODO: handle multiple receiver report blocks
-        DLOGS("unhandled packet type RTCP_PACKET_TYPE_RECEIVER_REPORT size %d", pRtcpPacket->payloadLength);
```
MushMal, line 95: No need to remove DLOGS

MushMal (approved):

src/source/PeerConnection/Rtcp.c
disa6302: Agreed. But, when compiling with `LOG_STREAMING` define, it still produces a spew. Which is not required at the time since we know multiple report blocks are not supported at the moment.

MushMal (approved):

src/source/PeerConnection/Rtcp.c
hassanctech: I would have preferred to keep it but at a lower level like INFO. WARN is clearly wrong since it looks like something went wrong when really nothing did.

src/source/PeerConnection/Rtcp.c
hassanctech: Is this actually incorrect/misleading?  Why not keep it as an info level log?

src/source/PeerConnection/Rtcp.c
disa6302: We can keep the logging, but it does not give anything useful to the customers, not even as INFO. So, I would prefer getting rid of it altogether.

src/source/PeerConnection/Rtcp.c
disa6302: It is not required because it does not provide any useful information. Even as an INFO level log, the logs would be flooded with these messages. INFO is useful in one way when customers do not have the luxury to have DEBUG/VERBOSE enabled in production. Considering that and the flooding this specific log generates, it is better to get rid of it.

hassanctech (approved):

Issue #1115, opened 2021-04-05
Easy get ICE_AGENT_STATE_DISCONNECTED when streaming over internet


nguyenduynghia:
It is fine when I connect on local network.
But when I tried to connect over internet, I got this state after a short time.
Log file only show that:
```
[2021-04-05 15:23:59]-DEBUG-stepStateMachine(): State Machine - Current state: 0x0000000000000010, Next state: 0x0000000000000020
[2021-04-05 15:23:59]-DEBUG-executeDisconnectedIceAgentState(): Ice agent detected disconnection. current state ICE_AGENT_STATE_READY. Last data received time 1617611008 s
```

Is there any way to find out what cause this state?

disa6302:
@nguyenduynghia , 

Please include more details: 

1. What is your set up? 
2. Does the viewer disconnect anytime? Or network drops? 
3. Please attach complete logs. 2 log lines do not give any information.

zhuker:
I started seeing this recently as well in certain setups, trying to narrow
it down

nguyenduynghia:
I want to view video of 4 IP cameras. So I create one WebRTC Channel and use 4 peer connections.
It works fine on LAN. When I view via internet, it display video for a short time and  then disconnect.
Here is log file.
[kvs.log](https://github.com/awslabs/amazon-kinesis-video-streams-webrtc-sdk-c/files/6277457/kvs.log)

zhuker:
@nguyenduynghia 
do you use `peerConnectionOnSenderBandwidthEstimation` API?
if so try applying this patch and let me know if you see any difference
```
diff --git a/src/source/PeerConnection/Rtcp.c b/src/source/PeerConnection/Rtcp.c
index de35f8e08..d20f68786 100644
--- a/src/source/PeerConnection/Rtcp.c
+++ b/src/source/PeerConnection/Rtcp.c
@@ -319,6 +319,8 @@ STATUS onRtcpTwccPacket(PRtcpPacket pRtcpPacket, PKvsPeerConnection pKvsPeerConn
     }
 
     if (duration > 0) {
+        MUTEX_UNLOCK(pKvsPeerConnection->twccLock);
+        locked = FALSE;
         pKvsPeerConnection->onSenderBandwidthEstimation(pKvsPeerConnection->onSenderBandwidthEstimationCustomData, sentBytes, receivedBytes,
                                                         sentPackets, receivedPackets, duration);
     }
```

[...]

PR #1120, opened 2021-04-08
prevent deadlock when calling onSenderBandwidthEstimation callback

zhuker:
*Issue #, if available:*
https://github.com/awslabs/amazon-kinesis-video-streams-webrtc-sdk-c/issues/1115

*Description of changes:*
call sender estimation callback with twcclock unlocked to prevent potential deadlock


disa6302:
I restarted the job. Looks like it is going through. Will wait it out to see.

PR #1121, opened 2021-04-13
Check the TWCC manager before use

Nomidia:
When disableSenderSideBandwidthEstimation is set to TRUE, it will cause Segmentation fault

*Issue #, if available:*

*Description of changes:*


disa6302 (approved):
Thanks for the fix!

PR #1148, opened 2021-05-20
Add more detail to setsockopt logs

kev-the-dev:
*Description of changes:*
Indicate which option was attempted in debug logs when setting setsockopt() in 2 more places, as it is in the other usage of 
setsockopt.

This allows users to see what socket options may be unavailable on their network.


disa6302 (approved):
Thanks!

hassanctech (approved):
lgtm

PR #1167, opened 2021-06-07
extend iot resource to multi-channel per iotthing.

ycyang1229:
*Issue #, if available:*
no
*Description of changes:*
extend iot resource to multi-channel per iotthing.

### DIFF
--- a/samples/Common.c
+++ b/samples/Common.c
@@ -693,6 +693,7 @@ STATUS createSampleConfiguration(PCHAR channelName, SIGNALING_CHANNEL_ROLE_TYPE 
 
 #ifdef IOT_CORE_ENABLE_CREDENTIALS
     PCHAR pIotCoreCredentialEndPoint, pIotCoreCert, pIotCorePrivateKey, pIotCoreRoleAlias, pIotCoreThingName;
+    CHK_ERR((pIotCoreThingName = getenv(IOT_CORE_THING_NAME)) != NULL, STATUS_INVALID_OPERATION, "AWS_IOT_CORE_THING_NAME must be set");
     CHK_ERR((pIotCoreCredentialEndPoint = getenv(IOT_CORE_CREDENTIAL_ENDPOINT)) != NULL, STATUS_INVALID_OPERATION,
             "AWS_IOT_CORE_CREDENTIAL_ENDPOINT must be set");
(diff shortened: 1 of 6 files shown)


disa6302 (approved):

samples/Common.c
```diff
@@ -693,6 +693,7 @@ STATUS createSampleConfiguration(PCHAR channelName, SIGNALING_CHANNEL_ROLE_TYPE
 
 #ifdef IOT_CORE_ENABLE_CREDENTIALS
     PCHAR pIotCoreCredentialEndPoint, pIotCoreCert, pIotCorePrivateKey, pIotCoreRoleAlias, pIotCoreThingName;
+    CHK_ERR((pIotCoreThingName = getenv(IOT_CORE_THING_NAME)) != NULL, STATUS_INVALID_OPERATION, "AWS_IOT_CORE_THING_NAME must be set");
```
unicornss, line 696: here we are checking that the iot thing name matches with the channel name but in the policy we are authz ing all channels isn't?

ycyang1229: Sorry for my ignorance, but I do not understand what you mentioned. I did not check the iot thing name matching with the channel name. Would you mind giving me more information?

MushMal: The codebase is using the macro versions - GETENV

samples/Common.c
```diff
@@ -693,6 +693,7 @@ STATUS createSampleConfiguration(PCHAR channelName, SIGNALING_CHANNEL_ROLE_TYPE
 
 #ifdef IOT_CORE_ENABLE_CREDENTIALS
```
MushMal, line 694: I really don't understand why it was decided to make this a compile time define. We spent so much effort to avoid any compile time definitions in the headers and especially in the source files. The whole reason for the platform-level integration in CommonDefs and the way the entire source tree is architected is to avoid as much as possible the conditional compilation which is terrible when it comes to readability and maintenance of the codebase.

MushMal (approved):
minor comments

samples/Common.c
ycyang1229: Yes, I am also a little confused here. I will provide the sample code for ecs credential provider later, and I do not know do I need to add one more definition?

samples/Common.c
ycyang1229: I changed all the getenv in the common.c. Thanks for your review. :)

PR #1182, opened 2021-06-28
use poll instead of select

zhuker:
*Description of changes:*

Use `poll` system call instead of `select`.
Select fails when socket fd value is larger than 1024. This happens when running as part of a process which has already opened a lot of descriptors.
This is caused by a design bug in glibc `select` implementation.

```
BUGS
       POSIX allows an implementation to define an upper limit, advertised via the constant FD_SETSIZE, on the  range  of  file  de‐
       scriptors that can be specified in a file descriptor set.  The Linux kernel imposes no fixed limit, but the glibc implementa‐
       tion makes fd_set a fixed-size type, with FD_SETSIZE defined as 1024, and the  FD_*()  macros  operating  according  to  that
       limit.  To monitor file descriptors greater than 1023, use poll(2) instead.
```

### DIFF
--- a/src/source/Ice/ConnectionListener.c
+++ b/src/source/Ice/ConnectionListener.c
@@ -216,6 +216,17 @@ CleanUp:
     return retStatus;
 }
 
+BOOL canReadFd(INT32 fd, struct pollfd* fds, INT32 nfds)
+{
+    INT32 i;
+    for (i = 0; i < nfds; i++) {
+        if (fds[i].fd == fd && (fds[i].revents & POLLIN) != 0) {
+            return TRUE;
+        }
+    }
+    return FALSE;
+}
+
 PVOID connectionListenerReceiveDataRoutine(PVOID arg)
 {
     STATUS retStatus = STATUS_SUCCESS;
@@ -226,8 +237,7 @@ PVOID connectionListenerReceiveDataRoutine(PVOID arg)
     UINT32 i, socketCount;
 
     INT32 nfds = 0;
-    fd_set rfds;
-    struct timeval tv;
+    struct pollfd rfds[CONNECTION_LISTENER_DEFAULT_MAX_LISTENING_CONNECTION];
     INT32 retval, localSocket;
     INT64 readLen;
     // the source address is put here. sockaddr_storage can hold either sockaddr_in or sockaddr_in6
@@ -243,12 +253,11 @@ PVOID connectionListenerReceiveDataRoutine(PVOID arg)
     /* Ensure that memory sanitizers consider
      * rfds initialized even if FD_ZERO is
      * implemented in assembly. */
-    MEMSET(&rfds, 0x00, SIZEOF(fd_set));
+    MEMSET(&rfds, 0x00, SIZEOF(rfds));
 
     srcAddr.isPointToPoint = FALSE;
 
     while (!ATOMIC_LOAD_BOOL(&pConnectionListener->terminate)) {
-        FD_ZERO(&rfds);
         nfds = 0;
 
         // Perform the socket connection gathering under the lock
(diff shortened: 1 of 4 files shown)


src/source/Ice/Network.h
```diff
@@ -117,6 +117,12 @@ INT32 getErrorCode(VOID);
  */
 PCHAR getErrorString(INT32);
 
+#ifdef _WIN32
+#define POLL WSAPoll
```
zhuker, line 121: https://docs.microsoft.com/en-us/windows/win32/api/winsock2/nf-winsock2-wsapoll

src/source/Ice/ConnectionListener.c
```diff
@@ -216,6 +216,17 @@ STATUS connectionListenerStart(PConnectionListener pConnectionListener)
     return retStatus;
 }
 
+static BOOL canReadFd(INT32 fd, struct pollfd* fds, INT32 nfds)
```
MushMal, line 219: I would remove static scope. The codebase assumes Ansi C

src/source/Ice/ConnectionListener.c
```diff
@@ -276,24 +287,16 @@ PVOID connectionListenerReceiveDataRoutine(PVOID arg)
...
         MUTEX_UNLOCK(pConnectionListener->lock);
 
-        // timeout select every SOCKET_WAIT_FOR_DATA_TIMEOUT_SECONDS seconds and check if terminate
-        // on linux tv need to be reinitialized after select is done.
-        tv.tv_sec = 0;
-        tv.tv_usec = CONNECTION_LISTENER_SOCKET_WAIT_FOR_DATA_TIMEOUT / HUNDREDS_OF_NANOS_IN_A_MICROSECOND;
-
         // blocking call until resolves as a timeout, an error, a signal or data received
-        retval = select(nfds, &rfds, NULL, NULL, &tv);
+        retval = POLL(rfds, nfds, CONNECTION_LISTENER_SOCKET_WAIT_FOR_DATA_TIMEOUT / HUNDREDS_OF_NANOS_IN_A_MILLISECOND);
```
MushMal, line 294: Might want to play with the value of CONNECTION_LISTENER_SOCKET_WAIT_FOR_DATA_TIMEOUT as it will impact the latency. Unfortunately, selecting very fast polling will impact the CPU/battery life while using slow polling will impact the latency. If it's actually decided to go with polling then I would imagine this number to be around 40ms

MushMal (commented):
We tried to avoid polling due to inherent impact on latency and battery life. I am curious as to how prevalent this case is? If this really is a big issue then perhaps we should consider a compile time different implementation. I know it's going to be messy but we do have a few good examples of taming the differences. One of the good examples is OpenSSL vs mbedTLS implementations where the low-level functionality is put behind a common interface.

src/source/Ice/ConnectionListener.c
zhuker: `poll` is a misnomer i think
internally it functions in the very same way as `select` except it has better designed input parameters
From linux man pages:
```
DESCRIPTION
       poll()  performs a similar task to select(2): it waits for one of a set of file descriptors to become ready to per‐
       form I/O.
```

zhuker: @MushMal 
> We tried to avoid polling due to inherent impact on latency and battery life.

See above
`poll` doesn't really poll :) it `select`s but with a nicer arguments :)

> I am curious as to how prevalent this case is?

well it is considered a bug to set FD_SET if localSocket int value is over 1024
so if it so happens that there are a lot of open file descriptors, the library wont function at all

MushMal: Familiarized myself with the API. I don't have any further arguments - poll it is!

MushMal (approved):
LGTM. Perhaps, remove the "static" scope

src/source/Ice/ConnectionListener.c
zhuker: ok

zhuker:
@disa6302 @hassanctech  do you know whats happening with the static build?
```
  The C compiler
    "/usr/bin/cc"
  is not able to compile a simple test program.
```

hassanctech:
> @disa6302 @hassanctech do you know whats happening with the static build?
> 
> ```
>   The C compiler
>     "/usr/bin/cc"
>   is not able to compile a simple test program.
> ```

Yeah so it appears to be a mismatch between a library version provided by travis ci docker and alpine which recently upgraded.  The manifest difference is you can see back when the compiler version was 10.2.1 everything worked fine now with 10.3.1 we get this error.  We have not been able to fix it yet and it's impacting our other PRs as well.  I suppose if a local static build work we can bypass this and still merge.  That said I haven't actually looked at the PR itself let me do that first :).

hassanctech (approved):
nice!

PR #1193, opened 2021-07-19
update libsrtp

hassanctech:
*Issue #, if available:*

*Description of changes:*


CMake/Dependencies/libsrtp-CMakeLists.txt
```diff
@@ -37,10 +35,6 @@ else()
   set(LIBSRTP_SHARED_LIBS ON)
 endif()
 
-if (NOT WIN32)
```
jdelapla, line 40: Not entirely sure why this section is needed, and I didn't see an explanation in the commit. Could you elaborate on this piece?

hassanctech: This was a patch that I deleted after verifying the version of the library I'm upgrading to already has those fixes.

hassanctech: In short I think at the time the patch was created it it would have made sense to include a comment as to why those fixes were needed.  But now that I'm removing them I don't see any need to put a comment referencing something that's been removed.  Perhaps I misunderstood?

disa6302 (approved):
Just adding a note: Static, Windows and Cross compile build issues due to libwebsockets will be fixed in subsequent PRs.

RobGries:
@disa6302 Has the cross compilation issue caused by the libwebsockets update in this PR been fixed?

PR #1203, opened 2021-07-27
create data channel sample

niyatim23:
Create a data channel sample
* Create a data channel from the viewer's side
* Communicate over the data channel by sending and receiving 
messages using `onopen` and `onnmessage` callbacks from master and viewer

### DIFF
--- a/samples/Common.c
+++ b/samples/Common.c
@@ -1,5 +1,6 @@
 #define LOG_CLASS "WebRtcSamples"
 #include "Samples.h"
+// #define MASTER_DATA_CHANNEL_MESSAGE "This message is from the KVS Master"
 
 PSampleConfiguration gSampleConfiguration = NULL;
 
--- a/samples/Common.c
+++ b/samples/Common.c
@@ -1,6 +1,5 @@
 #define LOG_CLASS "WebRtcSamples"
 #include "Samples.h"
-// #define MASTER_DATA_CHANNEL_MESSAGE "This message is from the KVS Master"
 
 PSampleConfiguration gSampleConfiguration = NULL;
 
@@ -23,11 +22,11 @@ VOID onDataChannelMessage(UINT64 customData, PRtcDataChannel pDataChannel, BOOL 
         DLOGI("DataChannel String Message: %.*s\n", pMessageLen, pMessage);
     }
     // Send a response to the message sent by the viewer
-    /*STATUS retStatus = STATUS_SUCCESS;
+    STATUS retStatus = STATUS_SUCCESS;
     retStatus = dataChannelSend(pDataChannel, FALSE, (PBYTE) MASTER_DATA_CHANNEL_MESSAGE, STRLEN(MASTER_DATA_CHANNEL_MESSAGE));
     if(retStatus != STATUS_SUCCESS) {
         DLOGI("[KVS Master] dataChannelSend(): operation returned status code: 0x%08x \n", retStatus);
-    }*/
+    }
 }
 
 VOID onDataChannel(UINT64 customData, PRtcDataChannel pRtcDataChannel)
--- a/samples/Common.c
+++ b/samples/Common.c
@@ -1,6 +1,5 @@
 #define LOG_CLASS "WebRtcSamples"
 #include "Samples.h"
-// #define MASTER_DATA_CHANNEL_MESSAGE "This message is from the KVS Master"
 
 PSampleConfiguration gSampleConfiguration = NULL;
 
--- a/samples/Common.c
+++ b/samples/Common.c
@@ -1,5 +1,6 @@
 #define LOG_CLASS "WebRtcSamples"
 #include "Samples.h"
+// #define MASTER_DATA_CHANNEL_MESSAGE "This message is from the KVS Master"
 
 PSampleConfiguration gSampleConfiguration = NULL;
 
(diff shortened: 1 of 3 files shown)


samples/Common.c
```diff
@@ -1,5 +1,6 @@
 #define LOG_CLASS "WebRtcSamples"
 #include "Samples.h"
+// #define MASTER_DATA_CHANNEL_MESSAGE "This message is from the KVS Master"
```
disa6302, line 3: Sugest moving this to `Sample.h` instead.

samples/Common.c
```diff
@@ -21,6 +22,12 @@ VOID onDataChannelMessage(UINT64 customData, PRtcDataChannel pDataChannel, BOOL
     } else {
         DLOGI("DataChannel String Message: %.*s\n", pMessageLen, pMessage);
     }
+    // Send a response to the message sent by the viewer
```
disa6302, line 25: Is the customer expected to uncomment this to run? I am wondering if we can encapsulate all this into the `ENABLE_DATA_CHANNEL` directive which exists in the repository. That way we can avoid have commented code blocks.

samples/kvsWebRTCClientViewer.c
```diff
@@ -165,6 +195,26 @@ INT32 main(INT32 argc, CHAR* argv[])
         goto CleanUp;
     }
 
+    PRtcDataChannel pDataChannel = NULL;
```
disa6302, line 198: Encapsulate this entire section this way:

```
#ifdef ENABLE_DATA_CHANNEL
<code block>
#endif

Checkout usage [here](https://github.com/awslabs/amazon-kinesis-video-streams-webrtc-sdk-c/blob/c4a13309b6559d19529ca8049a50d4e2b241564f/src/source/PeerConnection/PeerConnection.c#L143)

File: samples/kvsWebRTCClientViewer.c [hunk 4]
```diff
@@ -165,6 +195,26 @@ INT32 main(INT32 argc, CHAR* argv[])
...
+    printf("[KVS Viewer] Creating data channel...completed\n");
+
+    // Setting a callback for when the data channel is open
+    retStatus = dataChannelOnOpen(pDataChannel, (UINT64) &datachannelLocalOpenCount, dataChannelOnOpenCallback);
+    if(retStatus != STATUS_SUCCESS) {
+        printf("[KVS Viewer] dataChannelOnOpen(): operation returned status code: 0x%08x \n", retStatus);
+        goto CleanUp;
+    }
+    printf("[KVS Viewer] Data Channel open now...\n");
+
```
disa6302, line 217: We should look into cleaning up data channels at the application exit for which we require `dataChannelClose()` API. We can handle that in a new PR.

samples/Common.c
niyatim23: Moved #define MASTER_DATA_CHANNEL_MESSAGE and #define VIEWER_DATA_CHANNEL_MESSAGE to Samples.h

samples/Common.c
niyatim23: Added `ENABLE_DATA_CHANNEL` in necessary places

samples/kvsWebRTCClientViewer.c
niyatim23: Added ENABLE_DATA_CHANNEL encapsulation here

samples/kvsWebRTCClientViewer.c
```diff
@@ -2,6 +2,31 @@
 
 extern PSampleConfiguration gSampleConfiguration;
 
+// onMessage callback for a message received by the viewer on a data channel
+VOID dataChannelOnMessageCallback(UINT64 customData, PRtcDataChannel pDataChannel, BOOL isBinary, PBYTE pMessage, UINT32 pMessageLen)
+{
+    UNUSED_PARAM(customData);
+    STATUS retStatus = STATUS_SUCCESS;
```
hassanctech, line 9: `retStatus` is unused so it can be removed.

hassanctech: Looks like `pDataChannel` is unused as well.  If this is required to match the signature (I didn't check) then you can just mark it as `UNUSED_PARAM()`

samples/kvsWebRTCClientViewer.c
```diff
@@ -2,6 +2,31 @@
...
+        DLOGI("DataChannel String Message: %.*s\n", pMessageLen, pMessage);
+    }
+}
+
+// onOpen callback for the onOpen event of a viewer created data channel
+VOID dataChannelOnOpenCallback(UINT64 customData, PRtcDataChannel pDataChannel) {
+    DLOGI("New DataChannel has been opened %s \n", pDataChannel->name);
+    dataChannelOnMessage(pDataChannel, customData, dataChannelOnMessageCallback);
+    ATOMIC_INCREMENT((PSIZE_T) customData);
+    STATUS retStatus = STATUS_SUCCESS;
```
hassanctech, line 22: To adhere with C89 standard variables need to be declared the top of a scope block.  You will need to move the declaration of `retStatus` to the top of this function.

hassanctech:
Looks great overall, some minor tweaks necessary to pass CI.

samples/kvsWebRTCClientViewer.c
niyatim23: Moved the declaration of `retStatus` to the top of this function.

samples/kvsWebRTCClientViewer.c
niyatim23: Marked `pDataChannel` as `UNUSED_PARAM()`. Removed `retStatus`.

disa6302 (approved):

PR #1228, opened 2021-08-17
[FIX] When protocol in DCEP header of data channel open command is no…

ela34:
…t empty, there is a check preventing the data channel to be opened. Now protocol length is correctly handled to avoid check failing.

In this PR the value of the protocol string (if not empty) is not checked. May be it should be checked against the default value "stcp" and if not matching the data channel creation should fail. 

### DIFF
--- a/src/source/Sctp/Sctp.c
+++ b/src/source/Sctp/Sctp.c
@@ -305,14 +305,17 @@ STATUS handleDcepPacket(PSctpSession pSctpSession, UINT32 streamId, PBYTE data, 
     ENTERS();
     STATUS retStatus = STATUS_SUCCESS;
     UINT16 labelLength = 0;
+    UINT16 protocolLength = 0;
 
     // Assert that is DCEP of type DataChannelOpen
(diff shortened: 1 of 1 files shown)


jdelapla (approved):

src/source/Sctp/Sctp.c
```diff
@@ -305,14 +305,17 @@ STATUS handleDcepPacket(PSctpSession pSctpSession, UINT32 streamId, PBYTE data,
...
     STATUS retStatus = STATUS_SUCCESS;
     UINT16 labelLength = 0;
+    UINT16 protocolLength = 0;
 
     // Assert that is DCEP of type DataChannelOpen
     CHK(length > SCTP_DCEP_HEADER_LENGTH && data[0] == DCEP_DATA_CHANNEL_OPEN, STATUS_SUCCESS);
 
-    MEMCPY(&labelLength, data + 8, SIZEOF(UINT16));
-    putInt16((PINT16) &labelLength, labelLength);
+	MEMCPY(&labelLength, data + 8, SIZEOF(UINT16));
```
MushMal, line 313: Nice! Would have loved this file to have more meaningful comments referencing the specs

MushMal (approved):

Issue #1174, opened 2021-06-15
[BUG]  iceAgentLogNewCandidate() unable to decipher protocol from remote SDP


suggestedfixes:
**Logging**
5 2021-06-15 21:29:03 DEBUG   iceAgentLogNewCandidate(): New remote ice candidate discovered. Id: fZ+4XiMkI. Ip: 172.218.223.215:57470. Type: srflx. Protocol: UNKNOWN.
142021-06-15 21:29:03 DEBUG   iceAgentLogNewCandidate(): New remote ice candidate discovered. Id: fZ+4XiMkI. Ip: 172.218.223.215:57466. Type: srflx. Protocol: UNKNOWN.
:2021-06-15 21:29:03 DEBUG   iceAgentLogNewCandidate(): New remote ice candidate discovered. Id: fZ+4XiMkI. Ip: 34.215.44.31:49312. Type: relay. Protocol: NA.
292021-06-15 21:29:03 DEBUG   iceAgentLogNewCandidate(): New remote ice candidate discovered. Id: fZ+4XiMkI. Ip: 172.218.223.215:57468. Type: srflx. Protocol: UNKNOWN.
:03:192021-06-15 21:29:03 DEBUG   iceAgentLogNewCandidate(): New remote ice candidate discovered. Id: fZ+4XiMkI. Ip: 34.215.44.31:64620. Type: relay. Protocol: NA.
22021-06-15 21:29:03 DEBUG   iceAgentLogNewCandidate(): New remote ice candidate discovered. Id: fZ+4XiMkI. Ip: 34.222.141.63:58049. Type: relay. Protocol: NA.
7]2021-06-15 21:29:03 DEBUG   iceAgentLogNewCandidate(): New remote ice candidate discovered. Id: fZ+4XiMkI. Ip: 34.215.44.31:61676. Type: relay. Protocol: NA.
 I:2021-06-15 21:29:03 DEBUG   iceAgentLogNewCandidate(): New remote ice candidate discovered. Id: fZ+4XiMkI. Ip: 34.222.141.63:58736. Type: relay. Protocol: NA.
 2021-06-15 21:29:03 DEBUG   iceAgentLogNewCandidate(): New remote ice candidate discovered. Id: fZ+4XiMkI. Ip: 34.222.141.63:58209. Type: relay. Protocol: NA.
[...]

PR #1229, opened 2021-08-17
Log ice candidate protocol appropriately

disa6302:
*Issue #, if available:*
#1174 

*Description of changes:*
Previously, we never recorded remote candidate IP address. However, the SDP of remote candidate comprises of the protocol information. Since we did not record this information, all remote candidates being discovered would log protocol as N/A or UNKNOWN. This PR fixes this bug by storing the protocol for remote candidates. 

Additionally, we also modify the logic for local candidate protocol logging by relying on pSocketConnection instead of pTurnConnection to log for all types of candidates. Although our SDK supports only UDP, it is important it is logged regardless based on SDP.

Resolves: #1174 


disa6302:
New PR to set develop as base.

PR #1230, opened 2021-08-17
Log ice candidate protocol appropriately

disa6302:
Issue #, if available:
#1174

Description of changes:
Previously, we never recorded remote candidate IP address. However, the SDP of remote candidate comprises of the protocol information. Since we did not record this information, all remote candidates being discovered would log protocol as N/A or UNKNOWN. This PR fixes this bug by storing the protocol for remote candidates.

Additionally, we also modify the logic for local candidate protocol logging by relying on pSocketConnection instead of pTurnConnection to log for all types of candidates. Although our SDK supports only UDP, it is important it is logged regardless based on SDP.

Resolves: #1174

### DIFF
--- a/src/source/Ice/IceAgent.c
+++ b/src/source/Ice/IceAgent.c
@@ -327,6 +327,7 @@ STATUS iceAgentAddRemoteCandidate(PIceAgent pIceAgent, PCHAR pIceCandidateString
     PDoubleListNode pCurNode = NULL;
     SDP_ICE_CANDIDATE_PARSER_STATE state;
     ICE_CANDIDATE_TYPE iceCandidateType = ICE_CANDIDATE_TYPE_HOST;
+    CHAR remoteProtocol[MAX_PROTOCOL_LENGTH];
 
     CHK(pIceAgent != NULL && pIceCandidateString != NULL, STATUS_NULL_ARG);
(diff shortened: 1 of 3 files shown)


src/source/Ice/IceAgent.c
```diff
@@ -354,6 +355,7 @@ STATUS iceAgentAddRemoteCandidate(PIceAgent pIceAgent, PCHAR pIceCandidateString
                 STRTOUI32(curr, next, 10, &priority);
                 break;
             case SDP_ICE_CANDIDATE_PARSER_STATE_PROTOCOL:
+                STRNCPY(remoteProtocol, curr, tokenLen);
```
jdelapla, line 358: string is not null terminated, but below is appears that 'protocol' is meant to be a null terminated string.
Please add remoteProtocol[tokenLen] = '\0';

jdelapla: Also, need to add a check that tokenLen < MAX_PROTOCOL_LENGTH or we have an overwrite security risk

jdelapla (changes requested):

src/source/Ice/IceAgent.c
disa6302: `\0` initialized the string instead.

jdelapla (approved):

Issue #660, opened 2020-07-24
[BUG] Invalidate/Flush SignalingCache if the Channel is not found


bkneff:
**Describe the bug**
The gstreamer sample applications will automatically create a channel if it doesn't exist.  The application will also cache this information in the Signaling FileCache.  If you delete the channel, the application will still attempt to use the cache'd value in FileCache.

**SDK version number**
acc94940

**Open source building**
If it is a build issue, include 3rd party library version and steps to how you are building it

**To Reproduce**
1. Run the gstreamer sample and specify a channel name
2. Delete the channel via the console
3. Kill the gstreamer application
4. Re-run the gstreamer application
5. The application will error out and not re-create the channel

If you delete the FileCache and re-run it will work.

**Expected behavior**
The application will re-create the channel

**Screenshots**

**Desktop (please complete the following information):**

**Smartphone (please complete the following information):**

**Additional context**

MushMal:
We should ensure that if the cached operations fail they will have to run the non-cached version after the state machine retries.

disa6302:
Pending implementation.

disa6302:
Closing since this is not an issue in the newer version anymore. Exponential retry covers this case.

PR #1251, opened 2021-08-31
Return appropriate error code on non-success result from getChannelEndpoint call

dewei-verkada:
*Issue #, if available:*
N/A

*Description of changes:*
Currently if we get a non-successful HTTP response from call to `/getSignalingChannelEndpoint` (for example, a 403), a `STATUS_SUCCESS` is returned from this function. This makes it hard to debug errors since we will have assumed that the channel endpoint was successfully retrieved and stored into `pSignalingClient`.

I am not 100% on whether or not this is the right return code, please let me know if it isn't and I will update accordingly.

### DIFF
--- a/src/source/Signaling/LwsApiCalls.c
+++ b/src/source/Signaling/LwsApiCalls.c
@@ -898,7 +898,8 @@ STATUS getChannelEndpointLws(PSignalingClient pSignalingClient, UINT64 time)
     resultLen = pLwsCallInfo->callInfo.responseDataLen;
 
     // Early return if we have a non-success result
-    CHK((SERVICE_CALL_RESULT) ATOMIC_LOAD(&pSignalingClient->result) == SERVICE_CALL_RESULT_OK && resultLen != 0 && pResponseStr != NULL, retStatus);
+    CHK((SERVICE_CALL_RESULT) ATOMIC_LOAD(&pSignalingClient->result) == SERVICE_CALL_RESULT_OK && resultLen != 0 && pResponseStr != NULL,
+        STATUS_INVALID_API_CALL_RETURN_JSON);
 
     // Parse and extract the endpoints
     jsmn_init(&parser);


src/source/Signaling/LwsApiCalls.c
```diff
@@ -898,7 +898,8 @@ STATUS getChannelEndpointLws(PSignalingClient pSignalingClient, UINT64 time)
     resultLen = pLwsCallInfo->callInfo.responseDataLen;
 
     // Early return if we have a non-success result
-    CHK((SERVICE_CALL_RESULT) ATOMIC_LOAD(&pSignalingClient->result) == SERVICE_CALL_RESULT_OK && resultLen != 0 && pResponseStr != NULL, retStatus);
+    CHK((SERVICE_CALL_RESULT) ATOMIC_LOAD(&pSignalingClient->result) == SERVICE_CALL_RESULT_OK && resultLen != 0 && pResponseStr != NULL,
```
MushMal, line 901: Why do you need to return the error here? All these backend API calls return success even if the call to the backend fails itself to ensure the higher-level state machine iterates again and retries. The higher-level state machine uses the response to check for the retry or move to the next state

dewei-verkada: I received a 403 back from the call to `/getSignalingChannelEndpoint` (due to lacking access permissions in IAM in my setup). Since a success code assumes the `pSignalingClient` will have `channelEndpointWss` and `channelEndpointHttps` are set, this will lead to failure later when saving to cache here: https://github.com/awslabs/amazon-kinesis-video-streams-webrtc-sdk-c/blob/master/src/source/Signaling/FileCache.c#L180.

This is misleading the user to think that issue is with the cache rather than a permission error from the API call. The state machine won't progress anyway on a non-zero status due to this line: https://github.com/awslabs/amazon-kinesis-video-streams-webrtc-sdk-c/blob/master/src/source/Signaling/StateMachine.c#L490

MushMal: This is similar to the issue #660 

The main issue is that the caching assumes the result of STATUS_SUCCESS as an indicator of an actual API call success. The caching layer needs to be updated to make a proper distinction and only cache when the backend API call succeeds.

dewei-verkada: Not sure about what you mean by caching layer? It currently reads the `retStatus` to make decision to save to cache or not by calling `signalingCacheSaveToFile`, the only way I could see to make that distinction is by creating another status code (something like `STATUS_API_ERROR`).

MushMal: The fix should be similar to the state machine processing. If the status is OK and we have the data we need then we should proceed with caching. I think this should call for a proper fix. Not sure what the priority is. Perhaps AWS folks could comment.

dewei-verkada: @Sean-Der looks like you are the original author of this, suggestions on this?

Personally I'd prefer this to be explicit, even if it means returning a `STATUS_SUCCESS` here instead of `retStatus`.

MushMal: These were conscious choices to differentiate between critical failures caused from the client side driven by retStatus and anything that could actually be related to the backend. An alternative could have been to flow everything via mapping to retStatus and have a retry logic working only based on the retStatus.

PR #1268, opened 2021-09-13
Fix scripts/pare_status.py for non-status #defines

kev-the-dev:
*Description of changes:*
Update the regex in `scripts/pare_status.py` to only pick up on `#define` lines which begin with `STATUS_`, to exclude lines such as:

```
#define WINVER       0x0A00
```

Which involve a hex defined variable but should not be included in the ouput of this script


hassanctech (approved):
Thank you for your contribution!  LGTM

PR #1272, opened 2021-09-17
Modifying relay candidate will change to prflx candidate in a specifi…

ZhaoJianliuser:
…c network environment, resulting in the problem of being unable to connect

*Issue #, if available:*

Modifying relay candidate will change to prflx candidate in a specific network environment, resulting in the problem of being unable to connect.

*Description of changes:*
The simplest way is to modify only non relay candidate to the prflx type .

### DIFF
--- a/src/source/Ice/IceAgent.c
+++ b/src/source/Ice/IceAgent.c
@@ -2574,7 +2574,9 @@ STATUS iceAgentCheckPeerReflexiveCandidate(PIceAgent pIceAgent, PKvsIpAddress pI
         CHK(pIceCandidate == NULL, retStatus); // return early if duplicated
 
         findCandidateWithSocketConnection(pSocketConnection, pIceAgent->localCandidates, &pLocalIceCandidate);
-        pLocalIceCandidate->iceCandidateType = ICE_CANDIDATE_TYPE_PEER_REFLEXIVE;
+		if (pLocalIceCandidate->iceCandidateType != ICE_CANDIDATE_TYPE_RELAYED) {
+			pLocalIceCandidate->iceCandidateType = ICE_CANDIDATE_TYPE_PEER_REFLEXIVE;	
+		}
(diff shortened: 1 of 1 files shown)


ZhaoJianliuser:
> …c network environment, resulting in the problem of being unable to connect
> 
> _Issue #, if available:_
> 
> Modifying relay candidate will change to prflx candidate in a specific network environment, resulting in the problem of being unable to connect.
>

MushMal (approved):
Looks OK. Is there any scenario that we might be breaking with this?

ZhaoJianliuser:
> Looks OK. Is there any scenario that we might be breaking with this?

Like a scene I just met, the turn server is deployed in the company intranet,and configure EIP for external camera to access. The camera obtains the relay candidate through the allocation request, at this time, the relay candidate's address is the EIP,  but the app is on the company intranet, when the camera sends a binding request through the relay for connection check, the turn server address seen by the app is intranet address,which is diffrent from the EIP.


Another note is that I disabled the UDP protocol on the router connected to the camera, so that successful connection must be the relay candidate of the TLS protocol. After I made such modifications, I can connect normally.

MushMal:
> > Looks OK. Is there any scenario that we might be breaking with this?
> 
> Like a scene I just met, the turn server is deployed in the company intranet,and configure EIP for external camera to access. The camera obtains the relay candidate through the allocation request, at this time, the relay candidate's address is the EIP, but the app is on the company intranet, when the camera sends a binding request through the relay for connection check, the turn server address seen by the app is intranet address,which is diffrent from the EIP.
> 
> Another note is that I disabled the UDP protocol on the router connected to the camera, so that successful connection must be the relay candidate of the TLS protocol. After I made such modifications, I can connect normally.

@hassanctech can you also comment on this? Last time I checked the code was a while back. From what I see, this shouldn't impact the existing scenarios.

Issue #1284, opened 2021-10-01
[FEATURE] Flag to build without Signaling service / libwebsocket


kev-the-dev:
**Is your feature request related to a problem? Please describe.**

When packaging this library for other build systems (bazel, for example), the largest complexity has been seen in bundling libwebsockets, which requires lots of cmake custom work with generated headers and custom flags set by kinesis. libwebsockets is not used by the part of the library which implements [the webrtc spec](https://www.w3.org/TR/webrtc/), but only in parts implementing websocket-based Kinesis Signaling service. 

For applications which use a different transport to carry signaling messages (both for Kinesis or the much broader set of applications which use webrtc and lead to a larger community who can contribute to this library), building in libwebsockets and  the signaling service code is unnecessary and introduces significant build complexity and surface area for bugs.

**Describe the solution you'd like**

Add a cmake / compiler flag `-DNO_SIGNALING_SERVICE` (undefined / unset by default), which, when set by user's custom build systems or when invoking cmake, build without libwebsockets and the Signaling code.

**The main request is accepting a small <10 line patch I developed which places lines related to Signaling/websockets behind `#ifndef NO_SIGNALING_SERVICE`.** This has no impact on current functionality, but allows users to much easier use this library with non-cmake build systems and custom signaling systems and transport methods.

**Describe alternatives you've considered**

* Separating Signaling service code into a separate library, so that this library is strictly an implementation of the webrtc sepc. This is a much larger refactor, and is presumably in conflict with the goals of the authors/maintainers

**Additional context**

Somewhat related to #1235

PR #1285, opened 2021-10-01
Add compiler flag to not build signaling service / libwebsockets

kev-the-dev:
*Issue #, if available:*
#1284 

*Description of changes:*

Add a compiler flag (unset by default) which will disable certain components which require libwebsocket (namely the Signaling service code). On its own, it allows users to manually set this flag to build without these components for projects which struggle to integrate libwebsockets (using different architectures or build systems, for example).

It is intended as part of future work to add this to the CMake environment as well.


src/source/Include_i.h
```diff
@@ -149,11 +151,16 @@ STATUS generateJSONSafeString(PCHAR, UINT32);
 #include "Rtp/Codecs/RtpH264Payloader.h"
 #include "Rtp/Codecs/RtpOpusPayloader.h"
 #include "Rtp/Codecs/RtpG711Payloader.h"
+#include <poll.h>
```
MushMal, line 154: Where is this being used? Do you need it?

MushMal (commented):
Great idea! A comment about poll.h inclusion

src/source/Include_i.h
kev-the-dev: Yes, it doesn't compile without this change, as things like `pollfd` won't be defined in the include chain (when `-DNO_SIGNALING_SERVICE` is set. I assume this is something that is currently included by libwebsockets. I want to check about cross-platform concerns here (if libwebsockets simply included `poll.h` or if there is cross-platform polyfil)

kev-the-dev: To help me out, can I assume the CI will check for me if this is an issue for supported platforms? Or are there are some manual checks I may need to do for things like windows support (is windows supported?)?

MushMal: > To help me out, can I assume the CI will check for me if this is an issue for supported platforms? Or are there are some manual checks I may need to do for things like windows support (is windows supported?)?

Your build seem to have passed the checks which means that the target platforms in Travis CI have build and run OK.

@disa6302 could you comment on this please?

disa6302: @kev-the-dev , we still do support Windows. The reason the test is disabled is because some manual steps did not work in CI. With the manual steps for Windows highlighted [here](https://github.com/awslabs/amazon-kinesis-video-streams-webrtc-sdk-c#cmake-arguments), we would want this patch to work on Windows as well.

disa6302:
Thank you @kev-the-dev , 

We will merge this PR once we get a chance to test it out on Windows.

Issue #1300, opened 2021-10-15
[CONTRIBUTION]fix compile error: ‘POLLIN’ undeclared 


bigcat26:
PLEASE ADD THE APPROPRIATE TAG TO ALLOW BUCKETIZATION OF THE SOLUTIONS. IF THE TAG IS NOT AVAILABLE, DO NOT WORRY, WE WILL TAKE CARE OF IT!

** Describe the issue you are trying to solve **

fix compile errors cause by not including "poll.h":

```text
invalid use of undefined type ‘struct pollfd’
ConnectionListener.c:223:50: error: ‘POLLIN’ undeclared 
```

** Details of the changes **

add #include <poll.h> to Include_i.h

** Test cases **

** Additional context **

PR #1301, opened 2021-10-15
#1300 fix compile errors cause by not including "poll.h"

bigcat26:
*Issue #1300 

*Description of changes:*

add #include <poll.h> to Include_i.h


kev-the-dev:
Interesting overlap with #1285. Can you expand on the conditions in which you needed this? Are you using the normal CMake workflow?

PR #1314, opened 2021-10-29
log response, particularly in case of failure

hassanctech:
*Issue #, if available:*

*Description of changes:*

For example before if the request signature was too old we should just swallow that error and manifest as a timeout 0x0000000f in the sync call.  Now we have logging to indicate what the error is, for example:

```
2021-10-29 14:50:08 WARN    lwsHttpCallbackRoutine(): Received client http read response:  {"message":"Signature expired: 20211029T144008Z is now earlier than 20211029T144508Z (20211029T145008Z - 5 min.)"}
```

In this instance I tried to submit the API call 10 minutes after the time was set in the request header.

### DIFF
--- a/src/source/Signaling/LwsApiCalls.c
+++ b/src/source/Signaling/LwsApiCalls.c
@@ -111,7 +111,7 @@ INT32 lwsHttpCallbackRoutine(struct lws* wsi, enum lws_callback_reasons reason, 
             lwsl_hexdump_debug(pDataIn, dataSize);
 
             if (dataSize != 0) {
-                CHK(NULL != (pLwsCallInfo->callInfo.responseData = (PCHAR) MEMALLOC(dataSize)), STATUS_NOT_ENOUGH_MEMORY);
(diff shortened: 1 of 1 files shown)


jdelapla (dismissed):

src/source/Signaling/LwsApiCalls.c
```diff
@@ -111,9 +111,15 @@ INT32 lwsHttpCallbackRoutine(struct lws* wsi, enum lws_callback_reasons reason,
             lwsl_hexdump_debug(pDataIn, dataSize);
 
             if (dataSize != 0) {
-                CHK(NULL != (pLwsCallInfo->callInfo.responseData = (PCHAR) MEMALLOC(dataSize)), STATUS_NOT_ENOUGH_MEMORY);
+                CHK(NULL != (pLwsCallInfo->callInfo.responseData = (PCHAR) MEMALLOC(dataSize+1)), STATUS_NOT_ENOUGH_MEMORY);
```
disa6302, line 114: Why this change?

PR #1315, opened 2021-10-29
Added 1-20% jitter to expiration value stored by AwsCredentials struc…

jdelapla:
…t in producer-c.

Modified stepSignaling to include a check for expiration of credentials.
This is part of a larger effort to help customers with large fleets distribute the timings of their attempts to reconnect to our servers.

*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/src/source/Signaling/StateMachine.c
+++ b/src/source/Signaling/StateMachine.c
@@ -82,7 +82,7 @@ STATUS stepSignalingStateMachine(PSignalingClient pSignalingClient, STATUS statu
 
     // Fix-up the expired credentials transition
     // NOTE: Api Gateway might not return an error that can be interpreted as unauthorized to
-    // make the correct transition to auth integration state.
+    // Make the correct transition to auth integration state.
     if (status == STATUS_SERVICE_CALL_NOT_AUTHORIZED_ERROR ||
         (pSignalingClient->pAwsCredentials != NULL && pSignalingClient->pAwsCredentials->expiration < currentTime)) {
         // Set the call status as auth error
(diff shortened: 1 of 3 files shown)


MushMal:
I provided some feedback on 

https://github.com/awslabs/amazon-kinesis-video-streams-producer-c/pull/254#issuecomment-956453242

Also, the PIC solution to jitter has a threshold below which jitter is not applied. Is there a good reason to add jitter to IoT credential provider at all? What about the other providers? Perhaps a better approach is to leave the providers alone and handle the jitter on the "consumer" side - i.e. the signaling client in this case.

src/source/Signaling/StateMachine.c
```diff
@@ -82,11 +82,11 @@ STATUS stepSignalingStateMachine(PSignalingClient pSignalingClient, STATUS statu
...
     // Fix-up the expired credentials transition
     // NOTE: Api Gateway might not return an error that can be interpreted as unauthorized to
-    // make the correct transition to auth integration state.
+    // Make the correct transition to auth integration state.
     if (status == STATUS_SERVICE_CALL_NOT_AUTHORIZED_ERROR ||
-        (SERVICE_CALL_UNKNOWN == (SERVICE_CALL_RESULT) ATOMIC_LOAD(&pSignalingClient->result) &&
-         pSignalingClient->pAwsCredentials->expiration < currentTime)) {
+        (pSignalingClient->pAwsCredentials != NULL && pSignalingClient->pAwsCredentials->expiration < currentTime)) {
         // Set the call status as auth error
+        DLOGD("Expiration: %llu %llu", pSignalingClient->pAwsCredentials->expiration, currentTime);
```
ycyang1229, line 89: maybe replace %llu with PRIu64.

PR #1320, opened 2021-11-04
for non retriable failures, set the terminal exit status for state in…

hassanctech:
For issues such as signature expired, we don't want to keep retrying as that is unlikely to help things.  This bails out early in this case with the call failed status which is the terminal error condition for in the state machine so the lower level layers will not retry.

*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/src/source/Signaling/LwsApiCalls.c
+++ b/src/source/Signaling/LwsApiCalls.c
@@ -1063,7 +1063,7 @@ STATUS getIceConfigLws(PSignalingClient pSignalingClient, UINT64 time)
     ATOMIC_STORE(&pSignalingClient->result, (SIZE_T) pLwsCallInfo->callInfo.callResult);
     pResponseStr = pLwsCallInfo->callInfo.responseData;
     resultLen = pLwsCallInfo->callInfo.responseDataLen;
-    
+
     CHK_ERR(isCallResultFailureRetryable(&pLwsCallInfo->callInfo), STATUS_SIGNALING_GET_ICE_CONFIG_CALL_FAILED,
             "GetIceConfig API call failed with: %s and will not be retried.", pResponseStr);
 
--- a/src/source/Signaling/LwsApiCalls.c
+++ b/src/source/Signaling/LwsApiCalls.c
@@ -643,6 +643,10 @@ CleanUp:
     return retStatus;
 }
 
+BOOL isLwsCallResultFailureRetryable(PCallInfo pCallInfo) {
+    return (STRSTR(pCallInfo->responseData, "Signature expired") == NULL);
+}
+
 //////////////////////////////////////////////////////////////////////////
 // API calls
 //////////////////////////////////////////////////////////////////////////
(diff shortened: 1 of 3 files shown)


src/source/Signaling/LwsApiCalls.c
```diff
@@ -689,6 +693,11 @@ STATUS describeChannelLws(PSignalingClient pSignalingClient, UINT64 time)
     pResponseStr = pLwsCallInfo->callInfo.responseData;
     resultLen = pLwsCallInfo->callInfo.responseDataLen;
 
+    if (!isLwsCallResultFailureRetryable(&pLwsCallInfo->callInfo)) {
```
MushMal, line 696: Minor: Perhaps cleaner to use:

CHK_ERR(isCallResultFailureRetryable(&pLwsCallInfo->callInfo, STATUS_SIGNALING_DESCRIBE_CALL_FAILED, "Describe API call failed with %*.s and will not be retried.", resultLen, pResounseStr);

src/source/Signaling/LwsApiCalls.c
```diff
@@ -906,6 +921,8 @@ STATUS getChannelEndpointLws(PSignalingClient pSignalingClient, UINT64 time)
                                  SIGNALING_SERVICE_API_CALL_CONNECTION_TIMEOUT, SIGNALING_SERVICE_API_CALL_COMPLETION_TIMEOUT,
                                  DEFAULT_LOW_SPEED_LIMIT, DEFAULT_LOW_SPEED_TIME_LIMIT, pSignalingClient->pAwsCredentials, &pRequestInfo));
 
+    pRequestInfo->currentTime -= (10 * HUNDREDS_OF_NANOS_IN_A_MINUTE);
```
MushMal, line 924: Remove

src/source/Signaling/LwsApiCalls.c
```diff
@@ -643,6 +643,10 @@ STATUS lwsCompleteSync(PLwsCallInfo pCallInfo)
     return retStatus;
 }
 
+BOOL isLwsCallResultFailureRetryable(PCallInfo pCallInfo) {
```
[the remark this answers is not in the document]
hassanctech: Done

src/source/Signaling/LwsApiCalls.c
```diff
@@ -643,6 +643,10 @@ STATUS lwsCompleteSync(PLwsCallInfo pCallInfo)
...
+    return (STRSTR(pCallInfo->responseData, "Signature expired") == NULL);
```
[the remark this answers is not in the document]
hassanctech: It is in fact guaranteed to be null terminated since the last commit on this branch:

https://github.com/awslabs/amazon-kinesis-video-streams-webrtc-sdk-c/blob/release_1_5_1/src/source/Signaling/LwsApiCalls.c#L119

But I made the change anyway just to be safe in case that changes in the future.

src/source/Signaling/LwsApiCalls.c
hassanctech: Yeah I thought of that but I chose to not do that to avoid a duplicate log, I added a general error log in case the retStatus was not success, but I think including something specific to the call not being retried is fine, I made this change we can live with 2 log messages.

src/source/Signaling/LwsApiCalls.c
hassanctech: Thanks, left over from testing... removed.

MushMal (approved):

src/source/Signaling/LwsApiCalls.c
```diff
@@ -643,6 +643,10 @@ STATUS lwsCompleteSync(PLwsCallInfo pCallInfo)
     return retStatus;
 }
 
+BOOL isCallResultFailureRetryable(PCallInfo pCallInfo) {
+    return (STRNSTR(pCallInfo->responseData, "Signature expired", pCallInfo->responseDataLen) == NULL);
```
disa6302, line 647: Is "Signature expired" the only scenario where we would want to bail out? As a long term plan, we might want to translate these strings to status codes and create a list of retriable vs non retriable errors if that is not already happening. What do you think?

src/source/Signaling/LwsApiCalls.c
```diff
@@ -643,6 +643,10 @@ STATUS lwsCompleteSync(PLwsCallInfo pCallInfo)
     return retStatus;
 }
 
+BOOL isCallResultFailureRetryable(PCallInfo pCallInfo) {
```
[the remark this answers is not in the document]
hassanctech: I don't think there is an established spelling for this since neither is actually a dictionary word.  `Retryable` looks/feels more correct to me, but I will change it if you insist :).

src/source/Signaling/LwsApiCalls.c
hassanctech: Agreed, I expect that to come in future PRs.

disa6302 (approved):

PR #1323, opened 2021-11-09
Signaling state machine rework

niyatim23:
*Issue #, if available:*

*Description of changes:*
There are calls to `stepSignalingStateMachine` from `executeStateFn` and `getNextStateFn` functions for a lot of states in the signaling state machine. `stepSignalingStateMachine` calls `stepStateMachine` which also contains calls to `executeStateFn` and `getNextStateFn` functions. This causes recursion and could cause the stack to overflow.

- the `stepSignalingStateMachine` is replaced with `signalingStateMachineIterator`
- calls to the `stepSignalingStateMachine` from _Signaling.c_ and _LwsApiCalls.c_ have been replaced with calls to `signalingStateMachineIterator` with their respective final states
- the `stepUntil` and `continueOnReady` attributes have been removed from the signaling client since they are no longer needed to break out of or track the states. This is instead handled by having an `expiration` and a `final state` for the iterator

### DIFF
--- a/src/source/Signaling/Signaling.c
+++ b/src/source/Signaling/Signaling.c
@@ -37,7 +37,7 @@ STATUS createSignalingSync(PSignalingClientInfoInternal pClientInfo, PChannelInf
     CHK_STATUS(validateSignalingCallbacks(pSignalingClient, pCallbacks));
     CHK_STATUS(validateSignalingClientInfo(pSignalingClient, pClientInfo));
 
-    pSignalingClient->version = 1;
+    pSignalingClient->version = SIGNALING_CLIENT_CURRENT_VERSION;
 
     // Set invalid call times
     pSignalingClient->describeTime = INVALID_TIMESTAMP_VALUE;
@@ -626,7 +626,7 @@ STATUS signalingStoreOngoingMessage(PSignalingClient pSignalingClient, PSignalin
     CHK(pSignalingClient != NULL && pSignalingMessage != NULL, STATUS_NULL_ARG);
     MUTEX_LOCK(pSignalingClient->messageQueueLock);
     locked = TRUE;
-    // pOngoingCallInfo already null here
+
     CHK_STATUS(signalingGetOngoingMessage(pSignalingClient, pSignalingMessage->correlationId, pSignalingMessage->peerClientId, &pExistingMessage));
     CHK(pExistingMessage == NULL, STATUS_SIGNALING_DUPLICATE_MESSAGE_BEING_SENT);
     CHK_STATUS(stackQueueEnqueue(pSignalingClient->pMessageQueue, (UINT64) pSignalingMessage));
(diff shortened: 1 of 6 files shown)


src/source/Signaling/Signaling.c
```diff
@@ -37,6 +37,8 @@ STATUS createSignalingSync(PSignalingClientInfoInternal pClientInfo, PChannelInf
     CHK_STATUS(validateSignalingCallbacks(pSignalingClient, pCallbacks));
     CHK_STATUS(validateSignalingClientInfo(pSignalingClient, pClientInfo));
 
+    pSignalingClient->version = 1;
```
disa6302, line 40: Let us replace 1 with this to align with rest of the SDK:

`#define SIGNALING_CLIENT_CURRENT_VERSION      1`
Also, I dont see this being used anywhere. Since it is an internal structure, I dont see the significance of setting a version at the moment.

hassanctech (approved):

src/source/Signaling/Signaling.c
niyatim23: Made these changes in the latest commit

disa6302 (approved):

src/source/Signaling/Signaling.h
```diff
@@ -139,6 +139,9 @@ typedef struct {
  * Internal representation of the Signaling client.
  */
 typedef struct {
+    // Current version of the structure
```
MushMal, line 142: This is an internal struct- don't think you need a versioning on this one

src/source/Signaling/StateMachine.c
```diff
@@ -48,50 +48,48 @@ StateMachineState SIGNALING_STATE_MACHINE_STATES[] = {
...
-         pSignalingClient->pAwsCredentials->expiration < currentTime)) {
-        // Set the call status as auth error
-        ATOMIC_STORE(&pSignalingClient->result, (SIZE_T) SERVICE_CALL_NOT_AUTHORIZED);
-    }
+        retStatus = stepStateMachine(pSignalingClient->pStateMachine);
 
-    // Step the state machine
-    CHK_STATUS(stepStateMachine(pSignalingClient->pStateMachine));
+        CHK_STATUS(getStateMachineCurrentState(pSignalingClient->pStateMachine, &pState));
+        CHK(!(pState->state == finalState), STATUS_SUCCESS);
```
MushMal, line 91: CHK(pState->state != finalState, retStatus);

hassanctech (approved):
ok!

MushMal (commented):
This is an intricate change and I haven't had a chance to examine it deeply. The motivation of this change is a stack overflow. I would strongly suggest understanding the case where the stack would overflow. If there are large locals being allocated on the entire length of the stack then those should move to a heap memory allowing stack growth. 

This said, the only reason why we had a nested stack is due to the continuation model in async processing in KVS producer. As signaling is synchronous API ONLY currently, this is not needed. However, if at any stage there is an introduction of async APIs then this will be needed.

PR #1326, opened 2021-11-17
 Incorporating PIC state machine level retry changes into webrtc signaling state machine

kateyanurag:
…ling state machine

*Issue #, if available:*

*Description of changes:*
 Incorporating PIC state machine level retry changes into webrtc signaling state machine


disa6302 (commented):
Based on discussion:

There are 2 levels of retries:
1. Application level backoff for `createSignalingClientSync()`
2. Internal PIC level retry for state transitions based on failure (`PSignalingClientRetryStrategy`)

We want to have a single `createSignalingClientSync()` which handles application level retry and internal PIC state machinery retry as well. In order to consolidate this, we will have the backoff state structure as part of clientInfo, which will be set up in the application. However, we want to get rid of  `createSignalingClientSyncWithRetryStrategy()` and let PIC level retries to happen internally. Customers will invoke `createSignalingClientSync()` only once.

src/source/Signaling/Client.c
```diff
@@ -1,13 +1,37 @@
...
+STATUS validateSignalingClientRetryStrategy(PSignalingClientInfo pClientInfo) {
+    ENTERS();
+    STATUS retStatus = STATUS_SUCCESS;
+    PSignalingClientRetryStrategy pSignalingClientRetryStrategy;
+
+    CHK(pClientInfo != NULL, STATUS_NULL_ARG);
+
+    pSignalingClientRetryStrategy = &(pClientInfo->signalingClientRetryStrategy);
+
+    CHK(pSignalingClientRetryStrategy->retryStrategyType > KVS_RETRY_STRATEGY_DISABLED &&
```
disa6302, line 13: `pSignalingClientRetryStrategy->retryStrategyType > KVS_RETRY_STRATEGY_DISABLED` to return `STATUS_NULL_ARG` seems misleading. Maybe we can look into a more suitable status code? What is the behavior in PIC if this retry strategy is used? I am wondering if `STATUS_NOT_IMPLEMENTED` would be better.

src/source/Signaling/Client.c
```diff
@@ -16,7 +40,28 @@ STATUS createSignalingClientSync(PSignalingClientInfo pClientInfo, PChannelInfo
     MEMSET(&signalingClientInfoInternal, 0x00, SIZEOF(signalingClientInfoInternal));
     signalingClientInfoInternal.signalingClientInfo = *pClientInfo;
 
-    CHK_STATUS(createSignalingSync(&signalingClientInfoInternal, pChannelInfo, pCallbacks, pCredentialProvider, &pSignalingClient));
+    CHK_STATUS(validateSignalingClientRetryStrategy(pClientInfo));
```
disa6302, line 43: Same here. Do we want to pass the entire struct to validate retry strategy? No harm in passing it though.

disa6302 (commented):
Overall logic looks good to me. I have added some minor comments.

samples/Common.c
```diff
@@ -782,6 +803,8 @@ STATUS createSampleConfiguration(PCHAR channelName, SIGNALING_CHANNEL_ROLE_TYPE
     pSampleConfiguration->clientInfo.version = SIGNALING_CLIENT_INFO_CURRENT_VERSION;
     pSampleConfiguration->clientInfo.loggingLevel = logLevel;
     pSampleConfiguration->clientInfo.cacheFilePath = NULL; // Use the default path
+    CHK_STATUS(setupDefaultSignalingClientRetryStrategy(&pSampleConfiguration->clientInfo));
```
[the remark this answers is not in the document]
kateyanurag: Yes we need to pass entire client info.

Explanation:
I missed one change in common.c. I've added a new line - 
`pSignalingClientInfo->signalingClientCreationMaxRetryCount = MAX_CREATE_SIGNALING_CLIENT_RETRIES;`

This is for the outer retry while loop in createSignalingClientSync API. This parameter is defined in SignalingClientInfo struct. This parameter cannot go into SignalingClientRetryStrategy. SignalingClientRetryStrategy is basically KvsRetryStrategy defined in PIC.

```
typedef struct {
    UINT32 version;                                                
    CHAR clientId[MAX_SIGNALING_CLIENT_ID_LEN + 1];                
    UINT32 loggingLevel;                                  
    PCHAR cacheFilePath;             
    SignalingClientRetryStrategy signalingClientRetryStrategy;      
    UINT32 signalingClientCreationMaxRetryCount;              
} SignalingClientInfo, *PSignalingClientInfo;
```

src/source/Signaling/Client.c
kateyanurag: Agree. STATUS_NOT_IMPLEMENTED would be better. Fixing it. I did not find that earlier. If KVS_RETRY_STRATEGY_DISABLED is used, then no wait times will be injected on failures (current behavior).

src/source/Signaling/Client.c
kateyanurag: Yeah same reason. Need to make sure `pClientInfo->signalingClientCreationMaxRetryCount` is at least 1.

disa6302 (approved):

PR #1328, opened 2021-11-18
Fix issue with API call failures being treated as success

hassanctech:
cherry pick commits from release_1_5_1 to here, all tests pass locally

Fixes issue where previously any API call failure would return a success status code and we would bail out early with status success so no retries would happen until we would get to a latter state in the state machine which would fail with internal failure.

Also previously we were resetting call results to unknown before setting into the state machine, this has been removed.


src/source/Signaling/LwsApiCalls.c
```diff
@@ -111,9 +111,16 @@ INT32 lwsHttpCallbackRoutine(struct lws* wsi, enum lws_callback_reasons reason,
             lwsl_hexdump_debug(pDataIn, dataSize);
 
             if (dataSize != 0) {
-                CHK(NULL != (pLwsCallInfo->callInfo.responseData = (PCHAR) MEMALLOC(dataSize)), STATUS_NOT_ENOUGH_MEMORY);
+                CHK(NULL != (pLwsCallInfo->callInfo.responseData = (PCHAR) MEMALLOC(dataSize+1)), STATUS_NOT_ENOUGH_MEMORY);
```
MushMal, line 114: Minor: Spacing

src/source/Signaling/LwsApiCalls.c
```diff
@@ -111,9 +111,16 @@ INT32 lwsHttpCallbackRoutine(struct lws* wsi, enum lws_callback_reasons reason,
...
                 MEMCPY(pLwsCallInfo->callInfo.responseData, pDataIn, dataSize);
+                pLwsCallInfo->callInfo.responseData[dataSize] = '\0';
                 pLwsCallInfo->callInfo.responseDataLen = (UINT32) dataSize;
+
+                if (pLwsCallInfo->callInfo.callResult != SERVICE_CALL_RESULT_OK) {
+                    DLOGW("Received client http read response:  %s", pLwsCallInfo->callInfo.responseData);
```
MushMal, line 120: Minor: Perhaps change the text to differ from the success case. Something like "Client read failed with.."

src/source/Signaling/LwsApiCalls.c
```diff
@@ -632,6 +639,10 @@ STATUS lwsCompleteSync(PLwsCallInfo pCallInfo)
     return retStatus;
 }
 
+BOOL isCallResultFailureRetryable(PCallInfo pCallInfo) {
```
MushMal, line 642: If there is no plan to expand on this then this can be a macro. 

More importantly, checking for the text within the response message is a very brittle architecture. There should be a proper contractual signal for a particular failure.

src/source/Signaling/LwsApiCalls.c
```diff
@@ -753,6 +768,10 @@ STATUS describeChannelLws(PSignalingClient pSignalingClient, UINT64 time)
 
 CleanUp:
 
+    if (STATUS_FAILED(retStatus)) {
```
MushMal, line 771: There was a macro I believe called CHK_LOG_ERR or something similar.

src/source/Signaling/Signaling.c
```diff
@@ -863,10 +862,6 @@ STATUS describeChannel(PSignalingClient pSignalingClient, UINT64 time)
 
 CleanUp:
 
-    if (STATUS_FAILED(retStatus) && pSignalingClient != NULL) {
```
MushMal, line 866: Not sure what the reason is for removing these checks

src/source/Signaling/LwsApiCalls.c
hassanctech: I should provide some context so an upcoming PR after this will actually automatically correct for clock skew in the SDK and retry at which time this may be removed altogether its kind of here as an intermediate step right now.  aws sdks have auto clock skew correction that will be feature so we're adding it here as well to help customers who are having a hard time with ntp syncing their devices to make sure time is accurate.

src/source/Signaling/Signaling.c
hassanctech: It is very unclear to me why on failed status we would override the result value to UKKNOWN.  The call result has useful information regarding in what way the call failed.

MushMal: I can't recall the reason for this or where this is checked or what edge case this is used. I have not followed all the recent changes so perhaps this is a no-op impact right now. I would only suggest to check whether this is somehow used in retriable failure cases but those should be covered in the unit tests so if they pass then I think its great that this gets removed.

disa6302 (approved):

src/source/Signaling/LwsApiCalls.c
hassanctech: I will make this change as part of an upcoming PR.

src/source/Signaling/LwsApiCalls.c
hassanctech: I will make this change as part of an upcoming PR.

src/source/Signaling/LwsApiCalls.c
hassanctech: I will make this change as part of an upcoming PR.

PR #1341, opened 2021-12-03
Incorporating PIC state machine level retry changes into webrtc signaling state machine

kateyanurag:
Incorporating PIC state machine level retry changes into webrtc signaling state machine

*Issue #, if available:*

*Description of changes:*
Incorporating PIC state machine level retry changes into webrtc signaling state machine


src/source/Signaling/Signaling.c
```diff
@@ -268,6 +272,71 @@ STATUS freeSignaling(PSignalingClient* ppSignalingClient)
...
+    ENTERS();
+    STATUS retStatus = STATUS_SUCCESS;
+    PKvsRetryStrategyCallbacks pKvsRetryStrategyCallbacks = NULL;
+
+    CHK(pSignalingClient != NULL, STATUS_NULL_ARG);
+    pKvsRetryStrategyCallbacks = &(pSignalingClient->clientInfo.signalingStateMachineRetryStrategyCallbacks);
+
+    // If the callbacks for retry strategy are already set, then use that otherwise
+    // build the client with a default retry strategy.
+    if (pKvsRetryStrategyCallbacks->createRetryStrategyFn == NULL ||
```
disa6302, line 304: So even if one of the callbacks is NULL, we want to default, is it? What if we want to override only one of the callbacks?

disa6302 (commented):
Let us also not forget to update the producer commit once the dependencies are merged into develop.

src/source/Signaling/Client.c
```diff
@@ -23,6 +23,7 @@ STATUS createSignalingClientSync(PSignalingClientInfo pClientInfo, PChannelInfo
 CleanUp:
 
     if (STATUS_FAILED(retStatus)) {
+        DLOGE("Create signaling client API failed with return code ["PRIu64"]", retStatus);
```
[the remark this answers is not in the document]
kateyanurag: Will change it in the commit where I change Producer hash

src/source/Signaling/Signaling.c
kateyanurag: Just to keep it simple. 
Providing a provisioning to override just one callback while others being default sounds confusing. Also it will be unnecessarily complex to debug. Also this is intended mostly for unit tests. There should not be a need to override one and keep others default.

src/source/Signaling/Client.c
```diff
@@ -16,17 +71,45 @@ STATUS createSignalingClientSync(PSignalingClientInfo pClientInfo, PChannelInfo
     MEMSET(&signalingClientInfoInternal, 0x00, SIZEOF(signalingClientInfoInternal));
     signalingClientInfoInternal.signalingClientInfo = *pClientInfo;
 
-    CHK_STATUS(createSignalingSync(&signalingClientInfoInternal, pChannelInfo, pCallbacks, pCredentialProvider, &pSignalingClient));
+    CHK_STATUS(createRetryStrategyForCreatingSignalingClient(pClientInfo, &createSignalingClientRetryStrategy));
 
-    *pSignalingHandle = TO_SIGNALING_CLIENT_HANDLE(pSignalingClient);
+    signalingClientCreationMaxRetryCount = pClientInfo->signalingClientCreationMaxRetryAttempts;
+    while (TRUE) {
```
disa6302, line 77: Why are we doing a while(TRUE) and then checking for `signalingClientCreationMaxRetryCount` inside? Directly looping on the variable would avoid using break altogether.

tst/SignalingApiFunctionalityTest.cpp
```diff
@@ -257,6 +258,35 @@ TEST_F(SignalingApiFunctionalityTest, basicCreateConnectFree)
     EXPECT_EQ(STATUS_SUCCESS, freeSignalingClient(&signalingHandle));
 }
 
+TEST_F(SignalingApiFunctionalityTest, basicCreateCreateWithRetries)
```
disa6302, line 261: Nit: Naming seems weird. `basicCreate**Create**WithRetries` . Did you mean `basicCreateWithRetries`

src/source/Signaling/Client.c
disa6302: Ok. I get it. We want to make sure we dont wait after the last attempt. 
So, with, `while(signalingClientCreationMaxRetryCount >= 0)`, if `signalingClientCreationMaxRetryCount` fails, we would end up invoking exponentialBackOff unnecessarily after the last try. With this, we avoid doing that.

disa6302 (approved):

PR #1343, opened 2021-12-03
State machine preparation is decoupled from the constructor

jdelapla:
*Issue #, if available:*

*Description of changes:*
Sample has been adjusted to no longer regularly free/allocate the client object repeatedly as a means of resetting the state machine.
New PUBLIC_API 'Fetch' added to signaling work flow to act as a way to fetch all necessary assets before attempting to connect.


src/include/com/amazonaws/kinesis/video/webrtcclient/Include.h
```diff
@@ -1923,6 +1923,8 @@ PUBLIC_API STATUS signalingClientGetIceConfigInfoCount(SIGNALING_CLIENT_HANDLE,
  */
 PUBLIC_API STATUS signalingClientGetIceConfigInfo(SIGNALING_CLIENT_HANDLE, UINT32, PIceConfigInfo*);
 
+
```
jdelapla, line 1926: adding this description now

src/source/Signaling/StateMachine.c
```diff
@@ -79,8 +79,7 @@ STATUS signalingStateMachineIterator(PSignalingClient pSignalingClient, UINT64 e
         // NOTE: Api Gateway might not return an error that can be interpreted as unauthorized to
         // make the correct transition to auth integration state.
         if (retStatus == STATUS_SERVICE_CALL_NOT_AUTHORIZED_ERROR ||
-            (SERVICE_CALL_UNKNOWN == (SERVICE_CALL_RESULT) ATOMIC_LOAD(&pSignalingClient->result) &&
-             pSignalingClient->pAwsCredentials->expiration < currentTime)) {
+            (pSignalingClient->pAwsCredentials != NULL && pSignalingClient->pAwsCredentials->expiration < currentTime)) {
```
niyatim23, line 82: Are we handling `(SERVICE_CALL_UNKNOWN == (SERVICE_CALL_RESULT) ATOMIC_LOAD(&pSignalingClient->result)` somewhere else?

src/source/Signaling/StateMachine.c
```diff
@@ -273,6 +272,11 @@ STATUS executeGetTokenSignalingState(UINT64 customData, UINT64 time)
                                                                             SIGNALING_CLIENT_STATE_GET_CREDENTIALS));
     }
 
+    // reset credentials expiration if we already have one.
+    if (NULL == pSignalingClient->pAwsCredentials) {
+        pSignalingClient->pAwsCredentials = 0;
```
disa6302, line 277: If it is NULL, we are setting it to 0? What is the purpose of this?

src/source/Signaling/Signaling.c
```diff
@@ -364,6 +363,29 @@ STATUS signalingGetIceConfigInfo(PSignalingClient pSignalingClient, UINT32 index
...
+
+    CHK(pSignalingClient != NULL, STATUS_NULL_ARG);
+
+    // Check if we are already not connected
+    if (ATOMIC_LOAD_BOOL(&pSignalingClient->connected)) {
+        CHK_STATUS(terminateOngoingOperations(pSignalingClient));
+    }
+
+    // move to the fromGetToken() so we can move to the necessary step
+    setStateMachineCurrentState(pSignalingClient->pStateMachine, SIGNALING_STATE_GET_TOKEN);
```
disa6302, line 379: Could be a stretch, but want to make sure we are covering all basis and not missing any data corruption. If GET_TOKEN succeeds in the `createSignalingClient()` call, the service result would be `SERVICE_CALL_RESULT_OK` which is what we check for in `fromGetTokenSignalingState` while deciding state transitions. Just need to ensure there is not a weird scenario where the `pSignalingClient->result` changes to an error service code between `connect` and `fetch` calls, which might lead to the `GET_TOKEN` call being made again.

src/source/Signaling/StateMachine.c
```diff
@@ -778,9 +782,6 @@ STATUS fromDisconnectedSignalingState(UINT64 customData, PUINT64 pState)
 
     CHK(pSignalingClient != NULL && pState != NULL, STATUS_NULL_ARG);
 
-    // See if we need to retry first of all
-    CHK(pSignalingClient->pChannelInfo->reconnect, STATUS_SUCCESS);
```
disa6302, line 782: Purpose of this change?

disa6302 (commented):
The changes are made without taking into account the exponential backoff and recent changes made by @kateyanurag in https://github.com/awslabs/amazon-kinesis-video-streams-webrtc-sdk-c/pull/1341. With this split, where will the exponential retry logic sit? I am assuming it will sit in the fetch call, but how are we handling getToken() retries in the create call in that case? Would it be required to set up exponential backoff in both the calls now?

src/source/Signaling/StateMachine.c
jdelapla: The previous check required an UNKNOWN error code && an expiration. Now we only need an expiration. No need to have an UNKNOWN check elsewhere.

src/source/Signaling/StateMachine.c
jdelapla: Whoops, it's supposed to be pSignalingClient->pAwsCredentials->expiration = 0;

src/source/Signaling/StateMachine.c
jdelapla: It's a bug that leads to unnecessary state transitions. If pChannelInfo->reconnect is true, we'll jump to Ready even if we have an Auth error, which thenj will make the next ConnectSync have an auth error to then force it to jump to GET_TOKEN. The skip is unnecessary.

disa6302: Ok. In that case, shouldnt the jump be conditional? I think it is ok for now, but we will have to look into making this smart.

src/source/Signaling/Signaling.c
jdelapla: That result should only change from those exact calls. No opportunity to change it between them exists.

src/source/Signaling/Client.c
```diff
@@ -173,6 +177,64 @@ STATUS signalingClientConnectSync(SIGNALING_CLIENT_HANDLE signalingClientHandle)
...
+
+        pSignalingClient->clientInfo.signalingClientInfo.stateMachineRetryCountReadOnly = signalingClientInfoInternal.signalingClientInfo.stateMachineRetryCountReadOnly;
+
+        // Wait before attempting to create signaling client
+        CHK_STATUS(pSignalingClient->clientInfo.signalingStateMachineRetryStrategyCallbacks.executeRetryStrategyFn(&createSignalingClientRetryStrategy,
+                                                                                       &signalingClientCreationWaitTime));
+
+        DLOGV("Attempting to back off for [%lf] milliseconds before creating signaling client again. "
+              "Signaling client creation retry count [%d]",
+              retStatus, signalingClientCreationWaitTime / 1000.0, signalingClientCreationMaxRetryCount);
```
hassanctech, line 225: make `1000.0f` if you want to print as a `float`.

use `[%u]` for unsigned not `%d`.

src/source/Signaling/Client.c
```diff
@@ -173,6 +177,64 @@ STATUS signalingClientConnectSync(SIGNALING_CLIENT_HANDLE signalingClientHandle)
...
+    while (TRUE) {
+        retStatus = signalingFetchSync(pSignalingClient);
+        // NOTE: This will retry on all status codes except SUCCESS.
+        // This includes status codes for bad arguments, internal non-recoverable errors etc.
+        // Retrying on non-recoverable errors is useless, but it is quite complex to segregate recoverable
+        // and non-recoverable errors at this layer. So to simplify, we would retry on all non-success status codes.
+        // It is the application's responsibility to fix any validation/null-arg/bad configuration type errors.
+        CHK(retStatus != STATUS_SUCCESS, retStatus);
+
+        DLOGE("Create Signaling Sync API returned [0x%08x]  %d\n", retStatus, signalingClientCreationMaxRetryCount);
```
hassanctech, line 212: use `[%u]` for unsigned not `%d`.

hassanctech (approved):
lgtm!

hassanctech (approved):

PR #1344, opened 2021-12-06
add support for automatic clock skew detection and correction

hassanctech:
*Issue #, if available:*

*Description of changes:*  
Adding Support for Automatic Clock Skew Detection and Correction

The approach taken is to maintain a hashmap from state (which maps 1:1 to an API) to the clock skew offset (both + and - are supported, i.e. if the device time is behind or ahead of the server).  This means that there is a separate offset stored for each API, this approach is similar other aws sdks also do.  The down side is if the system time is off, each API call will fail once, at which point the offset will be stored into a hash map and internally the call result will be updated (from the standard 403 / forbidden received when the request signature time is too far ahead or behind the server time) to indicate which direction the offset is in time.  Next suppose during the execution of this code the system time is updated (perhaps via NTP or some other process) in this case when the request time is "corrected" from the offset in the hash table the API call will again fail with a 403, however while processing the response callback we compare the device now time with the server time in the HTTP response header, once the time on the device is "correct" this delta will now be tiny, we add a check for this case to *remove* the entry in the hash table to indicate that there is no longer an offset for this API so a correction doesn't need to be made to the request before signing it.  Both of these cases are tested and there are tests for a "fast" clock and a "slow" clock.


src/source/Signaling/LwsApiCalls.c
```diff
@@ -93,10 +101,45 @@ INT32 lwsHttpCallbackRoutine(struct lws* wsi, enum lws_callback_reasons reason,
...
+            getStateMachineCurrentState(pSignalingClient->pStateMachine, &pStateMachineState);
 
+            DLOGD("Connected with server response: %d", status);
             pLwsCallInfo->callInfo.callResult = getServiceCallResultFromHttpStatus((UINT32) status);
 
+            lws_hdr_copy(wsi, &dateHdrBuffer[0], MAX_DATE_HEADER_BUFFER_LENGTH, WSI_TOKEN_HTTP_DATE);
+
+            time(&td);
+
+            len = (unsigned int)lws_hdr_total_length(wsi, WSI_TOKEN_HTTP_DATE);
```
disa6302, line 113: `UINT64` instead of `unsigned int`?

src/source/Signaling/LwsApiCalls.c
```diff
@@ -18,17 +18,23 @@ INT32 lwsHttpCallbackRoutine(struct lws* wsi, enum lws_callback_reasons reason,
...
     PSingleListNode pCurNode;
-    UINT64 item;
+    UINT64 item, serverTime;
     UINT32 headerCount;
     UINT32 logLevel;
     PRequestHeader pRequestHeader;
     PSignalingClient pSignalingClient = NULL;
     BOOL locked = FALSE;
+    time_t td;
+    size_t len;
```
disa6302, line 34: `SIZE_T` instead

src/source/Signaling/LwsApiCalls.c
```diff
@@ -93,10 +101,45 @@ INT32 lwsHttpCallbackRoutine(struct lws* wsi, enum lws_callback_reasons reason,
 
         case LWS_CALLBACK_ESTABLISHED_CLIENT_HTTP:
             status = lws_http_client_http_response(wsi);
-            DLOGD("Connected with server response: %d", status);
+            getStateMachineCurrentState(pSignalingClient->pStateMachine, &pStateMachineState);
 
+            DLOGD("Connected with server response: %d", status);
             pLwsCallInfo->callInfo.callResult = getServiceCallResultFromHttpStatus((UINT32) status);
 
+            lws_hdr_copy(wsi, &dateHdrBuffer[0], MAX_DATE_HEADER_BUFFER_LENGTH, WSI_TOKEN_HTTP_DATE);
```
disa6302, line 109: What if the copy fails? I believe this call returns length if success and -1 if it fails. Can we do this instead?

`len = lws_hdr_copy(wsi, &dateHdrBuffer[0], MAX_DATE_HEADER_BUFFER_LENGTH, WSI_TOKEN_HTTP_DATE);`. This will knock out line 113 requirement. Correct me if I misunderstood this.

src/source/Signaling/LwsApiCalls.c
```diff
@@ -832,6 +924,12 @@ STATUS createChannelLws(PSignalingClient pSignalingClient, UINT64 time)
                                  SIGNALING_SERVICE_API_CALL_CONNECTION_TIMEOUT, SIGNALING_SERVICE_API_CALL_COMPLETION_TIMEOUT,
                                  DEFAULT_LOW_SPEED_LIMIT, DEFAULT_LOW_SPEED_TIME_LIMIT, pSignalingClient->pAwsCredentials, &pRequestInfo));
 
+    if (pSignalingClient->signalingClientCallbacks.getCurrentTimeFn != NULL) {
+        pRequestInfo->currentTime = pSignalingClient->signalingClientCallbacks.getCurrentTimeFn(pSignalingClient->signalingClientCallbacks.customData);
+    }
+
+    checkAndCorrectForClockSkew(pSignalingClient, pRequestInfo);
```
disa6302, line 931: Shouldnt we be checking for `STATUS` here? What happens in the event `hashTableGet` fails?

src/source/Signaling/Signaling.h
```diff
@@ -133,6 +147,7 @@ typedef struct {
     UINT64 connectTime;
     UINT64 cpApiLatency;
     UINT64 dpApiLatency;
+    PHashTable pEndpointToClockSkewHashMap;
```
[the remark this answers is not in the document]
hassanctech: Agree

src/source/Signaling/LwsApiCalls.c
hassanctech: I copied the usage from a sample or a test in lws, I can't find it now, this is similar:

https://github.com/warmcat/libwebsockets/blob/0e9842a0c0e4ab3d0bbf63536d7f9f4b5c3e7c8b/lib/roles/http/date.c#L161

Let me look more into this I'm not 100% sure yet if the return value of `lws_hdr_copy` is the same as what `lws-hdr_total_length` returns

src/source/Signaling/LwsApiCalls.c
hassanctech: I intentionally left it out, only one of two things happen, either its in the hash table (which means we apply the offset, there's no status failures possible in that route), other case is its not in the hash table in which case it CHK_STATUS will return some non success status but all that means is that there was no clock skew so nothing to correct.

src/source/Signaling/LwsApiCalls.c
hassanctech: I looked at the implmenetation and yes `lws_hdr_copy` returns exactly the same thing as `lws_hdr_total_length` so I'll make that simplification

disa6302 (approved):
Looks great!

PR #1345, opened 2021-12-07
Log git hash

disa6302:
*Issue #, if available:*

*Description of changes:*
This will log the git hash being used. Will be useful when customers attach logs and we would not have to request git hash to debug issues.


hassanctech (approved):
awesomeness!

PR #1362, opened 2021-12-31
Mark datachannel specific code explicitly

widgetii:
The library has ENABLE_DATA_CHANNEL CMake option which seems doesn't work as planned (I thought it meant to disable datachannel specific code that uses a rather large usrsctp library). It turns out that setting the option to ON is not sufficient because compiled code still has external links to routines in usrsctp library and doesn't compile without it.

In our project, we use the library on devices with constrained RAM and ROM (IP cameras) and we don't use datachannel functions at all. Disabling unnecessary functions saves about 320Kb of the resulting binary compiled statically. I hope this will be useful for others struggling with space on embedded devices.


ycyang1229:
@widgetii Maybe you want to take a look at this PR, https://github.com/awslabs/amazon-kinesis-video-streams-webrtc-sdk-c/pull/846.

widgetii:
Thank you for sharing, I missed it. It turns out this feature needs more than one person I guess. Anyway, the current codebase is broken because ENABLE_DATA_CHANNEL option does nothing actually.

disa6302:
@widgetii , 

Thank you for your PR. You are right. It does look like the flag has been pointless so far. I have restarted failed builds on travis. Our development strategy has changed. Can you open this PR against origin/develop branch please? For more information, follow this guide: https://github.com/awslabs/amazon-kinesis-video-streams-webrtc-sdk-c#development

niyatim23:
This change has been merged through another PR where I applied your commits. This PR has a lot of unintended commits which is why I could not merge this one. Thanks for your contribution

PR #1416, opened 2022-02-24
Migrate Doxygen to GitHub Actions

niyatim23:
*Issue #, if available:*

*Description of changes:*


.github/workflows/ci.yml
```diff
@@ -268,6 +268,7 @@ jobs:
           ulimit -c unlimited -S
           timeout --signal=SIGABRT 60m ./tst/webrtc_client_test
   sample-check:
+    if: github.repository == 'awslabs/amazon-kinesis-video-streams-webrtc-sdk-c'
```
disa6302, line 271: Why are we doing this? We can modify the script instead: https://github.com/awslabs/amazon-kinesis-video-streams-webrtc-sdk-c/blob/2dff44fea83a79336fda676ba2801d0d77417cd1/scripts/check-sample.sh#L3. Replace with checking for `AWS_REGION` or `AWS_ROLE_TO_ASSUME` or `AWS_ROLE_SESSION_NAME`?

.github/workflows/doxygen-gh-pages.yml
```diff
@@ -0,0 +1,28 @@
+name: Doxygen GitHub Pages Deploy Action WebRTC C SDK
+
+on:
+  push:
+    branches:
+      - master
```
disa6302, line 6: Include develop as well

disa6302 (approved):

.github/workflows/ci.yml
niyatim23: We are doing this because the job fails way before that on the "Configure AWS Credentials" step for forked branches. Even if we move that step right before the script execution, it would fail on this step because GitHub does not allow accessing secrets for forked branches at all

disa6302: Ok. Makes sense. How does setting repo work in itself? How does this detect a forked branch?

.github/workflows/doxygen-gh-pages.yml
niyatim23: Included in the latest commit

.github/workflows/ci.yml
niyatim23: I referred [this](https://github.community/t/stop-github-actions-running-on-a-fork/17965/2)

disa6302 (approved):

Issue #1437, opened 2022-03-24
A confusing key searching logic in JitterBuffer.c


northteam:
https://github.com/awslabs/amazon-kinesis-video-streams-webrtc-sdk-c/blob/2dff44fea83a79336fda676ba2801d0d77417cd1/src/source/PeerConnection/JitterBuffer.c#L223-L230
if `retStatus` is `STATUS_HASH_KEY_NOT_PRESENT`, then `hashValue` will be untouched, which is `0`. And `pCurPacket` will be `NULL` and cannot be dereferenced. So why treat `STATUS_HASH_KEY_NOT_PRESENT` as a success?

jdelapla:
Hello! 
The reason for this odd looking if statement is that we only want line 228 to exit on a truly critical internal error, like the hashtable doesn't exist. There is also an important check on line 216:
`216         CHK_STATUS(hashTableContains(pJitterBuffer->pPkgBufferHashTable, index, &hasEntry));`
This **should** make STATUS_HASH_KEY_NOT_PRESENT an unreachable error, unless we're dealing with some very obscure bug. 

As for the `NULL` outcome, you're correct that would be the case if this was the very first packet we've iterated through. I'll add a check for this scenario. In all other cases we would send a duplicate packet- also not ideal, I think the correct error handling would be to call `continue;` and go to the next index. I'll send a PR shortly.

PR #1443, opened 2022-03-30
Addressing error handling concerns for an edge case

jdelapla:
*Issue #, if available:*
https://github.com/awslabs/amazon-kinesis-video-streams-webrtc-sdk-c/issues/1437
*Description of changes:*

### DIFF
--- a/src/source/PeerConnection/JitterBuffer.c
+++ b/src/source/PeerConnection/JitterBuffer.c
@@ -221,13 +221,11 @@ STATUS jitterBufferInternalParse(PJitterBuffer pJitterBuffer, BOOL bufferClosed)
         } else {
             lastNonNullIndex = index;
             retStatus = hashTableGet(pJitterBuffer->pPkgBufferHashTable, index, &hashValue);
-            if (retStatus == STATUS_SUCCESS) {
-                retStatus = STATUS_SUCCESS;
-            } else if (retStatus == STATUS_HASH_KEY_NOT_PRESENT) {
+            if (retStatus == STATUS_HASH_KEY_NOT_PRESENT) {
                 // should be unreachable, this means hashTablContains() said we had it but hashTableGet() couldn't find it.
                 continue;
             } else {
-                CHK(FALSE, retStatus);
+                CHK_STATUS(retStatus);
             }
             pCurPacket = (PRtpPacket) hashValue;
             CHK(pCurPacket != NULL, STATUS_NULL_ARG);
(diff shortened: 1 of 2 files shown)


src/source/PeerConnection/JitterBuffer.c
```diff
@@ -221,12 +221,16 @@ STATUS jitterBufferInternalParse(PJitterBuffer pJitterBuffer, BOOL bufferClosed)
         } else {
             lastNonNullIndex = index;
             retStatus = hashTableGet(pJitterBuffer->pPkgBufferHashTable, index, &hashValue);
-            pCurPacket = (PRtpPacket) hashValue;
-            if (retStatus == STATUS_SUCCESS || retStatus == STATUS_HASH_KEY_NOT_PRESENT) {
+            if (retStatus == STATUS_SUCCESS) {
                 retStatus = STATUS_SUCCESS;
```
MushMal, line 225: This is a no-op. Perhaps you need the following

CHK(retStatus == STATUS_SUCCESS || retStatus == STATUS_HASH_KEY_NOT_PRESENT, retStatus);
if (retStatus == STATUS_HASH_KEY_NOT_PRESENT) {
    retStatus = STATUS_SUCCESS;
} else {
    // Do the rest of the function. Try to avoid loop controls within a larger loops such as break; and continue; or restructure the loop
}

src/source/PeerConnection/JitterBuffer.c
```diff
@@ -344,13 +350,8 @@ STATUS jitterBufferFillFrameData(PJitterBuffer pJitterBuffer, PBYTE pFrame, UINT
     CHK(pJitterBuffer != NULL && pFrame != NULL && pFilledSize != NULL, STATUS_NULL_ARG);
     for (; UINT16_DEC(index) != endIndex; index++) {
         hashValue = 0;
-        retStatus = hashTableGet(pJitterBuffer->pPkgBufferHashTable, index, &hashValue);
+        CHK_STATUS(hashTableGet(pJitterBuffer->pPkgBufferHashTable, index, &hashValue));
```
MushMal, line 353: This is a behavior change in case the key is not present. Would be nice to comment why we the missing hash key is not expected.

jdelapla: This is not a behavior change, it just changes which error code. If you returned STATUS_HASH_KEY_NOT_PRESENT then on line 354 you would return STATUS_NULL_ARG every time.

disa6302 (approved):

PR #1489, opened 2022-06-07
isolate Signaling related code

northteam:
If you don't need signaling functionality, you don't need to download or
compile its dependency like libwebsockets, kvsCommonLws.

*Issue #, if available:*
None

*Description of changes:*
This commit separate Signaling related code from the other part of the SDK. It give us a chance to remove libwebsockets/kvsCommonLws dependency, so that we can use some other signaling mechanism.

### DIFF
--- a/src/source/Include_i.h
+++ b/src/source/Include_i.h
@@ -52,8 +52,6 @@ extern "C" {
 #define INET6 1
 #include <usrsctp.h>
 
-#include <libwebsockets.h>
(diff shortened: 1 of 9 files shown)


jdelapla:
Signaling is a core part of the KVS WebRTC open-source project. After internal deliberation in KVS team, we have concluded to not take in this PR as it doesn't align with the project goals.

PR #1543, opened 2022-08-26
Fix the problem that the session is not released when handleOffer processing fails

Kelephant:
Fix the problem that the session is not released when handleOffer processing fails

### DIFF
--- a/samples/Common.c
+++ b/samples/Common.c
@@ -1288,9 +1288,6 @@ STATUS signalingMessageReceived(UINT64 customData, PReceivedSignalingMessage pRe
             CHK_STATUS(createSampleStreamingSession(pSampleConfiguration, pReceivedSignalingMessage->signalingMessage.peerClientId, TRUE,
                                                     &pSampleStreamingSession));
             pSampleStreamingSession->offerReceiveTime = GETTIME();
-            MUTEX_LOCK(pSampleConfiguration->streamingSessionListReadLock);
-            pSampleConfiguration->sampleStreamingSessionList[pSampleConfiguration->streamingSessionCount++] = pSampleStreamingSession;
-            MUTEX_UNLOCK(pSampleConfiguration->streamingSessionListReadLock);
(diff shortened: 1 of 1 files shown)


samples/Common.c
```diff
@@ -1304,7 +1301,11 @@ STATUS signalingMessageReceived(UINT64 customData, PReceivedSignalingMessage pRe
                 // NULL the pointer to avoid it being freed in the cleanup
                 pPendingMessageQueue = NULL;
             }
-
+            
+            MUTEX_LOCK(pSampleConfiguration->streamingSessionListReadLock);
```
MushMal, line 1305: I wonder if any of the code above expects the session to be in the session list. If so, I would recommend having a decrement in the cleanup clause. The cleanup could be predicated on whether the session has been incremented in the first place, whether there was an error condition and whether the offer was handled or not.

Kelephant: Yes. My changes are not perfect.
I want to add the session to the list only when the process is successful. 
When the processing fails, the invalid session will occupy the session list. As a result, the number of sessions will be greater than DEFAULT_MAX_CONCURRENT_STREAMING_SESSION, and a new connection cannot be established.

MushMal: Understandably. The fix is a good one but it's incomplete. I checked the code quickly and it's OK to move the streaming session list setting to the bottom of the clause like you've done. However, if there is an error prior to the adding to the list, the newly created streaming session will be leaked. Recommend having an boolean such as freeStreamingSession = TRUE; After the mutex is released in the change you've made, when the session is already added to the list, you set the freeStreamingSession = FALSE.

Later, in the cleanup, before returning, you can add the following cleanup clause

if (freeStreamingSession && pSampleStreamingSession != NULL) {
    freeSampleStreamingSession(pSampleStreamingSession);
}

niyatim23:
Closing in favor of the linked PR

Issue #1560, opened 2022-09-27
[BUG] Fix possible dead lock when freeSampleStreamingSession (getIceCandidatePairStatsCallback)


RufUsul:
Describe the bug
I found a possible deadlock: freeSampleStreamingSession() may deadlock in interaction with getIceCandidatePairStatsCallback()

SDK version number
1.7.3

Additional context
I have an fix and just wanted to post an issue before the PR. Bug and fix are very similar to #1448

disa6302:
Thank you! PR merged.

PR #1561, opened 2022-09-27
Fix deadlock getIceCandidatePairStatsCallback

RufUsul:
*Issue #1560*

*Description of changes:*
avoid deadlock via MUTEX_TRYLOCK and just try again later via timer

### DIFF
--- a/samples/Common.c
+++ b/samples/Common.c
@@ -864,10 +864,13 @@ STATUS getIceCandidatePairStatsCallback(UINT32 timerId, UINT64 currentTime, UINT
 
     pSampleConfiguration->rtcIceCandidatePairMetrics.requestedTypeOfStats = RTC_STATS_TYPE_CANDIDATE_PAIR;
 
-    // We need to execute this under the object lock due to race conditions that it could pose
-    MUTEX_LOCK(pSampleConfiguration->sampleConfigurationObjLock);
-    locked = TRUE;
-
+    // Use MUTEX_TRYLOCK to avoid possible dead lock when canceling timerQueue
+    if (!MUTEX_TRYLOCK(pSampleConfiguration->sampleConfigurationObjLock)) {
+        return retStatus;
+    } else {
+        locked = TRUE;
+    }
+    
     for (i = 0; i < pSampleConfiguration->streamingSessionCount; ++i) {
         if (STATUS_SUCCEEDED(rtcPeerConnectionGetMetrics(pSampleConfiguration->sampleStreamingSessionList[i]->pPeerConnection, NULL,
                                                          &pSampleConfiguration->rtcIceCandidatePairMetrics))) {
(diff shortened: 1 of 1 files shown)


samples/Common.c
```diff
@@ -864,9 +864,12 @@ STATUS getIceCandidatePairStatsCallback(UINT32 timerId, UINT64 currentTime, UINT
 
     pSampleConfiguration->rtcIceCandidatePairMetrics.requestedTypeOfStats = RTC_STATS_TYPE_CANDIDATE_PAIR;
 
-    // We need to execute this under the object lock due to race conditions that it could pose
-    MUTEX_LOCK(pSampleConfiguration->sampleConfigurationObjLock);
-    locked = TRUE;
+    // Use MUTEX_TRYLOCK to avoid possible dead lock when canceling timerQueue
+    if (!MUTEX_TRYLOCK(pSampleConfiguration->sampleConfigurationObjLock)) {
+        return retStatus;
```
MushMal, line 869: Per code conventions, I would recommend using the following construct
CHK(FALSE, retStatus);

samples/Common.c
```diff
@@ -864,9 +864,12 @@ STATUS getIceCandidatePairStatsCallback(UINT32 timerId, UINT64 currentTime, UINT
...
+    } else {
```
MushMal, line 870: No need for else clause - just set the locked = TRUE as you are locked at this stage.

MushMal (commented):
Stylistic comments. I am still not entirely sure what the precondition for the deadlock is but this change is good at any rate.

samples/Common.c
RufUsul: thought so but preferred to copy & paste the code from pregenerateCertTimerCallback()
should i change both?

samples/Common.c
RufUsul: thought so but preferred to copy & paste the code from pregenerateCertTimerCallback()
should i change both?

samples/Common.c
MushMal: I would recommend keeping the changes as local as possible. I don't see this as a major issue and for some reason there is less consistency in the samples than in the library code.

samples/Common.c
MushMal: Same comment. It's a minor issue that I am OK with. These elements were put in place and followed in the main library code for a reason of readability and better forward compatibility. I think it's up to you to decide.

MushMal (approved):

samples/Common.c
RufUsul: ok, then lets keep the copy & pasted code

samples/Common.c
RufUsul: same, then lets keep the copy & pasted code

disa6302 (approved):

PR #1577, opened 2022-10-19
Add the feature of "Storage for WebRTC."

ycyang1229:
*Issue #, if available:*

*Description of changes:*
Add the new feature of "Storage for WebRTC".


MushMal:
This is a massive change in terms of functionality added but there is zero documentation as to what's added to the backend API, etc.,. It would have been great if this project was run more like an open source project rather than just publicly exposed proprietary repo.

ycyang1229:
> This is a massive change in terms of functionality added but there is zero documentation as to what's added to the backend API, etc.,. It would have been great if this project was run more like an open source project rather than just publicly exposed proprietary repo.

@MushMal Sorry for late response. Right now, this new feature is evolving, so it is not one finalized version. I know it is bad not to have any explanation here. I will add one section into readme and also indicate the link of the official website documents, after the feature is finalized.

PR #1668, opened 2023-03-17
Fix the problem that the session is not released when handleOffer processing fails

niyatim23:
Issue #, if available:
https://github.com/awslabs/amazon-kinesis-video-streams-webrtc-sdk-c/pull/1543

*What was changed?*
Customer contribution to fix the problem that the session is not released when handleOffer processing fails.  

*Why was it changed?*
Previously the streamingSessionCount would be incremented right after creating the session irrespective of the success / failure of handleOffer. 

*How was it changed?*
This change moves the incrementing lines after the success of the session. The bool flag is introduced so that if the session fails, it can be freed.

Tested locally that the sample works. Forced handleOffer failure to make sure that the streaming session is freed if it cannot be used


benwkim (approved):

disa6302 (dismissed):

samples/Common.c
```diff
@@ -1411,6 +1414,10 @@ STATUS signalingMessageReceived(UINT64 customData, PReceivedSignalingMessage pRe
         freeMessageQueue(pPendingMessageQueue);
     }
 
+    if (freeStreamingSession && pSampleStreamingSession != NULL) {
```
disa6302, line 1417: If handleOffer fails, `freeStreamingSession` would be false which means this still wouldnt be freed right? Also, shouldnt we free sample streaming session in case of any call failure?

jdelapla: +1

samples/Common.c
```diff
@@ -1330,6 +1327,12 @@ STATUS signalingMessageReceived(UINT64 customData, PReceivedSignalingMessage pRe
                 pPendingMessageQueue = NULL;
             }
 
+            MUTEX_LOCK(pSampleConfiguration->streamingSessionListReadLock);
+            freeStreamingSession = TRUE;
```
jdelapla, line 1331: What is the purpose of this bool being between the mutex_lock? Isn't the purpose of this bool to check for goto jumps from the CHK macros?

samples/Common.c
```diff
@@ -1313,9 +1313,6 @@ STATUS signalingMessageReceived(UINT64 customData, PReceivedSignalingMessage pRe
             CHK_STATUS(createSampleStreamingSession(pSampleConfiguration, pReceivedSignalingMessage->signalingMessage.peerClientId, TRUE,
                                                     &pSampleStreamingSession));
             pSampleStreamingSession->offerReceiveTime = GETTIME();
-            MUTEX_LOCK(pSampleConfiguration->streamingSessionListReadLock);
-            pSampleConfiguration->sampleStreamingSessionList[pSampleConfiguration->streamingSessionCount++] = pSampleStreamingSession;
-            MUTEX_UNLOCK(pSampleConfiguration->streamingSessionListReadLock);
 
             CHK_STATUS(handleOffer(pSampleConfiguration, pSampleStreamingSession, &pReceivedSignalingMessage->signalingMessage));
```
jdelapla, line 1317: Here is handleOffer, shouldn't the bool be set somewhere in proximity to this function?

jdelapla (changes requested):

samples/Common.c
niyatim23: Addressed this by moving the bool right after creating the streaming session

samples/Common.c
niyatim23: Addressed

samples/Common.c
niyatim23: Addressed by moving it right after the creation

disa6302 (approved):

jdelapla (approved):

PR #1536, opened 2022-08-17
fix turn allocation taken long time when credential is incorrect

vc60er:
*Issue #, if available:*
turn allocation taken long time when credential is incorrect. It taken about 10s+ until the local candidate gathering timeout

*Description of changes:*
1. add trunStateFailedFn, called when turn state change to TURN_STATE_FAILED, and then update pNewCandidate state from ICE_CANDIDATE_STATE_NEW to ICE_CANDIDATE_STATE_INVALID
2. add stateTryCountMax and stateTryCount, limit the allocation times not more than stateTryCountMax

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.

PR #1674, opened 2023-03-27
update Candidate state to invalid when turn failed

vc60er:
*Issue #, if available:*
1. When pTurnConnection->state failed, turnConnectionTimerCallback did not stop correctly

**reason**: 
The state cannot be changed from TURN_STATE_CLEAN_UP to TURN_STATE_FAILED, causing turnConnectionTimerCallback cannot exit normally

For example, in the following steps, this problem will be reproduced:

1). When an error occurs in turnConnectionStepState, such as socketConnectionInitSecureConnection returning retStatus(0x15000020),
2). Then in turnConnectionStepState(line:1113), set pTurnConnection->state = TURN_STATE_FAILED;
3). In turnConnectionStepState(line:1090), set pTurnConnection->state = TURN_STATE_CLEAN_UP;
4). In turnConnectionStepState(line:1078), set pTurnConnection->state = TURN_STATE_FAILED;
5). When IceAgent state step into ICE_AGENT_STATE_READY, execute iceAgentReadyStateSetup(line:2134), call turnConnectionShutdown(line:1163), set pTurnConnection->stopTurnConnection=true
6). In turnConnectionStepState(line:1106), set pTurnConnection->state = TURN_STATE_CLEAN_UP;

As above:

Steps 3), 4), and 6) appear repeatedly, The value of pTurnConnection->state alway update to TURN_STATE_CLEAN_UP when turnConnectionStepState returned, but turnConnectionTimerCallback need it updat to TURN_STATE_FAILED then can exit itself(line:1314)

2. If the network is slow, or for some other reasons, the relayAddress is not obtained within 150ms, the process in state of TURN_STATE_ALLOCATION will fail retStatus(0x1500002a)(line:999)

**Reason:** 
Only 3 times of turn allocation are allowed now, but in reality is 3 times of allocation packets are allowed to be sent, and then it is check whether the allocation is successful. This process is equivalent determine whether the allocation is successful after 150ms (turnConnectionTimerCallback is executed every 50ms). Network latency is not considered. If there is a large delay on the network, it is difficult to complete the allocation within 150ms

some code:

```
pTurnConnection->stateTryCountMax = DEFAULT_TURN_ALLOCATION_MAX_TRY_COUNT; //3
pTurnConnection->stateTryCount = 0;

CHK(pTurnConnection->stateTryCount < pTurnConnection->stateTryCountMax, STATUS_TURN_CONNECTION_ALLOCAITON_FAILED);
```

*Description of changes:*

1. For fix Issue1: Add condition checking in turnConnectionStepState(line:1106): !ATOMIC_LOAD_BOOL(&pTurnConnection->shutdownComplete), to ensure the value pTurnConnection->state not update to TURN_STATE_CLEAN_UP repeatedly.

2. For fix Issue2: There is already a turn allocation timeout checking in turnConnectionStepState(line:1000):, so it is no longer necessary to checking the times of allocations to terminate the allocation process.

3. For fix [1536](https://github.com/awslabs/amazon-kinesis-video-streams-webrtc-sdk-c/pull/1536): When pTurnConnection->state failed, the corresponding process of update pIceCandidate->state to invalid is placed into iceAgentGatherCandidateTimerCallback instead of callback funtion, just like update it to valid when 
turnConnection get relay addres(line:1544), which is more elegant.

### DIFF
--- a/src/source/Ice/TurnConnectionStateMachine.c
+++ b/src/source/Ice/TurnConnectionStateMachine.c
@@ -108,7 +108,7 @@ STATUS stepTurnConnectionStateMachine(PTurnConnection pTurnConnection)
         retStatus = stepStateMachine(pTurnConnection->pStateMachine);
 
         if (STATUS_SUCCEEDED(retStatus) && ATOMIC_LOAD_BOOL(&pTurnConnection->stopTurnConnection) && pTurnConnection->state != TURN_STATE_NEW &&
-            pTurnConnection->state != TURN_STATE_CLEAN_UP) {
+            pTurnConnection->state != TURN_STATE_CLEAN_UP && !ATOMIC_LOAD_BOOL(&pTurnConnection->shutdownComplete)) {
             currentTime = GETTIME();
             pTurnConnection->state = TURN_STATE_CLEAN_UP;
             pTurnConnection->stateTimeoutTime = currentTime + DEFAULT_TURN_CLEAN_UP_TIMEOUT;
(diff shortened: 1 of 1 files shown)


vc60er:
> Hey @vc60er, thanks for the contribution! Could you fix the clang-format please?
> 
> ```
> Performing clang format compliance check....
> src/include/com/amazonaws/kinesis/video/webrtcclient/Include.h failed clang-format check.
> src/include/com/amazonaws/kinesis/video/webrtcclient/Stats.h failed clang-format check.
> src/source/Ice/IceAgent.c failed clang-format check.
> src/source/PeerConnection/SessionDescription.c failed clang-format check.
> src/source/Rtcp/RtcpPacket.h failed clang-format check.
> Error: Process completed with exit code 1.
> ```

@sirknightj done.

src/source/Ice/IceAgent.c
```diff
@@ -1543,6 +1543,9 @@ STATUS iceAgentGatherCandidateTimerCallback(UINT32 timerId, UINT64 currentTime,
                 /* Check if any relay address has been obtained. */
                 CHK_STATUS(updateCandidateAddress(pIceCandidate, &relayAddress));
                 CHK_STATUS(createIceCandidatePairs(pIceAgent, pIceCandidate, FALSE));
+            } else if (pIceCandidate->iceCandidateType == ICE_CANDIDATE_TYPE_RELAYED && pIceCandidate->pTurnConnection != NULL &&
```
jdelapla, line 1546: This check should come before the previous one. Additionally since this if statement and the else if above share 2 conditions please set them in a common if/else block.

PR #1691, opened 2023-04-11
Fix build dependencies off and add version requirements in readme

disa6302:
*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/README.md
+++ b/README.md
@@ -64,12 +64,20 @@ GStreamer is installed on your system.
 
 On Ubuntu and Raspberry Pi OS you can get the libraries by running
 ```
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -64,12 +64,20 @@ GStreamer is installed on your system.
...
 ```
 
 By default we download all the libraries from GitHub and build them locally, so should require nothing to be installed ahead of time.
 If you do wish to link to existing libraries you can use the following flags to customize your build.
 
+### Dependency requirements
+
+These would be applicable if the SDK is being linked with system dependencies instead of building from source by the SDK.
+`libmbedtls`: `>= 2.25.0 & < 3.x.x`
+`libopenssl`: `>= 1.1.1t`
```
hassanctech, line 77: We should be careful about openssl I don't think we actual need it to be at or near this version.  We picked the latest version with CVE's addressed + builds on arm64 for mac.  But strictly speaking at runtime this is not needed, probably any 1.1.1x should be fine.  I imagine some embedded devices where this is running may not be able to update the system library if a firmware revision isn't available and they still want to use the system openssl.

hassanctech (approved):
lgtm

PR #1699, opened 2023-04-21
Fix compilation of OpenSSL 1.1.1t on Raspberry Pi 4B with 64-bit

benmesander:
Description of changes:

Raspberry pi 4B has a 64-bit ARM cpu, but the camera module is not currently supported with a 64-bit userland. So the default raspbian configuration is a 64-bit kernel with a 32-bit userland. This is not correctly identified by OpenSSL version 1.1.1t as used in the Amazon Kinesis Video Streams with WebRTC SDK. It results in an attempt to use 64-bit assembly opcodes in a 32-bit assembler. So in this case if the BUILD_OPENSSL_PLATFORM is not set, set it to linux-armv4, which will cause OpenSSL to select the correct assembly module for the 32-bit Raspberry Pi userland.

OpenSSL issue related to this: https://github.com/openssl/openssl/issues/20804

### DIFF
--- a/CMake/Dependencies/libopenssl-CMakeLists.txt
+++ b/CMake/Dependencies/libopenssl-CMakeLists.txt
@@ -16,6 +16,22 @@ else()
     SET(OPENSSL_EXTRA ${OPENSSL_EXTRA} no-shared no-dso)
   endif()
 
+  # Raspberry pi 4 has a 64-bit ARM cpu, but the camera module is not
+  # currently supported with a 64-bit userland. So the default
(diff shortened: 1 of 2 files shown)


disa6302:
@benmesander ,

Thank you for your PR. This seems to be platform specific and we like to keep the build files as generic and customizable as possible. Instead of modifying the cmake file, can you add a one liner in the [Debug section](https://github.com/awslabs/amazon-kinesis-video-streams-webrtc-sdk-c#debug) highlighting what specific flag needs to be set for Raspberry Pi 4B platform/.

PR #1721, opened 2023-06-05
IoT CertificateId as Stream Name.

chenln1124:
*Issue #, if available:*

*Description of changes:*


niyatim23 (commented):
Hi @chenln1124, why do you intend to use certificate ID as stream name? We have a default stream name in case nothing is specified

niyatim23 (approved):

PR #1739, opened 2023-06-27
[BUG] Set a default user agent if not provided

disa6302:
*Issue #, if available:*

*Description of changes:*

What was changed?
-  By default, the user agent string for the WebRTC SDK was set to`AWS-SDK-KVS/...`. Set a WebRTC SDK specific default user agent.

Why was it changed?
- When trying to debug a problem that requires checking backend, it is impossible to understand the source of the request since this SDK depends on producer SDK. This change helps with the differentiation and simplify debugging

How was it changed?
- Setting the default format to: `AWS-WEBRTC-KVS-AGENT/<version>`
- Changed the CMake to set project version which is read to set the user agent string. This would have to be changed before every release

Testing:

- Added unit test to test out default agent string and supplied agent string
- Tested locally with samples

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -4,7 +4,8 @@ include(Utilities)
 include(CheckIncludeFiles)
 include(CheckFunctionExists)
 
-project(KinesisVideoWebRTCClient LANGUAGES C)
+# The version MUST be updated before every release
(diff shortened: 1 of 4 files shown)


hassanctech (approved):
lgtm

PR #1744, opened 2023-07-06
Connection timeout sleep

jdelapla:
**Description of changes:**
ConnectionListener destructor no longer waits an arbitrary amount of time before forcibly freeing all allocated objects.
Instead it will wait for the executing thread to exit. Added an internal pipe()/socketpair() in order to forcibly kick the executing thread out of poll() when the destructor is invoked.

**What was changed?**
The timeout is now a thread_join on the executing thread and a write to a pipe()/socketpair() was added before the join() to encourage the executing thread to hurry it up. Because Windows doesn't support socketpair() or pipe(), that feature is currently unix only. The consequence for windows is that it will have to wait a bit longer for poll() to timeout (currently 200ms).

**Why was it changed?**
The original implementation would crash if the arbitrary timeout was exceeded. We received customer reports that this was happening. They previously attempted to patch this by increasing the timeout, but to no avail.

**Testing:**
Edited existed unit tests to account for THREAD_JOIN instead of guessing a timeout


hassanctech:
@jdelapla Can you please add a high level overview of this change, like what problem it's solving and a brief description of the approach take.

disa6302:
> @jdelapla Can you please add a high level overview of this change, like what problem it's solving and a brief description of the approach take.

You can reference the template for PR here: https://github.com/awslabs/amazon-kinesis-video-streams-webrtc-sdk-c/pull/1739

jdelapla:
Updated the original description to aide the reviewers.

src/source/Ice/ConnectionListener.c
```diff
@@ -56,31 +62,35 @@ STATUS freeConnectionListener(PConnectionListener* ppConnectionListener)
...
-            }
+        MUTEX_LOCK(pConnectionListener->lock);
+        threadId = pConnectionListener->receiveDataRoutine;
+        MUTEX_UNLOCK(pConnectionListener->lock);
 
-            // Allow the thread to finish and exit
-            if (!threadTerminated) {
-                THREAD_SLEEP(KVS_ICE_SHORT_CHECK_DELAY);
-            }
-        } while (!threadTerminated && GETTIME() < timeToWait);
```
disa6302, line 75: Curious. What caused this block to crash with the timeout approach? Also since this is not available in Windows, there is a chance of crash on Windows correct?

src/source/Ice/ConnectionListener.c
```diff
@@ -56,31 +62,35 @@ STATUS freeConnectionListener(PConnectionListener* ppConnectionListener)
...
+        // TODO add support for windows socketpair
+        // This writes to the socketpair, kicking the POLL() out early,
+        // otherwise wait for the POLL to timeout
+#ifndef _WIN32
+        socketWrite(pConnectionListener->kickSocket[CONNECTION_LISTENER_KICK_SOCKET_WRITE], msg, strlen(msg));
```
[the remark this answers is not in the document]
disa6302: Oh also, socketWrite() would fail if createSocketPair() fails, correct? Since the kickSocket fd is -1?

src/source/Ice/ConnectionListener.c
jdelapla: 1. The crash was caused from the connectionListenerReceiveDataRoutine access the ConnectionListener object after the destructor has freed it. The thread would still be running after the timeout occurs.

2. No, the thread_join() is the crux of the fix, since it will have the destructor wait indefinitely for the thread to close. The part windows is missing is the socketpair() kicking poll() out early, so in the worst case windows will have to wait 200ms more.

src/source/Ice/ConnectionListener.c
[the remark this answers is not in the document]
jdelapla: createConnectionListener() would return an error if createSocketPair() fails which would free the object, so there would be a `ConnectionListener` to use.

I'll change strlen -> STRLEN

src/source/Ice/ConnectionListener.c
```diff
@@ -292,9 +301,20 @@ PVOID connectionListenerReceiveDataRoutine(PVOID arg)
...
+        if (nfds != 0) {
+            // TODO add support for socketpair() in windows
+            // This end of the socketpair has been added to the list of sockets polled
+            // in order to have a way to end the poll early from the destructor
+#ifndef _WIN32
+            rfds[nfds].fd = pConnectionListener->kickSocket[CONNECTION_LISTENER_KICK_SOCKET_LISTEN];
+            rfds[nfds].events = POLLIN;
+            rfds[nfds].revents = 0;
+            nfds++;
+#endif
```
hassanctech, line 314:
```
    struct pollfd rfds[CONNECTION_LISTENER_DEFAULT_MAX_LISTENING_CONNECTION];
```
The loop above iterates as follows:

```
for (i = 0, socketCount = 0; i < CONNECTION_LISTENER_DEFAULT_MAX_LISTENING_CONNECTION; i++)
```

Which means it is possible that `nfds` could end up equal to `CONNECTION_LISTENER_DEFAULT_MAX_LISTENING_CONNECTION` when the `for` loop exits, which means we would be trying to access out of bounds here.  

Perhaps a safe solution here is to define `rfds` as follows:

```
    struct pollfd rfds[CONNECTION_LISTENER_DEFAULT_MAX_LISTENING_CONNECTION + 1];
```

disa6302 (dismissed):

disa6302 (approved):

PR #1761, opened 2023-07-24
Include and use threadpool for signaling channel messages

jdelapla:
*Description of changes:*
Adding threadpool that is now supported from PIC. The threadpool is a part of the SignalingClient and is used to manage the threads handling Signaling Messages. 

### DIFF
--- a/README.md
+++ b/README.md
@@ -348,6 +348,11 @@ When building on MacOS M1, if the build fails while trying to build OpenSSL or W
 To build on a 32-bit Raspbian GNU/Linux 11 on 64-bit hardware, the OpenSSL library must be manually configured. This is due to the OpenSSL autoconfiguration script detecting 64-bit hardware and emitting 64-bit ARM assembly instructions which are not allowed in 32-bit executables. A 32-bit ARM version of OpenSSL can be configured by setting 32-bit ARM platform:
 `cmake .. -DBUILD_OPENSSL_PLATFORM=linux-armv4`
 
+### Threadpool for Signaling Channel messages
+The threadpool is enabled by default, and starts with 3 threads that it can increase up to 5 if all 3 are actively in use. To change these values to better match the resources of your use case
+please edit src/source/Signaling/Signaling.h defines `KVS_SIGNALING_THREADPOOL_MIN` and `KVS_SIGNALING_THREADPOOL_MAX`. You can also disable the threadpool to instead create and detach each thread
+to handle signaling messages by commenting out `KVS_USE_SIGNALING_CHANNEL_THREADPOOL`.
+
 ## Documentation
 All Public APIs are documented in our [Include.h](https://github.com/awslabs/amazon-kinesis-video-streams-webrtc-sdk-c/blob/master/src/include/com/amazonaws/kinesis/video/webrtcclient/Include.h), we also generate a [Doxygen](https://awslabs.github.io/amazon-kinesis-video-streams-webrtc-sdk-c/) each commit for easier navigation.
 
--- a/README.md
+++ b/README.md
@@ -350,7 +350,7 @@ To build on a 32-bit Raspbian GNU/Linux 11 on 64-bit hardware, the OpenSSL libra
 
 ### Threadpool for Signaling Channel messages
 The threadpool is enabled by default, and starts with 3 threads that it can increase up to 5 if all 3 are actively in use. To change these values to better match the resources of your use case
-please edit src/source/Signaling/Signaling.h defines `KVS_SIGNALING_THREADPOOL_MIN` and `KVS_SIGNALING_THREADPOOL_MAX`. You can also disable the threadpool to instead create and detach each thread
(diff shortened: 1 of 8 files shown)


README.md
```diff
@@ -348,6 +348,11 @@ When building on MacOS M1, if the build fails while trying to build OpenSSL or W
 To build on a 32-bit Raspbian GNU/Linux 11 on 64-bit hardware, the OpenSSL library must be manually configured. This is due to the OpenSSL autoconfiguration script detecting 64-bit hardware and emitting 64-bit ARM assembly instructions which are not allowed in 32-bit executables. A 32-bit ARM version of OpenSSL can be configured by setting 32-bit ARM platform:
 `cmake .. -DBUILD_OPENSSL_PLATFORM=linux-armv4`
 
+### Threadpool for Signaling Channel messages
+The threadpool is enabled by default, and starts with 3 threads that it can increase up to 5 if all 3 are actively in use. To change these values to better match the resources of your use case
```
disa6302, line 352: Do we want to make this configurable instead? Is it expected customers would not require more than 5?

src/source/Signaling/LwsApiCalls.c
```diff
@@ -2012,9 +2012,13 @@ STATUS receiveLwsMessage(PSignalingClient pSignalingClient, PCHAR pMessage, UINT
         DLOGW("Failed to validate the ICE server configuration received with an Offer");
     }
 
+#ifdef KVS_USE_SIGNALING_CHANNEL_THREADPOOL
```
disa6302, line 2015: Wondering if we really want to set this up in an ifdef block. What do you think about using threadpool always and set the MIN_MAX to 3,5 by default and allow we want this to be customer configurable? Because webrtc would create a min of 3 threads anyways - one for offer, one for answer and one per ice-candidate

README.md
jdelapla: That's why I've pointed out the location of the defines for customers to change in this README. I would have preferred to have placed them in a more common library but that would require a different constructor for slgnaling client that allows you to pass in a given threadpool. Possibly an alternative to explore later.

src/source/Signaling/LwsApiCalls.c
jdelapla: I've turned it on by default, but went with the ifdef block to preserve the previous implementation. I think a threadpool is better for most use cases, but if a user finds their system has the resources to handle bringing up and tearing down threads quickly then it threadpool is just needless overhead.

README.md
disa6302: Hmm, we can add the 2 parameters as part of SignalingClientInfo though. Having customers change the `#defs` and rebuild the SDK to be used as a dependency in another project gets messy. If we are allowing customers to configure this, having it as part of clientInfo is cleaner

src/include/com/amazonaws/kinesis/video/webrtcclient/Include.h
```diff
@@ -497,7 +497,7 @@ extern "C" {
 /**
  * Version of signaling client
  */
-#define SIGNALING_CLIENT_CURRENT_VERSION 1
+#define SIGNALING_CLIENT_CURRENT_VERSION 2
```
disa6302, line 500: Wouldnt need this update unless we are updating the SignalingClientInfo struct.

disa6302:
The implementation LGTM. Only minor comments.

src/include/com/amazonaws/kinesis/video/webrtcclient/Include.h
```diff
@@ -487,7 +487,7 @@ extern "C" {
 /**
  * Version of SignalingClientInfo structure
  */
-#define SIGNALING_CLIENT_INFO_CURRENT_VERSION 1
+#define SIGNALING_CLIENT_INFO_CURRENT_VERSION 2
```
disa6302, line 490: Since, we are updating the struct we would need to update the version table here: https://github.com/awslabs/amazon-kinesis-video-streams-webrtc-sdk-c/blob/develop/src/source/Signaling/Signaling.c#L586
Basically need to set the defaults for versions less than 2 and need internal copy. Signaling client info maintain an internal copy of all the variables, so need to do the same for this and so that the updated pSignalingClient version is passed around.

README.md
```diff
@@ -348,6 +348,11 @@ When building on MacOS M1, if the build fails while trying to build OpenSSL or W
 To build on a 32-bit Raspbian GNU/Linux 11 on 64-bit hardware, the OpenSSL library must be manually configured. This is due to the OpenSSL autoconfiguration script detecting 64-bit hardware and emitting 64-bit ARM assembly instructions which are not allowed in 32-bit executables. A 32-bit ARM version of OpenSSL can be configured by setting 32-bit ARM platform:
 `cmake .. -DBUILD_OPENSSL_PLATFORM=linux-armv4`
 
+### Threadpool for Signaling Channel messages
```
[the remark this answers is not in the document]
jdelapla: I have it pointing to defines that apply to the struct, and the defines are in samples/Samples.h. I think that's simpler for customers to find than scrolling through `createSampleConfiguration()`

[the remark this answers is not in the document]
disa6302: Took a while to load for me. LGTM.

disa6302 (dismissed):

disa6302 (approved):

PR #1759, opened 2023-07-21
Parse IP address if the hostname conforms to Public ipv4 DNS format

disa6302:
What was changed?

The function to extract IP address from hostname through DNS resolution is modified to parse IP address from hostname if possible and fallback on getaddrinfo() if the hostname format does not comply to format. Also added a unit test to test this under different inputs.

Why was it changed?

getaddrinfo has been noticed to take a long time to resolve or even be unreliable in smaller embedded devices. Having this in place helps in bypassing this call altogether if the IP address complies with AWS public IPv4 DNS format.

How was it changed?

Checks and parser logic is added to ensure we verify if parsing is a feasibility.

Testing:

- Local testing with samples through logs monitoring
- Testing via newly added unit tests

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.

PR #1763, opened 2023-07-25
introduce bypass ice servers DNS resolve feature

codingspirit:
*Issue #, if available:*

*Description of changes:*

The time spent on DNS resolving for KVS servers is frustrating in some network environment.

By resolving IP address from hostname string directly can reduce the time of DNS resolving and speed up the process of connection establishment.

### DIFF
--- a/samples/Common.c
+++ b/samples/Common.c
@@ -336,6 +336,48 @@ CleanUp:
     CHK_LOG_ERR(retStatus);
 }
 
+STATUS resolveIceServerUrl(PCHAR originalIceServerUrl, PCHAR resolvedIceServerUrl)
+{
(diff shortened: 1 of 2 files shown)


disa6302:
@codingspirit ,

What you are attempting to do is very similar to https://github.com/awslabs/amazon-kinesis-video-streams-webrtc-sdk-c/pull/1759, so closing this in favor of that.

Thank you for your contribution though.

codingspirit:
@disa6302 Just checked #1759 but personally I prefer not to add this logic inside webrtc lib itself. This is a "hack" feature so it should not be included in a webrtc lib designed for common usages.

 It's up to SDK team to make final decisions though.

Issue #1670, opened 2023-03-23
[BUG]Compile error with esp-idf toolchain on Windows


bigcat26:
Hi, I've encountered a problem while compiling the project using esp-idf.

It seems that the `MBEDTLS_CONFIG_FILE` define value did not pass to the compiler correctly on the Windows platform because it was quoted with a single quotation mark. So I replaced the single quotation mark with double qutation mark, it works.

```
[7/45] Building C object esp-idf/main/webrtc/aws-kvs-webrtc/amazon-kinesis-video-streams-webrtc-sdk-c/CMakeFiles/kvsWebrtcClient.dir/src/source/Ice/ConnectionListener.c.obj
FAILED: esp-idf/main/webrtc/aws-kvs-webrtc/amazon-kinesis-video-streams-webrtc-sdk-c/CMakeFiles/kvsWebrtcClient.dir/src/source/Ice/ConnectionListener.c.obj
ccache C:\Espressif\tools\riscv32-esp-elf\esp-2022r1-11.2.0\riscv32-esp-elf\bin\riscv32-esp-elf-gcc.exe -DCMAKE_DETECTED_CACERT_PATH -DDETECTED_GIT_HASH -DENABLE_DATA_CHANNEL -DKVS_CA_CERT_PATH=\"D:/code/esp32-aws-kvs-webrtc/certs/cert.pem\" -DKVS_USE_MBEDTLS -DMBEDTLS_CONFIG_FILE=\"mbedtls/esp_config.h\" -DSDK_VERSION=\"c5a259a84987bff6ccaf35e2b6b004fa4d83cfb7\" -ID:/code/esp32-aws-kvs-webrtc/build/config -ID:/code/esp32-aws-kvs-webrtc/main/webrtc/aws-kvs-webrtc/amazon-kinesis-video-streams-webrtc-sdk-c/open-source/include -ID:/code/esp32-aws-kvs-webrtc/main/webrtc/aws-kvs-webrtc/amazon-kinesis-video-streams-webrtc-sdk-c/src/include 
[...]

Issue #1657, opened 2023-03-08
[QUESTION] KVS Client Sample App sending the first signaling message takes up to 10 seconds


mehmetcy84:
**Environment**: 
Latest Kinesis repo 1.7.3, built on Windows 10 64 bit.
Note: I had to disable threadpool here for the build to succeed:
-DLWS_WITH_THREADPOOL=0 (in repo\CMake\Dependencies\libwebsockets-CMakeLists.txt)

**Issue**: 
The first signalling message sent from the client takes up to 10 seconds irrespective of its content/size.

-kvsWebRTCClientViewer sample is by default using TrickleICE=enabled.
-onIceCandidateHandler sends a signalling message for each ice candidate found.
-The first signalling message sent from the onIceCandidateHandler takes around 10 seconds, the subsequent messages take about 150 milliseconds each.
-If I disable TrickleICE, then the offer which contains the ICE candidates, being the first signalling message sent, takes around 10 seconds.

What might be the reason? Is there a way to remedy this?
Thanks and best regards,

P.S. I build the same repo on my MacOS (WITHOUT setting the DLWS_WITH_THREADPOOL=0)
and the sample client takes less than a millisecond to send signaling messages.
P.P.S I tried building WITH DLWS_WITH_THREADPOOL=0 on MacOS, but the MacOS sample client can still send all signaling messages within a millisecond.

niyatim23:
Hi @mehmetcy84, can you please use the develop branch to build on Windows now, try running the sample application without any modifications to the SDK and see if this issue still persists?

niyatim23:
Please use `develop` and refer to the `windows-openssl-msvc` CI build for steps to build correctly. Closing this issue due to no response. Feel free to reopen / open a new issue if this persists.

mehmetcy84:
Updated to Master 1.9. Using unmodified Master 1.9 for Windows 10 build.
Using the kvsWebRTCClientViewer.exe against the WebRTC Test Page running as master.

The first signalling message ever sent from the client to the master takes 10 seconds. This is the case if I enabled or disabled useTURN, or TrickleICE. For all 4 combinations for these 2 parameters, independent from whatever signaling message goes first, the first and only the first signalling message being sent always takes 10 seconds, and the following signalling messages, if there are any, takes negligible time in comparison.

Here are all the logs for these cases:
- TrickleICE=TRUE,  UseTURN=TRUE (unmodified default)
[...]

Issue #1624, opened 2023-01-27
[BUG] Unable to stream on windows 10 with sample applications


srinivasbakki:
I'm trying to run the sample master and gstreamer applications on windows 10. Both of them do not work. The JS works in my network, both master and viewer. I'm using the master branch. Please find the log attached for both mater application and gstreamer. 

[aws_webrtc_log.txt](https://github.com/awslabs/amazon-kinesis-video-streams-webrtc-sdk-c/files/10516943/aws_webrtc_log.txt)

niyatim23:
Hi @srinivasbakki, are you running the sample application as is or have you made any changes to it?
I see `2023-01-26 15:47:23 ERROR   signalingCacheLoadFromFile(): operation returned status code: 0x0000000a`, seems like this application does not have permissions to write to the file system.

srinivasbakki:
Hi, i ran the application as is. I didn't change anything. My interest is with Client application to stream H264. Didin't check the gstreamer application with below observation -  I saw in some thread about changes to connectionListenerReceiveDataRoutine :
comment  retval = POLL(rfds, nfds, CONNECTION_LISTENER_SOCKET_WAIT_FOR_DATA_TIMEOUT / HUNDREDS_OF_NANOS_IN_A_MILLISECOND); and simply return success. 
Next, in the same routine comment if (canReadFd(localSocket, rfds, nfds)) {
and just go ahead with the recv loop. Bingo, this works. 
But this is not clean solution
Please help

niyatim23:
Hi @srinivasbakki, can you please use the develop branch to build on Windows now and try running the sample application without any modifications to the SDK?

Issue #1613, opened 2023-01-06
[BUG] Custom VS2017 C++ MFC Application occasionally crashes at RtpH264Payloader.c


mehmetcy84:
Using Kinesis SDK 1.7.3, developing a C++ MFC App with VS2017 on Win 10 64-bit PC, having built and added the necessary libs to my application, I can display frames from our camera via Kinesis. However there are a number of random crashes in different places.

One of the crashes happens in RtpH264Playloader.c depayH264FromRtpPayload function.

`STATUS depayH264FromRtpPayload(PBYTE pRawPacket, UINT32 packetLength, PBYTE pNaluData, PUINT32 pNaluLength, PBOOL pIsStart)`

```
case STAP_B_INDICATOR:
            naluLength = 0;
            pCurPtr = pRawPacket + STAP_A_HEADER_SIZE;
            do {
                subNaluSize = getUnalignedInt16BigEndian(pCurPtr);
                pCurPtr += SIZEOF(UINT16);
                MEMCPY(pNaluData, start4ByteCode, SIZEOF(start4ByteCode));
                pNaluData += SIZEOF(start4ByteCode);
                MEMCPY(pNaluData, pCurPtr, subNaluSize);    // <--------------PROBLEM HERE
                pCurPtr += subNaluSize;
                pNaluData += subNaluSize;
                naluLength += SIZEOF(start4ByteCode) + subNaluSize;
            } while (subNaluSize > 0 && pCurPtr < pRawPacket + packetLength);
            DLOGS("STAP_B_INDICATOR starting packet %d len %d", isStartingPacket, naluLength);
            break;
```

Here, I receive an exception:

> Exception thrown at 0x00007FF91981122E (vcruntime140d.dll) in MyApp.exe: 0xC0000005: Access violation reading location 0x0000016E87285000.

When I check the local values, 

> packetLength: 224
> subNaluSize: 59196

hence in this line:
`MEMCPY(pNaluData, pCurPtr, subNaluSize);`
pCurPtr points to the first input argument pointer, which points to a buffer of length 224, and memcpy attempts to read 59196 bytes from it.
The fourth input/output argument pNaluLength has the value of 89572.

So the pNaluLength and subNaluSize calculation seems to be off. Has there been a similar issue/crash reported before?
How can we fix this?

EDIT:
[...]

PR #1764, opened 2023-07-27
Fix the Windows build in the CI

niyatim23:
*Issue #, if available:*
https://github.com/awslabs/amazon-kinesis-video-streams-webrtc-sdk-c/issues/1670
https://github.com/awslabs/amazon-kinesis-video-streams-webrtc-sdk-c/issues/1657
https://github.com/awslabs/amazon-kinesis-video-streams-webrtc-sdk-c/issues/1624
https://github.com/awslabs/amazon-kinesis-video-streams-webrtc-sdk-c/issues/1613
https://github.com/awslabs/amazon-kinesis-video-streams-webrtc-sdk-c/issues/1611
https://github.com/awslabs/amazon-kinesis-video-streams-webrtc-sdk-c/issues/1451
https://github.com/awslabs/amazon-kinesis-video-streams-webrtc-sdk-c/issues/1288

*Description of changes:*

What was changed?

Fixed the Windows CI build which has been broken since forever. Modified the dependency's (libwebsockets) CMake file to fix the same by introducing arguments that were missing

Why was it changed?

We were missing the `OPENSSL_LIBRARIES`, instead we specified the `SSL_LIBRARY` and `CRYPTO_LIBRARY` which worked on Linux and Unix systems but was breaking on Windows. Additionally, introduced a `LIST_SEPARATOR` as the `OPENSSL_LIBRARIES` contains two file paths separated by a semicolon which causes CMake to ignore them at runtime, so I have used `|` as a separator instead

How was it changed?

By modifying the CMakeLists files

Testing:

With the CI


sirknightj (approved):
Nice work!

CMakeLists.txt
```diff
@@ -31,6 +31,9 @@ option(THREAD_SANITIZER "Build with ThreadSanitizer." OFF)
 option(UNDEFINED_BEHAVIOR_SANITIZER "Build with UndefinedBehaviorSanitizer." OFF)
 option(LINK_PROFILER "Link gperftools profiler" OFF)
 
+set(EXT_PTHREAD_INCLUDE_DIR "" CACHE PATH "Path to the PThread dir")
```
disa6302, line 34: This would be set only for Windows right? Can we add the platform check?

.github/workflows/ci.yml
```diff
@@ -342,33 +342,51 @@ jobs:
...
+      - name: Build repository
+        shell: powershell
+        run: |
+          cd D:\a\webrtc
+          git config --system core.longpaths true
+          .github\build_windows.bat
+      # TODO: Fix these tests
+      # - name: Run tests
+      #   run: |
+      #     & "D:\a\webrtc\build\tst\webrtc_client_test.exe --gtest_filter='-DataChannelFunctionalityTest.*:IceApiTest.*:IceFunctionalityTest.*:PeerConnectionFunctionalityTest.*:SignalingApiFunctionalityTest.*:TurnConnectionFunctionalityTest.*:RtpFunctionalityTest.marshallUnmarshallH264Data:RtpFunctionalityTest.packingUnpackingVerifySameH264Frame:RtcpFunctionalityTest.onRtcpPacketCompound:RtcpFunctionalityTest.twcc3'"
```
disa6302, line 389: We are not running the unit tests? Will that be focussed on another PR? If yes, can we create a github issue to track it?

disa6302:
On another note, I think we are missing a windows CI with Mbedtls enabled. Something to track as well to include at some point. @niyatim23 , can we create a github issue for that too?

.github/workflows/ci.yml
```diff
@@ -342,33 +342,51 @@ jobs:
...
+          dir
+      - name: Install dependencies
+        shell: powershell
+        run: |
+          choco install gstreamer --version=1.16.2
+          choco install gstreamer-devel --version=1.16.2 
+          curl.exe -o C:\tools\pthreads-w32-2-9-1-release.zip ftp://sourceware.org/pub/pthreads-win32/pthreads-w32-2-9-1-release.zip
+          mkdir C:\tools\pthreads-w32-2-9-1-release\
+          Expand-Archive -Path C:\tools\pthreads-w32-2-9-1-release.zip -DestinationPath C:\tools\pthreads-w32-2-9-1-release    
+      # - name: Build libwebsockets from source
```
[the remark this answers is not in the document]
niyatim23: I left those in because we may need it in the future when we upgrade to a higher version of libwebsockets

CMakeLists.txt
niyatim23: Since we are allowing it to be provided from command line, I think it needs to exist like that. It is empty at first ("") and then when it is read from command line, it gets its value which is used later

disa6302: Its still a -D option right? If its other platforms, this blocks is just not visited and the variable is unused. So it should be ok to add this in the `if(WIN32)` block. It is more clear that way.

niyatim23: Makes sense, done!

niyatim23:
> On another note, I think we are missing a windows CI with Mbedtls enabled. Something to track as well to include at some point. @niyatim23 , can we create a github issue for that too?

Tracking:
1. Windows-mbedTLS: https://github.com/awslabs/amazon-kinesis-video-streams-webrtc-sdk-c/issues/1768
2. Windows-tests: https://github.com/awslabs/amazon-kinesis-video-streams-webrtc-sdk-c/issues/1767

.github/workflows/ci.yml
niyatim23: Tracking: https://github.com/awslabs/amazon-kinesis-video-streams-webrtc-sdk-c/issues/1767

disa6302 (approved):
Great work @niyatim23 !

PR #1783, opened 2023-08-23
testing release with media storage support

hassanctech:
*What was changed?*
- New feature, Support WebRTC Stream Ingestion
- Fixed bug in cache where we now will actually update an existing entry whereas before we only updated the existing entry if the cache wasn't full otherwise we would update the last entry which could result in .duplicate entries in cache.  
- Fixed another bug where if an arn was specified instead of a name the cache would store the arn in place of the name, then later when the name was known all subsequent cache entires have the name in place of the name, so there could be two entires for the same signaling channel, one indexed by the name and one by the arn

*Why was it changed?*
- New feature was added to support AWS Kinesis Video Streams WebRTC functionality

*How was it changed?*
- 4 new states were added to the state machine, DescribeMediaStorageConfiguration, JoinSession, JoinSessionWaiting, JoinSessionConnected.

- Existing tests modified, new tests added


src/source/Signaling/ChannelInfo.c
```diff
@@ -308,3 +340,95 @@ PCHAR getStringFromChannelRoleType(SIGNALING_CHANNEL_ROLE_TYPE type)
...
 }
+
+// https://docs.aws.amazon.com/kinesisvideostreams-webrtc-dg/latest/devguide/kvswebrtc-how-iam.html#kinesis-using-iam-arn-format
+// Example: arn:aws:kinesisvideo:region:account-id:channel/channel-name/code
+STATUS validateKvsSignalingChannelArnAndExtractChannelName(PChannelInfo pChannelInfo, PUINT16 pStart, PUINT16 pNumChars)
+{
+    UINT16 arnLength, currPosIndex = 0, channelNameLength = 0, channelNameStart = 0;
+    PCHAR partitionEnd, regionEnd, channelNameEnd;
+    PCHAR currPos;
+    UINT8 accountIdStart = 1, accountIdEnd = 12;
```
disa6302, line 351: Nit: Can we #def these values?

src/source/Signaling/ChannelInfo.c
```diff
@@ -308,3 +340,95 @@ PCHAR getStringFromChannelRoleType(SIGNALING_CHANNEL_ROLE_TYPE type)
...
+                        if (currPosIndex < arnLength && *(currPos + currPosIndex) == ARN_DELIMETER_CHAR) {
+                            currPosIndex++;
+
+                            regionEnd = STRNCHR(currPos + currPosIndex, arnLength - currPosIndex, ARN_DELIMETER_CHAR);
+
+                            if (regionEnd != NULL) {
+                                if (currPosIndex < arnLength && (regionEnd - currPos) < arnLength) {
+                                    currPosIndex = regionEnd - currPos;
+
+                                    if (currPosIndex + 36 <= arnLength) {
```
disa6302, line 380: #def 36 to clarify what it means.

src/source/Signaling/ChannelInfo.c
```diff
@@ -308,3 +340,95 @@ PCHAR getStringFromChannelRoleType(SIGNALING_CHANNEL_ROLE_TYPE type)
...
+    UINT16 arnLength, currPosIndex = 0, channelNameLength = 0, channelNameStart = 0;
+    PCHAR partitionEnd, regionEnd, channelNameEnd;
+    PCHAR currPos;
+    UINT8 accountIdStart = 1, accountIdEnd = 12;
+    UINT8 timeCodeStart = 0, timeCodeEnd = 12;
+
+    if (pChannelInfo != NULL && pChannelInfo->pChannelArn != NULL) {
+        arnLength = STRNLEN(pChannelInfo->pChannelArn, MAX_ARN_LEN);
+
+        if (arnLength >= SIGNALING_CHANNEL_ARN_MIN_LENGTH) {
```
disa6302, line 357: Wondering if we can convert this entire nested if block into switch cases. Why not run a while loop and then use enums in a switch block to parse? It would be cleaner if it can be achieved and also maintainable.
typedef enum {
    PARSE_START = 0,
    PARSE_ARN_PREFIX,
    PARSE_PARTITION,
    PARSE_SERVICE,
    PARSE_REGION,
    PARSE_ACCOUNT_ID,
  ......
    PARSE_DONE
} ARN_PARSE_STATE;

We do a similar thing for ICE candidate parsing from incoming SDP:
https://github.com/awslabs/amazon-kinesis-video-streams-webrtc-sdk-c/blob/develop/src/source/Ice/IceAgent.c#L9-L19
https://github.com/awslabs/amazon-kinesis-video-streams-webrtc-sdk-c/blob/develop/src/source/Ice/IceAgent.c#L359-L418

src/source/Signaling/LwsApiCalls.c
```diff
@@ -1468,6 +1479,186 @@ STATUS connectSignalingChannelLws(PSignalingClient pSignalingClient, UINT64 time
...
+    // Set the service call result
+    ATOMIC_STORE(&pSignalingClient->result, (SIZE_T) pLwsCallInfo->callInfo.callResult);
+    pResponseStr = pLwsCallInfo->callInfo.responseData;
+    resultLen = pLwsCallInfo->callInfo.responseDataLen;
+
+    // Early return if we have a non-success result
+    CHK((SERVICE_CALL_RESULT) ATOMIC_LOAD(&pSignalingClient->result) == SERVICE_CALL_RESULT_OK, STATUS_SIGNALING_LWS_CALL_FAILED);
+
+    // Perform some validation on the channel description
+    CHK(pSignalingClient->channelDescription.channelStatus != SIGNALING_CHANNEL_STATUS_DELETING, STATUS_SIGNALING_CHANNEL_BEING_DELETED);
```
disa6302, line 1543: Shouldnt we have a similar check at the start for this before we go ahead and construct the requestInfo? Why have it at the end?

src/source/Signaling/LwsApiCalls.c
```diff
@@ -1982,15 +2176,16 @@ STATUS receiveLwsMessage(PSignalingClient pSignalingClient, PCHAR pMessage, UINT
...
+                                                                                                        : SIGNALING_STATE_CONNECTED));
 
             CHK(FALSE, retStatus);
             break;
 
         case SIGNALING_MESSAGE_TYPE_OFFER:
-            CHK(pSignalingMessageWrapper->receivedSignalingMessage.signalingMessage.peerClientId[0] != '\0',
-                STATUS_SIGNALING_NO_PEER_CLIENT_ID_IN_MESSAGE);
-            // Explicit fall-through !!!
+            // CHK(pSignalingMessageWrapper->receivedSignalingMessage.signalingMessage.peerClientId[0] != '\0',
```
disa6302, line 2186: Nit: Remove comment. Also, we are not expecting a Peer client ID in the message? Is this specific to storage? What about non storage use cases?

src/source/Signaling/LwsApiCalls.h
```diff
@@ -11,8 +11,8 @@ extern "C" {
 #endif
 
 // Timeout values
-#define SIGNALING_SERVICE_API_CALL_CONNECTION_TIMEOUT (2 * HUNDREDS_OF_NANOS_IN_A_SECOND)
-#define SIGNALING_SERVICE_API_CALL_COMPLETION_TIMEOUT (5 * HUNDREDS_OF_NANOS_IN_A_SECOND)
+#define SIGNALING_SERVICE_API_CALL_CONNECTION_TIMEOUT (20 * HUNDREDS_OF_NANOS_IN_A_SECOND)
```
disa6302, line 14: Why the increase in the time out value?

tst/SignalingApiTest.cpp
```diff
@@ -40,6 +43,160 @@ TEST_F(SignalingApiTest, createValidateChannelInfo)
...
+    EXPECT_EQ(createValidateChannelInfo(&channelInfo, &pChannelInfo), STATUS_SUCCESS);
+    freeChannelInfo(&pChannelInfo);
+
+    channelInfo.pChannelArn = arn8;
+    EXPECT_EQ(createValidateChannelInfo(&channelInfo, &pChannelInfo), STATUS_SUCCESS);
+    freeChannelInfo(&pChannelInfo);
+}
+
+
+TEST_F(SignalingApiTest, testChannelArnsInValid)
```
disa6302, line 96: Nice!

src/source/Signaling/LwsApiCalls.c
hassanctech: This is set after the `lwsCompleteSync` call, so it needs to be after the actual API call.  We follow this same pattern for all Signaling API calls.

src/source/Signaling/ChannelInfo.c
hassanctech: Done

src/source/PeerConnection/Rtcp.c
```diff
@@ -340,7 +340,7 @@ STATUS onRtcpPacket(PKvsPeerConnection pKvsPeerConnection, PBYTE pBuff, UINT32 b
     UINT32 currentOffset = 0;
 
     CHK(pKvsPeerConnection != NULL && pBuff != NULL, STATUS_NULL_ARG);
-
+    DLOGD("**rtcp**");
```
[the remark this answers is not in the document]
hassanctech: It is helpful to see that rtc communication is still working otherwise you need verbose level logs.  I think it's ok for debug level?

src/source/Signaling/LwsApiCalls.c
hassanctech: It is specific to storage.

src/source/Signaling/LwsApiCalls.h
hassanctech: reverted

src/source/Signaling/ChannelInfo.c
hassanctech: done

src/source/Signaling/LwsApiCalls.c
```diff
@@ -1468,6 +1479,186 @@ STATUS connectSignalingChannelLws(PSignalingClient pSignalingClient, UINT64 time
...
+            pSignalingClient->signalingClientCallbacks.getCurrentTimeFn(pSignalingClient->signalingClientCallbacks.customData);
+    }
+
+    checkAndCorrectForClockSkew(pSignalingClient, pRequestInfo);
+
+    CHK_STATUS(createLwsCallInfo(pSignalingClient, pRequestInfo, PROTOCOL_INDEX_HTTPS,
+                                 &pLwsCallInfo)); //!< TBD. Accroding to the design document, the prefix of url will be webrtc://
+
+    if (pSignalingClient->mediaStorageConfig.storageStatus) {
+        pSignalingClient->diagnostics.joinSessionToOfferRecvTime = GETTIME();
```
stefankiesz, line 1529: Should we also set `pSignalingClient->diagnostics.joinSessionCallTime` to the current time here? I don't think that one is currently set anywhere.

src/source/Signaling/LwsApiCalls.c
```diff
@@ -2171,14 +2367,28 @@ PVOID receiveLwsMessageWrapper(PVOID args)
...
+    if (messageType == SIGNALING_MESSAGE_TYPE_OFFER) {
+        MUTEX_LOCK(pSignalingClient->offerSendReceiveTimeLock);
+        pSignalingClient->offerReceivedTime = GETTIME();
+        MUTEX_UNLOCK(pSignalingClient->offerSendReceiveTimeLock);
+
+        if (pSignalingClient->mediaStorageConfig.storageStatus) {
+            MUTEX_LOCK(pSignalingClient->jssWaitLock);
+            ATOMIC_STORE_BOOL(&pSignalingClient->offerReceived, TRUE);
+            DLOGI("Offer Received from JoinStorageSession Call.");
+            pSignalingClient->diagnostics.joinSessionToOfferRecvTime =
```
stefankiesz, line 2379: Should `PROFILE_WITH_START_TIME_OBJ` be used as is below for the SIGNALING_MESSAGE_TYPE_ANSWER case?

src/source/Signaling/LwsApiCalls.c
hassanctech: Seems to be set ok, in my local logs I see:
```
2023-10-18 03:47:44.983 PROFILE executeJoinStorageSessionState(): [Join Session call] Time taken: 843 ms
```

And it prints this:
```
    PROFILE_CALL_WITH_T_OBJ(retStatus = joinStorageSession(pSignalingClient, time), pSignalingClient->diagnostics.joinSessionCallTime,
                            "Join Session call");
```

Which is the `joinSessionCallTime`.

I think the current calculation is correct for `joinSessionToOfferRecvTime ` which stores the time immediately before calling join session then computes it as follows:

```
            pSignalingClient->diagnostics.joinSessionToOfferRecvTime =
                pSignalingClient->offerReceivedTime - pSignalingClient->diagnostics.joinSessionToOfferRecvTime;
```

Which looks like the right thing to do.  Are you seeing something strange in your logs for join session call time?

src/source/Signaling/LwsApiCalls.c
hassanctech: That macro itself calls GETTIME().  The issue is we need to acquire the jssWaitLock before we can set that value and I was afraid if there's lock contention then the actual time the offer was received may not get stored and we'll have an inaccurate value there, hence why I set the offer received time *before* acquiring the lock.

src/source/Signaling/LwsApiCalls.c
stefankiesz: Ah ok, I misunderstood what the `joinSessionCallTime` is tracking - thought it was the current time when JoinSession is called, but it's the time take until it's called, and now I see it being set there with the `PROFILE_CALL_WITH_T_OBJ`. Looks good!

disa6302 (approved):

.github/workflows/codecov.yml
```diff
@@ -1,37 +0,0 @@
...
-jobs:
-  linux-gcc-codecov:
-    runs-on: ubuntu-latest
-    permissions:
-      id-token: write
-      contents: read
-    steps:
-      - name: Fetch
-        uses: actions/checkout@v3
-        with:
```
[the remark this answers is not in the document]
niyatim23: >Repository YAML: The repo level settings for each repo, set and held in your Github/Gitlab/Bitbucket. This YAML file must be a stand-alone file and is not part of your [CI-level YAML](https://docs.codecov.com/docs/codecov-uploader).

https://docs.codecov.com/docs/codecov-yaml

PR #1792, opened 2023-08-30
DTLS socket no longer pre-preemptively closed

jdelapla:
*What was changed?*
Fix issue with aggressive nomination viewers causing master to potentially close the socket negotiating DTLS connection.

*Why was it changed?*
The race condition could lead to stalled out DTLS negotiations, causing discovery to take 5+ seconds longer.

*How was it changed?*
Changed the ICE statemachine Ready and Connected setups. Connected will only set DataSending pair in an aggressive nomination case, and Ready will no longer change the DataSending pair if it has already been assigned (because if it has, DTLS handshaked could be actively happening on that socket).

This means that in the current ICE state machine design DTLS negotiation will only occur after nomination, where before it would start earlier and the client would assume which pair the viewer would nominate.

Added two tests to confirm this behavior can no longer occur for aggressive nomination or standard ICE/nomination.


hassanctech (approved):
lgtm

disa6302 (approved):

Issue #1803, opened 2023-09-11
[QUESTION]The definition of KVS_USE_SIGNALING_CHANNEL_THREADPOOL is not found in the SDK build.


YOSI-yoshidayuji:
README.md says "The threadpool is enabled by default," But I can't found add_definitions(-DKVS_USE_SIGNALING_CHANNEL_THREADPOOL) in CMakeLists.txt

I modified the function 'createSignalingSync' in Signaling.c for verification purposes.
--
#ifdef KVS_USE_SIGNALING_CHANNEL_THREADPOOL
printf( "threadpool create !\n" );
    CHK_STATUS(threadpoolCreate(&pSignalingClient->pThreadpool, pClientInfo->signalingClientInfo.signalingMessagesMinimumThreads,
                                pClientInfo->signalingClientInfo.signalingMessagesMaximumThreads));
#else
printf( "No treadpool \n" );
#endif
--
When I build it normally, "No threadpool" is displayed. When I write add_definitions(-DKVS_USE_SIGNALING_CHANNEL_THREADPOOL), "threadpool create!" is displayed. 
Am I misunderstanding something?

disa6302:
@YOSI-yoshidayuji , 

It does seem we are missing the add definition in the CMake file. Thank you for pointing it out! 
Would you like to open a PR with the flag change?

YOSI-yoshidayuji:
Thank you for your response. 
According to the regulations of my organization, I do not wish to open in PR.

disa6302:
No problem :) 
Keeping this ticket open till we have the cmake fix merged. Thanks for reporting again!

niyatim23:
We have a new CMake flag for this now `ENABLE_KVS_THREADPOOL`. Currently only available in `develop`

PR #1806, opened 2023-09-13
Create a CMake flag to enable KVS Threadpool

niyatim23:
*Issue #, if available:*
https://github.com/awslabs/amazon-kinesis-video-streams-webrtc-sdk-c/issues/1803
Additionally, previously the `KVS_SIGNALING_THREADPOOL` was not enabled due to the `#define` being in a different file. This fix makes sure that the threadpool is enabled and disabled correctly. This also fixes all the tests by adding missing initializations for min and max number of threads.

*What was changed?*
A new CMake flag `ENABLE_KVS_THREADPOOL` is available to configure this feature from the CMake

*Why was it changed?*
We do not have a way to configure this feature from CMake as correctly pointed out by the customer in the issue

*How was it changed?*
Created a new CMake flag `ENABLE_KVS_THREADPOOL` and passed it to use in the SDK with `add_definitions`

Manually tested by adding logs to the block of code being enabled and disabled and made sure that the flag was operational in doing so

### DIFF
--- a/CMake/Dependencies/libkvsCommonLws-CMakeLists.txt
+++ b/CMake/Dependencies/libkvsCommonLws-CMakeLists.txt
@@ -6,7 +6,7 @@ include(ExternalProject)
 
 ExternalProject_Add(libkvsCommonLws-download
     GIT_REPOSITORY    https://github.com/awslabs/amazon-kinesis-video-streams-producer-c.git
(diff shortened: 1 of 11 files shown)


jdelapla:
wait on merging this until the CMake dependencies have been merged

disa6302:
We need an equivalent producer PR to update to the develop commit in PIC and use that commit now that the PIC PR is merged.

disa6302 (approved):

PR #1810, opened 2023-09-18
Threadpool context for WebRTC

disa6302:
*Issue #, if available:*

*What was changed?*
- Created a new ThreadPoolContext library (`libkvsWebRtcThreadpool`) that has a shared linkage to `libkvsWebrtcClient` and `libkvsWebrtcSignalingClient`

*Why was it changed?*
- With new upcoming changes, we would be using threadpool in the WebRTC library and signaling. We need a common threadpool for these 2 object graphs. 

*How was it changed?*
- Added a `ENABLE_THREADPOOL_USAGE` flag in CMake based on which threadpool would be used. If disabled, normal threads would be created as and when required.
- Created a new `libkvsWebRtcThreadpool` that has a shared linkage with the existing libraries
- Created a new context object as a threadpool wrapper
- Added env options to allow configuring MAX and MIN threads. To do so, run:
    1. `export AWS_KVS_WEBRTC_THREADPOOL_MIN_THREADS=<value>`
     2. `export AWS_KVS_WEBRTC_THREADPOOL_MAX_THREADS=<value>`

- Local build with `BUILD_STATIC_LIBS` set to `ON` and `OFF` to ensure threadpool works
- Added debug in threadpool context to verify that the threadpool object address is the same in WebRTC and signaling library - created a dummy thread to be pushed onto threadpool in `initKvsWebRtc()`.
- Ran the unit tests locally


src/include/com/amazonaws/kinesis/video/webrtcclient/Include.h
```diff
@@ -631,6 +631,26 @@ extern "C" {
  */
 #define SIGNALING_CONNECT_TIMEOUT (5 * HUNDREDS_OF_NANOS_IN_A_SECOND)
 
+/**
+ * Default minimum number of threads in the threadpool for the SDK
+ */
+#define THREADPOOL_MIN_THREADS 5
```
jdelapla, line 637: This feels high. It's 5 threads idling at all times.

src/source/Threadpool/ThreadpoolContext.h
```diff
@@ -0,0 +1,29 @@
...
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+////////////////////////////////////////////////////
+// Project include files
+////////////////////////////////////////////////////
+
+typedef struct {
+    PThreadpool pThreadpool;
```
jdelapla, line 18: This needs an initialized flag.

jdelapla (changes requested):

src/include/com/amazonaws/kinesis/video/webrtcclient/Include.h
disa6302: Do we want to make it 3? Ok with that. Considering there will peer connection and signaling, though 5 would be ok.

src/source/Threadpool/ThreadpoolContext.h
disa6302: Had it initially, but removed it because there is a NULL check on threadpool object in every API. If it is NULL, it is uninitialized, so found to have an initializer flag redundant. Any other reason you foresee we would need the flag?

disa6302: Done

jdelapla (approved):

PR #1812, opened 2023-09-21
WebRTC Client singleton and early STUN DNS resolution

disa6302:
*Issue #, if available:*

*What was changed?*
- Using threadpool to start STUN DNS resolution as soon as the library is initialized instead of waiting on an offer
- Created a WebRTC client singleton as a higher level object to hold items common to multiple peer connections

*Why was it changed?*
- STUN DNS resolution uses `getaddrinfo` which seems to take time in some systems. We start this after offer is received which is unnecessary. Given STUN URL is fixed, DNS resolution can start earlier in the process to save time on this after offer is received.

*How was it changed?*
- Created a singleton that would be available to store the resolved address. This is important since none of the peer connection objects would have been created until an offer is received.
- Using the threadpool to resolve address on a seperate thread
- Added a callback that would provide the IP address whenever the client needs it, namely during the ICE agent set up process
- If the IP address is not ready by the time it is required, we wait for the resolution to complete with a pass/fail status. However, due to an early start to the process, there will still be some latency improvements.
- Using a time based caching solution to ensure we go through resolution again if the value is read post the caching period (2 hours).

- Unit tests were run
- Enforced some failure situations and tested expiry to ensure refresh.

### DIFF
--- a/src/source/PeerConnection/PeerConnection.c
+++ b/src/source/PeerConnection/PeerConnection.c
@@ -7,8 +7,33 @@ static volatile ATOMIC_BOOL gKvsWebRtcInitialized = (SIZE_T) FALSE;
 // Function to get access to the Singleton instance
 PWebRtcClientContext getWebRtcClientInstance()
 {
-    static WebRtcClientContext w = {.pStunIpAddrCtx = NULL, .stunCtxlock = INVALID_MUTEX_VALUE, .isContextInitialized = FALSE};
+    static WebRtcClientContext w = {.pStunIpAddrCtx = NULL,
+                                    .stunCtxlock = INVALID_MUTEX_VALUE,
+                                    .usageSemaphore = INVALID_SEMAPHORE_HANDLE_VALUE,
+                                    .isSemAccessInitialized = FALSE,
+                                    .isContextInitialized = FALSE};
+    if (w.isContextInitialized) {
+        if (STATUS_FAILED(semaphoreAcquire(w.usageSemaphore, INFINITE_TIME_VALUE))) {
+            DLOGW("Failed to create the semaphore to control access to the client context instance");
+        } else {
+            ATOMIC_STORE_BOOL(&w.isSemAccessInitialized, TRUE);
+        }
+    } else {
+        DLOGI("WebRTC Client instance hasnt been initialized");
+    }
     return &w;
+}
+
+STATUS releaseHoldOnWebRtcClientInstance(PWebRtcClientContext pWebRtcClientContext)
+{
+    STATUS retStatus = STATUS_SUCCESS;
+    INT32 count;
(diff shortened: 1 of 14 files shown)


jdelapla (changes requested):

src/source/PeerConnection/PeerConnection.c
```diff
@@ -687,6 +718,113 @@ STATUS rtcpReportsCallback(UINT32 timerId, UINT64 currentTime, UINT64 customData
...
+PVOID resolveStunIceServerIp(PVOID args)
+{
+    UNUSED_PARAM(args);
+    BOOL locked = FALSE;
+    CHAR addressResolved[KVS_IP_ADDRESS_STRING_BUFFER_LEN + 1] = {'\0'};
+    PWebRtcClientContext pWebRtcClientContext = getWebRtcClientInstance();
+    PCHAR pRegion;
+    PCHAR pHostnamePostfix;
+
+    MUTEX_LOCK(pWebRtcClientContext->stunCtxlock);
```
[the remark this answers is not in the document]
disa6302: Hmm. Ok. I see the point. I have added it in. Will wait for CI to pass.

src/source/PeerConnection/PeerConnection.c
```diff
@@ -1442,6 +1602,45 @@ STATUS initKvsWebRtc(VOID)
...
+    DLOGI("Destroyed STUN IP object");
+    MUTEX_UNLOCK(pWebRtcClientContext->stunCtxlock);
+    /* End of handling STUN object */
+
+    if (IS_VALID_MUTEX_VALUE(pWebRtcClientContext->stunCtxlock)) {
+        MUTEX_FREE(pWebRtcClientContext->stunCtxlock);
+        pWebRtcClientContext->stunCtxlock = INVALID_MUTEX_VALUE;
+    }
+
+    ATOMIC_STORE_BOOL(&pWebRtcClientContext->isContextInitialized, FALSE);
```
jdelapla, line 1636: Move this to the start of the function

src/source/PeerConnection/PeerConnection.c
```diff
@@ -1442,6 +1602,45 @@ STATUS initKvsWebRtc(VOID)
     return retStatus;
 }
 
+STATUS cleanupWebRtcClientInstance()
+{
+    STATUS retStatus = STATUS_SUCCESS;
+    // Stun object cleanup
+    PWebRtcClientContext pWebRtcClientContext = getWebRtcClientInstance();
+    CHK_WARN(ATOMIC_LOAD_BOOL(&pWebRtcClientContext->isContextInitialized), STATUS_INVALID_OPERATION,
```
jdelapla, line 1610: Need to release hold.

jdelapla (changes requested):

src/source/PeerConnection/PeerConnection.c
disa6302: Using an isReleased flag to ensure we release hold in case the client context is not initialized.

jdelapla (approved):

PR #1823, opened 2023-10-06
Fix mbedtls build failure for 22.04

disa6302:
*Issue #, if available:*

*What was changed?*
- Added a new CI job for mbedtls to run on ubuntu 22.04 which has gcc11
- Added the `-Wno-error=stringop-overflow` for gcc mbedtls build

*Why was it changed?*
- MBedtls build fails on ubuntu 22.04 for the version we use. The MBEDTLS library fixes this in v2.28.0, but adding this flag until we can update to 2.28.0 of mbedtls library

*How was it changed?*
- Added the relevant changes in the CMakeLists file based on the compiler used

- Using the CI to confirm the build succeeds
- Also used an EC2 instance to verify the fix


hassanctech (approved):
LGTM.  Let's prioritize moving to 3.x soon after this, there appears to be a detailed migration guide to do this:  https://github.com/Mbed-TLS/mbedtls/blob/development/docs/3.0-migration-guide.md

PR #1824, opened 2023-10-09
Dtls handshake process optimization

disa6302:
*Issue #, if available:*

*What was changed?*
- The PR focusses on improving DTLS handshake time by moving to a state transition based approach to be more responsive to the incoming DTLS handshake messages instead of waiting 200 milliseconds to respond. (based on the timer implementation)

*Why was it changed?*
- DTLS over Openssl inherently takes some time based on the cryptographic suite chosen. The PR focusses on improving the existing DTLS handshake design to reduce latency from the SDK end so that the true latency is from the actual handshake process. We use the threadpool to perform the DTLS handshake and the optimization can be enabled by enabling THREADPOOL usage for the WebRTC SDK.

*How was it changed?*
- Start DTLS handshake on the threadpool based on the `ENABLE_KVS_THREADPOOL` flag. 
- Start DTLS handshake **ONLY AFTER** ICE agent has nominated the data sending candidate pair. This helps in avoiding missed DTLS packets that increasing handshake time
- Using a CVAR based signaling mechanism when an incoming DTLS packet is received. 
- Using refcounting to track access to the DTLS session object to ensure we do not delete the heap allocated objects until all references to the DTLS object are returned. 

- Added new unit tests to test out the new piece. Used the new unit test to also verify the DTLS handshake timing optimization

### DIFF
--- a/src/source/Crypto/Dtls_openssl.c
+++ b/src/source/Crypto/Dtls_openssl.c
@@ -10,6 +10,20 @@ INT32 dtlsCertificateVerifyCallback(INT32 preverify_ok, X509_STORE_CTX* ctx)
     return 1;
 }
 
+VOID acquireDtlsSession(PDtlsSession pDtlsSession)
+{
+    if (pDtlsSession != NULL) {
+        ATOMIC_INCREMENT(&pDtlsSession->objRefCount);
+    }
+}
+
+VOID releaseDtlsSession(PDtlsSession pDtlsSession)
+{
+    if (pDtlsSession != NULL) {
+        ATOMIC_DECREMENT(&pDtlsSession->objRefCount);
+    }
+}
+
 STATUS dtlsCertificateFingerprint(X509* pCertificate, PCHAR pBuff)
 {
     ENTERS();
@@ -42,6 +56,7 @@ STATUS dtlsTransmissionTimerCallback(UINT32 timerID, UINT64 currentTime, UINT64 
     LONG dtlsTimeoutRet = 0;
 
     CHK(pDtlsSession != NULL, STATUS_NULL_ARG);
+    acquireDtlsSession(pDtlsSession);
 
     MEMSET(&timeout, 0x00, SIZEOF(struct timeval));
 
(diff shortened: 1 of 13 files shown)


src/source/Crypto/Dtls_openssl.c
```diff
@@ -592,6 +757,7 @@ STATUS dtlsSessionIsInitFinished(PDtlsSession pDtlsSession, PBOOL pIsConnected)
 
     CHK(pDtlsSession != NULL && pIsConnected != NULL, STATUS_NULL_ARG);
```
jdelapla, line 758: releasing unacquired session. Since acquire has a null check internally, you can just place it before the null check to avoid this happening. It's happening in several functions right now.

jdelapla: We're missing basic invalid unit tests like passing in null values, otherwise this should have been found by CI.

jdelapla (changes requested):

src/source/PeerConnection/PeerConnection.c
```diff
@@ -488,7 +496,15 @@ VOID onIceConnectionStateChange(UINT64 customData, UINT64 connectionState)
             // wait until DTLS state changes to CONNECTED.
             //
             // Reference: https://w3c.github.io/webrtc-pc/#rtcpeerconnectionstate-enum
+#ifdef ENABLE_KVS_THREADPOOL
+#ifdef KVS_USE_OPENSSL
```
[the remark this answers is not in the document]
disa6302: `#if defined (ENABLE_KVS_THREADPOOL) && defined (KVS_USE_OPENSSL)` this somehow did not work. It did not honour the setting after `&&` and mbedtls builds failed. Hence, split it. But I will give it a shot again.

src/source/Crypto/Dtls_openssl.c
disa6302: I will add some unit tests. But `releaseDtlsSession` also has an internal NULL check. So we are neither acquiring nor releasing here, so we are safe. But, I will move the acquire before the NULL check in all functions to avoid any future issues due to code changes.

disa6302: That is correct. Misunderstood the context.

PR #1825, opened 2023-10-11
Ice turn sleepless state machine

jdelapla:
*Issue #, if available:*

*What was changed?*
- Ice Agent State Machine
- Turn connection state machine

*Why was it changed?*
To reduce time to first frame

*How was it changed?*
- Turn now has a proper state machine
- Ice and Turn state machines reside in a timer (as they did before), but are now kicked whenever a state transition is possible

Existing unit tests cover the changes, new functionality tests were added for the timer queue & state machine specific APIs


src/source/Crypto/Dtls_openssl.c
```diff
@@ -539,6 +541,7 @@ STATUS dtlsSessionHandshakeInThread(PDtlsSession pDtlsSession, BOOL isServer)
                         }
                         CHK_STATUS(dtlsCheckOutgoingDataBuffer(pDtlsSession));
```
disa6302, line 542: Let us remove this since we are checking this 2 lines down anyways.

src/source/Ice/IceAgent.c
```diff
@@ -2583,6 +2599,7 @@ STATUS handleStunPacket(PIceAgent pIceAgent, PBYTE pBuffer, UINT32 bufferLen, PS
             if (pIceCandidatePair->state != ICE_CANDIDATE_PAIR_STATE_SUCCEEDED) {
                 DLOGD("Pair succeeded! %s %s", pIceCandidatePair->local->id, pIceCandidatePair->remote->id);
                 pIceCandidatePair->state = ICE_CANDIDATE_PAIR_STATE_SUCCEEDED;
+                // checkIceAgentStateMachine(pIceAgent);
```
disa6302, line 2602: Remove comment?

src/source/Ice/IceAgent.c
```diff
@@ -2112,6 +2120,10 @@ STATUS iceAgentNominatingStateSetup(PIceAgent pIceAgent)
 
     pIceAgent->stateEndTime = GETTIME() + pIceAgent->kvsRtcConfiguration.iceCandidateNominationTimeout;
 
+    MUTEX_UNLOCK(pIceAgent->lock);
+    locked = FALSE;
+    checkIceAgentStateMachine(pIceAgent);
```
disa6302, line 2125: In `checkIceAgentStateMachine`, we use `pIceAgent->pStateMachine`, which could be updated in multiple threads right? So, doesnt it need to happen within the mutex locks? Or we dont see this event happening?

src/source/Ice/TurnConnection.c
```diff
@@ -1226,137 +988,100 @@ BOOL turnConnectionGetRelayAddress(PTurnConnection pTurnConnection, PKvsIpAddres
...
     STATUS retStatus = STATUS_SUCCESS, sendStatus = STATUS_SUCCESS;
-    PTurnConnection pTurnConnection = (PTurnConnection) customData;
-    BOOL locked = FALSE, stopScheduling = FALSE;
     PTurnPeer pTurnPeer = NULL;
     PStunAttributeAddress pStunAttributeAddress = NULL;
     PStunAttributeChannelNumber pStunAttributeChannelNumber = NULL;
-    PStunAttributeLifetime pStunAttributeLifetime = NULL;
     UINT32 i = 0;
 
+    // turn mutex is assumed to be locked.
```
disa6302, line 999: I see this comment, but we do not end up locking in the places where this API is being used in the state machine. Am I missing where the locking is happening?

jdelapla: The lock is occurring above the state machine APIs, in step state machine.

src/source/Ice/IceAgent.c
jdelapla: The pattern isn't consistent between TURN and ICE state machines, but in all the from* function of the ice agent state machine locks occur where relevant. The check function does not edit the pStateMachine.

src/source/Crypto/Dtls_openssl.c
jdelapla: I think the one that was added was added incorrectly from a merge. Should I instead remove the new one since this bracket is inside an error check?

src/source/Ice/TurnConnection.c
disa6302: Ok I see it. Thanks.

disa6302 (dismissed):

src/source/Ice/IceAgent.c
jdelapla: done

src/source/Ice/TurnConnectionStateMachine.c
```diff
@@ -0,0 +1,708 @@
...
+    UINT64 state = TURN_STATE_CHECK_SOCKET_CONNECTION;
+
+    CHK(pTurnConnection != NULL && pState != NULL, STATUS_NULL_ARG);
+
+    if (pTurnConnection->state == TURN_STATE_CLEAN_UP || pTurnConnection->state == TURN_STATE_FAILED) {
+        *pState = pTurnConnection->state;
+        CHK(FALSE, STATUS_SUCCESS);
+    }
+
+    if (socketConnectionIsConnected(pTurnConnection->pControlChannel)) { // pTurnConnection->state == TURN_STATE_CHECK_SOCKET_CONNECTION) {
```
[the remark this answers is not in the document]
jdelapla: done

disa6302 (approved):

PR #1833, opened 2023-10-13
Fix frame indexing in master sample to handle SRTP connection not ready yet

stefankiesz:
*Issue #, if available:*

*What was changed?*
For the master sample, these changes will reset the frame file index back to the first index when the SRTP connection is not ready to ensure the first frame sent is a key frame. The master GStreamer sample has the addition of a log line to indicate that frames were dropped. Also, using a logger was added to the GStreamer sample to replace printf() calls.

*Why was it changed?*
The frame file index for the master sample is incremented for every writeFrame() call, even for writeFrame() attempts before the SRTP connection is ready. Such frames are dropped, but the index is still incremented, resulting in non keyframes being sent as the first successful frames. Logging replaced printf() calls to be more versatile and consistent with the rest of our samples.

*How was it changed?*
By updating the sample files.

Using the existing CI tests to build and run samples.

<br>

### DIFF
--- a/samples/kvsWebrtcClientMasterGstSample.c
+++ b/samples/kvsWebrtcClientMasterGstSample.c
@@ -180,7 +180,7 @@ PVOID sendGstreamerAudioVideo(PVOID args)
                                                     pSampleConfiguration->rtspUri);
 
                     if (stringOutcome > RTSP_PIPELINE_MAX_CHAR_COUNT) {
-                        printf("[KVS GStreamer Master] ERROR: rtsp uri entered exceeds maximum allowed length set by RTSP_PIPELINE_MAX_CHAR_COUNT\n");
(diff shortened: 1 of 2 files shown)


samples/kvsWebrtcClientMasterGstSample.c
```diff
@@ -180,7 +182,7 @@ PVOID sendGstreamerAudioVideo(PVOID args)
                                                     pSampleConfiguration->rtspUri);
 
                     if (stringOutcome > RTSP_PIPELINE_MAX_CHAR_COUNT) {
-                        printf("[KVS GStreamer Master] ERROR: rtsp uri entered exceeds maximum allowed length set by RTSP_PIPELINE_MAX_CHAR_COUNT\n");
+                        DLOGE("[KVS GStreamer Master] ERROR: rtsp uri entered exceeds maximum allowed length set by RTSP_PIPELINE_MAX_CHAR_COUNT\n");
```
disa6302, line 185: No /n required for DLOGx

disa6302 (approved):

PR #1840, opened 2023-10-25
Openssl 3.1.3

disa6302:
*Issue #, if available:*

*What was changed?*
Added support for OpenSSL 3.x version in the SDK

*Why was it changed?*
OpenSSL 1.1.1x went EOL in September 2023 and some of the APIs that the SDK uses are deprecated which mostly show up as warnings. This means these APIs could be removed in the future. To ensure we are up to date, this PR introduces support for OpenSSL 3.x, while also maintaining support for older OpenSSL versions.

*How was it changed?*
- Used `OPENSSL_VERSION_NUMBER` define to switch the APIs used where relevant. 
- Added a new cmake property: `BUILD_OLD_OPENSSL_VERSION` to allow building older version from source (`1.1.1w`) if `3.1.3` causes issues. System dependencies could also be used instead if applications/platforms require using older version, however, this property covers situations for applications that build the dependencies from source in our SDK
- Added CI jobs for different platforms to test of older OpenSSL version

- Local testing was done for RSA certificate generation and EC certificate generation to confirm the new APIs work
- For MD5, the hash was logged to compare the values between 1.1.1x and 3.x version
- Existing benchmark was used to verify no performance degradation in the handshake process.
- CI was used to fix build failures on Windows and ARM cross compilation that came with the upgrade


src/source/Crypto/Crypto.h
```diff
@@ -8,10 +8,32 @@ extern "C" {
...
-#define KVS_RSA_F4                  RSA_F4
-#define KVS_MD5_DIGEST_LENGTH       MD5_DIGEST_LENGTH
-#define KVS_SHA1_DIGEST_LENGTH      SHA_DIGEST_LENGTH
+#define KVS_RSA_F4             RSA_F4
+#define KVS_MD5_DIGEST_LENGTH  MD5_DIGEST_LENGTH
+#define KVS_SHA1_DIGEST_LENGTH SHA_DIGEST_LENGTH
+
+// https://www.openssl.org/docs/man3.1/man3/MD5.html
+#if (OPENSSL_VERSION_NUMBER >= 0x30000000L)
+#define KVS_MD5_DIGEST(m, mlen, ob)                                                                                                                  \
```
disa6302, line 17: Not sure why the formatting looks like this. But locally it is fine and clang-format passed locally and on CI.

.github/workflows/ci.yml
```diff
@@ -466,6 +466,32 @@ jobs:
         run:  |
           cd build
           timeout --signal=SIGABRT 60m ./tst/webrtc_client_test
```
niyatim23, line 468: Should we also add a macos-old-openssl build?

CMake/Dependencies/libwebsockets-CMakeLists.txt
```diff
@@ -51,7 +51,7 @@ ExternalProject_Add(project_libwebsockets
         -DLWS_WITH_STATIC=1
         -DLWS_WITH_SHARED=${LWS_WITH_SHARED}
         -DLWS_WITH_MBEDTLS=${LWS_WITH_MBEDTLS}
-        -DLWS_WITH_MINIMAL_EXAMPLES=1
+        -DLWS_WITH_MINIMAL_EXAMPLES=0
```
hassanctech, line 54: Why are we changing this?

src/source/Crypto/Dtls_openssl.c
```diff
@@ -104,26 +104,48 @@ STATUS createCertificateAndKey(INT32 certificateBits, BOOL generateRSACertificat
...
+        CHK_ERR(EVP_PKEY_keygen_init(pctx) > 0, STATUS_CERTIFICATE_GENERATION_FAILED, "Error initializing EC keygen");
+        CHK_ERR(EVP_PKEY_CTX_set_ec_paramgen_curve_nid(pctx, OBJ_sn2nid("prime256v1")) > 0, STATUS_CERTIFICATE_GENERATION_FAILED,
+                "Error setting curve name");
+        CHK_ERR(EVP_PKEY_keygen(pctx, ppPkey) > 0, STATUS_CERTIFICATE_GENERATION_FAILED, "EC Key generation errored");
+    }
+    EVP_PKEY_CTX_free(pctx);
+#else
+    RSA* pRsa = NULL;
     if (generateRSACertificate) {
+        BIGNUM* pBne = NULL;
```
hassanctech, line 135: Defining this here instead of at the start breaks the C89 / old compiler compliance we have enforced in our SDK.  These need to be moved back up to the top of the function.

src/source/Crypto/Dtls_openssl.c
```diff
@@ -104,26 +104,48 @@ STATUS createCertificateAndKey(INT32 certificateBits, BOOL generateRSACertificat
...
         CHK((pRsa = RSA_new()) != NULL, STATUS_CERTIFICATE_GENERATION_FAILED);
         CHK(RSA_generate_key_ex(pRsa, certificateBits, pBne, NULL) != 0, STATUS_CERTIFICATE_GENERATION_FAILED);
         CHK((EVP_PKEY_assign_RSA(*ppPkey, pRsa)) != 0, STATUS_CERTIFICATE_GENERATION_FAILED);
+        if (pBne != NULL) {
+            BN_free(pBne);
+        }
         pRsa = NULL;
     } else {
+        UINT32 eccGroup = 0;
+        EC_KEY* eccKey = NULL;
```
hassanctech, line 148: Same

.github/workflows/ci.yml
disa6302: I probably missed it. I will add it.

CMake/Dependencies/libwebsockets-CMakeLists.txt
disa6302: We dont need to build minimal examples for LWS build. It just adds on to the build time.

src/source/Crypto/Dtls_openssl.c
disa6302: Hmm. The reason I added it here was because Mac was complaining about unused variables. I can move it on the top and #if #else there too then.

src/source/Crypto/Dtls_openssl.c
```diff
@@ -193,12 +212,16 @@ STATUS createSslCtx(PDtlsSessionCertificateInfo pCertificates, UINT32 certCount,
 
     CHK(pSslCtx != NULL, STATUS_SSL_CTX_CREATION_FAILED);
 
-    // Version greater than or equal to 1.0.2
-#if (OPENSSL_VERSION_NUMBER >= 0x10002000L)
+    // Version less than 1.1.0
```
niyatim23, line 215: Shouldn't this be >= 1.1.1?

niyatim23 (dismissed):
Nice!

src/source/Crypto/Crypto.h
```diff
@@ -8,10 +8,32 @@ extern "C" {
...
+            mdctx = EVP_MD_CTX_new();                                                                                                                \
+            EVP_DigestInit_ex(mdctx, md, NULL);                                                                                                      \
+            EVP_DigestUpdate(mdctx, m, mlen);                                                                                                        \
+            EVP_DigestFinal_ex(mdctx, ob, NULL);                                                                                                     \
+            EVP_MD_CTX_free(mdctx);                                                                                                                  \
+            EVP_MD_free((EVP_MD*) md);                                                                                                               \
+        }                                                                                                                                            \
```
hassanctech, line 31: I see in the docs it says:

>Ignoring failure returns of EVP_DigestInit_ex(), EVP_DigestInit_ex2(), or EVP_DigestInit() can lead to undefined behavior on subsequent calls updating or finalizing the EVP_MD_CTX such as the EVP_DigestUpdate() or EVP_DigestFinal() functions. The only valid calls on the EVP_MD_CTX when initialization fails are calls that attempt another initialization of the context or release the context.

There is an example here:
```
mdctx = EVP_MD_CTX_new();
    if (!EVP_DigestInit_ex2(mdctx, md, NULL)) {
        printf("Message digest initialization failed.\n");
        EVP_MD_CTX_free(mdctx);
        exit(1);
    }
    if (!EVP_DigestUpdate(mdctx, mess1, strlen(mess1))) {
        printf("Message digest update failed.\n");
        EVP_MD_CTX_free(mdctx);
        exit(1);
    }
    if (!EVP_DigestUpdate(mdctx, mess2, strlen(mess2))) {
        printf("Message digest update failed.\n");
        EVP_MD_CTX_free(mdctx);
        exit(1);
    }
    if (!EVP_DigestFinal_ex(mdctx, md_value, &md_len)) {
        printf("Message digest finalization failed.\n");
        EVP_MD_CTX_free(mdctx);
        exit(1);
    }
    EVP_MD_CTX_free(mdctx);
```

Taken from here:  https://www.openssl.org/docs/man3.0/man3/EVP_DigestInit_ex2.html

We should confirm if we need to call `EVP_DigestInit_ex` or `EVP_DigestInit_ex2`.  Regardless the error checking applies to both.

src/source/Crypto/Dtls_openssl.c
```diff
@@ -104,24 +104,48 @@ STATUS createCertificateAndKey(INT32 certificateBits, BOOL generateRSACertificat
...
     EC_KEY* eccKey = NULL;
-    UINT64 certSn;
-
+#endif
     CHK(ppCert != NULL && ppPkey != NULL, STATUS_NULL_ARG);
     CHK((*ppPkey = EVP_PKEY_new()) != NULL, STATUS_CERTIFICATE_GENERATION_FAILED);
     CHK_STATUS(dtlsFillPseudoRandomBits((PBYTE) &certSn, SIZEOF(UINT64)));
 
+#if (OPENSSL_VERSION_NUMBER >= 0x30000000L)
+    DLOGI("Detected openssl version greater than 3.0.0");
```
hassanctech, line 122: "greater than or equal to"

src/source/Crypto/Dtls_openssl.c
```diff
@@ -104,24 +104,48 @@ STATUS createCertificateAndKey(INT32 certificateBits, BOOL generateRSACertificat
...
+        DLOGI("Detected older version");
         CHK((pBne = BN_new()) != NULL, STATUS_CERTIFICATE_GENERATION_FAILED);
         CHK(BN_set_word(pBne, KVS_RSA_F4) != 0, STATUS_CERTIFICATE_GENERATION_FAILED);
-
         CHK((pRsa = RSA_new()) != NULL, STATUS_CERTIFICATE_GENERATION_FAILED);
         CHK(RSA_generate_key_ex(pRsa, certificateBits, pBne, NULL) != 0, STATUS_CERTIFICATE_GENERATION_FAILED);
         CHK((EVP_PKEY_assign_RSA(*ppPkey, pRsa)) != 0, STATUS_CERTIFICATE_GENERATION_FAILED);
+        if (pBne != NULL) {
+            BN_free(pBne);
+        }
```
hassanctech, line 148: This still should be in CleanUp -- all the above CHKs if they fail will skip this free call.

hassanctech (changes requested):
Need to properly handle failure return on `EVP_DigestInit_ex()` or else we will have undefined behavior:

>Ignoring failure returns of EVP_DigestInit_ex(), EVP_DigestInit_ex2(), or EVP_DigestInit() can lead to undefined behavior on subsequent calls updating or finalizing the EVP_MD_CTX such as the EVP_DigestUpdate() or EVP_DigestFinal() functions. The only valid calls on the EVP_MD_CTX when initialization fails are calls that attempt another initialization of the context or release the context.

From: https://www.openssl.org/docs/man3.0/man3/EVP_DigestInit_ex.html

src/source/Crypto/Dtls_openssl.c
disa6302: My bad. Missed moving that to cleanup when I moved the bne out of this block.

src/source/Crypto/Crypto.h
hassanctech: It looks like this was still not addressed?

disa6302: I missed removing this macro altogether. Done. Also  about `EVP_DigestInit_ex` vs `EVP_DigestInit_ex2`. ex2 takes in a params options that we can use to configure the hashing operation, but we are not doing that. So `EVP_DigestInit_ex` is ok for now.

src/source/Crypto/Dtls_openssl.c
```diff
@@ -193,12 +245,16 @@ STATUS createSslCtx(PDtlsSessionCertificateInfo pCertificates, UINT32 certCount,
...
+#if (OPENSSL_VERSION_NUMBER < 0x10100000L)
     SSL_CTX_set_ecdh_auto(pSslCtx, TRUE);
-#else
+
+    // Version less than 3.0.0 and greater than 1.1.0
+#elif (OPENSSL_VERSION_NUMBER < 0x30000000L)
     CHK((ecdh = EC_KEY_new_by_curve_name(NID_X9_62_prime256v1)) != NULL, STATUS_SSL_CTX_CREATION_FAILED);
     CHK(SSL_CTX_set_tmp_ecdh(pSslCtx, ecdh) == 1, STATUS_SSL_CTX_CREATION_FAILED);
+#else
+    DLOGI("ECDH enabled by default in 3.0.0. Nothing to do");
```
hassanctech, line 257: Is this intended change in behavior?  Previously for openssl 1.0.0 we should call:

```
    CHK((ecdh = EC_KEY_new_by_curve_name(NID_X9_62_prime256v1)) != NULL, STATUS_SSL_CTX_CREATION_FAILED);
    CHK(SSL_CTX_set_tmp_ecdh(pSslCtx, ecdh) == 1, STATUS_SSL_CTX_CREATION_FAILED);
```

With this change with openssl 1.0.0 we call:

```
SSL_CTX_set_ecdh_auto(pSslCtx, TRUE);
```

disa6302: Yes. It is intended. Post 1.1.0, I do not see the auto API. Its deprecated. 
https://www.openssl.org/docs/man1.1.1/man3/

Before we would invoke:
```
SSL_CTX_set_ecdh_auto(pSslCtx, TRUE);
 CHK((ecdh = EC_KEY_new_by_curve_name(NID_X9_62_prime256v1)) != NULL, STATUS_SSL_CTX_CREATION_FAILED);
    CHK(SSL_CTX_set_tmp_ecdh(pSslCtx, ecdh) == 1, STATUS_SSL_CTX_CREATION_FAILED);
```

I just rearranged to make sure we dont call a deprecated API for versions >1.1.0

disa6302: Ok. I might have been wrong here. I just opened 1.0.0 on codespaces and could not find the API when I grep'ed it. `SSL_CTX_set_ecdh_auto` seems to have been introduced in 1.0.2. I will fix this.

src/source/Crypto/Dtls_openssl.c
```diff
@@ -193,12 +245,16 @@ STATUS createSslCtx(PDtlsSessionCertificateInfo pCertificates, UINT32 certCount,
...
+// https://www.openssl.org/docs/man1.0.2/man3/
+// https://www.openssl.org/docs/man1.1.1/man3/
+// Version < 3.0.0
+#if (OPENSSL_VERSION_NUMBER < 0x30000000L)
     CHK((ecdh = EC_KEY_new_by_curve_name(NID_X9_62_prime256v1)) != NULL, STATUS_SSL_CTX_CREATION_FAILED);
     CHK(SSL_CTX_set_tmp_ecdh(pSslCtx, ecdh) == 1, STATUS_SSL_CTX_CREATION_FAILED);
+#else
+    // https://www.openssl.org/docs/man3.0/man3/EC_KEY_new_by_curve_name.html -- indicates that EC_KEY_new_by_curve_name and SSL_CTX_set_tmp_ecdh are
+    // deprecated https://www.openssl.org/docs/man1.1.1/man3/SSL_CTX_set1_groups.html
+    CHK(SSL_CTX_set1_groups_list(pSslCtx, "prime256v1") == 1, STATUS_SSL_CTX_CREATION_FAILED);
```
hassanctech, line 257: Ok so I see that you're using the new API `SSL_CTX_set1_groups_list` for open ssl >= 3.0.0.  But where the diff now is compared to our existing code is for all versions higher than 1.0.2 (so 1.0.2 <= x < 3.0.0), really we did not enforce the upper bound before but for all intents in purposes for theses versions we were previously using `SSL_CTX_set_ecdh_auto`.  That I want to understand why are we changing the behavior for existing versions of openssl that's already been tried and tested, was there a bug with this or what is the reason for this change?  IMO unless there is a good reason this change should only impact when the version is >= 3.0.0 and otherwise it should not touch the existing behavior at all, unless there is a good reason to do that.  I'm not following what that reason is.

disa6302: `SSL_CTX_set_ecdh_auto` is primarily used for automatic curve selection during SSL/TLS handshakes, making it convenient for situations where you want to delegate curve selection to OpenSSL. On the other hand, `EC_KEY_new_by_curve_name` is used when you need to explicitly specify the ECC curve for ECC key generation and want more control over the curve selection process. I am not sure what `SSL_CTX_set_ecdh_auto` selects by default, but given we are controlling the curve selection pre 1.0.2, it only made sense to do the same post 1.0.2 since the API is available. Another reason for the change was I wanted to avoid having multiple ifdefs that make the code buggy when it is unnecessary. I do understand the point of the existing code already be tested, but the functions I have included for versions between 1.0.2 and 3.0.0 are also tested as being part of version <1.0.2. Without this change, the ifdefs look like:

```
#if (OPENSSL_VERSION_NUMBER >= 0x10002000L && OPENSSL_VERSION_NUMBER < 0x3000000L)
    SSL_CTX_set_ecdh_auto(pSslCtx, TRUE);
#elif (OPENSSL_VERSION_NUMBER < 0x10002000L)
    CHK((ecdh = EC_KEY_new_by_curve_name(NID_X9_62_prime256v1)) != NULL, STATUS_SSL_CTX_CREATION_FAILED);
    CHK(SSL_CTX_set_tmp_ecdh(pSslCtx, ecdh) == 1, STATUS_SSL_CTX_CREATION_FAILED);
#elif (OPENSSL_VERSION_NUMBER >= 0x3000000L)
    CHK(SSL_CTX_set1_groups_list(pSslCtx, "prime256v1") == 1, STATUS_SSL_CTX_CREATION_FAILED);
#endif
```
Let me know if you still feel strongly about this change and I can switch it to this block. But, given `EC_KEY_new_by_curve_name` and `SSL_CTX_set_tmp_ecdh` are already tested for versions less than 1.0.2, I did not see the harm in using the same for 1.0.2 to 3.0.0.

hassanctech: I did some digging and I found `SSL_CTX_set_ecdh_auto` was entirely removed in openssl 1.1.0 [https://github.com/openssl/openssl/issues/1437#issuecomment-238828306].  Then it was added back as a dummy no-op macro for backwards compat like a week later in 1.1.0a (https://github.com/openssl/openssl/commit/2ecb9f2d18614fb7b7b42830a358b7163ed43221).  So I suppose we've actually been getting lucky if someone was using this against 1.1.0 it wouldn't compile.

I cannot seem to find the manpages for openssl 1.0.0 to verify if `EC_KEY_new_by_curve_name` existed then, and also what about the curve itself, has `NID_X9_62_prime256v1` always existed.  Let me just see if I can find some official references if not I'll look through the source.

disa6302: Yeah. There doesnt seem to be official man pages pre-1.0.2. But I just checked out the branch to check if the API existed.

sirknightj:
Support for openssl-3.5.7 was added in https://github.com/awslabs/amazon-kinesis-video-streams-webrtc-sdk-c/pull/2354 (supersedes this), released in the 1.20.0 version

PR #1846, opened 2023-11-14
Time-to-first-frame metrics benchmarking using datachannel

niyatim23:
*Issue #, if available:*

*What was changed?*
- Introduced a new feature which allows sending peerConnectionMetrics, signalingMetrics, iceAgentMetrics, and datachannelMetrics to the viewer
- Modified the dataChannelOnMessage to send a series to timestamped JSON messages to the viewer to calculate the e2e time taken by the data channel.

*Why was it changed?*
It would be easier to visualize these on the viewer in a timeline chart to understand which calls / steps are time-consuming

*How was it changed?*
- Introduced a new CMake flag, which, when enabled, allows sending these metrics in multiple JSON messages to the viewer via the datachannel. 
- Datachannel metrics: Datachannel related e2e metrics has been introduced as a part of this feature. The datachannel first receives a message of the following format with t1 attached, to which the master responds back with the same message attaching t2. In response to this, the viewer sends the same message back with t3 and so on until t5. The viewer is responsible for attaching t1, t3, t5. The master is responsible for t2 and t4. (Master e2e time: t4 - t2, Viewer e2e time: t3 - t1)
```
let dataChannelLatencyCalcMessage = {
    content: 'Opened data channel by viewer',
    t1: '',
    t2: '',
    t3: '',
    t4: '',
    t5: ''
}
```

Tested these changes locally by running the master with a JS viewer and making sure the following messages sent by the master are received by the viewer
Screenshot 2023-11-14 at 9 11 16 AM

This can be visualized on the viewer as:
Screenshot 2023-11-14 at 9 10 02 AM

Error messages:
Screenshot 2023-12-19 at 8 57 39 PM
Screenshot 2023-12-19 at 8 58 20 PM
Screenshot 2023-12-19 at 8 55 50 PM

Regular mode
Screenshot 2023-12-19 at 9 06 27 PM


samples/Samples.h
```diff
@@ -131,6 +152,17 @@ typedef struct {
     UINT32 logLevel;
 } SampleConfiguration, *PSampleConfiguration;
 
+#ifdef ENABLE_SENDING_METRICS_TO_VIEWER
+typedef struct {
+    CHAR content[100];
+    CHAR t1[20];
```
disa6302, line 158: Can we make these names more descriptive? t1-t5 is pretty vague.

src/source/Signaling/Signaling.c
```diff
@@ -380,10 +380,10 @@ STATUS signalingSendMessageSync(PSignalingClient pSignalingClient, PSignalingMes
     CHK_STATUS(sendLwsMessage(pSignalingClient, pSignalingMessage->messageType, pSignalingMessage->peerClientId, pSignalingMessage->payload,
                               pSignalingMessage->payloadLen, pSignalingMessage->correlationId, 0));
     if (pSignalingMessage->messageType == SIGNALING_MESSAGE_TYPE_ANSWER) {
-        PROFILE_WITH_START_TIME_OBJ(pSignalingClient->offerTime, pSignalingClient->diagnostics.offerToAnswerTime, "Offer to answer time");
+        PROFILE_WITH_START_END_TIME_OBJ(pSignalingClient->offerTime, pSignalingClient->answerTime, "Offer to answer time");
     }
     if (pSignalingMessage->messageType == SIGNALING_MESSAGE_TYPE_OFFER) {
-        pSignalingClient->offerTime = GETTIME();
+        pSignalingClient->offerTime = GETTIME() / HUNDREDS_OF_NANOS_IN_A_MILLISECOND;
```
disa6302, line 386: Lets not do the conversion here. These should all be just `GETTIME()` and the conversion should happen in the macro. Any reason why we have modified the timestamp to be in milliseconds by default in multiple places?

src/include/com/amazonaws/kinesis/video/webrtcclient/Stats.h
```diff
@@ -593,20 +607,17 @@ typedef struct {
                                      //!< In all of these cases the error callback (if specified) will be called.
     UINT32 numberOfReconnects;       //!< Number of reconnects in the session
     UINT32 apiCallRetryCount;        //!< Number of retries due to API call failures in the state machine
-    UINT64 getTokenCallTime;         //!< Time (ms) taken to get credentials for signaling
```
disa6302, line 596: You are removing members of a public struct which will break backward compatibility. Unless there is a strong reason to make this change, the existing implementations should not change. If the goal is to record start and end time, let us record them in the new members and not change any existing calculations. Same applies to wherever we are removing public members.

src/source/Ice/IceAgent.c
```diff
@@ -599,7 +599,7 @@ STATUS iceAgentStartGathering(PIceAgent pIceAgent)
     CHK(!ATOMIC_LOAD_BOOL(&pIceAgent->agentStartGathering), retStatus);
 
     ATOMIC_STORE_BOOL(&pIceAgent->agentStartGathering, TRUE);
-    pIceAgent->candidateGatheringStartTime = GETTIME();
+    pIceAgent->candidateGatheringProcessStartTime = GETTIME() / HUNDREDS_OF_NANOS_IN_A_MILLISECOND;
```
disa6302, line 602: Let us not do the GETTIME() conversion to milliseconds here. We should use GETTIME() and only convert wherever the calculation is done. Comment applicable wherever this is being done in this PR.

disa6302 (changes requested):

.github/workflows/ci.yml
```diff
@@ -49,7 +49,29 @@ jobs:
       - name: Run tests
         run:  |
           cd build
-          ./tst/webrtc_client_test
+          ./tst/webrtc_client_test    
+  mac-os-dc-metrics-build-gcc:
```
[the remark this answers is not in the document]
niyatim23: The new build is running with the cmake flag. Since the CI was not running the code that I added I introduced that build. Should I remove it?

CMakeLists.txt
```diff
@@ -19,6 +19,7 @@ option(BUILD_LIBSRTP_DESTINATION_PLATFORM "If buildng LibSRTP what is the destin
 option(BUILD_SAMPLE "Build available samples" ON)
 option(ENABLE_DATA_CHANNEL "Enable support for data channel" ON)
 option(ENABLE_KVS_THREADPOOL "Enable support for KVS thread pool in signaling" ON)
+option(ENABLE_SENDING_METRICS_TO_VIEWER "Enable sending master-side metrics to the viewer as a json string via the datachannel" OFF)
```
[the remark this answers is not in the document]
niyatim23: This was discussed and agreed upon with James since this change should not work unless this flag is enabled as it is only intended for customers who want to benchmark. If this flag is enabled, the C SDK only accepts json messages from the other peer, attaches timestamps to them and responds with the timestamps from the signaling, ice agent, gathering phases

src/include/com/amazonaws/kinesis/video/webrtcclient/Stats.h
hassanctech: +1.  We should not remove these.

CMakeLists.txt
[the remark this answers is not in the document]
disa6302: I agree that the changes in this PR for sending profiling data via data channel should not happen unless an option is enabled. But, IMO it shouldnt be a CMake option given the core of the logic is in the callback that is implemented in the samples. CMake flags should be reserved for core SDK and should not be used for samples unless there is no other way to achieve the functionality. Is there a reason we cannot achieve setting this option by having a bool flag in one of the sample structs and pass that in as customData to `onDataChannelMessage`?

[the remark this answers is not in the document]
niyatim23: Removed as discussed

samples/Common.c
```diff
@@ -2,6 +2,9 @@
 #include "Samples.h"
 
 PSampleConfiguration gSampleConfiguration = NULL;
+CHAR pPeerConnectionMetricsMessage[MAX_PEER_CONNECTION_METRICS_MESSAGE_SIZE];
```
disa6302, line 5: Cant we have these in `pSampleStreamingSession` since you already are using it in the callback? 
As a rule of thumb, we should avoid using globals unless it is absolutely necessary.

samples/Common.c
```diff
@@ -34,18 +37,123 @@ STATUS signalingCallFailed(STATUS status)
...
-    UNUSED_PARAM(customData);
-    if (isBinary) {
-        DLOGI("DataChannel Binary Message");
+    STATUS retStatus = STATUS_SUCCESS;
+    UINT32 i, strLen, tokenCount;
+    CHAR pMessageSend[SIZEOF(DataChannelMessage)];
+    PCHAR json;
+    PSampleStreamingSession pSampleStreamingSession = (PSampleStreamingSession) customData;
+    PSampleConfiguration pSampleConfiguration = pSampleStreamingSession->pSampleConfiguration;
+    DataChannelMessage dataChannelMessage;
```
disa6302, line 46: We should have a NULL check here. If NULL, we should send the other end a message saying `unable to send metrics` or something like that.

samples/Common.c
```diff
@@ -34,18 +37,123 @@ STATUS signalingCallFailed(STATUS status)
...
+
+        if (tokenCount > 1) {
+            CHK(tokens[0].type == JSMN_OBJECT, STATUS_INVALID_API_CALL_RETURN_JSON);
+            DLOGI("DataChannel json message: %.*s\n", pMessageLen, pMessage);
+
+            for (i = 1; i < tokenCount; i++) {
+                if (compareJsonString(json, &tokens[i], JSMN_STRING, (PCHAR) "content")) {
+                    strLen = (UINT32) (tokens[i + 1].end - tokens[i + 1].start);
+                    STRNCPY(dataChannelMessage.content, json + tokens[i + 1].start, tokens[i + 1].end - tokens[i + 1].start);
+                } else if (compareJsonString(json, &tokens[i], JSMN_STRING, (PCHAR) "timestamp1")) {
```
disa6302, line 63: What do timestamp1,2,3 mean? Can we be more descriptive? Is the reason we are not being descriptive because the json can be in any order?

src/include/com/amazonaws/kinesis/video/webrtcclient/Include.h
```diff
@@ -46,6 +46,23 @@ extern "C" {
         DLOGP("[%s] Time taken: %" PRIu64 " ms", msg, (GETTIME() - (t)) / HUNDREDS_OF_NANOS_IN_A_MILLISECOND);                                       \
     } while (FALSE)
 
+#define PROFILE_CALL_WITH_START_END_T_OBJ(f, s, e, d, msg)                                                                                           \
+    do {                                                                                                                                             \
+        s = GETTIME() / HUNDREDS_OF_NANOS_IN_A_MILLISECOND;                                                                                          \
```
disa6302, line 51: Let us not calculate MILLISECONDS here. These should only be timestamp recording. The millisecond conversion should happen in `d= e - s`. Also, just a nit. But let us be consistent with naming across the macros and switch to t1, t2, etc.

src/source/Signaling/LwsApiCalls.c
```diff
@@ -2393,6 +2393,13 @@ PVOID receiveLwsMessageWrapper(PVOID args)
     }
     // Calling client receive message callback if specified
     if (pSignalingClient->signalingClientCallbacks.messageReceivedFn != NULL) {
+        if (messageType == SIGNALING_MESSAGE_TYPE_OFFER) {
```
disa6302, line 2396: Why are we adding this section here? Cant we reuse: `pSignalingClient->offerReceivedTime = GETTIME();` which is on line 2376? Same with answer.

src/include/com/amazonaws/kinesis/video/webrtcclient/Include.h
niyatim23: That is because we need to send it across in milliseconds, I moved all the conversion outside of the macros to inside as per your previous comments

disa6302: Ok yeah. My bad then. We are sending start and end timestamps and that is why we are doing this.

src/source/Signaling/LwsApiCalls.c
niyatim23: Removed it. I didn't notice that earlier because it was not present in the SignalingClientStats that I'm using later

samples/Common.c
niyatim23: Sounds good! Moved inside streamingSession

samples/Common.c
niyatim23: Added a NULL check!

samples/Common.c
niyatim23: Renamed everything here and in the JS SDK

samples/Common.c
```diff
@@ -34,18 +34,150 @@ STATUS signalingCallFailed(STATUS status)
...
+    UINT32 i, strLen, tokenCount;
+    CHAR pMessageSend[MAX_DATA_CHANNEL_METRICS_MESSAGE_SIZE];
+    PCHAR json;
+    PSampleStreamingSession pSampleStreamingSession = (PSampleStreamingSession) customData;
+    PSampleConfiguration pSampleConfiguration = pSampleStreamingSession->pSampleConfiguration;
+    DataChannelMessage dataChannelMessage;
+    jsmn_parser parser;
+    jsmntok_t tokens[MAX_JSON_TOKEN_COUNT];
+
+    CHK(pMessage != NULL && pDataChannel != NULL, STATUS_NULL_ARG);
```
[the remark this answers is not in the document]
niyatim23: Added, thanks for the suggestion!

samples/Common.c
```diff
@@ -34,18 +34,153 @@ STATUS signalingCallFailed(STATUS status)
...
 {
-    UNUSED_PARAM(customData);
-    if (isBinary) {
-        DLOGI("DataChannel Binary Message");
+    STATUS retStatus = STATUS_SUCCESS;
+    UINT32 i, strLen, tokenCount;
+    CHAR pMessageSend[MAX_DATA_CHANNEL_METRICS_MESSAGE_SIZE];
+    PCHAR json;
+    PSampleStreamingSession pSampleStreamingSession = (PSampleStreamingSession) customData;
+    PSampleConfiguration pSampleConfiguration = pSampleStreamingSession->pSampleConfiguration;
```
disa6302, line 42: If `pSampleStreamingSession` is NULL, then this would segfault. Assign only after the check.

niyatim23: Done!

niyatim23: Also added screenshots from the viewer showing that it received these messages in the PR description

disa6302 (approved):
LGTM. Let us wait on CI to pass.

disa6302 (approved):

PR #1854, opened 2023-11-27
Asynchronous get ice config

jdelapla:
*What was changed?*
GetIceServerConfig can now be called asynchronously when threadpool is enabled, and is no longer called by the fetch() API, and has to be directly invoked through refresh or gather ice config info.

*Why was it changed?*
GetIceServerConfig has a long time to resolution and is not involved in finding srflx candidates, however its previous sequential and synchronous design after receiving an offer slowed down time to first frame.

*How was it changed?*
Added to a threadpool, ice agent was changed to allow for relay candidates to be discovered after initial object creation.

Test added, multiple tests changed, and tested on CI & manually.


samples/Common.c
```diff
@@ -356,12 +356,56 @@ VOID onIceCandidateHandler(UINT64 customData, PCHAR candidateJson)
     CHK_LOG_ERR(retStatus);
 }
 
+// TODO this should all be in a higher webrtccontext layer above PeerConnection
+// Placing it here now since this is where all the current webrtccontext functions are placed
+typedef struct {
```
disa6302, line 361: Nit: Move to Sample.h

src/include/com/amazonaws/kinesis/video/webrtcclient/Include.h
```diff
@@ -1573,6 +1573,8 @@ typedef struct {
  */
 PUBLIC_API STATUS createPeerConnection(PRtcConfiguration, PRtcPeerConnection*);
 
+PUBLIC_API STATUS addConfigToServerList(PRtcPeerConnection*, PIceConfigInfo);
```
disa6302, line 1576: Since it is a public API, better to be specific: `addIceConfigToServerList`

src/source/Ice/IceAgent.c
```diff
@@ -613,9 +676,10 @@ STATUS iceAgentStartGathering(PIceAgent pIceAgent)
         PROFILE_CALL_WITH_T_OBJ(CHK_STATUS(iceAgentInitSrflxCandidate(pIceAgent)), pIceAgent->iceAgentProfileDiagnostics.srflxCandidateSetUpTime,
                                 "Srflx candidates setup time");
     }
-
+#if 0
```
disa6302, line 679: Remove this section instead of #if block since we moved this to `AddConfig` now.

src/source/PeerConnection/PeerConnection.h
```diff
@@ -176,6 +176,7 @@ STATUS onFrameDroppedFunc(UINT64, UINT16, UINT16, UINT32);
 VOID onSctpSessionOutboundPacket(UINT64, PBYTE, UINT32);
 VOID onSctpSessionDataChannelMessage(UINT64, UINT32, BOOL, PBYTE, UINT32);
 VOID onSctpSessionDataChannelOpen(UINT64, UINT32, PBYTE, UINT32);
+PVOID asyncGetIceConfigInfoWrapper(PVOID);
```
disa6302, line 179: Unused: remove

src/include/com/amazonaws/kinesis/video/webrtcclient/Include.h
disa6302: Also, can we add a simple API test in our unit tests with valid/invalid args?

samples/Common.c
```diff
@@ -356,12 +356,56 @@ VOID onIceCandidateHandler(UINT64 customData, PCHAR candidateJson)
...
+             *
+             * It's recommended to not pass too many TURN iceServers to configuration because it will slow down ice gathering in non-trickle mode.
+             */
+            CHK_STATUS(signalingClientGetIceConfigInfo(data->signalingClientHandle, i, &pIceConfigInfo));
+            CHECK(uriCount < MAX_ICE_SERVERS_COUNT);
+            uriCount += pIceConfigInfo->uriCount;
+            CHK_STATUS(addConfigToServerList(data->ppRtcPeerConnection, pIceConfigInfo));
+        }
+    }
+    *(data->pUriCount) += uriCount;
```
disa6302, line 395: Is there a reason to do this assignment given we are freeing `data` in the Cleanup anyways?

samples/Common.c
```diff
@@ -356,12 +356,56 @@ VOID onIceCandidateHandler(UINT64 customData, PCHAR candidateJson)
...
+             * if configuration.iceServers[uriCount + 1].urls is "turn:ip:port?transport=udp" then ICE will try TURN over UDP
+             * if configuration.iceServers[uriCount + 1].urls is "turn:ip:port?transport=tcp" then ICE will try TURN over TCP/TLS
+             * if configuration.iceServers[uriCount + 1].urls is "turns:ip:port?transport=udp", it's currently ignored because sdk dont do TURN
+             * over DTLS yet. if configuration.iceServers[uriCount + 1].urls is "turns:ip:port?transport=tcp" then ICE will try TURN over TCP/TLS
+             * if configuration.iceServers[uriCount + 1].urls is "turn:ip:port" then ICE will try both TURN over UDP and TCP/TLS
+             *
+             * It's recommended to not pass too many TURN iceServers to configuration because it will slow down ice gathering in non-trickle mode.
+             */
+            CHK_STATUS(signalingClientGetIceConfigInfo(data->signalingClientHandle, i, &pIceConfigInfo));
+            CHECK(uriCount < MAX_ICE_SERVERS_COUNT);
```
disa6302, line 390: I think this check should happen after uriCount is updated. `uriCount` would be 0 here anyways, so it would pass.

samples/Common.c
```diff
@@ -430,6 +443,39 @@ STATUS initializePeerConnection(PSampleConfiguration pSampleConfiguration, PRtcP
...
+             */
+            CHK_STATUS(signalingClientGetIceConfigInfo(pSampleConfiguration->signalingClientHandle, i, &pIceConfigInfo));
+            CHECK(uriCount < MAX_ICE_SERVERS_COUNT);
+            uriCount += pIceConfigInfo->uriCount;
+            CHK_STATUS(addConfigToServerList(ppRtcPeerConnection, pIceConfigInfo));
+        }
+#endif
+    }
+
+    pSampleConfiguration->iceUriCount = uriCount + 1;
```
disa6302, line 478: `uriCount` is a local variable, so the count would end up getting updated for sync/async case. For async case, the URI count is updated in the thread. So need to move this count value into the else block.

samples/Common.c
```diff
@@ -430,6 +443,39 @@ STATUS initializePeerConnection(PSampleConfiguration pSampleConfiguration, PRtcP
...
+
+    if (pSampleConfiguration->useTurn) {
+#ifdef ENABLE_KVS_THREADPOOL
+        AsyncGetIceStruct* pAsyncData = NULL;
+
+        pAsyncData = (AsyncGetIceStruct*) MEMCALLOC(1, SIZEOF(AsyncGetIceStruct));
+        pAsyncData->signalingClientHandle = pSampleConfiguration->signalingClientHandle;
+        pAsyncData->ppRtcPeerConnection = ppRtcPeerConnection;
+        pAsyncData->pUriCount = &(pSampleConfiguration->iceUriCount);
+        CHK_STATUS(peerConnectionAsync(asyncGetIceConfigInfo, (PVOID) pAsyncData));
```
disa6302, line 455: I think we should update `pSampleConfiguration->iceUriCount` and then MEMFREE(pAsyncData) in this function instead of within the thread function. We update `pUriCount` there, but then immediately free it, so, `pSampleConfiguration-> iceUriCount ` would not have the right value.

jdelapla:
This was supposed to be a Draft PR, not yet ready for review. Apologies if I misclicked.

samples/Common.c
jdelapla: `pUriCount` is dereferenced and incremented, which updates `pSampleConfiguration->iceUriCount` 's value. We have to wait to MEMFREE(pAsyncData) until it's actually been used in the thread function, or we could end up accessing already freed memory.

samples/Common.c
jdelapla: pUriCount is a pointer to an int outside the scope of this struct.

samples/Common.c
jdelapla: Unless max ice servers is changed

disa6302 (approved):

PR #1861, opened 2023-11-29
GCC CI Build Fixes

stefankiesz:
*Issue #, if available:*

*What was changed?*
Updated the GCC path set in the Mac GCC CI build.

*Why was it changed?*
"gcc" is mapped to CLANG on Mac, so the build was using CLANG rather than GCC compiler.

*How was it changed?*
The full path to the GCC and G++ compiler was identified, and it replaced the "gcc" and "g++" in the respective environment variables.

The CI is to run to verify that the correct compiler is used and that the test is passing.

### DIFF
--- a/src/source/Crypto/Tls_openssl.c
+++ b/src/source/Crypto/Tls_openssl.c
@@ -214,7 +214,9 @@ STATUS tlsSessionPutApplicationData(PTlsSession pTlsSession, PBYTE pData, UINT32
     }
 
     wBioGetMemDataRet = BIO_get_mem_data(SSL_get_wbio(pTlsSession->pSsl), &wBioBuffer);
-    CHK_ERR(wBioGetMemDataRet >= 0, STATUS_SEND_DATA_FAILED, "BIO_get_mem_data failed");
+    CHK_ERR(wBioGetMemDataRet != 0, STATUS_SEND_DATA_FAILED, "BIO_get_mem_data failed with: no BIO was connected to the SSL object.");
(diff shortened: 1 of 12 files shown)


stefankiesz:
Note to self: add comments in code explaining changes before merging.

src/source/Crypto/Tls_openssl.c
```diff
@@ -212,8 +213,11 @@ STATUS tlsSessionPutApplicationData(PTlsSession pTlsSession, PBYTE pData, UINT32
         }
     }
 
-    wBioDataLen = (SIZE_T) BIO_get_mem_data(SSL_get_wbio(pTlsSession->pSsl), &wBioBuffer);
-    CHK_ERR(wBioDataLen >= 0, STATUS_SEND_DATA_FAILED, "BIO_get_mem_data failed");
+    wBioGetMemDataRet = BIO_get_mem_data(SSL_get_wbio(pTlsSession->pSsl), &wBioBuffer);
+    CHK_ERR(wBioGetMemDataRet != 0, STATUS_SEND_DATA_FAILED, "BIO_get_mem_data failed with: no BIO was connected to the SSL object.");
```
stefankiesz, line 217: Note: "0" is an error, was being treated as a non-error prior to these changes

stefankiesz: > BIO_get_mem_data() returns the total number of bytes available on success, 0 if b is NULL, or a negative value in case of other errors.

PR #1869, opened 2023-12-05
fix compiling issue caused by "stringop-overflow" in old gcc

codingspirit:
*Issue #, if available:*
GCC introduced "-Wstringop-overflow" after 7.0. Older GCC like 5.4 will complain "no option -Wstringop-overflow".

See also:
- https://github.com/aws-samples/amazon-kinesis-video-streams-media-interface/issues/113
- https://gcc.gnu.org/gcc-7/changes.html

*What was changed?*
Introducing a GCC version check to add this flag only after GCC 7.0.

*Why was it changed?*

*How was it changed?*


disa6302:
Thank you @codingspirit . We have pulled in your commits into #1872 to introduce a new CI build and verify rest of the CI runs fine.

PR #1872, opened 2023-12-06
codingspirit PR #1869

stefankiesz:
Running CI tests for https://github.com/awslabs/amazon-kinesis-video-streams-webrtc-sdk-c/pull/1869

*Issue #, if available:*
[Copied from original PR]
GCC introduced "-Wstringop-overflow" after 7.0. Older GCC like 5.4 will complain "no option -Wstringop-overflow".

[Copied from original PR]
See also:
- https://github.com/aws-samples/amazon-kinesis-video-streams-media-interface/issues/113
- https://gcc.gnu.org/gcc-7/changes.html

*What was changed?*
- [Copied from original PR]
Introducing a GCC version check to add this flag only after GCC 7.0.

- Added an mbedtls-ubuntu-gcc-4_4-build CI test to test this fix and catch such issues in the future
- Changed the original PR to still set DMBEDTLS_USER_CONFIG_FILE for GCC version less than 7.0

*Why was it changed?*
- These changes were made to solve the above issue
- The mbedtls config file has to be set in order for the build to succeed

*How was it changed?*
- The original contributor made the above changes
- A fix was added after testing with the added CI test

A CI test was added to test the above changes.


CMakeLists.txt
```diff
@@ -115,10 +115,14 @@ elseif(USE_MBEDTLS)
     message(STATUS "Detected clang")
     set(CMAKE_C_FLAGS "-I${CMAKE_CURRENT_SOURCE_DIR}/configs -DMBEDTLS_USER_CONFIG_FILE=\"<config_mbedtls.h>\" ${CMAKE_C_FLAGS}")
     set(CMAKE_CXX_FLAGS "-I${CMAKE_CURRENT_SOURCE_DIR}/configs -DMBEDTLS_USER_CONFIG_FILE=\"<config_mbedtls.h>\" ${CMAKE_CXX_FLAGS}")
-  else()
+  elseif("${CMAKE_C_COMPILER_ID}" MATCHES "GNU")
```
disa6302, line 118: We only exclusively test GCC and Clang. So it should be ok to just use `else()`.

stefankiesz: Added an else clause with the following message:
`message(FATAL_ERROR "Detected unsupported compiler: ${CMAKE_C_COMPILER_ID}")`

disa6302 (approved):

PR #1874, opened 2023-12-07
Add missing NULL check for session token

niyatim23:
*Issue #, if available:*

*What was changed?*
Adding a missing null check before checking anything related to the string

*Why was it changed?*
If that check is missing and the env var is not set, the sample segfaults

*How was it changed?*
It was changed by adding a null check before checking anything related to the string

Tested locally that even if the value is not, the sample does not segfault


hassanctech (approved):
lgtm

PR #1880, opened 2023-12-08
Modify compilation errors,when IOT_CORE_ENABLE_CREDENTIALS is enabled

nianshao163:
*Issue #, if available:*

*What was changed?*

*Why was it changed?*

*How was it changed?*

*What testing was done for the changes?*

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.

PR #1882, opened 2023-12-08
Remove pIotCoreCertificateId usage from samples.

disa6302:
*Issue #, if available:*

Opening for #1880 to let CI run. 

*What was changed?*
Fixing compilation error when `IOT_CORE_ENABLE_CREDENTIALS` enabled.

*Why was it changed?*
To ensure the build runs clean. The issue does not persist in past releases and is only localized to 1.9.0. We do not want to modify the channel name that is being set in the command line option in our samples. If certificate Id is required, users can modify the sample to fit their use case.

*How was it changed?*
It was fixed by removing the usage of `pIotCoreCertificateId`. 

Tested locally and confirmed media shows up on JS test page by generating a new thing following the instructions here:

https://docs.aws.amazon.com/kinesisvideostreams/latest/dg/how-iot.html#


hassanctech (approved):
lgtm

PR #1889, opened 2023-12-13
No more detaching threads in tests

jdelapla:
*What was changed?*
Thread detach in the tests were changed to be joinable, joined, and then the callback removed at the end of the function connectTwoPeers and similar operations throughout the test.

*Why was it changed?*
CI was experiencing TSAN failures.

*How was it changed?*
Explained in "what was changed"

CI was used to confirm the changes fixed the issue

### DIFF
--- a/tst/PeerConnectionFunctionalityTest.cpp
+++ b/tst/PeerConnectionFunctionalityTest.cpp
@@ -60,6 +60,8 @@ TEST_F(PeerConnectionFunctionalityTest, connectTwoPeersWithDelay)
     RtcSessionDescriptionInit sdp;
     SIZE_T connectedCount = 0;
     PRtcPeerConnection offerPc = NULL, answerPc = NULL;
+    PeerContainer offer;
+    PeerContainer answer;
 
     MEMSET(&configuration, 0x00, SIZEOF(RtcConfiguration));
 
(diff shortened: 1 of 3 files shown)


disa6302 (approved):

tst/PeerConnectionFunctionalityTest.cpp
```diff
@@ -714,6 +758,16 @@ TEST_F(PeerConnectionFunctionalityTest, noLostFramesAfterConnected)
         THREAD_SLEEP(HUNDREDS_OF_NANOS_IN_A_MILLISECOND);
     }
 
+    this->lock.lock();
```
sirknightj, line 761: We should utilize try/catch blocks to guarantee an unlock. Thread join can throw an exception.

tst/WebRTCClientTestFixture.h
```diff
@@ -58,6 +58,9 @@ class WebRtcClientTestBase : public ::testing::Test {
     UINT32 mRtpPacketCount;
     UINT32 mUriCount = 0;
     SIGNALING_CLIENT_HANDLE mSignalingClientHandle;
+    std::vector<std::thread> threads;
+    std::mutex lock;
```
sirknightj, line 62: Can we add a comment somewhere explaining what does this lock guard?

tst/PeerConnectionFunctionalityTest.cpp
sirknightj: Might also want to looked into timed mutex for timed lock with timeouts

hassanctech (approved):
lgtm

PR #1897, opened 2024-01-09
CMAKE_BUILD_TYPE=Release build

disa6302:
*Issue #, if available:*

*What was changed?*
- Enforcing default BUILD_TYPE to `Release`.
- Removed SPRINTF usage throughout the SDK to ensure buffer overflow crash safety
- Updated the SDP attribute value to 512 from 256 since one of the attributes that is generated is a combination of streamId and trackId which adds to a max string size of 512 characters, causing buffer overflow situation.
- Revise character array sizes in the SDP struct and remove unused fields. 

*Why was it changed?*
- The default `CMAKE_BUILD_TYPE` if unset is platform dependent. It is important we enforce the default build type to `Release` for this SDK and its dependencies to ensure the library does not include debug symbols, increasing the size of them.

*How was it changed?*
- The CMake file checks if a `CMAKE_BUILD_TYPE` is defined. If not, the SDK builds release mode.
- The total SDP structure size changed from 802 KB to 850 KB because of SDP attribute value size increase.

- Tested by looking at the CMakeCache.txt file to confirm `Release` build binaries and libraries were generated.
- Also tested by explicitly setting `CMAKE_BUILD_TYPE` to `Debug` to confirm debug version of libraries and binaries are built.


src/source/Ice/TurnConnection.c
```diff
@@ -1104,7 +1104,7 @@ STATUS turnConnectionGetLongTermKey(PCHAR username, PCHAR realm, PCHAR password,
     CHK(username[0] != '\0' && realm[0] != '\0' && password[0] != '\0' && bufferLen >= KVS_MD5_DIGEST_LENGTH, STATUS_INVALID_ARG);
     CHK((STRLEN(username) + STRLEN(realm) + STRLEN(password)) <= ARRAY_SIZE(stringBuffer) - 2, STATUS_INVALID_ARG);
 
-    SPRINTF(stringBuffer, "%s:%s:%s", username, realm, password);
+    SNPRINTF(stringBuffer, SIZEOF(stringBuffer), "%s:%s:%s", username, realm, password);
```
hassanctech, line 1107: We should check `SNPRINTF` return value to see if there was truncation.  In that case we need to determine the correct behavior, i.e. is that a fatal error and should we bail out / return an error.

src/source/PeerConnection/SessionDescription.c
```diff
@@ -313,7 +313,7 @@ PCHAR fmtpForPayloadType(UINT64 payloadType, PSessionDescription pSessionDescrip
     CHAR payloadStr[MAX_SDP_ATTRIBUTE_VALUE_LENGTH];
 
     MEMSET(payloadStr, 0x00, MAX_SDP_ATTRIBUTE_VALUE_LENGTH);
-    SPRINTF(payloadStr, "%" PRId64, payloadType);
+    SNPRINTF(payloadStr, SIZEOF(payloadStr), "%" PRId64, payloadType);
```
hassanctech, line 316: Add checks for SNPRINTF return value, for example see usages in `iceCandidateSerialize`.  Most cases if we failed to write the whole thing we return an error / bail out.

hassanctech (approved):
lgtm

PR #1913, opened 2024-02-02
Add SDP component as submodule after re-factoring

ActoryOu:
*Issue #, if available:*

*What was changed?*

- Add [SDP component](https://github.com/awslabs/amazon-kinesis-video-streams-sdp) as external project, and update SDP code to call APIs of component.

*Why was it changed?*

- To apply SDP re-factoring component into dev branch. [SDP component](https://github.com/awslabs/amazon-kinesis-video-streams-sdp) is separated out as independent library.

*How was it changed?*

- Add [SDP component](https://github.com/awslabs/amazon-kinesis-video-streams-sdp) as external project, import it via [ExternalProject_Add](https://github.com/awslabs/amazon-kinesis-video-streams-webrtc-sdk-c/pull/1913/files#diff-393cbfcde8b40b23a46fba42a476f4f7a64a170c34660147bbb6f66f97107500R12), and update SDP code to call APIs of component.

- Verified master demo on Ubuntu
- Verified webrtc_client_test and pass all test cases.


hassanctech:
In this repository we have taken the decision to not use git submodules.  The producer c / pic libraries present in the current `.gitmodules` is outdated.  

The way we build dependencies is inside the root `CMake` folder we have a `Dependencies` folder and inside there we have a `lib${libname}-CMakeLists.txt`  inside here you can see we do a external project add.  Then in the root `CMakeLists.txt` file you'll see a BUILD_DEPENDENCIES section (which is true by default) we call `build_dependency` which a function we've defined inside `CMake/Utilities.cmake` you'll see we have a list of supported libs there so you will need to add the new sdp package there and you'll need to create a file like `libkvsSdp-CMakeLists.txt` or however you want to name it and follow that same pattern.

src/source/Sdp/Serialize.c
```diff
@@ -1,47 +1,68 @@
...
-                                    pSDPOrigin->sessionId, pSDPOrigin->sessionVersion, pSDPOrigin->sdpConnectionInformation.networkType,
-                                    pSDPOrigin->sdpConnectionInformation.addressType, pSDPOrigin->sdpConnectionInformation.connectionAddress);
-
-        CHK(*ppOutputData == NULL || ((*pBufferSize - *pTotalWritten) >= currentWriteSize), STATUS_BUFFER_TOO_SMALL);
-        *pTotalWritten += currentWriteSize;
-        if (*ppOutputData != NULL) {
-            *ppOutputData += currentWriteSize;
-        }
+        origin.pUserName = pSDPOrigin->userName;
+        origin.userNameLength = STRLEN(pSDPOrigin->userName);
```
hassanctech, line 37: Let's always use `STRNLEN` instead of `STRLEN` for safety.

ActoryOu: Thanks for comment!
I replaced almost all STRLEN with STRNLEN. The only one left is `deserializeSessionDescription()` in Deserialize.c to calculate the length of sdpBytes, which we don't know the buffer size in this function.

src/source/Sdp/Deserialize.c
```diff
@@ -1,37 +1,48 @@
...
+#define LOG_CLASS "SDPDeserialize"
 #include "../Include_i.h"
+#include "sdp_deserializer.h"
 
-STATUS parseMediaName(PSessionDescription pSessionDescription, PCHAR pch, UINT32 lineLen)
+STATUS parseMediaName(PSessionDescription pSessionDescription, PCHAR mediaValue, size_t mediaValueLength)
 {
     ENTERS();
     STATUS retStatus = STATUS_SUCCESS;
+    size_t copyLength;
```
hassanctech, line 9: use `SIZE_T` macro instead of `size_t`

ActoryOu: Updated via commit 4711d2d9a8eaa0117ed8f63c7b8b957ac3a1d283.

ActoryOu:
Note that the SDP repo link is not official one for now. Do NOT merge this PR for now. Will update the repo link after [SDP PR#2](https://github.com/awslabs/amazon-kinesis-video-streams-sdp/pull/2) merged.

src/source/Sdp/Deserialize.c
ActoryOu: I got this compile error in [mac-os-build-clang](https://github.com/awslabs/amazon-kinesis-video-streams-webrtc-sdk-c/actions/runs/7927210884/job/21643215938?pr=1913) and [mac-os-build-gcc](https://github.com/awslabs/amazon-kinesis-video-streams-webrtc-sdk-c/actions/runs/7927210884/job/21643215315?pr=1913) if I set size_t to SIZE_T. Looks like they're set to different data structures in PIC repo.

```
2024-02-16T07:31:48.7756110Z /Users/runner/work/amazon-kinesis-video-streams-webrtc-sdk-c/amazon-kinesis-video-streams-webrtc-sdk-c/src/source/Sdp/Deserialize.c:111:82: error: incompatible pointer types passing 'SIZE_T *' (aka 'unsigned long long *') to parameter of type 'size_t *' (aka 'unsigned long *') [-Werror,-Wincompatible-pointer-types]
2024-02-16T07:31:48.7759080Z         sdpResult = SdpDeserializer_GetNext(&ctx, &type, (const CHAR**) &pValue, &valueLength);
2024-02-16T07:31:48.7760320Z                                                                                  ^~~~~~~~~~~~
2024-02-16T07:31:48.7762560Z /Users/runner/work/amazon-kinesis-video-streams-webrtc-sdk-c/amazon-kinesis-video-streams-webrtc-sdk-c/open-source/include/kvssdp/sdp_deserializer.h:22:47: note: passing argument to parameter 'pValueLength' here
2024-02-16T07:31:48.7764560Z                                      size_t * pValueLength );
2024-02-16T07:31:48.7765460Z                                               ^
2024-02-16T07:31:48.7784270Z 1 error generated.
```

ActoryOu: It passes with casting (cast to size_t). Let me know if you have other concern.

ActoryOu: It's still failing on [mac-os-build-clang](https://github.com/awslabs/amazon-kinesis-video-streams-webrtc-sdk-c/actions/runs/7927508659/job/21646981674) when using SIZE_T. Revert this back to size_t for now.

moninom1:
Can you please update the change description and also add little description of serialisation and deserialisation APIs being updated?

src/source/Sdp/Deserialize.c
```diff
@@ -1,37 +1,48 @@
...
+#define LOG_CLASS "SDPDeserialize"
 #include "../Include_i.h"
+#include "kvssdp/sdp_deserializer.h"
 
-STATUS parseMediaName(PSessionDescription pSessionDescription, PCHAR pch, UINT32 lineLen)
+STATUS parseMediaName(PSessionDescription pSessionDescription, PCHAR mediaValue, SIZE_T mediaValueLength)
 {
     ENTERS();
     STATUS retStatus = STATUS_SUCCESS;
+    SIZE_T copyLength;
```
moninom1, line 9: copyLength - Can we make the variable name more meaningful, like mediaLength or something similar.

ActoryOu: Updated

ActoryOu:
> Can you please update the change description?

Updated. Thanks for remind.

src/source/Sdp/Deserialize.c
```diff
@@ -74,74 +96,80 @@ STATUS deserializeSessionDescription(PSessionDescription pSessionDescription, PC
...
 
-        if (lineLen > 0 && curr[lineLen - 1] == '\r') {
-            lineLen--;
-        }
+    for (; sdpResult == SDP_RESULT_OK;) {
+        sdpResult = SdpDeserializer_GetNext(&ctx, &type, (const CHAR**) &pValue, (size_t*) &valueLength);
 
-        if (0 == STRNCMP(curr, SDP_MEDIA_NAME_MARKER, (ARRAY_SIZE(SDP_MEDIA_NAME_MARKER) - 1))) {
-            CHK_STATUS(parseMediaName(pSessionDescription, curr, lineLen));
+        if (sdpResult == SDP_RESULT_OK) {
```
moninom1, line 114: I think this if can be removed, as this is expected behavior and flow will anyway continue.

src/source/Sdp/Deserialize.c
```diff
@@ -74,74 +96,80 @@ STATUS deserializeSessionDescription(PSessionDescription pSessionDescription, PC
...
 
-    while ((next = STRNCHR(curr, tail - curr, '\n')) != NULL) {
-        lineLen = (UINT32) (next - curr);
+    sdpResult = SdpDeserializer_Init(&ctx, sdpBytes, STRLEN(sdpBytes));
+    CHK(sdpResult == SDP_RESULT_OK, sdpResult);
 
-        if (lineLen > 0 && curr[lineLen - 1] == '\r') {
-            lineLen--;
-        }
+    for (; sdpResult == SDP_RESULT_OK;) {
```
moninom1, line 111: Using both SDP_RESULT_OK (sdpResult) and STATUS_SUCCESS (retStatus) can be a bit confusing, should we stick to using only one of them throughout, as i think the value is same.

ActoryOu: Even though they're same value, SDP_RESULT_OK is returned from SDP library and STATUS_SUCCESS is used in whole webrtc sdk. IMHO we shouldn't mix them.

src/source/Sdp/Deserialize.c
ActoryOu: I put it here because we break in all other conditions. Otherwise it would be like below, which is basically same. I prefer to put SDP_RESULT_OK in first condition because it's the highest possible return value from SDP library. Then we can speed this up with this sequence.
```
if (sdpResult == SDP_RESULT_MESSAGE_END) {
    /* Reset return value when done. */
    retStatus = STATUS_SUCCESS;
    break;
} else if (sdpResult != SDP_RESULT_OK) {
    retStatus = sdpResult;
    break;
} else {
    /* Do nothing. */
}
```

src/source/Sdp/Deserialize.c
moninom1: Should we have a mapping to map STUN library errors to webrtc SDK error ? I am fine with keeping it as it as you suggested, but give it a thought if it can be improved.

ActoryOu: Updated

src/include/com/amazonaws/kinesis/video/webrtcclient/Include.h
```diff
@@ -107,22 +107,40 @@ extern "C" {
...
+#define STATUS_SDP_INFORMATION_ERROR                      STATUS_SDP_BASE + 0x00000007
+#define STATUS_SDP_MEDIA_NAME_ERROR                       STATUS_SDP_BASE + 0x00000008
+#define STATUS_SDP_ORIGIN_ERROR                           STATUS_SDP_BASE + 0x00000009
+#define STATUS_SDP_PHONE_NUMBER_ERROR                     STATUS_SDP_BASE + 0x0000000A
+#define STATUS_SDP_TIME_DECRYPTION_ERROR                  STATUS_SDP_BASE + 0x0000000B
+#define STATUS_SDP_TIMEZONE_ERROR                         STATUS_SDP_BASE + 0x0000000C
+#define STATUS_SDP_URI_ERROR                              STATUS_SDP_BASE + 0x0000000D
+#define STATUS_SDP_VERSION_ERROR                          STATUS_SDP_BASE + 0x0000000E
+#define STATUS_SDP_ATTRIBUTE_MAX_EXCEEDED                 STATUS_SDP_BASE + 0x0000000F
+#define STATUS_SDP_UNKNOWN_ERROR                          STATUS_SDP_BASE + 0x00000010
```
moninom1, line 126: Ideally UNKOWN error should be the last one. So these are all the new error we added in our SDP library?

ActoryOu: Yes, the error codes below unknow error are all able to be mapped to SDP library error code.

src/include/com/amazonaws/kinesis/video/webrtcclient/Include.h
```diff
@@ -107,22 +107,40 @@ extern "C" {
...
+#define STATUS_SDP_BAD_PARAM                              STATUS_SDP_BASE + 0x00000011
```
moninom1, line 127: Instead of adding this, shouldn't it be mapped to STATUS_INVALID_ARG

ActoryOu: Thanks for suggestion. Updated for this.

moninom1 (dismissed):

src/source/Sdp/Deserialize.c
```diff
@@ -74,74 +99,76 @@ STATUS deserializeSessionDescription(PSessionDescription pSessionDescription, PC
...
 
-    while ((next = STRNCHR(curr, tail - curr, '\n')) != NULL) {
-        lineLen = (UINT32) (next - curr);
+    sdpResult = SdpDeserializer_Init(&ctx, sdpBytes, STRLEN(sdpBytes));
+    CHK(sdpResult == SDP_RESULT_OK, convertSdpErrorCode(sdpResult));
 
-        if (lineLen > 0 && curr[lineLen - 1] == '\r') {
-            lineLen--;
-        }
+    for (; sdpResult == SDP_RESULT_OK;) {
```
hassanctech, line 114: seems odd to use `for` loop here instead of `while`?

src/source/Sdp/Serialize.c
```diff
@@ -1,160 +1,182 @@
...
 
-CleanUp:
+    retStatus = convertSdpErrorCode(sdpResult);
 
     LEAVES();
     return retStatus;
 }
 
-STATUS serializeOrigin(PSdpOrigin pSDPOrigin, PCHAR* ppOutputData, PUINT32 pTotalWritten, PUINT32 pBufferSize)
+static STATUS serializeOrigin(SdpSerializerContext_t* pCtx, PSdpOrigin pSDPOrigin)
```
hassanctech, line 22: For what reason are we making these functions static?

ActoryOu: These functions are only used inside Serialize.c. It's more secure to keep it in file scope with static keyword IMO. Or does this function open for external use?

src/source/Sdp/Deserialize.c
ActoryOu: You're right! Great catch! Updated to while loop.

src/source/Sdp/Serialize.c
hassanctech: I suppose previously nothing blocked it from being available for external use.  I like the idea of making it static.

moninom1 (approved):

PR #1914, opened 2024-02-05
Reduce hash table size for SDP transceivers

disa6302:
*Issue #, if available:*

*What was changed?*
- Reduce the size of the following hash tables:
   1. `pSeenTransceivers`: This is populated with only codecs supported by the SDK. It was originally set to allow 10000 entries leading to a hash table size of 480 KB. However, the size should be adapted to max codecs supported by SDK since the table will not have duplicate entries. With this PR, the size is set to MAX(Max_codecs_supported, min_hashtable_count), where, min_hashtable_count is 16.
   2. `pUnknownCodecPayloadTypesTable`: This is populated with the first unknown codec encountered in the m section of the SDP. It was originally set to allow 10000 entries leading to a hash table size of 480KB.  However, the maximum entries in the hash table is equal to the maximum m lines the SDK supports which is 5 currently. This hash table is purely used to create a dummy m section entry to match the m lines between offer and answer. If this number exceeds the maximum of 5, the SDK returns a `STATUS_SESSION_DESCRIPTION_MAX_MEDIA_COUNT` error anyways. With this PR, the size is set to MAX(number_of_media_sections, min_hashtable_count), where, min_hashtable_count is 16.
   3. `pUnknownCodecRtpmapTable`: This is populated with the first unknown rtpmap encountered in the m section of the SDP. It was originally set to allow 10000 entries leading to a hash table size of 480KB.  However, the maximum entries in the hash table is equal to the maximum m lines the SDK supports which is 5 currently. This hash table is purely used to create a dummy m section entry to match the m lines between offer and answer. If this number exceeds the maximum of 5, the SDK returns a `STATUS_SESSION_DESCRIPTION_MAX_MEDIA_COUNT` error anyways. With this PR, the size is set to MAX(number_of_media_sections, min_hashtable_count), where, min_hashtable_count is 16.

*Why was it changed?*
- Having 3 hash tables with sizes 480KB each temporarily increases the memory consumption by 1.4MB. By making the sizes reasonable, we avoid this sudden unnecessary peak.
- Post this PR, the allocation of each hash table would be 784 bytes and the total allocation would drop from 1.4MB to 2.3KB. 

Before the change:
Screenshot 2024-02-06 at 10 58 11 AM

After the change:
Screenshot 2024-02-06 at 11 50 51 AM

*How was it changed?*
- Added a new enum RTC_CODEC_MAX, leaving it unassigned to track max codecs supported by SDK
- Switch to `hashTableCreateWithParams` from `hashTableCreate`
- Dynamic hash table size decision

- Added a unit test to confirm fake transceivers are created
- Also tested that `STATUS_SESSION_DESCRIPTION_MAX_MEDIA_COUNT` is encountered to prove the answer sdp is not generated.

### DIFF
--- a/src/source/PeerConnection/SessionDescription.c
+++ b/src/source/PeerConnection/SessionDescription.c
@@ -910,8 +910,8 @@ STATUS populateSessionDescriptionMedia(PKvsPeerConnection pKvsPeerConnection, PS
         }
     } else {
         pDtlsRole = DTLS_ROLE_ACTIVE;
-        CHK_STATUS(hashTableCreate(&pUnknownCodecPayloadTypesTable));
-        CHK_STATUS(hashTableCreate(&pUnknownCodecRtpmapTable));
+        CHK_STATUS(hashTableCreateWithParams(CODEC_RTPMAP_PAYLOAD_TYPES_HASH_TABLE_BUCKET_COUNT, CODEC_RTPMAP_PAYLOAD_TYPES_HASH_TABLE_BUCKET_LENGTH, &pUnknownCodecPayloadTypesTable));
+        CHK_STATUS(hashTableCreateWithParams(CODEC_RTPMAP_PAYLOAD_TYPES_HASH_TABLE_BUCKET_COUNT, CODEC_RTPMAP_PAYLOAD_TYPES_HASH_TABLE_BUCKET_LENGTH, &pUnknownCodecRtpmapTable));
 
         // this function creates a list of transceivers corresponding to each m-line and adds it answerTransceivers
         // if an m-line does not have a corresponding transceiver created by the user, we create a fake transceiver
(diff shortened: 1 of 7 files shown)


hassanctech:
Excellent work @disa6302 

Would it be possible to include a graph showing the memory spike (pre-change) and then post-change no spike?

disa6302:
>Would it be possible to include a graph showing the memory spike (pre-change) and then post-change no spike?

@hassanctech , looking at the instruments tool, it seems to be difficult to capture the spike in the graph. However, I have added the screenshots of the allocation value change. Let me know if this does not work.

src/source/PeerConnection/SessionDescription.c
```diff
@@ -886,6 +886,7 @@ STATUS populateSessionDescriptionMedia(PKvsPeerConnection pKvsPeerConnection, PS
     PCHAR pDtlsRole = NULL;
     PHashTable pUnknownCodecPayloadTypesTable = NULL, pUnknownCodecRtpmapTable = NULL;
     UINT32 unknownCodecHashTableKey = 0;
+    UINT32 unknownHashTableBucketCount;
```
jdelapla, line 889: Initialize to 0.

jdelapla (commented):
1 small nit, however I was also curious if there is a place where items are inserted into the table that should now have an error check for when the table is at max? Since we're making the table much smaller I feel that is a prudent step we should take. Does it already exist?

jdelapla (approved):

hassanctech (approved):
lgtm

PR #1915, opened 2024-02-06
Ensure remote description does not live in memory for viewer duration

disa6302:
*Issue #, if available:*

*What was changed?*
- Fixed the code where the SDK allocates remoteDescription member along with peer connection object, causing it to live in memory for the duration of the peer connection

*Why was it changed?*
- The SDP object takes up 800KB of memory to store a remote/local description. This leads to unnecessary memory usage, when the remote description is not required to remain in memory once the answer is generated. In the future, say the SDK resends the answer suspecting the viewer did not receive it, the SDK would not regenerate the answer, but would resend the same answer. A reconstruction would not be required and is not expected. 

*How was it changed?*
- Switched the peer connection member from : `SessionDescription remoteSessionDescription;` to `PSessionDescription pRemoteSessionDescription;` to allocate remote description separately. 
- Ensured the internal logic was fixed to perform certain checks only if the SDK is acting as a viewer.

Doing so ensures the ~800 KB of memory lives in memory for a short time.

Following are the graphs showing the persistent memory allocation in the viewer run duration.

Before this change: As seen, 832KiB persists till viewer disconnects
Screenshot 2024-02-08 at 12 00 04 PM

After this change: Only 1.5KiB of memory resides till viewer disconnects
Screenshot 2024-02-08 at 12 05 13 PM

- Tested master sample with viewer JS to confirm successful peer connection
- Tested viewer sample with master JS to confirm successful peer connection
- CI run

### DIFF
--- a/src/source/PeerConnection/SessionDescription.c
+++ b/src/source/PeerConnection/SessionDescription.c
@@ -413,9 +413,10 @@ STATUS populateSingleMediaSection(PKvsPeerConnection pKvsPeerConnection, PKvsRtp
         CHK_STATUS(hashTableGet(pUnknownCodecPayloadTypesTable, unknownCodecHashTableKey, &payloadType));
     } else {
         CHK_STATUS(hashTableGet(pKvsPeerConnection->pCodecTable, pRtcMediaStreamTrack->codec, &payloadType));
-        currentFmtp = fmtpForPayloadType(payloadType, &(pKvsPeerConnection->remoteSessionDescription));
+        if (!pKvsPeerConnection->isOffer) {
+            currentFmtp = fmtpForPayloadType(payloadType, pRemoteSessionDescription);
+        }
     }
-
     if (pRtcMediaStreamTrack->kind == MEDIA_STREAM_TRACK_KIND_VIDEO) {
         if (pRtcMediaStreamTrack->codec == RTC_CODEC_H264_PROFILE_42E01F_LEVEL_ASYMMETRY_ALLOWED_PACKETIZATION_MODE) {
             retStatus = hashTableGet(pKvsPeerConnection->pRtxTable, RTC_RTX_CODEC_H264_PROFILE_42E01F_LEVEL_ASYMMETRY_ALLOWED_PACKETIZATION_MODE,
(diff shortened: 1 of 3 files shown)


hassanctech (commented):
Since the improvement here is intended to be long lived not just a spike, can we include a graph with the overall RAM utilization before and after this change.  Here I would expect to see a sudden drop with the change compared to before because previously we did not free for the entire duration of the viewer.

disa6302:
> Since the improvement here is intended to be long lived not just a spike, can we include a graph with the overall RAM utilization before and after this change. Here I would expect to see a sudden drop with the change compared to before because previously we did not free for the entire duration of the viewer.

Updated the description. While the peak is not visible (covered by the many other allocations that happen around the same time), the graphs indicates how much of the peer connection object actually lives till viewer disconnection (832KiB vs 1.5KiB)

stefankiesz (approved):

jdelapla (approved):

hassanctech (approved):
lgtm

src/source/PeerConnection/SessionDescription.c
```diff
@@ -555,11 +556,14 @@ STATUS populateSingleMediaSection(PKvsPeerConnection pKvsPeerConnection, PKvsRtp
...
-    // check all session attribute lines to see if a line with mid is present. If it is present, copy its content and break
-    for (i = 0; i < pRemoteSessionDescription->mediaDescriptions[mediaSectionId].mediaAttributesCount; i++) {
-        if (STRCMP(pRemoteSessionDescription->mediaDescriptions[mediaSectionId].sdpAttributes[i].attributeName, MID_KEY) == 0) {
-            STRCPY(remoteSdpAttributeValue, pRemoteSessionDescription->mediaDescriptions[mediaSectionId].sdpAttributes[i].attributeValue);
-            break;
+
+    if (!pKvsPeerConnection->isOffer) {
+        // check all session attribute lines to see if a line with mid is present. If it is present, copy its content and break
+        for (i = 0; i < pRemoteSessionDescription->mediaDescriptions[mediaSectionId].mediaAttributesCount; i++) {
+            if (STRCMP(pRemoteSessionDescription->mediaDescriptions[mediaSectionId].sdpAttributes[i].attributeName, MID_KEY) == 0) {
```
sirknightj, line 563: Probably want to use `STRNCMP` for safety

sirknightj: And STRNCPY below

PR #1921, opened 2024-02-08
Rb size config

disa6302:
*Issue #, if available:*

*What was changed?*
- Reduced default size of rolling buffer to about half of original to reduce default memory allocation
- Added new members to allow configuring rolling buffer size per transceiver
- Updated readme about usage details

*Why was it changed?*
- Our SDK was allocating rolling buffer worth 5 seconds of media at bitrate of 10Mbps which increases the rolling buffer size which is very large. Ideally, the bitrate depends on the media track and the codec being used. Audio tracks typically have lower bitrates than video and within audio/video, different codecs have different typical bitrates. Given this, the values should be configurable. By controlling the capacity, the SDK essentially controls the number of RTP packets the buffer holds references to.

*How was it changed?*
- Added new fields `rollingBufferDurationSec` and `rollingBufferBitratebps` in `RtcRtpTransceiverInit` struct. Applications can set these values up while setting up the tracks. An example usage pattern is provided in the README and in our samples with values set to default.
- Lowered the default rolling buffer duration to 1 seconds and bitrate to 5Mbps. 
- If the values are not provided or is less than minimum allowed, the SDK sets the defaults
- While calculating the minimum values, the consideration was to calculate a size to store 1 RTP packet. Given this, a 100 ms + 100 kbps combination leads to allowing storing reference to only 1 RTP packet.
- Post this change, using the master sample with JS viewer, the default values based peak heap memory consumption was observed to be reduced by 1.7MB.

Before the change -- peak heap memory consumption = 6.68MB
Screenshot 2024-02-06 at 4 20 42 PM

After the change: (with default 3 seconds, 5 Mbps for video, 3 seconds, 1000Kbps for audio) -- peak heap memory consumption = 4.98MB
Screenshot 2024-02-09 at 12 59 49 PM

With sample settings (H.264 : 3 seconds, 4 Mbps and Opus: 3 seconds, 510 Kbps) -- peak heap memory consumption = 4.82MB
Screenshot 2024-02-09 at 11 17 13 AM

- Added a specific unit test to test different input combinations
- Ran the sample with different values to confirm it works fine.

### DIFF
--- a/samples/Common.c
+++ b/samples/Common.c
@@ -407,6 +407,8 @@ STATUS initializePeerConnection(PSampleConfiguration pSampleConfiguration, PRtcP
     SNPRINTF(configuration.iceServers[0].urls, MAX_ICE_CONFIG_URI_LEN, KINESIS_VIDEO_STUN_URL, pSampleConfiguration->channelInfo.pRegion,
              pKinesisVideoStunUrlPostFix);
 
+    configuration.rollingBufferDurationInSec = 0.0f;
+    configuration.rollingBufferBitrateInMBps = 0.0f;
     // Check if we have any pregenerated certs and use them
     // NOTE: We are running under the config lock
     retStatus = stackQueueDequeue(pSampleConfiguration->pregeneratedCertificates, &data);
--- a/samples/Common.c
+++ b/samples/Common.c
@@ -552,8 +552,6 @@ STATUS createSampleStreamingSession(PSampleConfiguration pSampleConfiguration, P
     videoTrack.kind = MEDIA_STREAM_TRACK_KIND_VIDEO;
     videoTrack.codec = RTC_CODEC_H264_PROFILE_42E01F_LEVEL_ASYMMETRY_ALLOWED_PACKETIZATION_MODE;
     videoRtpTransceiverInit.direction = RTC_RTP_TRANSCEIVER_DIRECTION_SENDRECV;
-    videoRtpTransceiverInit.rollingBufferDurationSec = 1;
-    videoRtpTransceiverInit.rollingBufferBitrateInMbps = 1;
     STRCPY(videoTrack.streamId, "myKvsVideoStream");
     STRCPY(videoTrack.trackId, "myVideoTrack");
     CHK_STATUS(addTransceiver(pSampleStreamingSession->pPeerConnection, &videoTrack, &videoRtpTransceiverInit,
@@ -566,8 +564,6 @@ STATUS createSampleStreamingSession(PSampleConfiguration pSampleConfiguration, P
     audioTrack.kind = MEDIA_STREAM_TRACK_KIND_AUDIO;
     audioTrack.codec = RTC_CODEC_OPUS;
     audioRtpTransceiverInit.direction = RTC_RTP_TRANSCEIVER_DIRECTION_SENDRECV;
-    audioRtpTransceiverInit.rollingBufferDurationSec = 1;
-    audioRtpTransceiverInit.rollingBufferBitrateInMbps = 1;
     STRCPY(audioTrack.streamId, "myKvsVideoStream");
     STRCPY(audioTrack.trackId, "myAudioTrack");
     CHK_STATUS(addTransceiver(pSampleStreamingSession->pPeerConnection, &audioTrack, &audioRtpTransceiverInit,
--- a/samples/Common.c
+++ b/samples/Common.c
@@ -552,8 +552,10 @@ STATUS createSampleStreamingSession(PSampleConfiguration pSampleConfiguration, P
     videoTrack.kind = MEDIA_STREAM_TRACK_KIND_VIDEO;
     videoTrack.codec = RTC_CODEC_H264_PROFILE_42E01F_LEVEL_ASYMMETRY_ALLOWED_PACKETIZATION_MODE;
     videoRtpTransceiverInit.direction = RTC_RTP_TRANSCEIVER_DIRECTION_SENDRECV;
-    videoRtpTransceiverInit.rollingBufferDurationSec = 1;
-    videoRtpTransceiverInit.rollingBufferBitratebps = 5 * 1024 * 1024;
+    videoRtpTransceiverInit.rollingBufferDurationSec = 3;
+    // Considering 4 Mbps for 720p (which is what our samples use). This is for H.264.
+    // The value could be different for other codecs.
+    videoRtpTransceiverInit.rollingBufferBitratebps = 4 * 1024 * 1024;
     STRCPY(videoTrack.streamId, "myKvsVideoStream");
     STRCPY(videoTrack.trackId, "myVideoTrack");
     CHK_STATUS(addTransceiver(pSampleStreamingSession->pPeerConnection, &videoTrack, &videoRtpTransceiverInit,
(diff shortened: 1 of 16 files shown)


samples/Common.c
```diff
@@ -564,6 +566,8 @@ STATUS createSampleStreamingSession(PSampleConfiguration pSampleConfiguration, P
     audioTrack.kind = MEDIA_STREAM_TRACK_KIND_AUDIO;
     audioTrack.codec = RTC_CODEC_OPUS;
     audioRtpTransceiverInit.direction = RTC_RTP_TRANSCEIVER_DIRECTION_SENDRECV;
+    audioRtpTransceiverInit.rollingBufferDurationSec = 1;
+    audioRtpTransceiverInit.rollingBufferBitratebps = 5 * 1024 * 1024;
```
hassanctech, line 570: For audio transceiver from my research even for professional studio quality recording we will not exceed 1.5Mbps.  I think we should use that as a default in the sample.

hassanctech:
Very nice !!!

samples/Common.c
disa6302: I have switched the default to be based on the codec. For opus, based on the official website: https://opus-codec.org/#:~:text=It%20can%20scale%20from%20low,)%20to%2048%20kHz%20(fullband) --> it seems the range is 6-51kbps

For H.264, given we use 720p in our GStreamer sample and based on multiple resources I looked at, the range seems to be 1.5-4 Mbps. So set the value to 4Mbps.

I updated the graphs in the description to show memory consumption with these settings too.

I also increased default buffer duration to 3 seconds to be safe.

src/include/com/amazonaws/kinesis/video/webrtcclient/Include.h
```diff
@@ -1489,6 +1488,9 @@ typedef struct {
  */
 typedef struct {
     RTC_RTP_TRANSCEIVER_DIRECTION direction; //!< Transceiver direction - SENDONLY, RECVONLY, SENDRECV
+    DOUBLE rollingBufferDurationSec;         //!< Duration of media that needs to be buffered (in seconds). The lowest allowed is 0.1 seconds (100ms)
+    DOUBLE rollingBufferBitratebps; //!< Expected bitrate of media (In bits/second). It is used to determine the buffer capacity. The lowest allowed
```
stefankiesz, line 1492: Would be nice for this comment to specify that it's the "maximum" expected bitrate as is mentioned in the ReadMe section.

src/source/PeerConnection/PeerConnection.c
```diff
@@ -1472,20 +1473,39 @@ STATUS addTransceiver(PRtcPeerConnection pPeerConnection, PRtcMediaStreamTrack p
...
     UINT32 ssrc = (UINT32) RAND(), rtxSsrc = (UINT32) RAND();
     RTC_RTP_TRANSCEIVER_DIRECTION direction = RTC_RTP_TRANSCEIVER_DIRECTION_SENDRECV;
+    DOUBLE rollingBufferDurationSec = DEFAULT_ROLLING_BUFFER_DURATION_IN_SECONDS;
+    DOUBLE rollingBufferBitratebps = DEFAULT_EXPECTED_VIDEO_BIT_RATE;
     RtcMediaStreamTrack videoTrack;
+
+    CHK(pKvsPeerConnection != NULL, STATUS_NULL_ARG);
+
+    // Set defaults for audio and video track. This can be modified to values set up with pRtcRtpTransceiverInit
+    // if so.
```
stefankiesz, line 1483: Not fully understanding this comment's wording. Is it the values of the defaults that can be modified or the variables that they are assigned to? And what is the "if so" referring to?

src/source/PeerConnection/SessionDescription.c
```diff
@@ -295,8 +295,10 @@ STATUS setTransceiverPayloadTypes(PHashTable codecTable, PHashTable rtxTable, PD
             }
         }
 
-        CHK_STATUS(createRtpRollingBuffer(DEFAULT_ROLLING_BUFFER_DURATION_IN_SECONDS * HIGHEST_EXPECTED_BIT_RATE / 8 / DEFAULT_MTU_SIZE,
-                                          &pKvsRtpTransceiver->sender.packetBuffer));
+        UINT64 size = (UINT64) (pKvsRtpTransceiver->rollingBufferDurationSec * pKvsRtpTransceiver->rollingBufferBitratebps / 8 / DEFAULT_MTU_SIZE);
```
stefankiesz, line 298: (nit) can this variable have a more descriptive name like bufferSize or bufferedPackets - although this is already explained in the log one line below, so not too important.

My justification for still renaming is for when searching for a particular variable in all files when debugging something, can be helpful to have names be more unique to have less generic names like "size" to sift through.

stefankiesz (approved):
Nice! Approved as all of my comments are about non-functional, optional changes.

src/source/PeerConnection/PeerConnection.c
disa6302: Oops. I should have removed that line.

stefankiesz (dismissed):

stefankiesz (approved):

src/source/PeerConnection/Rtp.c
```diff
@@ -31,6 +31,28 @@ STATUS createKvsRtpTransceiver(RTC_RTP_TRANSCEIVER_DIRECTION direction, PKvsPeer
...
+    }
+    if (rollingBufferBitratebps < MIN_EXPECTED_BIT_RATE) {
+        if (pRtcMediaStreamTrack->kind == MEDIA_STREAM_TRACK_KIND_VIDEO) {
+            DLOGW("Rolling buffer duration set to less than 100 KBps for video. Setting to default %d Bps", DEFAULT_EXPECTED_VIDEO_BIT_RATE);
+            pKvsRtpTransceiver->rollingBufferBitratebps = DEFAULT_EXPECTED_VIDEO_BIT_RATE;
+        } else if (pRtcMediaStreamTrack->kind == MEDIA_STREAM_TRACK_KIND_AUDIO) {
+            DLOGW("Rolling buffer duration set to less than 100 KBps for audio. Setting to default %d Bps", DEFAULT_EXPECTED_AUDIO_BIT_RATE);
+            pKvsRtpTransceiver->rollingBufferBitratebps = DEFAULT_EXPECTED_AUDIO_BIT_RATE;
+        } else {
+            DLOGW("Rolling buffer duration set to less than 100 KBps for unknown codec. Setting to default %d Bps", DEFAULT_EXPECTED_AUDIO_BIT_RATE);
```
sirknightj, line 48: Careful with caps. Capital B usually represents bytes.

Also minor: Ki as prefix since increments of 1024 instead of 1000

src/source/PeerConnection/SessionDescription.c
```diff
@@ -295,8 +295,11 @@ STATUS setTransceiverPayloadTypes(PHashTable codecTable, PHashTable rtxTable, PD
             }
         }
 
-        CHK_STATUS(createRtpRollingBuffer(DEFAULT_ROLLING_BUFFER_DURATION_IN_SECONDS * HIGHEST_EXPECTED_BIT_RATE / 8 / DEFAULT_MTU_SIZE,
-                                          &pKvsRtpTransceiver->sender.packetBuffer));
+        UINT64 rollingBufferCapacity =
+            (UINT64) (pKvsRtpTransceiver->rollingBufferDurationSec * pKvsRtpTransceiver->rollingBufferBitratebps / 8 / DEFAULT_MTU_SIZE);
```
sirknightj, line 299: May want to extract 8 out into PIC or something to say we're converting from bits to bytes. Also recommend adding BYTES to the DEFAULT_MTU_SIZE variable name.

stefankiesz (approved):

hassanctech (approved):
nice work @disa6302 !

PR #1934, opened 2024-02-23
Refactor TWCC

disa6302:
*Issue #, if available:*

*What was changed?*
- Refactored TWCC implementation to reduce memory consumption. 

*Why was it changed?*
- The current TWCC implementation consumes 1.5MB of memory and is enabled by default. While this provides a good cache locality, the SDK need not store 65536 RTP packet information in memory since TWCC reports come in periodically (every few milliseconds based on tests). Even if TWCC reports are not received, the RTP packet information should be removed from memory on a rolling time window basis. 

Before this PR, the max heap allocation observed is 4.05 MB

Screenshot 2024-02-23 at 12 00 43 PM

After this PR,  the max heap allocation observed is 2.53MB

Screenshot 2024-02-23 at 12 01 00 PM

As seen, the implementation change leads to almost getting rid of all the allocation (1.5MB) in the best case scenario.

*How was it changed?*
- Switched to using a hash map of TWCC info packets from an array for 65536 TWCC packet. The changes also make sure the TWCC info packets are deleted once the packet is confirmed as received as per TWCC reports. Current implementation did not need that
- The rolling window in the earlier implementation used a stack queue. However, since the TWCC info packets always lived in memory, this was not necessary. The current implementation uses a variable to track last checked packet in the rolling window.

- Ran the sample by disabling rolling window related changes to confirm deletion of TWCC info packets as and when TWCC packets are received.
- Ran the sample by disabling the deletion post receive changes to confirm rolling window approach works as expected.
- Modified the unit test to accommodate for the new implementation.


src/source/PeerConnection/PeerConnection.c
```diff
@@ -992,6 +994,7 @@ STATUS freePeerConnection(PRtcPeerConnection* ppPeerConnection)
     PDoubleListNode pCurNode = NULL;
     UINT64 item = 0;
     UINT64 startTime;
+    UINT32 count;
```
stefankiesz, line 997: (nit) Can we make count variable name more descriptive

src/source/PeerConnection/PeerConnection.c
```diff
@@ -1770,47 +1776,86 @@ STATUS deinitKvsWebRtc(VOID)
...
+{
+    ENTERS();
+    STATUS retStatus = STATUS_SUCCESS;
+    UINT16 updatedSeqNum;
+    PTwccRtpPacketInfo tempTwccRtpPktInfo = NULL;
+    UINT64 ageOfOldest, firstRtpTime;
+    UINT64 value;
+    BOOL isCheckComplete = FALSE;
+
+    CHK(pKvsPeerConnection != NULL && pRtpPacket != NULL, STATUS_NULL_ARG);
```
stefankiesz, line 1790: Might also need to check for null `pKvsPeerConnection->pTwccManager` here

jdelapla:
> After this PR, the max heap allocation observed is 2.53MB

Is this with TWCC enabled?

disa6302:
> > After this PR, the max heap allocation observed is 2.53MB
> 
> Is this with TWCC enabled?

Yes. It is with TWCC enabled. Not disabling TWCC by default in the PR.

src/source/PeerConnection/PeerConnection.c
jdelapla: initialize

README.md
```diff
@@ -348,6 +348,22 @@ createLwsIotCredentialProvider(
...
 
+## Use of TWCC
+In order to listen in on TWCC reports, the application must set up a callback using the `peerConnectionOnSenderBandwidthEstimation` API. In our samples, it is set up like this:
+
+```
+    CHK_STATUS(peerConnectionOnSenderBandwidthEstimation(pSampleStreamingSession->pPeerConnection, (UINT64) pSampleStreamingSession,
+                                                         sampleSenderBandwidthEstimationHandler));
+```
+
+Note that TWCC is enabled by default in the SDK. In order to disable it, set the `disableSenderSideBandwidthEstimation` flag to TRUE. For example,
```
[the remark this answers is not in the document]
disa6302: Agreed. Disabled by default.

src/source/PeerConnection/PeerConnection.c
```diff
@@ -1065,12 +1069,19 @@ STATUS freePeerConnection(PRtcPeerConnection* ppPeerConnection)
     }
 
     if (pKvsPeerConnection->pTwccManager != NULL) {
+        MUTEX_LOCK(pKvsPeerConnection->twccLock);
+        twccLocked = TRUE;
+        hashTableGetCount(pKvsPeerConnection->pTwccManager->pTwccRtpPktInfosHashTable, &twccHashTableCount);
+        DLOGI("Number of TWCC info packets in memory: %d", twccHashTableCount);
+        hashTableIterateEntries(pKvsPeerConnection->pTwccManager->pTwccRtpPktInfosHashTable, 0, freeHashEntry);
```
hassanctech, line 1076: `hashTableGetCount` and `hashTableIterateEntries` return a `STATUS` so we should check for success.

Once you add that, then you will need to make sure to add an `if locked` section in the CleanUp as @jdelapla suggested.

disa6302: Done. Already did the if(locked) check before freeing the twcc mutex.

README.md
```diff
@@ -348,6 +348,22 @@ createLwsIotCredentialProvider(
 freeIotCredentialProvider(&pSampleConfiguration->pCredentialProvider);
 ```
 
+## Use of TWCC
+In order to listen in on TWCC reports, the application must set up a callback using the `peerConnectionOnSenderBandwidthEstimation` API. In our samples, it is set up like this:
+
+```
```
sirknightj, line 354: Let's add the language specifier "```c" to add syntax highlighting.

jdelapla: How does that change the formatting of this README section?

jdelapla (approved):

README.md
sirknightj: Before:

image


After:

image

Also, if we can remove the indent on the left as well, less text will flow off-screen.

disa6302: Given it is a C repo, I am not sure why syntax highlighting makes a difference. Can you clarify what we are achieving with this change?

hassanctech: Yeah I agree the syntax highlighting looks much nicer and we should remove the wasted white space.  We are doing this in most places I think it was just missed in this section.

README.md
```diff
@@ -354,6 +354,22 @@ createLwsIotCredentialProvider(
...
+    CHK_STATUS(peerConnectionOnSenderBandwidthEstimation(pSampleStreamingSession->pPeerConnection, (UINT64) pSampleStreamingSession,
+                                                         sampleSenderBandwidthEstimationHandler));
+```
+
+Note that TWCC is disabled by default in the SDK samples. In order to enable it, set the `disableSenderSideBandwidthEstimation` flag to FALSE. For example,
+
+```
+RtcConfiguration configuration;
+configuration.kvsRtcConfiguration.disableSenderSideBandwidthEstimation = FALSE;
+```
```
hassanctech, line 370: Add syntax highlighting here as well.

hassanctech (dismissed):
lgtm

src/source/PeerConnection/PeerConnection.c
```diff
@@ -1065,12 +1068,15 @@ STATUS freePeerConnection(PRtcPeerConnection* ppPeerConnection)
     }
 
     if (pKvsPeerConnection->pTwccManager != NULL) {
+        MUTEX_LOCK(pKvsPeerConnection->twccLock);
```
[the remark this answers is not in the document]
hassanctech: The unlock in cleanup is still missing.

src/source/PeerConnection/Rtcp.c
```diff
@@ -277,44 +311,75 @@ STATUS parseRtcpTwccPacket(PRtcpPacket pRtcpPacket, PTwccManager pTwccManager)
 STATUS onRtcpTwccPacket(PRtcpPacket pRtcpPacket, PKvsPeerConnection pKvsPeerConnection)
 {
     STATUS retStatus = STATUS_SUCCESS;
-    PTwccManager twcc;
+    PTwccManager pTwccManager = NULL;
     BOOL locked = FALSE;
-    BOOL empty = TRUE;
     UINT64 sn = 0;
```
hassanctech, line 316: `sn` can be `UINT16` since we're setting it to `prevReportedBaseSeqNum` which is already `UINT16`

Also can we change the name of `sn` to be a bit more descriptive, something like `baseSeqNum` should be fine and that is consistent with what it is called in other functions.

tst/RtcpFunctionalityTest.cpp
```diff
@@ -295,37 +295,75 @@ TEST_F(RtcpFunctionalityTest, onpli)
...
+    UNUSED_PARAM(rxBytes);
+    UNUSED_PARAM(txPacketsCnt);
+    UNUSED_PARAM(rxPacketsCnt);
+    UNUSED_PARAM(duration);
+    return;
+}
+
 static void parseTwcc(const std::string& hex, const uint32_t expectedReceived, const uint32_t expectedNotReceived)
 {
+    PRtcPeerConnection pRtcPeerConnection = nullptr;
```
hassanctech, line 311: nullptr is C++, let's use NULL as is standard in our code base.

tst/RtcpFunctionalityTest.cpp
```diff
@@ -295,37 +295,75 @@ TEST_F(RtcpFunctionalityTest, onpli)
...
+    PKvsPeerConnection pKvsPeerConnection;
```
hassanctech, line 312: Might as well init this to NULL as well

tst/RtcpFunctionalityTest.cpp
```diff
@@ -295,37 +295,75 @@ TEST_F(RtcpFunctionalityTest, onpli)
...
+
+    EXPECT_EQ(STATUS_SUCCESS, createPeerConnection(&config, &pRtcPeerConnection));
+    pKvsPeerConnection = reinterpret_cast<PKvsPeerConnection>(pRtcPeerConnection);
+    EXPECT_EQ(STATUS_SUCCESS, peerConnectionOnSenderBandwidthEstimation(pRtcPeerConnection, 0,
+                                                                        testBwHandler));
+
+    UINT16 baseSeqNum = getUnalignedInt16BigEndian(rtcpPacket.payload + 8);
+    UINT16 pktCount = TWCC_PACKET_STATUS_COUNT(rtcpPacket.payload);
+
+    for(auto i = baseSeqNum; i < baseSeqNum + pktCount; i++) {
```
hassanctech, line 336: Let's not use auto, the standard across this code base is to declare variables at the top of the function (to be C89 compatible) explicitly.  Should be `UINT16` in this case.

tst/RtcpFunctionalityTest.cpp
hassanctech: nvm just noticed this is in the tests which are C++, ignore

tst/RtcpFunctionalityTest.cpp
hassanctech: nvm just noticed this is in the tests which are C++, ignore.

tst/RtcpFunctionalityTest.cpp
hassanctech: Just noticed this is in the tests, which are C++, ignore.

src/source/PeerConnection/Rtcp.c
disa6302: Done

src/source/PeerConnection/PeerConnection.c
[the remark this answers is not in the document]
disa6302: I think I lost it in rebase. Added back the change.

hassanctech (dismissed):
nice!

hassanctech (approved):
lgtm

jdelapla (approved):

PR #1935, opened 2024-02-26
Fix signaling cache and GStreamer sample build on Windows

niyatim23:
*Issue #, if available:*

*What was changed?*
Modified the Windows CI and deserialization logic in Signaling caching

*Why was it changed?*
Signaling cache was broken on Windows. This change would fix it and some tests that were disabled on Windows

*How was it changed?*
By modifying the readFile in the PIC layer and fixing the deserialize function on Windows by introducing an additional check

The tests pass in the CI. Ran the sample application to make sure the cache is being used instead of the describe call

### DIFF
--- a/src/source/Signaling/FileCache.c
+++ b/src/source/Signaling/FileCache.c
@@ -38,7 +38,8 @@ STATUS deserializeSignalingCacheEntries(PCHAR cachedFileContent, UINT64 fileSize
     pCurrent = cachedFileContent;
     remainingSize = (UINT32) fileSize;
     /* detect end of file */
-    while (STRNLEN(pCurrent, MAX_SERIALIZED_SIGNALING_CACHE_ENTRY_LEN * 32) > 0 && remainingSize > MAX_SIGNALING_CACHE_ENTRY_TIMESTAMP_STR_LEN) {
+    while (STRNLEN(pCurrent, MAX_SERIALIZED_SIGNALING_CACHE_ENTRY_LEN * MAX_SIGNALING_CACHE_ENTRY_COUNT) > 0 &&
+           remainingSize > MAX_SIGNALING_CACHE_ENTRY_TIMESTAMP_STR_LEN) {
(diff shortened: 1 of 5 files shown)


src/source/Signaling/FileCache.c
```diff
@@ -38,7 +38,7 @@ STATUS deserializeSignalingCacheEntries(PCHAR cachedFileContent, UINT64 fileSize
     pCurrent = cachedFileContent;
     remainingSize = (UINT32) fileSize;
     /* detect end of file */
-    while (remainingSize > MAX_SIGNALING_CACHE_ENTRY_TIMESTAMP_STR_LEN) {
+    while (STRLEN(pCurrent) > 0 && remainingSize > MAX_SIGNALING_CACHE_ENTRY_TIMESTAMP_STR_LEN) {
```
hassanctech, line 41: Can we use `STRNLEN` instead of `STRLEN` since the latter is unsafe.

niyatim23: Done

src/source/Signaling/FileCache.c
```diff
@@ -38,7 +38,7 @@ STATUS deserializeSignalingCacheEntries(PCHAR cachedFileContent, UINT64 fileSize
     pCurrent = cachedFileContent;
     remainingSize = (UINT32) fileSize;
     /* detect end of file */
-    while (remainingSize > MAX_SIGNALING_CACHE_ENTRY_TIMESTAMP_STR_LEN) {
+    while (STRNLEN(pCurrent, MAX_SERIALIZED_SIGNALING_CACHE_ENTRY_LEN * 32) > 0 && remainingSize > MAX_SIGNALING_CACHE_ENTRY_TIMESTAMP_STR_LEN) {
```
hassanctech, line 41: Instead of `32` use `MAX_SIGNALING_CACHE_ENTRY_COUNT` because it could change.

hassanctech (approved):
LGTM

PR #1937, opened 2024-02-28
Filter out non-operational and loopback interfaces

disa6302:
*Issue #, if available:*

*What was changed?*
- Add check to filter out gathering interfaces that are inactive or loopback on windows.

*Why was it changed?*
- This was done to ensure the SDK does not spend cycles on gathering interfaces unnecessarily and it could lead to time improvement if the getAddress operation is slow on Windows.

*How was it changed?*
- Added a check to check for interfaces activity and loopback.

- Tested the sample to confirm original functionality does not break.

### DIFF
--- a/src/source/Ice/Network.c
+++ b/src/source/Ice/Network.c
@@ -35,48 +35,51 @@ STATUS getLocalhostIpAddresses(PKvsIpAddress destIpList, PUINT32 pDestIpListLen,
     CHK(retWinStatus == ERROR_SUCCESS, STATUS_GET_LOCAL_IP_ADDRESSES_FAILED);
 
     for (aa = adapterAddresses; aa != NULL && ipCount < destIpListLen; aa = aa->Next) {
-        char ifa_name[BUFSIZ];
-        memset(ifa_name, 0, BUFSIZ);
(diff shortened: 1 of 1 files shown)


hassanctech (approved):
LGTM

jdelapla (approved):

src/source/Ice/Network.c
```diff
@@ -35,48 +35,51 @@ STATUS getLocalhostIpAddresses(PKvsIpAddress destIpList, PUINT32 pDestIpListLen,
...
-                DLOGI("Callback set to allow network interface filtering");
-                // The callback evaluates to a FALSE if the application is interested in black listing an interface
-                if (filter(customData, ifa_name) == FALSE) {
-                    filterSet = FALSE;
-                } else {
-                    filterSet = TRUE;
+        // Skip inactive interfaces and loop back interfaces
+        if (aa->OperStatus == IfOperStatusUp && aa->IfType != IF_TYPE_SOFTWARE_LOOPBACK) {
+            char ifa_name[BUFSIZ];
+            memset(ifa_name, 0, BUFSIZ);
```
sirknightj, line 41: Minor but we should be using the macros from pic (MEMSET)

PR #1944, opened 2024-03-08
Move gst blk inside GST_FOUND

disa6302:
*Issue #, if available:*

*What was changed?*
- Moving the pkg-config file search if only GST is found for windows.

*Why was it changed?*
- Without this change, windows customers would be mandated to install gstreamer which is not right.

*How was it changed?*
- Moved the block to within the `if(GST_FOUND)` block

- CI runs to confirm everything works as expected.


stefankiesz (approved):

hassanctech (approved):
lgtm

PR #1953, opened 2024-03-19
Viewer fix

disa6302:
*Issue #, if available:*

*What was changed?*
- Added new flags to monitor relay candidate gathering. 

*Why was it changed?*
- In case of viewer connection, the getIceServerConfig resolves after candidate gathering callback exits due to which relay candidates do not get added to the list, causing the gathering callback to assume that all candidates in the local candidate list are addressed and reported. This causes connections requiring TURN to fail. Adding new checks to ensure we stop gathering callback only if candidate pair is nominated or the relay candidates are added to the list fixes this problem. Forcing TURN does not catch this issue because the timer callback waits for the local candidate list to have some entry.

*How was it changed?*
- Added a flag for 2 cases: 
   1. `addedRelayCandidate`: This is set once relay candidate entries are created in `addIceServerConfig`. 
   2. `stopGathering`: This is set once a pair is nominated. It is useful in cases where the connection with STUN/host candidates is already established and the SDK need not wait on relay candidates to be generated.

- Confirmed existing unit tests passed
- Ran the JS SDK on phone and laptop with Viewer C SDK to confirm relay connection gets established and non-relay too
- Ran the JS SDK on phone and laptop with Master C SDK to confirm existing functionality is not broken.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -5,10 +5,7 @@ include(CheckIncludeFiles)
 include(CheckFunctionExists)
 
 # The version MUST be updated before every release
-project(KinesisVideoWebRTCClient VERSION 1.10.0 LANGUAGES C)
-
-
-
+project(KinesisVideoWebRTCClient VERSION 1.10.1 LANGUAGES C)
 
 # User Flags
 option(ADD_MUCLIBC "Add -muclibc c flag" OFF)
--- a/src/source/Ice/IceAgent.c
+++ b/src/source/Ice/IceAgent.c
@@ -328,11 +328,7 @@ STATUS iceAgentAddConfig(PIceAgent pIceAgent, PIceConfigInfo pIceConfigInfo)
             DLOGE("Failed to parse ICE servers");
         }
     }
-    //    if(!locked) {
-    //        MUTEX_LOCK(pIceAgent->lock);
-    //        locked = TRUE;
     ATOMIC_STORE_BOOL(&pIceAgent->addedRelayCandidate, TRUE);
-//    }
 CleanUp:
     CHK_LOG_ERR(retStatus);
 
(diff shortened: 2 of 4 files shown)


src/source/Ice/IceAgent.c
```diff
@@ -47,6 +47,7 @@ STATUS createIceAgent(PCHAR username, PCHAR password, PIceAgentCallbacks pIceAge
     ATOMIC_STORE_BOOL(&pIceAgent->shutdown, FALSE);
     ATOMIC_STORE_BOOL(&pIceAgent->restart, FALSE);
     ATOMIC_STORE_BOOL(&pIceAgent->processStun, TRUE);
+    ATOMIC_STORE_BOOL(&pIceAgent->addedRelayCandidate, FALSE);
```
niyatim23, line 50: Shouldn't we also initialize `stopGathering` to `FALSE` here?

hassanctech: +1

CMakeLists.txt
```diff
@@ -5,10 +5,7 @@ include(CheckIncludeFiles)
 include(CheckFunctionExists)
 
 # The version MUST be updated before every release
-project(KinesisVideoWebRTCClient VERSION 1.10.0 LANGUAGES C)
-
-
-
+project(KinesisVideoWebRTCClient VERSION 1.10.1 LANGUAGES C)
```
cheersalam, line 8: Is versioning automated? if not can you please create a backlog item.

niyatim23 (approved):
Lgtm! I tested C-SDK viewer, ICE_TRANSPORT_POLICY_ALL with JS-master, forced TURN and saw a seamless connection. I also tested this on Windows. Need to increase the timeout there because Windows is a bit slow: `configuration.kvsRtcConfiguration.iceLocalCandidateGatheringTimeout = (20 * HUNDREDS_OF_NANOS_IN_A_SECOND);`

hassanctech (approved):
lgtm

PR #1955, opened 2024-03-22
New unit tests

disa6302:
*Issue #, if available:*

*What was changed?*
- Added some missing API and functional unit tests.

*Why was it changed?*
- To improve coverage and fix any potential bugs in the process.

*How was it changed?*
- Added some unit tests.
- Fixed a few bugs
   - NULL check fix in metrics
   - Log only if `getIpAddrStr` succeeds while logging failing socket - segfault without this change
   - Update rolling buffer member instead of local variable in default setting case

- Ensuring CI runs clean on all unit tests. Expecting coverage to have increased from 75% to 80%.


src/source/Metrics/Metrics.c
```diff
@@ -9,7 +9,7 @@ STATUS getIceCandidatePairStats(PRtcPeerConnection pRtcPeerConnection, PRtcIceCa
     STATUS retStatus = STATUS_SUCCESS;
     BOOL locked = FALSE;
     PIceAgent pIceAgent = NULL;
-    CHK((pRtcPeerConnection != NULL || pRtcIceCandidatePairStats != NULL), STATUS_NULL_ARG);
+    CHK((pRtcPeerConnection != NULL && pRtcIceCandidatePairStats != NULL), STATUS_NULL_ARG);
```
sirknightj, line 12: Perhaps we can just change this to if "either are null" rather than "if both are not null". (more readable that way)

tst/MetricsApiTest.cpp
```diff
@@ -156,6 +156,71 @@ TEST_F(MetricsApiTest, webRtcIceCandidateGetMetrics)
...
 
+TEST_F(MetricsApiTest, webRtcTransportGetMetrics)
+{
+    RtcConfiguration configuration;
+    PRtcPeerConnection pRtcPeerConnection = NULL;
+    RtcStats rtcTransportMetrics;
+
+    MEMSET(&configuration, 0x00, SIZEOF(RtcConfiguration));
+    STRNCPY(configuration.iceServers[0].urls, (PCHAR) "stun:stun.kinesisvideo.us-west-2.amazonaws.com:443", MAX_ICE_CONFIG_URI_LEN);
+    STRNCPY(configuration.iceServers[0].credential, (PCHAR) "", MAX_ICE_CONFIG_CREDENTIAL_LEN);
```
sirknightj, line 167: I saw this macro that we can use instead:
https://github.com/awslabs/amazon-kinesis-video-streams-pic/blob/65e38dac9b30523d43a57bc009d679e627b58d9a/src/common/include/com/amazonaws/kinesis/video/common/CommonDefs.h#L878

src/source/Metrics/Metrics.c
```diff
@@ -87,9 +88,11 @@ STATUS getIceServerStats(PRtcPeerConnection pRtcPeerConnection, PRtcIceServerSta
 {
     STATUS retStatus = STATUS_SUCCESS;
     BOOL locked = FALSE;
-    PIceAgent pIceAgent = ((PKvsPeerConnection) pRtcPeerConnection)->pIceAgent;
+    PIceAgent pIceAgent = NULL;
```
sirknightj, line 91: nice

tst/SignalingApiTest.cpp
```diff
@@ -9,11 +9,57 @@ namespace webrtcclient {
...
+    MEMSET(testChannel, 0, MAX_CHANNEL_NAME_LEN+1);
+
+    append = rand() % TEST_CHANNEL_COUNT;
+    SNPRINTF(testWssEndpoint, SIZEOF(testWssEndpoint), "%s%d", "testWssEndpoint", append);
+    SNPRINTF(testHttpsEndpoint, SIZEOF(testHttpsEndpoint),"%s%d", "testHttpsEndpoint", append);
+    SNPRINTF(testRegion, SIZEOF(testRegion),"%s%d", "testRegion", append);
+    SNPRINTF(testChannelArn, SIZEOF(testChannelArn),"%s%d", "testChannelArn", append);
+    SNPRINTF(testChannel, SIZEOF(testChannel),"%s%d", "testChannel", append);
+
+    EXPECT_EQ(STATUS_SUCCESS, createFileIfNotExist(DEFAULT_CACHE_FILE_PATH));
```
sirknightj, line 47: Do we also need to clear it just in case it exists from a previous test or something

disa6302: Done.

tst/MetricsApiTest.cpp
disa6302: Done

src/source/Metrics/Metrics.c
disa6302: That was the check. But `CHK` has a !(cond), so using `&&`

tst/sdp_config.h
```diff
@@ -0,0 +1,25 @@
...
+
+#ifndef __KINESIS_VIDEO_WEBRTC_CLIENT_SDP_SDP_CONFIG__
+#define __KINESIS_VIDEO_WEBRTC_CLIENT_SDP_SDP_CONFIG__
+
+#include <inttypes.h>
+
+/**
+ * @brief SDP print format for uint64_t.
+ */
+#define SDP_PRINT_FMT_UINT64 PRIu64
```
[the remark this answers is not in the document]
disa6302: Its a requirement to pass the build by the newly added SDP component. We have this dummy file in configs too. Creating one exclusively for tst directory to separate out the source and tst components.

tst/StunApiTest.cpp
```diff
@@ -213,14 +213,141 @@ TEST_F(StunApiTest, appendStunAttributeValidityTests)
...
+    PStunPacket pStunPacket;
+    PStunAttributeHeader pAttribute;
+    CHAR errorPhrase[128];
+    CHAR password[256];
+    BYTE buffer[10000];
+    UINT32 size = 0;
+    BOOL found;
+
+    STRCPY(password, "test password");
+    STRCPY(errorPhrase, "Sample phrase");
```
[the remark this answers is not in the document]
disa6302: Done

tst/SignalingApiTest.cpp
disa6302: The file removal already exists in the start of the function. And I am directly writing the entry to the file.

src/source/Metrics/Metrics.c
```diff
@@ -58,8 +58,9 @@ STATUS getIceCandidateStats(PRtcPeerConnection pRtcPeerConnection, BOOL isRemote
 {
     STATUS retStatus = STATUS_SUCCESS;
     BOOL locked = FALSE;
-    PIceAgent pIceAgent = ((PKvsPeerConnection) pRtcPeerConnection)->pIceAgent;
-    CHK((pRtcPeerConnection != NULL || pRtcIceCandidateStats != NULL), STATUS_NULL_ARG);
+    PIceAgent pIceAgent = NULL;
+    CHK((pRtcPeerConnection != NULL && pRtcIceCandidateStats != NULL), STATUS_NULL_ARG);
+    pIceAgent = ((PKvsPeerConnection) pRtcPeerConnection)->pIceAgent;
```
sirknightj, line 63: Note to self: [KvsPeerConnection.pIceAgent](https://github.com/awslabs/amazon-kinesis-video-streams-webrtc-sdk-c/blob/f1073439213db4aa4b4b7cc1423609d920dd5489/src/source/PeerConnection/PeerConnection.h#L79) appears to be [guaranteed non-null](https://github.com/awslabs/amazon-kinesis-video-streams-webrtc-sdk-c/blob/f1073439213db4aa4b4b7cc1423609d920dd5489/src/source/PeerConnection/PeerConnection.c#L950).

sirknightj (approved):

tst/sdp_config.h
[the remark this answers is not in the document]
hassanctech: Why do we need a specific one for tst directory?  Since the existing one is already at the configs/ directory and not in source, I think it would be much cleaner (and avoid code duplication) to simply reference the same file in the cmakelists for both the SDK and the test project (use target_include_directories).  I understand separating out configs in which the tests will override something but in this case we actually don't want the tests to override this as this is supposed to work for all platforms.

src/source/Signaling/FileCache.h
```diff
@@ -38,6 +38,7 @@ typedef struct {
 STATUS deserializeSignalingCacheEntries(PCHAR, UINT64, PSignalingFileCacheEntry, PUINT32, PCHAR);
 STATUS signalingCacheLoadFromFile(PCHAR, PCHAR, SIGNALING_CHANNEL_ROLE_TYPE, PSignalingFileCacheEntry, PBOOL, PCHAR);
 STATUS signalingCacheSaveToFile(PSignalingFileCacheEntry, PCHAR);
+STATUS createFileIfNotExist(PCHAR);
```
hassanctech, line 41: Nice!

hassanctech (approved):
awesome work!

sirknightj (approved):

PR #1957, opened 2024-03-27
Sample TWCC implementation

disa6302:
*Issue #, if available:*

*What was changed?*
- Added TWCC sample to integrate with gstreamer pipelines

*Why was it changed?*
- To provide an example demonstration of TWCC usage

*How was it changed?*
- Included a new structure to store twcc metadata in sample
- Using `g_object_set` and `g_object_get` to change bitrate

- Used `tc` on ubuntu with testsrc to introduce packet loss % at specified times to confirm better playback quality (less freezing, faster recovery because of adaptive bitrate) compared to when not using TWCC.


samples/Common.c
```diff
@@ -704,24 +715,70 @@ VOID sampleBandwidthEstimationHandler(UINT64 customData, DOUBLE maximumBitrate)
...
+        audioBitrate = (UINT64) MAX(audioBitrate * (1.0f - pSampleStreamingSession->twccMetadata.averagePacketLoss / 100.0f), MIN_AUDIO_BITRATE_BPS);
+    }
+
+    MUTEX_LOCK(pSampleStreamingSession->twccMetadata.updateLock);
+    locked = TRUE;
+    // Update the session with the new bitrate and adjustment time
+    pSampleStreamingSession->twccMetadata.newVideoBitrate = videoBitrate;
+    pSampleStreamingSession->twccMetadata.newAudioBitrate = audioBitrate;
+    if (locked) {
+        MUTEX_UNLOCK(pSampleStreamingSession->twccMetadata.updateLock);
```
cheersalam, line 772: We are doing lock/unlock multiple times and only doing assignments in between, there is not much computations. Single lock and unlock should suffice.

samples/Common.c
```diff
@@ -704,24 +715,70 @@ VOID sampleBandwidthEstimationHandler(UINT64 customData, DOUBLE maximumBitrate)
...
-          txBytes, txPacketsCnt, rxBytes, rxPacketsCnt, duration / 10000ULL);
+    UINT8 percentLost = (txPacketsCnt > 0) ? (lostPacketsCnt * 100 / txPacketsCnt) : 0;
+    SampleStreamingSession* pSampleStreamingSession = (SampleStreamingSession*) customData;
+    if (pSampleStreamingSession == NULL) {
+        DLOGW("Invalid streaming session (NULL object)");
+        return;
+    }
+
+    // Calculate packet loss
+    DOUBLE currentPacketLoss = (DOUBLE) percentLost;
```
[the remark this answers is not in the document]
cheersalam: UINT8 typecasted to DOUBLE?

samples/Common.c
```diff
@@ -704,24 +715,70 @@ VOID sampleBandwidthEstimationHandler(UINT64 customData, DOUBLE maximumBitrate)
...
-        bitrate *= 1.02f;
-    } else if (percentLost > 5) {
-        // decrease encoder bitrate by packet loss percent
-        bitrate *= (1.0f - percentLost / 100.0f);
-    }
-    // otherwise keep bitrate the same
-
-    DLOGV("received sender bitrate estimation: suggested bitrate %u sent: %u bytes %u packets received: %u bytes %u packets in %lu msec", bitrate,
-          txBytes, txPacketsCnt, rxBytes, rxPacketsCnt, duration / 10000ULL);
+    UINT8 percentLost = (txPacketsCnt > 0) ? (lostPacketsCnt * 100 / txPacketsCnt) : 0;
```
cheersalam, line 722: why percentLost is not double?

samples/Common.c
```diff
@@ -704,24 +715,70 @@ VOID sampleBandwidthEstimationHandler(UINT64 customData, DOUBLE maximumBitrate)
...
+    locked = TRUE;
+    videoBitrate = pSampleStreamingSession->twccMetadata.currentVideoBitrate;
+    audioBitrate = pSampleStreamingSession->twccMetadata.currentAudioBitrate;
+    if (locked) {
+        MUTEX_UNLOCK(pSampleStreamingSession->twccMetadata.updateLock);
+        locked = FALSE;
+    }
+
+    if (pSampleStreamingSession->twccMetadata.averagePacketLoss <= 5) {
+        // increase encoder bitrate by 5 percent with a cap at MAX_BITRATE
```
cheersalam, line 751: Is this correlation of avg packet loss with bitrate a recommendation from a spec? How did we deduced this calculation?

samples/Samples.h
```diff
@@ -72,6 +72,12 @@ extern "C" {
 #define MAX_SIGNALING_CLIENT_METRICS_MESSAGE_SIZE 736 // strlen(SIGNALING_CLIENT_METRICS_JSON_TEMPLATE) + 20 * 10
 #define MAX_ICE_AGENT_METRICS_MESSAGE_SIZE        113 // strlen(ICE_AGENT_METRICS_JSON_TEMPLATE) + 20 * 2
 
+#define ADJUSTMENT_INTERVAL_SECONDS 1 * HUNDREDS_OF_NANOS_IN_A_SECOND
+#define MIN_VIDEO_BITRATE_KBPS      512     // Unit kilobits/sec. Value could change based on codec.
+#define MAX_VIDEO_BITRATE_KBPS      2048000 // Unit kilobits/sec. Value could change based on codec.
```
cheersalam, line 77: Are these values from a sample video?

cheersalam, line 77: max is in Gbps?

samples/Common.c
```diff
@@ -704,24 +715,70 @@ VOID sampleBandwidthEstimationHandler(UINT64 customData, DOUBLE maximumBitrate)
...
+        // decrease encoder bitrate by average packet loss percent, with a cap at MIN_BITRATE
+        videoBitrate = (UINT64) MAX(videoBitrate * (1.0f - pSampleStreamingSession->twccMetadata.averagePacketLoss / 100.0f), MIN_VIDEO_BITRATE_KBPS);
+    }
+
+    if (pSampleStreamingSession->twccMetadata.averagePacketLoss <= 5) {
+        // increase encoder bitrate by 5 percent with a cap at MAX_BITRATE
+        audioBitrate = (UINT64) MIN(audioBitrate * 1.05f, MAX_AUDIO_BITRATE_BPS);
+    } else {
+        // decrease encoder bitrate by average packet loss percent, with a cap at MIN_BITRATE
+        audioBitrate = (UINT64) MAX(audioBitrate * (1.0f - pSampleStreamingSession->twccMetadata.averagePacketLoss / 100.0f), MIN_AUDIO_BITRATE_BPS);
```
cheersalam, line 763: Audio has higher priority over Video? Is there any industry recommendation?

samples/Samples.h
```diff
@@ -168,6 +175,16 @@ typedef struct {
 
 typedef VOID (*StreamSessionShutdownCallback)(UINT64, PSampleStreamingSession);
 
+typedef struct {
+    MUTEX updateLock;
+    UINT64 lastAdjustmentTimeMs;
+    UINT64 currentVideoBitrate;
```
cheersalam, line 181: This is in Kbps? Can we add Kbps to name to avoid any confusion?

samples/Common.c
```diff
@@ -704,24 +715,70 @@ VOID sampleBandwidthEstimationHandler(UINT64 customData, DOUBLE maximumBitrate)
...
+
+    if (pSampleStreamingSession->twccMetadata.averagePacketLoss <= 5) {
+        // increase encoder bitrate by 5 percent with a cap at MAX_BITRATE
+        videoBitrate = (UINT64) MIN(videoBitrate * 1.05f, MAX_VIDEO_BITRATE_KBPS);
+    } else {
+        // decrease encoder bitrate by average packet loss percent, with a cap at MIN_BITRATE
+        videoBitrate = (UINT64) MAX(videoBitrate * (1.0f - pSampleStreamingSession->twccMetadata.averagePacketLoss / 100.0f), MIN_VIDEO_BITRATE_KBPS);
+    }
+
+    if (pSampleStreamingSession->twccMetadata.averagePacketLoss <= 5) {
```
cheersalam, line 758: During your testing, did you observed any oscillations between good and bad bitrate? Did it got stable after a packet loss?

disa6302: Yeah. I ran tests with tc to change packet loss rates and then switching it off at times too. I saw a better playback recovery and stability compared to without twcc.

samples/Samples.h
disa6302: Not setting units because the unit is gstreamer encoder dependent. In most cases, video is in kbps and audio in bps, but do not want to assume. The actual value is read from with the gstreamer apis and that is what is modified. Since this is in the sample, customers are free to change these.

samples/Common.c
disa6302: There is no priority, calculation is for both video and audio.

samples/Samples.h
disa6302: Essentially, yes. The limit is for gstreamer x264enc element and the unit is kbps. Hence the comment that these values could change based on the codec being used. The limit is necessary to make sure gstreamer does not fail with critical warnings being unable to set a value beyond this.

disa6302: Its for x264enc. Does not matter what the source is. I am setting min to 512kbps since it is the value we set in our samples. As such gstreamer min for x264enc is 100kbps if I not wrong.

samples/Common.c
disa6302: There is no spec recommendation on how the values should be set, neither is there any spec on how the algorithm should be implemented. This is an adaptation of the google congestion loss based algorithm. I set it to 5% to allow a decent increase with good bitrate. This callback can be modified to change these values based on customer set up and tolerances. It is an example implementation.

samples/Common.c
disa6302: Done.

samples/Common.c
disa6302: You are right there is no computation and just assignment. But, this lock is being used in the gstreamer callback while sending frame, so just did not want to block it while updating. Hence the locking only while updating critical members.

samples/Common.c
```diff
@@ -704,24 +715,70 @@ VOID sampleBandwidthEstimationHandler(UINT64 customData, DOUBLE maximumBitrate)
...
+
+    UINT64 currentTimeMs = GETTIME();
+    UINT64 timeDiff = currentTimeMs - pSampleStreamingSession->twccMetadata.lastAdjustmentTimeMs;
+    if (timeDiff < ADJUSTMENT_INTERVAL_SECONDS) {
+        // Too soon for another adjustment
+        return;
+    }
+
+    MUTEX_LOCK(pSampleStreamingSession->twccMetadata.updateLock);
+    locked = TRUE;
```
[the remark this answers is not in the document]
disa6302: The value is read in gstreamer callback and updated here.

samples/Common.c
```diff
@@ -549,10 +548,16 @@ STATUS createSampleStreamingSession(PSampleConfiguration pSampleConfiguration, P
...
     ATOMIC_STORE_BOOL(&pSampleStreamingSession->terminateFlag, FALSE);
     ATOMIC_STORE_BOOL(&pSampleStreamingSession->candidateGatheringDone, FALSE);
-
+    MEMSET(&pSampleStreamingSession->twccMetadata, 0x00, SIZEOF(TwccMetadata));
     pSampleStreamingSession->peerConnectionMetrics.peerConnectionStats.peerConnectionStartTime = GETTIME() / HUNDREDS_OF_NANOS_IN_A_MILLISECOND;
     // Flag to enable SDK to calculate selected ice server, local, remote and candidate pair stats.
     pSampleConfiguration->enableIceStats = FALSE;
+    pSampleConfiguration->disableTwcc = FALSE;
+
+    if (!pSampleConfiguration->disableTwcc) {
```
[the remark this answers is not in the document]
disa6302: Yes. Because we are setting to FALSE. But customer could change this to TRUE and the lock will get allocated for no reason. They might not remove the lock allocation.

samples/Common.c
```diff
@@ -704,24 +715,60 @@ VOID sampleBandwidthEstimationHandler(UINT64 customData, DOUBLE maximumBitrate)
...
+    MUTEX_LOCK(pSampleStreamingSession->twccMetadata.updateLock);
+    videoBitrate = pSampleStreamingSession->twccMetadata.currentVideoBitrate;
+    audioBitrate = pSampleStreamingSession->twccMetadata.currentAudioBitrate;
+
+    if (pSampleStreamingSession->twccMetadata.averagePacketLoss <= 5) {
+        // increase encoder bitrate by 5 percent with a cap at MAX_BITRATE
+        videoBitrate = (UINT64) MIN(videoBitrate * 1.05f, MAX_VIDEO_BITRATE_KBPS);
+    } else {
+        // decrease encoder bitrate by average packet loss percent, with a cap at MIN_BITRATE
+        videoBitrate = (UINT64) MAX(videoBitrate * (1.0f - pSampleStreamingSession->twccMetadata.averagePacketLoss / 100.0f), MIN_VIDEO_BITRATE_KBPS);
```
sirknightj, line 750: Any particular reason float is used instead of double?

samples/Common.c
```diff
@@ -704,24 +715,60 @@ VOID sampleBandwidthEstimationHandler(UINT64 customData, DOUBLE maximumBitrate)
 VOID sampleSenderBandwidthEstimationHandler(UINT64 customData, UINT32 txBytes, UINT32 rxBytes, UINT32 txPacketsCnt, UINT32 rxPacketsCnt,
```
sirknightj, line 715: Can we add a comment here explaining what the sample implementation does at a high level

e.g/ sent bitrate is inversely proportional to the packet loss %

samples/Common.c
disa6302: Nah. It does not matter because we convert to uint64 anyways. But removed the f to avoid confusion.

samples/Common.c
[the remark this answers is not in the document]
cheersalam: it is your local variable. I still didn't get the purpose of this variable.

samples/Common.c
cheersalam: It is locked twice in a function. And between two locks there is no other function call and no time consuming computations. You can save some CPU cycles by locking once. Lock at beginning and release in the end.

samples/Common.c
```diff
@@ -409,13 +409,12 @@ STATUS initializePeerConnection(PSampleConfiguration pSampleConfiguration, PRtcP
     configuration.kvsRtcConfiguration.iceSetInterfaceFilterFunc = NULL;
 
     // disable TWCC
-    configuration.kvsRtcConfiguration.disableSenderSideBandwidthEstimation = TRUE;
+    configuration.kvsRtcConfiguration.disableSenderSideBandwidthEstimation = pSampleConfiguration->disableTwcc;
```
sirknightj, line 412: Having `!` here is confusing due to double negative, can we make the sample configuration positive. This variable is always used inverted.

We can keep "disableSenderSideBandwidthEstimation" to avoid changing existing. But since `disableTwcc` is a new property in the sample configuration, we can invert it for easier to read.

samples/Common.c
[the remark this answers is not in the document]
sirknightj: Can we enable through an env variable?
I am thinking it will make it faster to test what it is like with / without TWCC. Or building 2 executables one with enabled and disabled.

samples/Common.c
```diff
@@ -701,27 +714,66 @@ VOID sampleBandwidthEstimationHandler(UINT64 customData, DOUBLE maximumBitrate)
...
+
+    if (pSampleStreamingSession->twccMetadata.averagePacketLoss <= 5) {
+        // increase encoder bitrate by 5 percent with a cap at MAX_BITRATE
+        videoBitrate = (UINT64) MIN(videoBitrate * 1.05, MAX_VIDEO_BITRATE_KBPS);
+    } else {
+        // decrease encoder bitrate by average packet loss percent, with a cap at MIN_BITRATE
+        videoBitrate = (UINT64) MAX(videoBitrate * (1.0 - pSampleStreamingSession->twccMetadata.averagePacketLoss / 100.0), MIN_VIDEO_BITRATE_KBPS);
+    }
+
+    if (pSampleStreamingSession->twccMetadata.averagePacketLoss <= 5) {
```
sirknightj, line 758: Since these are the same condition, can we place this into the same if/else?

```
if () {
  adjust video
  adjust audio
} else {
  adjust video
  adjust audio
}
```

samples/Samples.h
```diff
@@ -72,6 +72,12 @@ extern "C" {
 #define MAX_SIGNALING_CLIENT_METRICS_MESSAGE_SIZE 736 // strlen(SIGNALING_CLIENT_METRICS_JSON_TEMPLATE) + 20 * 10
 #define MAX_ICE_AGENT_METRICS_MESSAGE_SIZE        113 // strlen(ICE_AGENT_METRICS_JSON_TEMPLATE) + 20 * 2
 
+#define ADJUSTMENT_INTERVAL_SECONDS 1 * HUNDREDS_OF_NANOS_IN_A_SECOND
```
sirknightj, line 75: A bit confusing that `ADJUSTMENT_INTERVAL_SECONDS = 1 * HUNDREDS_OF_NANOS_IN_A_SECOND`

README.md
```diff
@@ -502,6 +502,32 @@ To disable threadpool, run `cmake .. -DENABLE_KVS_THREADPOOL=OFF`
...
+The importance of TWCC in WebRTC lies in its ability to ensure efficient use of available network bandwidth while minimizing the negative impacts of network congestion. By monitoring the delivery of packets across the network, TWCC helps identify bottlenecks and adjust the media transmission rates accordingly. 
+This dynamic approach to congestion control is essential for preventing degradation in call quality, such as pixelation, stuttering, or drops in audio and video streams, especially in environments with fluctuating network conditions. To learn more about TWCC, you can refer to the [RFC draft](https://datatracker.ietf.org/doc/html/draft-holmer-rmcat-transport-wide-cc-extensions-01)
+
+In order to enable TWCC usage in the SDK, 2 things need to be set up:
+
+1. Set the `disableSenderSideBandwidthEstimation` to FALSE. In our samples, the value is set using `disableTwcc` flag in `pSampleConfiguration`
```
sirknightj, line 513: Enable

samples/Common.c
```diff
@@ -414,13 +414,13 @@ STATUS initializePeerConnection(PSampleConfiguration pSampleConfiguration, PRtcP
     configuration.kvsRtcConfiguration.iceSetInterfaceFilterFunc = NULL;
 
     // disable TWCC
-    configuration.kvsRtcConfiguration.disableSenderSideBandwidthEstimation = TRUE;
+    configuration.kvsRtcConfiguration.disableSenderSideBandwidthEstimation = !(pSampleConfiguration->enableTwcc);
+    DLOGI("Disable: %s", configuration.kvsRtcConfiguration.disableSenderSideBandwidthEstimation ? "Disabled" : "Enabled");
```
sirknightj, line 418: Maybe we can have more descriptive log of what is disabled

samples/Samples.h
```diff
@@ -83,6 +83,11 @@ extern "C" {
 #define MAX_SIGNALING_CLIENT_METRICS_MESSAGE_SIZE 736 // strlen(SIGNALING_CLIENT_METRICS_JSON_TEMPLATE) + 20 * 10
 #define MAX_ICE_AGENT_METRICS_MESSAGE_SIZE        113 // strlen(ICE_AGENT_METRICS_JSON_TEMPLATE) + 20 * 2
 
+
+#define ADJUSTMENT_INTERVAL_MS 5 * HUNDREDS_OF_NANOS_IN_A_SECOND
```
sirknightj: Why multiply by HUNDREDS_OF_NANOS_IN_A_SECOND?

samples/Common.c
```diff
@@ -554,10 +554,16 @@ STATUS createSampleStreamingSession(PSampleConfiguration pSampleConfiguration, P
 
     ATOMIC_STORE_BOOL(&pSampleStreamingSession->terminateFlag, FALSE);
     ATOMIC_STORE_BOOL(&pSampleStreamingSession->candidateGatheringDone, FALSE);
-
+    MEMSET(&pSampleStreamingSession->twccMetadata, 0x00, SIZEOF(TwccMetadata));
```
hassanctech, line 557: We don't need this right I see that in side `createSampleStreamingSession` we call:

```
    pSampleStreamingSession = (PSampleStreamingSession) MEMCALLOC(1, SIZEOF(SampleStreamingSession));
```

And since `enableTwcc` is a struct member (non-pointer) it would be guaranteed to set to zero.

disa6302: Good catch! Removed.

samples/Common.c
hassanctech: One thing we've missed on when implementing features like this that have possibly multiple toggles instead of exposing all the toggles to the customer (which we do some times) OR hard coding those values (which we also do sometimes) is to instead have like 2-3 profiles / strategies.  


Examining this whole feature at a high level I'm thinking if instead of even exposing TWCC as an option would it better customer experience to instead expose options like:

`AGGRESSIVE_PACKET_LOSS`, `MODERATE_PACKET_LOSS`, `LOW_PACKET_LOSS`

Perhaps in these options the first two we enable TWCC and we set different values for average packet loss threshold and for low packet loss we just don't do TWCC at all? Basically I'm thinking about a higher level profile/mode the SDK runs in and we hide the fine grained control of this packet loss behind that?

samples/Common.c
```diff
@@ -554,10 +554,16 @@ STATUS createSampleStreamingSession(PSampleConfiguration pSampleConfiguration, P
...
     pSampleStreamingSession->peerConnectionMetrics.peerConnectionStats.peerConnectionStartTime = GETTIME() / HUNDREDS_OF_NANOS_IN_A_MILLISECOND;
     // Flag to enable SDK to calculate selected ice server, local, remote and candidate pair stats.
     pSampleConfiguration->enableIceStats = FALSE;
+    pSampleConfiguration->enableTwcc = TRUE;
```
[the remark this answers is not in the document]
hassanctech: I see in that case doesn't it make sense to have this toggle inside `createSampleConfiguration` instead of this function since the enableTwcc flag is part of the sample configuration??

And in that case we should have a 1-line comment above where we set it that this toggle is used to enable/disable twcc with a link to a page that explains what twcc is (ideally a rfc or something we have reasonably certainty the url won't change).  I suppose I have the same question about the `enableIceStats` why don't we set that inside `createSampleConfiguration` ?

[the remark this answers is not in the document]
disa6302: Moved both. Added TWCC related details in the README.

samples/Common.c
hassanctech: Since this PR is for the sample illustration of the feature and what I'm talking about is core to the SDK itself, we can consider that as a future enhancement.

hassanctech (approved):
lgtm

samples/Common.c
```diff
@@ -706,27 +716,62 @@ VOID sampleBandwidthEstimationHandler(UINT64 customData, DOUBLE maximumBitrate)
...
+        return;
+    }
+
+    // Calculate packet loss
+    pSampleStreamingSession->twccMetadata.averagePacketLoss =
+        EMA_ACCUMULATOR_GET_NEXT(pSampleStreamingSession->twccMetadata.averagePacketLoss, ((DOUBLE) percentLost));
+
+    currentTimeMs = GETTIME();
+    timeDiff = currentTimeMs - pSampleStreamingSession->twccMetadata.lastAdjustmentTimeMs;
+    if (timeDiff < TWCC_BITRATE_ADJUSTMENT_INTERVAL_SECONDS) {
```
sirknightj, line 743: The unit here should be millis, timediff is in millis

samples/kvsWebRTCClientMasterGstSample.c
```diff
@@ -168,38 +198,38 @@ PVOID sendGstreamerAudioVideo(PVOID args)
...
-                    pipeline =
-                        gst_parse_launch("videotestsrc is-live=TRUE ! queue ! videoconvert ! videoscale ! video/x-raw,width=1280,height=720 ! "
-                                         "videorate ! video/x-raw,framerate=25/1 ! "
-                                         "x264enc bframes=0 speed-preset=veryfast bitrate=512 byte-stream=TRUE tune=zerolatency ! "
-                                         "video/x-h264,stream-format=byte-stream,alignment=au,profile=baseline ! appsink sync=TRUE emit-signals=TRUE "
-                                         "name=appsink-video",
-                                         &error);
+                    senderPipeline = gst_parse_launch(
+                        "videotestsrc pattern=ball is-live=TRUE ! queue ! videoconvert ! video/x-raw,width=1280,height=720,framerate=25/1 ! "
+                        "x264enc name=sampleVideoEncoder bframes=0 speed-preset=veryfast bitrate=512 byte-stream=TRUE tune=zerolatency ! "
```
sirknightj, line 203: Name param isn't necessary

sirknightj: Can we add a clock overlay w/ date + time on it?

samples/kvsWebRTCClientMasterGstSample.c
```diff
@@ -168,38 +198,38 @@ PVOID sendGstreamerAudioVideo(PVOID args)
...
                 }
                 case DEVICE_SOURCE: {
-                    pipeline = gst_parse_launch("autovideosrc ! queue ! videoconvert ! videoscale ! video/x-raw,width=1280,height=720 ! "
-                                                "videorate ! video/x-raw,framerate=25/1 ! "
-                                                "x264enc bframes=0 speed-preset=veryfast bitrate=512 byte-stream=TRUE tune=zerolatency ! "
-                                                "video/x-h264,stream-format=byte-stream,alignment=au,profile=baseline ! appsink sync=TRUE "
-                                                "emit-signals=TRUE name=appsink-video",
-                                                &error);
+                    senderPipeline = gst_parse_launch(
+                        "autovideosrc ! queue ! videoconvert ! video/x-raw,width=1280,height=720,framerate=25/1 ! "
```
sirknightj, line 211: Will want video rate and video scale in case camera doesn't support that framerate or resolution

disa6302: Done. I thought this already existed. Not sure if some rebase messed it up.

samples/kvsWebRTCClientMasterGstSample.c
disa6302: Done. Going to keep naming because we use it in modifying.

samples/Common.c
disa6302: It essentially means we need a 1 second difference in millisecond format. I changed it to clarify, but this does not change the functionality in anyway. Still means 1 second.

sirknightj (dismissed):

samples/Common.c
```diff
@@ -706,27 +716,61 @@ VOID sampleBandwidthEstimationHandler(UINT64 customData, DOUBLE maximumBitrate)
     DLOGV("received bitrate suggestion: %f", maximumBitrate);
 }
 
+// Sample callback for TWCC. Average packet is calculated with EMA. If average packet lost is <= 5%,
```
unicornss, line 719: exponential moving avg??

unicornss (commented):
Could you add the scenarios that were used to test this sample? It will be helpful if you can add any statistics or graphs from those testing.

README.md
```diff
@@ -396,21 +396,32 @@ createLwsIotCredentialProvider(
...
+
+The importance of TWCC in WebRTC lies in its ability to ensure efficient use of available network bandwidth while minimizing the negative impacts of network congestion. By monitoring the delivery of packets across the network, TWCC helps identify bottlenecks and adjust the media transmission rates accordingly. This dynamic approach to congestion control is essential for preventing degradation in call quality, such as pixelation, stuttering, or drops in audio and video streams, especially in environments with fluctuating network conditions.
+
+To learn more about TWCC, check [TWCC spec](https://datatracker.ietf.org/doc/html/draft-holmer-rmcat-transport-wide-cc-extensions-01)
+
+### Enabling TWCC support
+
+TWCC is enabled by default in the SDK samples (via `pSampleConfiguration->enableTwcc`) flag. In order to disable it, set this flag to `FALSE`.
```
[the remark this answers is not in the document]
disa6302: I have added the details in the readme already. Section on the 2 things that need to be set to use TWCC.

samples/Common.c
disa6302: Yes.

hassanctech (approved):
lgtm

samples/Common.c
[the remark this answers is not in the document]
hassanctech: thank you!

sirknightj (approved):

PR #1972, opened 2024-04-18
[Bug] GStreamer sample

niyatim23:
*Issue #, if available:*

*What was changed?*
- Removed the receiver from the GstMaster sample
- PTS fix
- EOS fix

*Why was it changed?*
It was removed because it segfaults with the JS viewer because the JS viewer modifies caps mid-streaming. It is not supported by matroskamux in GStreamer. It has been broken but didn't manifest itself earlier because autovideosink in the previous pipeline did not work at all

*How was it changed?*
Removing the receiver from the GStreamer sample application as it was non-functional

- Ran the GstSample with JS viewer and not seeing a segfault
- Ran the GstSample master with GstSample viewer and made sure the audio and video played properly

### DIFF
--- a/README.md
+++ b/README.md
@@ -309,8 +309,8 @@ To run:
 ./samples/kvsWebrtcClientViewerGstSample <channelName> <mediaType>
 ```
 
-##### Note:
-Our GStreamer samples leverage [MatroskaMux](https://gstreamer.freedesktop.org/documentation/matroska/matroskamux.html?gi-language=c) to receive media from its peer and save it to a file. However, MatroskaMux is designed for scenarios where the media's format remains constant throughout streaming. Unfortunately, this is not the case while using certain browsers as peers with ([AWS KVS JS SDK](https://awslabs.github.io/amazon-kinesis-video-streams-webrtc-sdk-js/examples/index.html)), so receiving media from the browsers andf writing it to filesink via MatroskaMux is not supported by our existing GStreamer samples. When the media's format changes mid-streaming (referred to as "caps changes"), MatroskaMux encounters limitations and is unable to handle these changes, resulting in an error message like:
+##### Known issues:
+Our GStreamer samples leverage [MatroskaMux](https://gstreamer.freedesktop.org/documentation/matroska/matroskamux.html?gi-language=c) to receive media from its peer and save it to a file. However, MatroskaMux is designed for scenarios where the media's format remains constant throughout streaming. When the media's format changes mid-streaming (referred to as "caps changes"), MatroskaMux encounters limitations, its behavior cannot be predicted and it may be unable to handle these changes, resulting in an error message like:
 
 ```shell
 matroskamux matroska-mux.c:1134:gst_matroska_mux_video_pad_setcaps:<mux> error: Caps changes are not supported by Matroska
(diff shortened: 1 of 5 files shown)


README.md
```diff
@@ -309,6 +309,14 @@ To run:
 ./samples/kvsWebrtcClientViewerGstSample <channelName> <mediaType>
 ```
 
+##### Note:
+Our GStreamer samples leverage [MatroskaMux](https://gstreamer.freedesktop.org/documentation/matroska/matroskamux.html?gi-language=c) to receive media from its peer and save it to a file. However, MatroskaMux is designed for scenarios where the media's format remains constant throughout streaming. Unfortunately, this is not the case while using certain browsers as peers with ([AWS KVS JS SDK](https://awslabs.github.io/amazon-kinesis-video-streams-webrtc-sdk-js/examples/index.html)), so receiving media from the browsers andf writing it to filesink via MatroskaMux is not supported by our existing GStreamer samples. When the media's format changes mid-streaming (referred to as "caps changes"), MatroskaMux encounters limitations and is unable to handle these changes, resulting in an error message like:
```
disa6302: > so receiving media from the browsers and writing it to filesink via MatroskaMux is not supported by our existing GStreamer samples

Let us reword to state that there could be an issue if media caps changed mid-streaming, in which case the application could encounter `matroskamux matroska-mux.c:1134:gst_matroska_mux_video_pad_setcaps:<mux> error: Caps changes are not supported by Matroska` error. Let us not mention browsers specifically since it could happen with really anything.

niyatim23: Done! Thanks

samples/kvsWebRTCClientMasterGstSample.c
```diff
@@ -326,8 +326,6 @@ INT32 main(INT32 argc, CHAR* argv[])
 
     pSampleConfiguration->videoSource = sendGstreamerAudioVideo;
     pSampleConfiguration->mediaType = SAMPLE_STREAMING_VIDEO_ONLY;
-    pSampleConfiguration->receiveAudioVideoSource = receiveGstreamerAudioVideo;
```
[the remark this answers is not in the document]
niyatim23: Done. Don't know how I forgot this

samples/kvsWebRTCClientMasterGstSample.c
```diff
@@ -326,7 +326,7 @@ INT32 main(INT32 argc, CHAR* argv[])
...
+    pSampleConfiguration->receiveAudioVideoSource = NULL;
```
stefankiesz, line 329: Can we move the null initialization into the `createSampleConfiguration` function

stefankiesz (changes requested):

samples/kvsWebRTCClientMasterGstSample.c
niyatim23: If we do that, we would be changing it for other samples that use this as well and we don't want to do that is why we are doing this outside

stefankiesz: Don't they all set it after invoking create though?

niyatim23: Yes, but they set it to different receivers

disa6302 (dismissed):

stefankiesz (approved):

samples/GstAudioVideoReceiver.c
```diff
@@ -16,22 +19,28 @@ VOID onGstVideoFrameReady(UINT64 customData, PFrame pFrame)
...
-    CHK_ERR(buffer != NULL, STATUS_NULL_ARG, "Buffer allocation failed");
+    if (!eos) {
+        buffer = gst_buffer_new_allocate(NULL, pFrame->size, NULL);
+        CHK_ERR(buffer != NULL, STATUS_NULL_ARG, "Buffer allocation failed");
+
+        DLOGV("Video frame size: %d, presentationTs: %llu", pFrame->size, presentationTsIncrement);
+
+        GST_BUFFER_DTS(buffer) = presentationTsIncrement;
+        GST_BUFFER_PTS(buffer) = presentationTsIncrement;
+        GST_BUFFER_DURATION(buffer) = gst_util_uint64_scale(1, GST_SECOND, DEFAULT_FPS_VALUE);
```
sirknightj, line 30: Can you add some comments explaining what this does?

disa6302 (approved):

PR #1975, opened 2024-04-29
Add RTP component as submodule after re-factoring 

moninom1:
*Issue #, if available:*

*What was changed?*
- Add [RTP component](https://github.com/awslabs/amazon-kinesis-video-streams-rtp) as external project, and update RTP code to call APIs of component.

*Why was it changed?*
- To apply RTP re-factoring component into dev branch. [RTP component](https://github.com/awslabs/amazon-kinesis-video-streams-rtp) is separated out as independent library.

*How was it changed?*
- Add [RTP component](https://github.com/awslabs/amazon-kinesis-video-streams-rtp) as external project, import it via ExternalProject_Add, and update RTP code to call APIs of component.

- Verified master demo on Ubuntu
- Verified serialization & deserialization with webrtc_client_test and pass all test cases.
- Verified packetization & depacketization with unit test framework in [RTP component](https://github.com/awslabs/amazon-kinesis-video-streams-rtp) 


src/source/Rtp/RtpPacket.c
```diff
@@ -246,9 +261,13 @@ STATUS setBytesFromRtpPacket(PRtpPacket pRtpPacket, PBYTE pRawPacket, UINT32 pac
     ENTERS();
     STATUS retStatus = STATUS_SUCCESS;
     PRtpPacketHeader pHeader = &pRtpPacket->header;
-    UINT32 packetLengthNeeded = 0;
-    PBYTE pCurPtr = pRawPacket;
-    UINT8 i;
+    size_t packetLengthNeeded = 0;
```
hassanctech, line 264: Use `SIZE_T`

moninom1: I used SIZE_T earlier but seem like in PIC the correct type for macOS is not defined. Hence it was failing in MAC CI builds
error: incompatible pointer types passing 'SIZE_T *' (aka 'unsigned long long *') to parameter of type 'size_t *' (aka 'unsigned long *')

hassanctech: Ah I see the problem. On MacOS `size_t` is `unsigned long long` on gcc and `unsigned long` with clang compiler.  Even though the failing job says in the title gcc, if you look at the output it's actually using clang.  We map SIZE_T to unsigned long long , hence it if any of the following macros are defined

```
#if defined __GNUC__ || defined __GNUG__
#if defined __x86_64__ || defined __ppc64__ || __aarch64__
```

clang also defines `__GNUC__`, I will fix that and post the update, I'm surprised this did not fail in PIC tests.

moninom1: Should we merge this PR? Or wait for your fix?

hassanctech (approved):
nice work !!

src/source/Rtp/RtpPacket.c
hassanctech: Yeah let's not block this, looks good, I've approved.

disa6302 (approved):

src/source/Rtp/RtpPacket.c
```diff
@@ -168,45 +169,59 @@ STATUS setRtpPacketFromBytes(PBYTE rawPacket, UINT32 packetLength, PRtpPacket pR
...
     UINT32 ssrc;
     PUINT32 csrcArray = NULL;
+    PUINT32 extensionPayloadWord = NULL;
     UINT16 extensionProfile = 0;
     UINT16 extensionLength = 0;
     PBYTE extensionPayload = NULL;
-    UINT32 currOffset = 0;
-
-    CHK(pRtpPacket != NULL && rawPacket != NULL, STATUS_NULL_ARG);
+    UINT32 word;
```
shubnil, line 180: I know this is a bit of knit picking. I think it will be good if we initialize all or none. Here some of the variables are initialized and some are not. Please do it in the other functions as well.

moninom1: I think webrtc SDK is not following the rule to initialising all the variables, hence only the variables whose values are required was initialised and others are initialised before use (the new variables). Hence i will keep it as it is for now, as i do not want to update existing variables.

PR #1977, opened 2024-04-30
Static build prodc

disa6302:
*Issue #, if available:*

*What was changed?*
- Fixed static building issue of SDK with mbedtls encountered on Ubuntu and added a CI for the same.

*Why was it changed?*
- Without this change, it seems the `libkvsCommonLws` library generated was a shared one despite setting static flag. This caused failure (complaining about compiling a shared version of libmbedtls) when trying to link mbedtls static library to shared `libkvsCommonLws` causing build failure, which surfaced this discrepancy. Same issue is not encountered with openssl, however, if static flag is set, we should make sure producer C is built static too.

*How was it changed?*
- Passing in `BUILD_STATIC_LIBS` setting to producer C.
- Added a new build only CI to build mbedtls static on ubuntu.

- Tested on Ubuntu 22.04 EC2 instance and added CI for the same.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -351,6 +351,41 @@ jobs:
         run:  |
           cd build
           timeout --signal=SIGABRT 60m ./tst/webrtc_client_test
+
+  mbedtls-ubuntu-gcc-static:
+    runs-on: ubuntu-latest
+    env:
+      AWS_KVS_LOG_LEVEL: 2
+    permissions:
+      id-token: write
+      contents: read
+    steps:
+      - name: Clone repository
+        uses: actions/checkout@v4
+      - name: Configure AWS Credentials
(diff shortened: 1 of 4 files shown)


.github/workflows/ci.yml
```diff
@@ -351,6 +350,39 @@ jobs:
         run:  |
           cd build
           timeout --signal=SIGABRT 60m ./tst/webrtc_client_test
+
+  mbedtls-ubuntu-gcc-static:
```
sirknightj, line 354: This just verifies that it builds, but I don't see any checks (e.g. objdump) to verify that the binary is _actually_ static or whether it is able to run the binary.

disa6302: The build fails. That is the verification. With mbedtls, if the library types are mixed on ubuntu, the build is not successful.

PR #1985, opened 2024-05-10
CMake: Allow to pass CFLAGS into a build dependency, when using a emb…

thomas-roos:
CMake: Allow to pass CFLAGS into a build dependency, when using a embedded linux sdk

This is because env CMAKE_TOOLCHAIN_FILE is pointing to a OEToolchainConfig.cmake file doing: set( CMAKE_C_FLAGS $ENV{CFLAGS} CACHE STRING "" FORCE )

*Issue #, if available:*

*What was changed?*

*Why was it changed?*

*How was it changed?*

### DIFF
--- a/CMake/Utilities.cmake
+++ b/CMake/Utilities.cmake
@@ -43,6 +43,14 @@ function(build_dependency lib_name)
     return()
   endif()
 
+  message(STATUS "CFLAGS environment variable before setting to CMAKE_C_FLAGS: $ENV{CFLAGS}")
+  # save original cflags to reset after build is finished for this build dependency
+  set(ORIG_CFLAGS $ENV{CFLAGS})
+
+  # set CFLAGS env variable since this is the only way to pass CFLAGS into a build dependency if a cross compile sdk is used (eg. Yocto)
+  set(ENV{CFLAGS} ${CMAKE_C_FLAGS})
+  message(STATUS "CFLAGS after setting to CMAKE_C_FLAGS: $ENV{CFLAGS}")
+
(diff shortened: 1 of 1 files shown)


thomas-roos:
Would still love to see this merged. As described `set CFLAGS env variable since this is the only way to pass CFLAGS into a build dependency if a cross compile sdk is used (eg. Yocto)`

sirknightj:
I'm wondering about where in the CMakeLists for the Dependencies, there are sometimes extra flags added.

https://github.com/awslabs/amazon-kinesis-video-streams-webrtc-sdk-c/blob/08549862075603242c269f3fcbf602403203d424/CMake/Dependencies/libusrsctp-CMakeLists.txt#L14

https://github.com/awslabs/amazon-kinesis-video-streams-webrtc-sdk-c/blob/08549862075603242c269f3fcbf602403203d424/CMake/Dependencies/libkvsCommonLws-CMakeLists.txt#L25

With the current changes, it would set the `CFLAGS` without these variables (which probably won't work for all use cases).

And in `libsrtp`, it seems the SDK explicitly sets `CFLAGS` for `autoconf`:

https://github.com/awslabs/amazon-kinesis-video-streams-webrtc-sdk-c/blob/cf817bc5d18f3e4bd499c6b0f9a68c6f4d7e01de/CMake/Dependencies/libsrtp-CMakeLists.txt#L20-L21

There might be some in OpenSSL as well.

----

I would suggest adding a conditional (w/ cmake option) for this feature. Let me know if you have other thoughts. Thanks!

thomas-roos:
Closing, when needed working on a better solution, as you figured this will not be compatible with the way it is actually used. Probably.

PR #2000, opened 2024-05-29
Fix MID_NTP macro definition

aggarg:
*Issue #, if available:*

*What was changed?*
Updated the MID_NTP macro definition to use the parameter passed to it.

*Why was it changed?*
The macro definition was not using the parameter passed to it. It was working so far because the variable name used in the macro definition happened to be same as the one at the call site.

*How was it changed?*
Updated the macro definition to use the parameter passed.

The change is pretty small and does not warrant any testing. 


hassanctech (approved):
nice catch!

moninom1 (approved):

PR #2013, opened 2024-06-14
Cloudwatch integration demo

disa6302:
*Issue #, if available:*

*What was changed?*

*Why was it changed?*

*How was it changed?*


cloudwatch-integ/kvsWebRTCClientViewerCloudwatch.cpp
```diff
@@ -0,0 +1,374 @@
+#include <aws/core/Aws.h>
+#include "../samples/Samples.h"
+#include "Cloudwatch.h"
+
+extern PSampleConfiguration gSampleConfiguration;
+
+#ifdef ENABLE_DATA_CHANNEL
+// onMessage callback for a message received by the viewer on a data channel
+VOID dataChannelOnMessageCallback(UINT64 customData, PRtcDataChannel pDataChannel, BOOL isBinary, PBYTE pMessage, UINT32 pMessageLen)
```
niyatim23, line 9: Can we move these to `lib/DataChannelHandling.c`?

cloudwatch-integ/kvsWebRTCClientViewerCloudwatch.cpp
```diff
@@ -0,0 +1,374 @@
...
+    if (isBinary) {
+        DLOGI("DataChannel Binary Message");
+    } else {
+        DLOGI("DataChannel String Message: %.*s", pMessageLen, pMessage);
+    }
+}
+
+// onOpen callback for the onOpen event of a viewer created data channel
+VOID dataChannelOnOpenCallback(UINT64 customData, PRtcDataChannel pDataChannel)
+{
```
niyatim23, line 22: Can we move these to `lib/DataChannelHandling.c`?

samples/kvsWebRTCClientMaster.c
```diff
@@ -139,23 +139,6 @@ INT32 main(INT32 argc, CHAR* argv[])
...
-CleanUp:
-
-    if (pSize != NULL) {
-        *pSize = (UINT32) size;
-    }
-
-    return retStatus;
-}
-
 PVOID sendVideoPackets(PVOID args)
```
niyatim23, line 142: Can we also move this to `lib/Media.c`?

disa6302: These are sample specific options customers might want to set up. So moving it to `lib` might not make sense. Any definition which is very specific to customer use case like these - it makes sense to let them be in the sample itself.

cloudwatch-integ/kvsWebRTCClientViewerCloudwatch.cpp
disa6302: I am inclined to letting this be in the samples since viewer is creating the data channel and setting the operation on incoming data. Let me know if you see a strong reason to move this into lib/

cloudwatch-integ/CloudwatchMonitoring.cpp
```diff
@@ -0,0 +1,409 @@
...
+
+    datum.SetMetricName("ExitStatus");
+    datum.SetValue(1.0);
+    datum.SetUnit(Aws::CloudWatch::Model::StandardUnit::Count);
+
+    datum.AddDimensions(statusDimension);
+
+    this->push(datum);
+}
+
```
niyatim23, line 139: Is there a reason `VOID CloudwatchMonitoring::pushSignalingRoundtripStatus(STATUS retStatus)`, `VOID CloudwatchMonitoring::pushSignalingRoundtripLatency(UINT64 delay, Aws::CloudWatch::Model::StandardUnit unit)`, `VOID CloudwatchMonitoring::pushSignalingConnectionDuration(UINT64 duration, Aws::CloudWatch::Model::StandardUnit unit)`, `VOID CloudwatchMonitoring::pushSignalingInitDelay(UINT64 delay, Aws::CloudWatch::Model::StandardUnit unit)`, `VOID CloudwatchMonitoring::pushICEHolePunchingDelay(UINT64 delay, Aws::CloudWatch::Model::StandardUnit unit)` are not included here but are in the canary?

scripts/generate-iot-credential.sh
```diff
@@ -65,43 +67,23 @@ cat > iot-policy-document.json <<EOF
...
-curl --silent 'https://www.amazontrust.com/repository/SFSRootCAG2.pem' --output cacert.pem
-
-
-export AWS_IOT_CORE_CREDENTIAL_ENDPOINT=$(cat iot-credential-provider.txt)
-export AWS_IOT_CORE_CERT=$(pwd)"/"$iotCert
-export AWS_IOT_CORE_PRIVATE_KEY=$(pwd)"/"$iotPrivateKey
-export AWS_IOT_CORE_ROLE_ALIAS=$iotRoleAlias
-export AWS_IOT_CORE_THING_NAME=$thingName
-        
+aws --profile default iot create-policy --policy-name $iotPolicyName --policy-document 'file://iot-policy-document.json'
```
niyatim23, line 87: The following lines are commented out but present in this file in the demos repository:
```
# Step 3: Create and Configure the X.509 Certificate
# Create the certificate to which you must attach the policy for IoT that you created above.
# aws --profile default  iot create-keys-and-certificate --set-as-active --certificate-pem-outfile $iotCert --public-key-outfile $iotPublicKey --private-key-outfile $iotPrivateKey > certificate
# # Attach the policy for IoT (KvsCameraIoTPolicy created above) to this certificate.
# aws --profile default  iot attach-policy --policy-name $iotPolicyName --target $(jq --raw-output '.certificateArn' certificate)
# # Attach your IoT thing (kvs_example_camera_stream) to the certificate you just created:
# aws --profile default  iot attach-thing-principal --thing-name $thingName --principal $(jq --raw-output '.certificateArn' certificate)
# # In order to authorize requests through the IoT credentials provider, you need the IoT credentials endpoint which is unique to your AWS account ID. You can use the following command to get the IoT credentials endpoint.
# aws --profile default  iot describe-endpoint --endpoint-type iot:CredentialProvider --output text > iot-credential-provider.txt
# # In addition to the X.509 cerficiate created above, you must also have a CA certificate to establish trust with the back-end service through TLS. You can get the CA certificate using the following command:
# curl --silent 'https://www.amazontrust.com/repository/SFSRootCAG2.pem' --output cacert.pem
```
Is there a reason we excluded it from here?

cloudwatch-integ/kvsWebRTCClientMasterCloudwatch.cpp
```diff
@@ -0,0 +1,521 @@
...
+        CHK_STATUS(setUpCredentialProvider(pSampleConfiguration, USE_IOT));
+
+        pSampleConfiguration->forceTurn = FORCE_TURN_ONLY;
+        pSampleConfiguration->enableMetrics = ENABLE_METRICS;
+        pSampleConfiguration->channelInfo.useMediaStorage = ENABLE_STORAGE;
+        pSampleConfiguration->audioCodec = AUDIO_CODEC;
+        pSampleConfiguration->videoCodec = VIDEO_CODEC;
+
+        if(pSampleConfiguration->channelInfo.useMediaStorage) {
+            pSampleConfiguration->audioSource = sendRealAudioPackets;
```
niyatim23, line 419: What's the reason for using real source only when storage is enabled?

samples/lib/Common.c
```diff
@@ -0,0 +1,980 @@
...
+{
+    UNUSED_PARAM(sigNum);
+    if (gSampleConfiguration != NULL) {
+        ATOMIC_STORE_BOOL(&gSampleConfiguration->interrupted, TRUE);
+        CVAR_BROADCAST(gSampleConfiguration->cvar);
+    }
+}
+
+STATUS signalingCallFailed(STATUS status)
+{
```
niyatim23, line 30: If this isn't expected to be modified by customers later, would it be more appropriate to move it to `SignalingMsgHandler` and rename the file to something else?

cloudwatch-integ/kvsWebRTCClientViewerCloudwatch.cpp
niyatim23: I suggested that because I thought the purpose to have lib/ was to modularize Common.c but I see that it is to retain the pieces used by the customer as well

cloudwatch-integ/CloudwatchMonitoring.cpp
disa6302: There isnt going to be a signaling specific canary. These are integrated in profiling metrics now.

scripts/generate-iot-credential.sh
disa6302: Moved to a new file (cert_setup.sh). Retained these as comments during dev and never removed it.

cloudwatch-integ/kvsWebRTCClientMasterCloudwatch.cpp
disa6302: Media server backend does not allow dummy frames.

samples/lib/Common.c
disa6302: Retained here since it isnt a callback. SignalingMsgHandler is specific to SDP message handling.

PR #2018, opened 2024-06-25
v1.11.0 Release

disa6302:
*Issue #, if available:*

*What was changed?*
- New release removing `addConfigToServerList` API that was introduced in 1.10.0.

*Why was it changed?*
- To avoid enforcing samples to use this API to generate relay candidates. 

Additional summary of changes:

Commit Hash | Description
-- | --
4cd8eed6fa388d6d1d6596e0bd396cd402f08ae0 | Edit to CMakeLists to properly build static version of executable and libs
6bc0d2dc4fd4de585117e3c1c01dbb188f6fbe2e | 1.  Update mbedtls from 2.25.0 to 2.28.5 2. Properly INSTALL threadpool libs 3. Remove checkIceAgentStateMachine and checkTurnConnectionStateMachine APIs (which in where the `timerQueueKick` calls were being made.
4b5ed4749f4ccc6abdcf07f2c6e7e5808cecd8cd | Update mbedtls to 2.28.8 since it is the latest in the 2.28 LTS track
e40fab5f8c38a34abade9ce6c2c161b60c030fa5 | Fixes build for mips32 on old compilers [was necessary for gcc 5.4 cross toolchain with uclibc], being tracked in GitHub lws repo as well but this is the current workaround that works:  https://github.com/warmcat/libwebsockets/issues/3163
ad2922a492b274845f14366c9af6f4d6db0378d6 | Propagate the mbedtls found versions to lws cmake (as suggested by lws project itself) necssary when you have a local install as well and that one gets picked up first.  Also added cmake command line argument to specify CA CERT PATH (but this didn't work, fixed in next commit below).
38e942c835fba43993859dd857bcecbfa09f26ad | Properly add -DKVS_CA_CERT_PATH argument for cmake so a custom one can be specified (all embedded devices will need this), tested on Ingenic T31

*How was it changed?*
- Reverted the commits that introduced this change. Restored unit tests too.

- Ran samples under different config to confirm nothing is broken.
- Unit test


CMakeLists.txt
```diff
@@ -163,6 +167,17 @@ if(BUILD_DEPENDENCIES)
       build_dependency(mbedtls ${BUILD_ARGS})
   endif()
 
+  # This step is necessary because the next set of dependencies have a dependency on
+  # mbedtls/openssl and the find_package command  populates certain cmake variables which
+  # are needed to pass to the cmake commands for the next set of dependencies
+  if (USE_OPENSSL)
```
stefankiesz, line 175: We should be able to remove the existing `find_package` calls on the crypto libs that start down on line 262 right?

.github/workflows/ci.yml
```diff
@@ -5,6 +5,7 @@ on:
     branches:
       - develop
       - master
+      - 1.10.3
```
stefankiesz, line 8: I think need to revert this before merging.

.github/workflows/codecov.yml
```diff
@@ -4,6 +4,7 @@ on:
     branches:
       - develop
       - master
+      - 1.10.3
```
stefankiesz, line 7: I think need to revert this before merging.

.github/workflows/codeql-analysis.yml
```diff
@@ -2,7 +2,8 @@ name: "CodeQL"
 
 on:
   push:
-    branches: [ develop, master ]
+    branches: [ develop, master, 1.10.3 ]
```
stefankiesz, line 5: I think need to revert this before merging.

CMake/Dependencies/libkvsCommonLws-CMakeLists.txt
```diff
@@ -6,10 +6,12 @@ include(ExternalProject)
 
 ExternalProject_Add(libkvsCommonLws-download
     GIT_REPOSITORY    https://github.com/awslabs/amazon-kinesis-video-streams-producer-c.git
-    GIT_TAG           v1.5.2
+    GIT_TAG           link-mbedtls-properly
```
stefankiesz, line 9: Should merge this fix on Producer C SDK and change this to master before we merge this into master.

stefankiesz, line 9:
```diff
target_link_libraries(kvsCommonLws
+           ${PRODUCER_CRYPTO_LIBRARY}
            ${OPENSSL_CRYPTO_LIBRARY}
            ${OPENSSL_SSL_LIBRARY}
            ${LIBWEBSOCKETS_LIBRARIES}
```

Looking at the cmake 1-line change in this Producer C branch, its seems we should only link the `${OPENSSL_CRYPTO_LIBRARY} ${OPENSSL_SSL_LIBRARY}` libraries if not opting for MbedTLS - or better yet only have the `PRODUCER_CRYPTO_LIBRARY` one which should cover both cases:

```
set(PRODUCER_CRYPTO_LIBRARY
        OpenSSL::Crypto
        OpenSSL::SSL)
if (USE_MBEDTLS)
    set(CPRODUCER_COMMON_TLS_OPTION KVS_USE_MBEDTLS)
    set(PRODUCER_CRYPTO_LIBRARY
            MbedTLS
            MbedCrypto)
endif()
```

So I propose the following change:

```diff
target_link_libraries(kvsCommonLws
            ${PRODUCER_CRYPTO_LIBRARY}
-           ${OPENSSL_CRYPTO_LIBRARY}
-           ${OPENSSL_SSL_LIBRARY}
            ${LIBWEBSOCKETS_LIBRARIES}
```
And with a nit variable name change that makes more sense as there are multiple libraries and to be more aligned with how we also named `LIBWEBSOCKETS_LIBRARIES`:

```diff
target_link_libraries(kvsCommonLws
-           ${PRODUCER_CRYPTO_LIBRARY}
+           ${PRODUCER_CRYPTO_LIBRARIES}
            ${LIBWEBSOCKETS_LIBRARIES}
```

CMakeLists.txt
```diff
@@ -379,7 +399,7 @@ endif()
 if(ENABLE_KVS_THREADPOOL)
   file(GLOB THREADPOOL_SOURCE_FILES "src/source/Threadpool/*.c")
   add_library(kvsWebRtcThreadpool ${LINKAGE} ${THREADPOOL_SOURCE_FILES})
-  target_link_libraries(kvsWebRtcThreadpool PRIVATE kvspicUtils)
+  target_link_libraries(kvsWebRtcThreadpool PUBLIC kvspicUtils)
```
stefankiesz, line 402: All consuming libraries of **kvsWebRtcThreadpool** (A.K.A. `EXTRA_DEPS`) already link with **kvspicUtils**, so seems PRIVATE linking would be more appropriate here since we don not need to expose the consuming libraries to **kvspicUtils** _through_ kvsWebRtcThreadpool. Is there another reason this change was made?

stefankiesz (commented):
Reviewed the build changes portion of the PR. Will do WebRTC changes next.

CMake/Dependencies/libkvsCommonLws-CMakeLists.txt
cheersalam: Want to highlight that `link-mbedtls-properly` branch is temporary. We need to do a release and update tag here again

CMake/Dependencies/libmbedtls-CMakeLists.txt
```diff
@@ -21,7 +21,7 @@ message(STATUS "C flags here are ${CMAKE_C_FLAGS}")
 ExternalProject_Add(
   project_libmbedtls
   GIT_REPOSITORY  https://github.com/ARMmbed/mbedtls.git
-  GIT_TAG         v2.25.0
+  GIT_TAG         v2.28.8
```
cheersalam, line 24: Latest is 3.6.1. Why 2.28.8?

CMakeLists.txt
```diff
@@ -102,7 +102,13 @@ message(STATUS "Kinesis Video WebRTC Client path is ${KINESIS_VIDEO_WEBRTC_CLIEN
 message(STATUS "dependencies install path is ${OPEN_SRC_INSTALL_PREFIX}")
 
 # pass ca cert location to sdk
-add_definitions(-DKVS_CA_CERT_PATH="${CMAKE_SOURCE_DIR}/certs/cert.pem")
+
+if(NOT DEFINED KVS_CA_CERT_PATH)
```
cheersalam, line 106: How was it working, before this addition?

samples/Common.c
```diff
@@ -345,51 +345,13 @@ VOID onIceCandidateHandler(UINT64 customData, PCHAR candidateJson)
...
-}
-
 STATUS initializePeerConnection(PSampleConfiguration pSampleConfiguration, PRtcPeerConnection* ppRtcPeerConnection)
 {
     ENTERS();
     STATUS retStatus = STATUS_SUCCESS;
     RtcConfiguration configuration;
-#ifndef ENABLE_KVS_THREADPOOL
-    UINT32 i, j, maxTurnServer = 1;
+    UINT32 i, j, iceConfigCount, uriCount = 0, maxTurnServer = 1;
```
cheersalam, line 353: nit:
maxTurnServer = 1; Should be a macro or const static

src/source/Ice/TurnConnectionStateMachine.c
```diff
@@ -60,30 +60,6 @@ PCHAR turnConnectionGetStateStr(UINT64 state)
...
-    BOOL transitionReady = FALSE;
-
-    CHK(pTurnConnection != NULL && pTurnConnection->pStateMachine != NULL, STATUS_NULL_ARG);
-
-    // if a state transition is ready, tell the timer to kick the timer
-    CHK_STATUS(checkForStateTransition(pTurnConnection->pStateMachine, &transitionReady));
-
-    if (transitionReady) {
-        // dangerous to have any mutexes locked by timerqueue when entering this function
-        CHK_STATUS(timerQueueKick(pTurnConnection->timerQueueHandle, pTurnConnection->timerCallbackId));
```
cheersalam, line 76: Is this Kick added to some other place?

CMake/Dependencies/libmbedtls-CMakeLists.txt
hassanctech: We never tested with the latest major version bump, but 2.28.x is an LTS track so I updated to the latest.

CMakeLists.txt
hassanctech: Customers could specify an environment variable as well, or they would make the change in the CMakeLists themselves manually, or when you're running from the sample place it was built, the path will be right.

src/source/Ice/TurnConnectionStateMachine.c
hassanctech: No it's been removed from all the places where it was.

CMake/Dependencies/libkvsCommonLws-CMakeLists.txt
hassanctech: Right

.github/workflows/codeql-analysis.yml
hassanctech: Eventually this will go into main and will need to be rebased against main anyway there it's develop and main only.  But yeah 1.10.3 needs to be removed from here.

CMakeLists.txt
hassanctech: No because this is in the if build dependencies section, so this needs to run in both cases and since it's idempotent I kept it there.

CMake/Dependencies/libkvsCommonLws-CMakeLists.txt
hassanctech: Yes I thought of this as well, I have no idea why it was set up the way it is, however the openssl cmake variables will be blank in this case so it will be ignored by the target link libraries (I verified this).  Preference though would be to do as you suggested as the final change.

src/source/Ice/TurnConnectionStateMachine.c
unicornss: if this is the change caused the Axis P3354 not work properly due to the contention-  let  us add the details on the locks/threads?

src/source/Ice/IceAgentStateMachine.c
```diff
@@ -29,28 +29,6 @@ StateMachineState ICE_AGENT_STATE_MACHINE_STATES[] = {
 
 UINT32 ICE_AGENT_STATE_MACHINE_STATE_COUNT = ARRAY_SIZE(ICE_AGENT_STATE_MACHINE_STATES);
 
-STATUS checkIceAgentStateMachine(PIceAgent pIceAgent)
```
sirknightj, line 32: Could you share in the PR description or comment, why these were removed?

CMakeLists.txt
```diff
@@ -105,6 +105,8 @@ message(STATUS "dependencies install path is ${OPEN_SRC_INSTALL_PREFIX}")
 
 if(NOT DEFINED KVS_CA_CERT_PATH)
   add_definitions(-DKVS_CA_CERT_PATH="${CMAKE_SOURCE_DIR}/certs/cert.pem")
+else()
+  add_definitions(-DKVS_CA_CERT_PATH="${KVS_CA_CERT_PATH}")
```
sirknightj, line 109: Note for later: Let's see if we can make this a runtime parameter option as well.

CMake/Dependencies/libwebsockets-CMakeLists.txt
```diff
@@ -38,6 +38,7 @@ ExternalProject_Add(project_libwebsockets
         -DCMAKE_INSTALL_PREFIX=${OPEN_SRC_INSTALL_PREFIX}
         -DCMAKE_C_FLAGS=${CMAKE_C_FLAGS}
         -DCMAKE_BUILD_TYPE=${CMAKE_BUILD_TYPE}
+        -DDISABLE_WERROR=ON
```
sirknightj, line 41: Could you share which error we're seeing here?

Can't we also make this one ON only the failing builds (mac), conditional?

src/source/Ice/TurnConnection.c
```diff
@@ -1033,6 +1033,8 @@ STATUS checkTurnPeerConnections(PTurnConnection pTurnConnection)
     PStunAttributeChannelNumber pStunAttributeChannelNumber = NULL;
     UINT32 i = 0;
 
+    UNUSED_PARAM(sendStatus);
```
sirknightj, line 1035: Can we add a check log error for this line?

The value is not being checked below.

src/source/Sctp/Sctp.c
```diff
@@ -365,6 +365,8 @@ INT32 onSctpInboundPacket(struct socket* sock, union sctp_sockstore addr, PVOID
     if (data != NULL) {
         free(data);
     }
-
+    if (STATUS_FAILED(retStatus)) {
```
sirknightj, line 368: Could you add a log check error here please?

src/source/Signaling/LwsApiCalls.c
```diff
@@ -1493,6 +1494,9 @@ STATUS joinStorageSessionLws(PSignalingClient pSignalingClient, UINT64 time)
     PCHAR pResponseStr;
     UINT32 resultLen;
 
+    UNUSED_PARAM(pResponseStr);
+    UNUSED_PARAM(pLwsCallInfo);
+    UNUSED_PARAM(resultLen);
```
sirknightj, line 1499: These seem like something we should log. Why else are we having these variables be stored locally?

CMakeLists.txt
stefankiesz: Sounds good, might want to consider breaking up the build_deps case to handle the before find_package logic and the after find_package logic.

src/source/Ice/IceAgentStateMachine.c
hassanctech: Done

CMake/Dependencies/libwebsockets-CMakeLists.txt
hassanctech: Added in description, but it's not on mac, this fails on mips gcc older toolchains and I think it's specific with uclibc toolchains (this was required to build in Ingenic T31) with gcc 5.4 uclibc toolchain. Details and link to lws issue has been added to the description.

src/source/Ice/TurnConnection.c
hassanctech: Agree, we should log this.

unicornss (approved):
Approving the timer-queue changes based on the benchmarking in SoCs 

commit/4cd8eed6fa388d6d1d6596e0bd396cd402f08ae0 	 commit/6bc0d2dc4fd4de585117e3c1c01dbb188f6fbe2e 	 commit/4b5ed4749f4ccc6abdcf07f2c6e7e5808cecd8cd 	 
commit/e40fab5f8c38a34abade9ce6c2c161b60c030fa5 	 
commit/ad2922a492b274845f14366c9af6f4d6db0378d6 	 commit/38e942c835fba43993859dd857bcecbfa09f26ad

sirknightj (dismissed):
Approving the commits between
4cd8eed6fa388d6d1d6596e0bd396cd402f08ae0 and
38e942c835fba43993859dd857bcecbfa09f26ad

sirknightj:
This PR includes bumping MBED TLS, which currently does not build previously on my Macbook (https://github.com/awslabs/amazon-kinesis-video-streams-webrtc-sdk-c/tree/master, commit `ec75ba1`)
```
Cloning into 'project_libmbedtls'...
HEAD is now at 1c54b5410 Merge pull request #786 from ARMmbed/dev/yanesca/mbedtls-2.25.0r0-pr
CMake Deprecation Warning at CMakeLists.txt:23 (cmake_minimum_required):
  Compatibility with CMake < 3.5 will be removed from a future version of
  CMake.

  Update the VERSION argument <min> value or use a ...<max> suffix to tell
  CMake that the project does not need compatibility with older versions.


/work/open-source/libmbedtls/build/src/project_libmbedtls/library/bignum.c:1547:29: error: variable 't' set but not used [-Werror,-Wunused-but-set-variable]
    mbedtls_mpi_uint c = 0, t = 0;
                            ^
1 error generated.
gmake[5]: *** [library/CMakeFiles/mbedcrypto.dir/build.make:174: library/CMakeFiles/mbedcrypto.dir/bignum.c.o] Error 1
gmake[4]: *** [CMakeFiles/Makefile2:166: library/CMakeFiles/mbedcrypto.dir/all] Error 2
gmake[3]: *** [Makefile:136: all] Error 2
gmake[2]: *** [CMakeFiles/project_libmbedtls.dir/build.make:87: build/src/project_libmbedtls-stamp/project_libmbedtls-build] Error 2
gmake[1]: *** [CMakeFiles/Makefile2:83: CMakeFiles/project_libmbedtls.dir/all] Error 2
gmake: *** [Makefile:91: all] Error 2
CMake Error at CMake/Utilities.cmake:75 (message):
  CMake step for libmbedtls failed: 2
Call Stack (most recent call first):
  CMakeLists.txt:163 (build_dependency)
```

After trying out this PR, it builds successfully.
```
2024-09-20 08:12:18,351 [INFO] [100%] Built target project_libmbedtls
```

.github/workflows/ci.yml
```diff
@@ -5,7 +5,6 @@ on:
     branches:
       - develop
       - master
```
sirknightj, line 7: These should be main now instead of master

sirknightj: Fixed in #2059

stefankiesz (approved):

sirknightj (approved):

unicornss (approved):
Release based on the performance tests review.

Issue #2025, opened 2024-07-05
[Bug]: Wrong SDP answer when the codec doesn't match with viewer


Nomidia:
### Please confirm you have already done the following

- [X] I have searched the repository for related/existing bug reports
- [X] I have all the details the issue requires

### Please answer the following prompt

- [X] This issue is replicable using the unmodified sample application

### Describe the bug

I switch to the latest version, and set the video codec to h265. When the viewer doesn't not support H264, the device will create a fake transceiver. But it creates a wrong media attribute of rtpmap.

### Expected Behavior

It create a valid sdp answer.

### Current Behavior

 Here is is the SDP answer:
m=video 9 UDP/TLS/RTP/SAVPF 96
c=IN IP4 127.0.0.1
a=candidate:2 1 udp 2130706431 192.168.2.106 60377 typ host raddr 0.0.0.0 rport 0 generation 0 network-cost 999
a=msid:fakeStream fakeTrack
a=ssrc:806117966 cname:QH4BKBHw9jx3IknE
a=ssrc:806117966 msid:fakeStream fakeTrack
a=ssrc:806117966 mslabel:fakeStream
a=ssrc:806117966 label:fakeTrack
a=rtcp:9 IN IP4 0.0.0.0
a=ice-ufrag:n376
a=ice-pwd:c+R0wbpkxp/O9ny1pB427UyH
a=ice-options:trickle
a=fingerprint:sha-256 A6:93:AC:8B:13:93:BC:7F:D4:6A:B2:75:AD:B6:18:A2:0A:DA:14:1E:96:B1:15:CF:84:69:DC:1B:5C:D7:76:1F
a=setup:active
a=mid:1
a=inactive
a=rtcp-mux
a=rtcp-rsize
a=rtpmap:4294967295 VP8/90000
a=ssrc:806117966 cname:QH4BKBHw9jx3IknE
a=ssrc:806117966 msid:fakeStream fakeTrack
a=rtcp-fb:4294967295 goog-remb
a=rtcp-fb:4294967295 transport-cc

The m-line is 96, but rtpmap is 4294967295, 4294967295 = 0xffffffff.
In populateSingleMediaSection, after hashTableGet, the value of payloadType is been modified.

[...]

PR #2033, opened 2024-07-25
fix bug 32 bit system reading 8 bytes for pointer results in overwriting 4 bytes of payloadType when writing rtpMapValue

hassanctech:
*Issue #, if available:*
https://github.com/awslabs/amazon-kinesis-video-streams-webrtc-sdk-c/issues/2025

*What was changed?*
In the case viewer does not support a codec the master supports on 32 bit issues a bug was fixed which could cause a garbage value to be set in the sdp answer for the rtpmap attribute.

*Why was it changed?*
Customer reported issue, results in incorrect/buggy behavior

*How was it changed?*
The `hashTableGet` fills in a UINT64 so it is expecting 8 bytes, a char* was being cast to this, but a pointer on a 32 bit system is only 4 bytes wide. This was resolved by creating an intermediate 64 bit int to hold the value then cast it back down.

Will request customer to test and confirm.  If I can get my hands on a 32 bit system I will try it out.
Linter that check's the PR description was modified to just use the linux utility `wc`.  Otherwise the current check breaks if you have any special characters in the description, since the description supports markdown there are many characters that when present would previously report a 0 char count when in fact they were not 0.  I tested the commands in my shell to confirm.

### DIFF
--- a/src/source/PeerConnection/SessionDescription.c
+++ b/src/source/PeerConnection/SessionDescription.c
@@ -415,13 +415,13 @@ STATUS populateSingleMediaSection(PKvsPeerConnection pKvsPeerConnection, PKvsRtp
 {
     ENTERS();
     STATUS retStatus = STATUS_SUCCESS;
-    UINT64 payloadType, rtxPayloadType;
+    UINT64 payloadType, rtxPayloadType, rtpMapValue;
     BOOL containRtx = FALSE;
     BOOL directionFound = FALSE;
     UINT32 i, remoteAttributeCount, attributeCount = 0;
     PRtcMediaStreamTrack pRtcMediaStreamTrack = &(pKvsRtpTransceiver->sender.track);
     PSdpMediaDescription pSdpMediaDescriptionRemote;
-    PCHAR currentFmtp = NULL, rtpMapValue = NULL;
+    PCHAR currentFmtp = NULL, rtpMapValueChar = NULL;
     CHAR remoteSdpAttributeValue[MAX_SDP_ATTRIBUTE_VALUE_LENGTH];
     INT32 amountWritten = 0;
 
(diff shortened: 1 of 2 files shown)


src/source/PeerConnection/SessionDescription.c
```diff
@@ -795,11 +795,12 @@ STATUS populateSingleMediaSection(PKvsPeerConnection pKvsPeerConnection, PKvsRtp
             attributeCount++;
         }
     } else if (pRtcMediaStreamTrack->codec == RTC_CODEC_UNKNOWN) {
-        CHK_STATUS(hashTableGet(pUnknownCodecRtpmapTable, unknownCodecHashTableKey, (PUINT64) &rtpMapValue));
+        CHK_STATUS(hashTableGet(pUnknownCodecRtpmapTable, unknownCodecHashTableKey, &rtpMapValue));
```
stefankiesz, line 798: Not fully understanding this. If a char* and a UINT64* are both the same size of 32 bits on 32-bit systems, and a char is still just 1 byte on both 32 and 64 bit systems, then why would we expect different behavior between the two systems?

stefankiesz (approved):
Nice, lgtm! Please confirm that lint changes do not break functionality and add how they were tested in the description.

src/source/PeerConnection/SessionDescription.c
hassanctech: What is stated there is correct.  While a char** is what is passed to the function, inside `hashTableGet` we dereferenced it before setting it (char** and char* are the same size, they are both pointers). The value that's assigned is a char* and it's assigned a UINT64 which is the bug, but since it was cast to a PUINT64 we will attempt (and succeed) to write the full 8 bytes, which will overwrite the half of `payloadType` and produce the garbage value.

niyatim23 (approved):

src/source/PeerConnection/SessionDescription.c
```diff
@@ -795,11 +795,12 @@ STATUS populateSingleMediaSection(PKvsPeerConnection pKvsPeerConnection, PKvsRtp
...
+        rtpMapValueChar = (PCHAR) rtpMapValue;
         STRCPY(pSdpMediaDescription->sdpAttributes[attributeCount].attributeName, "rtpmap");
         amountWritten =
             SNPRINTF(pSdpMediaDescription->sdpAttributes[attributeCount].attributeValue,
-                     SIZEOF(pSdpMediaDescription->sdpAttributes[attributeCount].attributeValue), "%" PRId64 " %s", payloadType, rtpMapValue);
+                     SIZEOF(pSdpMediaDescription->sdpAttributes[attributeCount].attributeValue), "%" PRId64 " %s", payloadType, rtpMapValueChar);
```
cheersalam, line 803: Why not directly typecast to (PCHAR)?

src/source/PeerConnection/SessionDescription.c
stefankiesz: I agree with that, but isn't it that the only casting being done is to a char** and the char* is just assigned a value?

PR #2053, opened 2024-09-10
IceAgent partial timerqueue shutdown fix and the turnconnection sending data after shutdown

sirknightj:
What was changed?*
1. Stop all three timer queues and shutdown all the turn connections when 'iceAgentShutdown' is called.
2. Add missing error log in 'freeIceAgent'.
3. Stop sending data after turn 'shutdown' was called.

*Why was it changed?*
1. When timer queue cancel is called, sometimes it takes more than the allotted timeout to stop. When this occurs, CHK will jump to cleanup before stopping the other timer queues. Even after 'iceAgentShutdown' was called (via 'freeSampleStreamingSession'), we sometimes saw the ICE candidate connectivity checks continuing. We often see this log line in the logs, which indicates that the partial cleanup occurred:
```log
iceAgentShutdown(): TurnConnection shutdown did not complete within 1 seconds
```

2. To make debugging easier and more robust.
3. Noticed that after 'turnConnectionShutdown' was called, packets were still being sent.

*How was it changed?*
1. [CHK](https://github.com/awslabs/amazon-kinesis-video-streams-pic/blob/65e38dac9b30523d43a57bc009d679e627b58d9a/src/common/include/com/amazonaws/kinesis/video/common/CommonDefs.h#L1157) was changed to [CHK_LOG_ERR](https://github.com/awslabs/amazon-kinesis-video-streams-pic/blob/65e38dac9b30523d43a57bc009d679e627b58d9a/src/common/include/com/amazonaws/kinesis/video/common/CommonDefs.h#L1226), which will ensure that when iceAgentShutdown is called, all of the timer queues will be shutdown and all of the turn connections too.
2. Added LOG_CHK_ERR.
3. In 'turnConnectionSendData', we check the value of the 'stopTurnConnection' sentinel, which is changed to 'true' when it is in the process of shutting down.

* Used our standard benchmarking tools to check for any regressions in peer connection connected rate and time to first frame.
* When run on a constrained device, we saw an increase in the performance for peer connection connected success rate.

### DIFF
--- a/src/source/Ice/TurnConnection.c
+++ b/src/source/Ice/TurnConnection.c
@@ -728,7 +728,7 @@ STATUS turnConnectionSendData(PTurnConnection pTurnConnection, PBYTE pBuf, UINT3
 
     CHK(pTurnConnection != NULL && pDestIp != NULL, STATUS_NULL_ARG);
     CHK(pBuf != NULL && bufLen > 0, STATUS_INVALID_ARG);
-    CHK(!ATOMIC_EXCHANGE_BOOL(&pIceAgent->shutdown, TRUE), retStatus);
+    CHK(!ATOMIC_LOAD_BOOL(&pTurnConnection->stopTurnConnection), STATUS_SUCCESS); // Do not send data after stop
 
     MUTEX_LOCK(pTurnConnection->lock);
     locked = TRUE;


src/source/Ice/IceAgent.c
```diff
@@ -766,18 +768,19 @@ STATUS iceAgentShutdown(PIceAgent pIceAgent)
     CHK(pIceAgent != NULL, STATUS_NULL_ARG);
     CHK(!ATOMIC_EXCHANGE_BOOL(&pIceAgent->shutdown, TRUE), retStatus);
 
+    // Shutdown all the timers during iceAgentShutdown; do not partially shut down timers.
     if (pIceAgent->iceAgentStateTimerTask != MAX_UINT32) {
-        CHK_STATUS(timerQueueCancelTimer(pIceAgent->timerQueueHandle, pIceAgent->iceAgentStateTimerTask, (UINT64) pIceAgent));
+        CHK_LOG_ERR(timerQueueCancelTimer(pIceAgent->timerQueueHandle, pIceAgent->iceAgentStateTimerTask, (UINT64) pIceAgent));
```
hassanctech, line 773: This is good so we do not abort trying to cancel the other timers.  However you should also *set* the `retStatus` so the overall function can still return the ERROR status to the caller to notify it that the `iceAgentShutdown` operation did not fully succeed.

Perhaps something like this:

```
retStatus = timerQueueCancelTimer(pIceAgent->timerQueueHandle, pIceAgent->iceAgentStateTimerTask, (UINT64) pIceAgent));
CHK_LOG_ERR(retStatus)
```

Note that `CHK_LOG_ERR` is a no-op if the retStatus passed to it is a STATUS_SUCCESS.

stefankiesz (dismissed):
Very nice!

src/source/Ice/IceAgent.c
sirknightj: One adjustment to that suggestion is that we'll also need to not set the status back to a success after something took too long, so an additional check will be needed.

PR #2070, opened 2024-10-25
Move offer from stack to heap

sirknightj:
*What was changed?*
- #1915 was rebased to latest `develop`, see the original PR for details.
- The remote description's lifetime is now shortened to only when it's needed. It does not live throughout the entire PeerConnection anymore.
  - In master mode, it lives between `setRemoteDescription` (receive offer) and `createAnswer`
  - In viewer mode, it lives between `setRemoteDescription` (received answer)
  - If for whatever reason, the user does not call the method in charge of freeing this memory, it is cleaned up in `freePeerConnection`.
- SDK-Application API remains unchanged. SDK handles the memory management for the SDP offer.
- Add `CHK_LOG_ERROR` and `ENTERS()` and `LEAVES()` in missing places in PeerConnection.c
- Add missing free's to make sure remote description is freed even when there's a goto cleanup due to an error.
- Add missing null checks to some callocs (failed to allocate due to oom)
- Add additional null checks to some methods.

*Why was it changed?*
- Reduce memory while the peer is connected. After responding to the remote sdp, it is not used anymore.

*How was it changed?*
- PeerConnection struct (SDK internal) stores a pointer to heap instead of it living within the struct.

- Unit tests pass.
- Check valgrind output of master and viewer samples with the JS page: All heap blocks freed -- no leaks are possible.
- Checked the master sample frames (H.264/Opus) on Ubuntu and Mac with the JS viewer and was able to see the video and hear the audio.


unicornss:
Refactor to allocate remote session description on heap.

src/source/PeerConnection/PeerConnection.c
```diff
@@ -876,6 +933,8 @@ PVOID resolveStunIceServerIp(PVOID args)
     }
     releaseHoldOnInstance(pWebRtcClientContext);
     DLOGD("Exiting from stun server IP resolution thread");
+
```
unicornss, line 936: is the log required?

src/source/PeerConnection/PeerConnection.h
```diff
@@ -105,6 +105,8 @@ typedef struct {
 
     MUTEX peerConnectionObjLock;
 
+    // If the local description is an SDP offer.
```
unicornss, line 108: /nit local session description

src/source/PeerConnection/SessionDescription.c
```diff
@@ -428,9 +428,10 @@ STATUS populateSingleMediaSection(PKvsPeerConnection pKvsPeerConnection, PKvsRtp
         CHK_STATUS(hashTableGet(pUnknownCodecPayloadTypesTable, unknownCodecHashTableKey, &payloadType));
     } else {
         CHK_STATUS(hashTableGet(pKvsPeerConnection->pCodecTable, pRtcMediaStreamTrack->codec, &payloadType));
-        currentFmtp = fmtpForPayloadType(payloadType, &(pKvsPeerConnection->remoteSessionDescription));
+        if (!pKvsPeerConnection->isOffer) {
+            currentFmtp = fmtpForPayloadType(payloadType, pRemoteSessionDescription);
```
unicornss, line 432: is non-null check handled?

unicornss (approved):

src/source/PeerConnection/SessionDescription.c
sirknightj: We actually can't because fmtpForPayloadType doesn't return a STATUS. We can add the null check here though.

sirknightj: The null check is already present on line 1059. I will add another one here for redundancy.

sirknightj: It seems like a lot of NULL checks are missing for these internal methods like `fmtpForPayloadType` (these methods are not publicly accessible). I'm going to timebox adding more null checks and we can take that up in a future PR.

src/source/PeerConnection/PeerConnection.c
sirknightj: It looks like this was added for the Debug level, whereas ENTERS and LEAVES are streaming level. I will keep the log there as it shouldn't cause any issues.

src/source/PeerConnection/PeerConnection.c
```diff
@@ -1059,6 +1122,10 @@ STATUS freePeerConnection(PRtcPeerConnection* ppPeerConnection)
         SAFE_MEMFREE(pKvsPeerConnection->pTwccManager);
     }
 
+    // Master mode: this should be freed once answer is created, user might not invoke `createAnswer`
+    // Viewer mode: this should be freed when `setRemoteDescription`, but user might not call it
```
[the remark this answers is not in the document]
sirknightj: I like that

src/source/PeerConnection/PeerConnection.h
sirknightj: Adjusted

src/source/PeerConnection/SessionDescription.c
```diff
@@ -163,6 +165,8 @@ STATUS setPayloadTypesFromOffer(PHashTable codecTable, PHashTable rtxTable, PSes
     PCHAR fmtp;
     UINT64 fmtpScore, bestFmtpScore;
 
+    CHK(codecTable != NULL && rtxTable != NULL && pSessionDescription != NULL, STATUS_NULL_ARG);
```
stefankiesz, line 168: It's not immediately clear whether `codecTable` and `rtxTable` should be null checked here since they are not ever dereferenced in this `setPayloadTypesFromOffer` function. The check for `pSessionDescription` should definitely be here since it is dereferenced here and will segfault if it is null. Disallowing the other two pointers from being null, on the other hand, has the potential for changing intended behavior. I don't see that it would change the behavior in this case, but we might want to consider a less impactful change for this PR.

sirknightj: This is an internal SDK function (not publicly accessible) and also this function is only used in 1 place: in `setRemoteDescription`.

The method uses both hash tables, and is replacing the elements inside of them:

`hashTableContains(codecTable), hashTableGet(codecTable), hashTableUpsert(codecTable), and hashTableUpsert(rtxTable)` 

Even though hashTable* methods have their own null check, I think it doesn't make sense to allow a null hash table as it will bail the method anyway once it calls that method. (`STATUS_NULL_ARG` will be returned regardless if I added this check or not, but this check makes it immediately obvious that we don't want a null hash table.)

We can revisit making the condition stronger in a future PR, as both tables look to have additional requirements which need some more investigating.

stefankiesz: > I think it doesn't make sense to allow a null hash table as it will bail the method anyway once it calls that method

I agree in this case - although it could change the method behavior, the null check is appropriate here.

stefankiesz (approved):

PR #2071, opened 2024-10-25
Add job to run the sample and check for memory leak

sirknightj:
*Issue #, if available:*
N/A

*What was changed?*
- GitHub actions CI. Run the master and viewer samples both with `valgrind `to check for memory leaks. Instead of kill with signal 9, it interrupts the process (signal 2) and also checks for graceful exit.

*Why was it changed?*
- Reduce the number of times I need to manually run this check. Automating this check will also give confidence to future developers that their change did not introduce a memory leak.

*How was it changed?*
- Script overview:
  - Disable logs by setting log level to 7 (silent).
  - Run the master and viewer samples in the background with valgrind. The valgrind output is captured into two different files, one for each.
  - After 30 seconds, interrupt the viewer (disconnect)
  - After 30 more seconds, interrupt the master.
  - Wait for processes to exit & capture their exit status.
  - After 10 seconds, kill the master and viewer processes. Killing the processes will make them exit with non-zero exit status.
  - Check if both exited with status 0 (whether they exit gracefully on their own after interrupt).
  - Check the valgrind output files and verify that no memory leaks are detected.

- Run it on the GitHub actions runner. Take a look at the image: image

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -45,7 +45,7 @@ jobs:
           cmake .. -DBUILD_TEST=TRUE -DCOMPILER_WARNINGS=TRUE
           make
       - name: Run tests
-        run:  |
+        run: |
           cd build
           ./tst/webrtc_client_test
   mac-os-build-gcc:
@@ -71,7 +71,7 @@ jobs:
           cmake .. -DBUILD_TEST=TRUE -DCOMPILER_WARNINGS=TRUE
           make
       - name: Run tests
-        run:  |
+        run: |
           cd build
           ./tst/webrtc_client_test
   mac-os-m1-build-clang:
@@ -100,7 +100,7 @@ jobs:
           sh -c 'cmake .. -DBUILD_TEST=TRUE -DCMAKE_C_COMPILER=$(brew --prefix llvm@15)/bin/clang -DCMAKE_CXX_COMPILER=$(brew --prefix llvm@15)/bin/clang++'
           make
       - name: Run tests
-        run:  |
+        run: |
           cd build
           ./tst/webrtc_client_test
   static-build-mac:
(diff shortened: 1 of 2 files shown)


sirknightj:
Output when I added a memory leak for verification:
image

.github/workflows/ci.yml
```diff
@@ -668,3 +668,78 @@ jobs:
...
+          sudo sh -c 'echo 0 > /proc/sys/net/ipv6/conf/all/disable_ipv6'
+          sudo add-apt-repository -y ppa:ubuntu-toolchain-r/test
+          sudo add-apt-repository 'deb http://archive.ubuntu.com/ubuntu/ jammy main'
+          sudo add-apt-repository 'deb http://archive.ubuntu.com/ubuntu/ jammy universe'
+          sudo apt-get -q update
+          sudo apt-get -y install libcurl4-openssl-dev valgrind
+      - name: Build repository
+        run: |
+          mkdir build && cd build
+          cmake .. -DUSE_OPENSSL=OFF -DUSE_MBEDTLS=ON
```
stefankiesz, line 699: Want to confirm that using MbedTLS rather than the default crypto lib is intentional here.

.github/workflows/ci.yml
```diff
@@ -668,3 +668,78 @@ jobs:
...
+          ) &
+          
+          wait $PID_MASTER
+          EXIT_STATUS_MASTER=$?
+          wait $PID_VIEWER
+          EXIT_STATUS_VIEWER=$?
+          
+          # Check exit statuses to determine if the interrupt was successful
+          if [ $EXIT_STATUS_MASTER -ne 0 ] || [ $EXIT_STATUS_VIEWER -ne 0 ]; then
+            echo "Process did not exit gracefully."
```
unicornss, line 733: print the err status?

unicornss (approved):

.github/workflows/ci.yml
sirknightj: Yes, for this pull request. We can add more combinations in the future.

.github/workflows/ci.yml
sirknightj: I don't think it will add much additional value, but I can add it, sure.

PR #2011, opened 2024-06-12
Default stack size change and Rollin gbuffer config rework

disa6302:
*Issue #, if available:*

*What was changed?*
- Setting default stack size parameter to 0 (this means threads created with THREAD_CREATE [will use platform specific thread stack size](https://github.com/awslabs/amazon-kinesis-video-streams-pic/blob/develop/src/utils/src/Thread.c#L178-L184))
- Reworked the rolling buffer config setting to move from setting fields in the `RtcRtpTransceiverInit` struct (which causes garbage values for config if the application does not memset the struct, leading to a crash) to surfacing an API to set up the values. The parameters are now internal to the SDK and can only be set up via this API call. 
- Cleaned up readme to have memory optimization related section

*Why was it changed?*
- To ensure that devices do not run into stack overflow issues due to default 64KB constraint
- To ensure the SDK is protected from garbage values

*How was it changed?*
- Set KVS_STACK_SIZE to 0 in cmake if no value is provided explicitly.
- `createRollingBufferConfig` API

*What testing was done for the changes?*
- Added unit tests to test the new code
- Added logging of the parameters to confirm the values are what they are supposed to be
- Unit test run.

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.

PR #1924, opened 2024-02-15
Thread stack sizes

jdelapla:
*What was changed?*
1. PIC dependency update.
2. Reduced minimum threads for threadpool context.
3. Default thread stack sizes.
4. Additionally custom stack sizes were introduced for connection listener (higher than default) and media kickoff thread (lower than default).
5. Default stack size can be changed by compile time variable provided to CMake

*Why was it changed?*
To reduce memory usage

*How was it changed?*
Mostly by PIC level changes, and using the new API for specifying thread stack sizes.

*What testing was done for the changes?*
Manual testing on unix testing on host - host connections for extended periods of time, and CI.

Additional notes, from using `pmap -x` on current develop on Ubuntu 20.04, the following values are found:
**IDLE**
```
Address           Kbytes     RSS   Dirty Mode  Mapping
total kB         1033452   10280    3368
```
**60 seconds into a call**
```
Address           Kbytes     RSS   Dirty Mode  Mapping
total kB         1033452   12124    5212
```

Compared to with this PR:
**IDLE**
```
Address           Kbytes     RSS   Dirty Mode  Mapping
total kB          542516   10116    3240
```
**60 seconds into a call**
```
Address           Kbytes     RSS   Dirty Mode  Mapping
total kB          608120   12844    5968
```

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.

PR #2073, opened 2024-10-29
Add option to specify stack size to pthread_create

sirknightj:
*What was changed?*
- See #1924 and #2011 for the original PRs.
- Add an option `KVS_STACK_SIZE` to specify the thread stack size of threads. By default, THREAD_CREATE (defined in PIC, which calls pthread_create), will use the system default settings (`ulimit -s`). These changes allow the user to set the stack sizes for threads created by the SDK without needing to change system defaults.
- For backwards compatibility, if the option is not present, the SDK will continue to use the system default stack sizes.
- Users can also control stack size for individual threads by using `THREAD_CREATE_WITH_PARAMS` API instead of `THREAD_CREATE`. (Added in [PIC#243](https://github.com/awslabs/amazon-kinesis-video-streams-pic/pull/243))

*Why was it changed?*
- Reduce unused stack memory.

*How was it changed?*
-  `KVS_STACK_SIZE` option will get passed to kvsCommonLws (Producer-C), who forwards it to PIC, which contains the thread creation definitions.

- See the original pull requests for more details.


CMakeLists.txt
```diff
@@ -52,6 +52,15 @@ execute_process(
 add_definitions(-DSDK_VERSION=\"${GIT_COMMIT_HASH}\")
 add_definitions(-DVERSION_STRING=\"${PROJECT_VERSION}\")
 add_definitions(-DDETECTED_GIT_HASH)
+if(NOT KVS_STACK_SIZE OR KVS_STACK_SIZE STREQUAL "")
```
stefankiesz, line 55: Rather than checking for an empty string, can we validate that it is a numeric value with something like:
```
if(KVS_STACK_SIZE MATCHES "^[+-]?[0-9]+$")
```

sirknightj: I'm thinking of removing this section actually. Because when I ran Producer-C with the thread stacksize changes, I was able to pass it in via producer-c's CMake command:
```
$ cmake .. -DBUILD_DEPENDENCIES=OFF -DKVS_DEFAULT_STACK_SIZE=256 -DCONSTRAINED_DEVICE=ON
...
CMake Error at dependency/libkvspic/kvspic-src/CMakeLists.txt:62 (message):
  Conflicting parameters: KVS_DEFAULT_STACK_SIZE and CONSTRAINED_DEVICE
  cannot both be set.

-- Configuring incomplete, errors occurred!
```

Which would indicate this section is redundant.

sirknightj: Removed this part and left it to PIC to determine it.

sirknightj:
Need to change from `develop` to v1.5.4 after it's released.

unicornss (approved):

README.md
```diff
@@ -159,6 +159,9 @@ You can pass the following options to `cmake ..`.
 * `-DENABLE_KVS_THREADPOOL` -- Enable the KVS threadpool which is off by default.
 * `-DENABLE_STATS_CALCULATION_CONTROL` -- Enable the runtime control of ICE agent stats calculations.
 
+You can pass the following options to `cmake ..`. These options get propagated to PIC.
```
stefankiesz, line 162: Extra sentence I believe:

> You can pass the following options to `cmake ..`.

sirknightj: Edited

stefankiesz (approved):

PR #2074, opened 2024-11-01
Memory Optimization - ICE Agent Stats (#1947)

stefankiesz:
* Original PR: #1947

*Issue #, if available:*

*What was changed?*
* Ice agent stats are now back to being enabled by default.
* Added `ENABLE_STATS_CALCULATION_CONTROL` CMake option to support enabling/disabling stats at runtime.
* Added unit tests to test these new paths.
* Updated ReadMe to reflect the new changes.

*Why was it changed?*
* Leaving stats calculations to be on by default for backwards-compatibility purposes.
* To ensure the new `enableIceStats` value is used as intended, a compile-time flag is required to enable its use for run-time control over enabling/disabling stats.
  * This is to avoid the risk of using garbage value for `enableIceStats` in the case that it was not initialized.
  * This will keep the SDK backwards-compatible with applications.

*How was it changed?*
* `#ifdef`'s were added to check for the `ENABLE_STATS_CALCULATION_CONTROL` flag before using the `enableIceStats` value.

* Allowing the new unit tests to pass in the CI.
* Locally testing all 4 possible paths:
  1. `ENABLE_STATS_CALCULATION_CONTROL`=FALSE and `enableIceStats`=TRUE
  2. `ENABLE_STATS_CALCULATION_CONTROL`=FALSE and `enableIceStats`=FALSE
  3. `ENABLE_STATS_CALCULATION_CONTROL`=TRUE and `enableIceStats`=TRUE
  4. `ENABLE_STATS_CALCULATION_CONTROL`=TRUE and `enableIceStats`=FALSE


src/include/com/amazonaws/kinesis/video/webrtcclient/Include.h
```diff
@@ -1204,6 +1199,8 @@ typedef struct {
     BOOL disableSenderSideBandwidthEstimation; //!< Disable TWCC feedback based sender bandwidth estimation, enabled by default.
                                                //!< You want to set this to TRUE if you are on a very stable connection and want to save 1.2MB of
                                                //!< memory
+    BOOL enableIceStats; //!< Control whether ICE agent stats are to be calculated. ENABLE_STATS_CALCULATION_CONTROL compiler flag must be defined
```
sirknightj, line 1202: This parameter should also be if-defined.

src/source/Ice/IceAgent.c
```diff
@@ -107,25 +113,40 @@ STATUS createIceAgent(PCHAR username, PCHAR password, PIceAgentCallbacks pIceAge
...
-                switch (pIceAgent->iceServers[pIceAgent->iceServersCount].transport) {
-                    case KVS_SOCKET_PROTOCOL_UDP:
-                        STRCPY(pIceAgent->rtcIceServerDiagnostics[i].protocol, ICE_TRANSPORT_TYPE_UDP);
-                        break;
-                    case KVS_SOCKET_PROTOCOL_TCP:
-                        STRCPY(pIceAgent->rtcIceServerDiagnostics[i].protocol, ICE_TRANSPORT_TYPE_TCP);
-                        break;
-                    default:
-                        MEMSET(pIceAgent->rtcIceServerDiagnostics[i].protocol, 0, SIZEOF(pIceAgent->rtcIceServerDiagnostics[i].protocol));
+                if (!statsControlEnabled || (statsControlEnabled && pIceAgent->kvsRtcConfiguration.enableIceStats)) {
```
sirknightj, line 116: The second `statsControlEnabled` is redundant.
```suggestion
                if (!statsControlEnabled || pIceAgent->kvsRtcConfiguration.enableIceStats) {
```

src/source/Ice/IceAgent.c
```diff
@@ -107,25 +113,40 @@ STATUS createIceAgent(PCHAR username, PCHAR password, PIceAgentCallbacks pIceAge
...
-                STRCPY(pIceAgent->rtcIceServerDiagnostics[i].url, pRtcConfiguration->iceServers[i].urls);
                 pIceAgent->iceServersCount++;
             } else {
                 DLOGE("Failed to parse ICE servers");
             }
         }
     }
 
+    // Note: The conditional is emphasizing that statsControlEnabled must be TRUE in order to use kvsRtcConfiguration.enableIceStats.
+    if (!statsControlEnabled || (statsControlEnabled && pIceAgent->kvsRtcConfiguration.enableIceStats)) {
```
sirknightj, line 140: Same as above.

src/source/Ice/IceAgent.c
```diff
@@ -2522,14 +2582,16 @@ STATUS handleStunPacket(PIceAgent pIceAgent, PBYTE pBuffer, UINT32 bufferLen, PS
...
-                    hashTableGetCount(pIceAgent->requestTimestampDiagnostics, &count);
+                if (pIceAgent->pRtcIceServerDiagnostics[pIceCandidate->iceServerIndex] != NULL) {
+                    // Update round trip time for server reflexive candidate
+                    pIceAgent->pRtcIceServerDiagnostics[pIceCandidate->iceServerIndex]->totalResponsesReceived++;
+                    // Transaction ID count be same for candidates coming from same interface, which means there would only
+                    // be one entry. It is not necessary to update a return sttaus since it is not indicative of a failure
+                    if ((hashTableGet(pIceAgent->requestTimestampDiagnostics, checkSum, &requestSentTime)) == STATUS_SUCCESS) {
+                        pIceAgent->pRtcIceServerDiagnostics[pIceCandidate->iceServerIndex]->totalRoundTripTime += GETTIME() - requestSentTime;
+                        CHK_STATUS(hashTableRemove(pIceAgent->requestTimestampDiagnostics, checkSum));
+                        hashTableGetCount(pIceAgent->requestTimestampDiagnostics, &count);
```
sirknightj, line 2593: Check return status of `hashTableGetCount`

src/source/Metrics/Metrics.c
```diff
@@ -56,20 +65,29 @@ STATUS getIceCandidateStats(PRtcPeerConnection pRtcPeerConnection, BOOL isRemote
...
     CHK((pRtcPeerConnection != NULL || pRtcIceCandidateStats != NULL), STATUS_NULL_ARG);
     MUTEX_LOCK(pIceAgent->lock);
     locked = TRUE;
-    PRtcIceCandidateDiagnostics pRtcIceCandidateDiagnostics = &pIceAgent->rtcSelectedRemoteIceCandidateDiagnostics;
-    if (!isRemote) {
-        pRtcIceCandidateDiagnostics = &pIceAgent->rtcSelectedLocalIceCandidateDiagnostics;
-        STRCPY(pRtcIceCandidateStats->relayProtocol, pRtcIceCandidateDiagnostics->relayProtocol);
-        STRCPY(pRtcIceCandidateStats->url, pRtcIceCandidateDiagnostics->url);
+    CHK_WARN(!statsControlEnabled || pIceAgent->kvsRtcConfiguration.enableIceStats, STATUS_SUCCESS, "ICE stats not enabled");
+    PRtcIceCandidateDiagnostics pRtcIceCandidateDiagnostics = pIceAgent->pRtcSelectedRemoteIceCandidateDiagnostics;
```
sirknightj, line 78: Move declaration up to the top of the function

src/source/Metrics/Metrics.c
```diff
@@ -56,20 +65,29 @@ STATUS getIceCandidateStats(PRtcPeerConnection pRtcPeerConnection, BOOL isRemote
...
+
     CHK((pRtcPeerConnection != NULL || pRtcIceCandidateStats != NULL), STATUS_NULL_ARG);
     MUTEX_LOCK(pIceAgent->lock);
     locked = TRUE;
-    PRtcIceCandidateDiagnostics pRtcIceCandidateDiagnostics = &pIceAgent->rtcSelectedRemoteIceCandidateDiagnostics;
-    if (!isRemote) {
-        pRtcIceCandidateDiagnostics = &pIceAgent->rtcSelectedLocalIceCandidateDiagnostics;
-        STRCPY(pRtcIceCandidateStats->relayProtocol, pRtcIceCandidateDiagnostics->relayProtocol);
-        STRCPY(pRtcIceCandidateStats->url, pRtcIceCandidateDiagnostics->url);
+    CHK_WARN(!statsControlEnabled || pIceAgent->kvsRtcConfiguration.enableIceStats, STATUS_SUCCESS, "ICE stats not enabled");
```
sirknightj, line 77: Should this be an error? Since there are no stats being calculated, so the out-value `pRtcIceCandidateStats` will remain NULL.

tst/MetricsApiTest.cpp
```diff
@@ -89,7 +89,117 @@ TEST_F(MetricsApiTest, webRtcIceCandidateGetMetrics)
...
+    STRNCPY(configuration.iceServers[0].credential, (PCHAR) "", MAX_ICE_CONFIG_CREDENTIAL_LEN);
+    STRNCPY(configuration.iceServers[0].username, (PCHAR) "", MAX_ICE_CONFIG_USER_NAME_LEN);
 
+    STRNCPY(configuration.iceServers[1].urls, (PCHAR) "turns:54.202.170.151:443?transport=tcp", MAX_ICE_CONFIG_URI_LEN);
+    STRNCPY(configuration.iceServers[1].credential, (PCHAR) "username", MAX_ICE_CONFIG_CREDENTIAL_LEN);
+    STRNCPY(configuration.iceServers[1].username, (PCHAR) "password", MAX_ICE_CONFIG_USER_NAME_LEN);
+    configuration.kvsRtcConfiguration.enableIceStats = FALSE;
+    ASSERT_EQ(STATUS_SUCCESS, createPeerConnection(&configuration, &pRtcPeerConnection));
+
+    rtcIceMetrics.rtcStatsObject.iceServerStats.iceServerIndex = 1;
```
sirknightj, line 181: Should we also test index 0, 1 and 2 like in the other test?

tst/MetricsApiTest.cpp
```diff
@@ -89,7 +89,117 @@ TEST_F(MetricsApiTest, webRtcIceCandidateGetMetrics)
...
+    EXPECT_EQ(STATUS_SUCCESS, rtcPeerConnectionGetMetrics(pRtcPeerConnection, NULL, &rtcIceMetrics));
+    EXPECT_EQ(0, rtcIceMetrics.rtcStatsObject.iceServerStats.port);
+    EXPECT_EQ('\0', rtcIceMetrics.rtcStatsObject.iceServerStats.url[0]);
+    EXPECT_EQ('\0', rtcIceMetrics.rtcStatsObject.iceServerStats.protocol[0]);
```
sirknightj, line 185: I would think this is a bit misleading because it seems like WebRTC is not sending anything. Even though it would be in this case just no stats are calculated.

src/source/Ice/IceAgent.c
```diff
@@ -2522,14 +2582,16 @@ STATUS handleStunPacket(PIceAgent pIceAgent, PBYTE pBuffer, UINT32 bufferLen, PS
...
-                if ((hashTableGet(pIceAgent->requestTimestampDiagnostics, checkSum, &requestSentTime)) == STATUS_SUCCESS) {
-                    pIceAgent->rtcIceServerDiagnostics[pIceCandidate->iceServerIndex].totalRoundTripTime += GETTIME() - requestSentTime;
-                    CHK_STATUS(hashTableRemove(pIceAgent->requestTimestampDiagnostics, checkSum));
-                    hashTableGetCount(pIceAgent->requestTimestampDiagnostics, &count);
+                if (pIceAgent->pRtcIceServerDiagnostics[pIceCandidate->iceServerIndex] != NULL) {
+                    // Update round trip time for server reflexive candidate
+                    pIceAgent->pRtcIceServerDiagnostics[pIceCandidate->iceServerIndex]->totalResponsesReceived++;
+                    // Transaction ID count be same for candidates coming from same interface, which means there would only
+                    // be one entry. It is not necessary to update a return sttaus since it is not indicative of a failure
+                    if ((hashTableGet(pIceAgent->requestTimestampDiagnostics, checkSum, &requestSentTime)) == STATUS_SUCCESS) {
```
sirknightj, line 2590: Minor. You can use STATUS_SUCCEEDED(x) instead of == STATUS_SUCCESS

tst/MetricsApiTest.cpp
stefankiesz: I don't think that's necessary as the test is currently for basic IceStatsControl enabling/disabling functionality. We can add more extensive tests in general as a separate PR item.

src/source/Ice/IceAgent.c
```diff
@@ -1084,12 +1083,12 @@ STATUS createIceCandidatePairs(PIceAgent pIceAgent, PIceCandidate pIceCandidate,
     PIceCandidatePair pIceCandidatePair = NULL;
     BOOL freeObjOnFailure = TRUE;
     PIceCandidate pCurrentIceCandidate = NULL;
-    BOOL statsControlEnabled = FALSE;
+    BOOL disableStats = FALSE;
```
sirknightj, line 1086: I would suggest negating this to `enableStats`. Would make it easier to read and follow (avoiding double-negatives)

src/source/Metrics/Metrics.c
```diff
@@ -19,7 +19,9 @@ STATUS getIceCandidatePairStats(PRtcPeerConnection pRtcPeerConnection, PRtcIceCa
     pIceAgent = ((PKvsPeerConnection) pRtcPeerConnection)->pIceAgent;
     MUTEX_LOCK(pIceAgent->lock);
     locked = TRUE;
-    CHK_WARN(!statsControlEnabled || pIceAgent->kvsRtcConfiguration.enableIceStats, STATUS_SUCCESS, "ICE stats are not enabled.");
+#ifdef ENABLE_STATS_CALCULATION_CONTROL
+    CHK_WARN(pIceAgent->kvsRtcConfiguration.enableIceStats, STATUS_INVALID_OPERATION, "ICE stats not enabled");
```
sirknightj, line 23: Nice :)

CMakeLists.txt
```diff
@@ -115,6 +116,10 @@ if (ENABLE_KVS_THREADPOOL)
   add_definitions(-DENABLE_KVS_THREADPOOL)
 endif()
 
+if (ENABLE_STATS_CALCULATION_CONTROL)
```
unicornss, line 119: Would it be helpful to include a note explaining that this modification is for memory optimization, along with an estimate of the memory savings if this feature were disabled?

src/include/com/amazonaws/kinesis/video/webrtcclient/Stats.h
```diff
@@ -267,15 +277,14 @@ typedef struct {
...
-    DOMString url;                        //!< For local candidates this is the URL of the ICE server from which the candidate was obtained
-    DOMString transportId;                //!< Not used currently. ID of object that was inspected for RTCTransportStats
-    CHAR address[IP_ADDR_STR_LENGTH + 1]; //!< IPv4 or IPv6 address of the candidate
-    DOMString protocol;                   //!< Valid values: UDP, TCP
-    DOMString relayProtocol;              //!< Protocol used by endpoint to communicate with TURN server. (Only for local candidate)
-                                          //!< Valid values: UDP, TCP, TLS
-    INT32 priority;                       //!< Computed using the formula in https://tools.ietf.org/html/rfc5245#section-15.1
-    INT32 port;                           //!< Port number of the candidate
-    DOMString candidateType;              //!< Type of local/remote ICE candidate
+    DOMString url;                               //!< For local candidates this is the URL of the ICE server from which the candidate was obtained
```
unicornss, line 280: Is it necessary to use DOMString here, or would it be possible to use a character array like CHAR address[IP_ADDR_STR_LENGTH + 1] instead?

CMakeLists.txt
stefankiesz: I have "Disabling these stats may lead to reductions in memory use." in the ReadMe. Will leave this comment open so we can consider adding a memory saving estimation for all of these, perhaps after more testing is done?

src/include/com/amazonaws/kinesis/video/webrtcclient/Stats.h
stefankiesz: I believe we want to follow the spec linked in the comment above this struct definition:
https://www.w3.org/TR/webrtc-stats/#icecandidate-dict

src/source/Ice/IceAgent.c
stefankiesz: I initially was going to do just that and name it `enableStats`, but it seemed having both an `enableStats` local BOOL and an `enableIceStats` member - one of which is FALSE by default and the other not even compiled by default - might be confusing. I will rename it to a negated `doStatCalcs`.

src/include/com/amazonaws/kinesis/video/webrtcclient/Stats.h
sirknightj: Domstring is a character array with 255 length:

https://github.com/awslabs/amazon-kinesis-video-streams-webrtc-sdk-c/blob/30f52d18e647747de1ec29ad6f6d3f62128f5494/src/include/com/amazonaws/kinesis/video/webrtcclient/Stats.h#L62-L73

sirknightj: But, since the URL length was changed in another spot in this PR:

```
    CHAR url[MAX_ICE_CONFIG_URI_LEN + 1];   //!< STUN/TURN server URL
```

We should remain consistent with the URL length

stefankiesz: Per offline convo, moving all such member size/type changes to a separate PR.

src/source/Ice/IceAgent.c
```diff
@@ -254,6 +273,12 @@ STATUS freeIceAgent(PIceAgent* ppIceAgent)
         freeTransactionIdStore(&pIceAgent->pStunBindingRequestTransactionIdStore);
     }
 
+    hashTableFree(pIceAgent->requestTimestampDiagnostics);
```
sirknightj, line 276: CHK_LOG_ERR for hashtablefree

sirknightj (approved):

unicornss (approved):

unicornss (approved):

PR #2077, opened 2024-11-08
Memory Optimization - Rolling Buffer

stefankiesz:
Original PRs:
* https://github.com/awslabs/amazon-kinesis-video-streams-webrtc-sdk-c/pull/1921
* https://github.com/awslabs/amazon-kinesis-video-streams-webrtc-sdk-c/pull/2011

*Issue #, if available:*

*What was changed?*
(besides the changes in the above PRs)
* Reverted unrelated changes.
* Added missing variable name-change.
* Minor fix-ups.
* Some comments added to explain code changes.

*Why was it changed?*
* To bring in rolling buffer related memory optimization changes.

*How was it changed?*
* By modifying the code as described above.

* Built and ran test locally through LLDB.
* Allow for CI to pass.


tst/SdpApiTest.cpp
```diff
@@ -1702,6 +1712,7 @@ TEST_P(SdpApiTest_SdpMatch, populateSingleMediaSection_TestH264Fmtp)
     MEMSET(&rtcConfiguration, 0x00, SIZEOF(RtcConfiguration));
     MEMSET(&track1, 0x00, SIZEOF(RtcMediaStreamTrack));
     MEMSET(&rtcSessionDescriptionInit, 0x00, SIZEOF(RtcSessionDescriptionInit));
+    MEMSET(&rtcRtpTransceiverInit, 0x00, SIZEOF(RtcRtpTransceiverInit));
```
sirknightj, line 1715: One concern I have is that not all applications zero this input parameter (transceiver settings), would suggest looking into a way that doesn't require this

src/source/PeerConnection/Rtp.h
```diff
@@ -52,6 +64,8 @@ typedef struct {
     UINT32 jitterBufferSsrc;
     PJitterBuffer pJitterBuffer;
 
+    PRollingBufferConfig pRollingBufferConfig;
```
sirknightj, line 67: My main concern is that this might contain junk values since the user is supposed to create this struct in application and pass to SDK if it's not initialized. Would suggest either:
- Creating a new versioned struct (e.g. advancedSettings) with a cmake flag toggle to pass this rollingBufferConfig and any additional future fields.
- Creating a new constructor for peerconnection (e.g. peerconnectionWithRollingBufferConfig or createPeerConnectionWithAdvancedSettings) to allow passing this new input variable

src/source/PeerConnection/Rtp.c
```diff
@@ -49,12 +49,63 @@ STATUS createKvsRtpTransceiver(RTC_RTP_TRANSCEIVER_DIRECTION direction, PKvsPeer
...
+    } else {
+        if (pRtcMediaStreamTrack->kind == MEDIA_STREAM_TRACK_KIND_VIDEO) {
+            DLOGW("Rolling buffer bitrate does not fit range (%lf bps - %lf bps) for video. Setting to default %lf bps", MIN_EXPECTED_BIT_RATE,
+                  MAX_EXPECTED_BIT_RATE, DEFAULT_EXPECTED_VIDEO_BIT_RATE);
+            pKvsRtpTransceiver->pRollingBufferConfig->rollingBufferBitratebps = DEFAULT_EXPECTED_VIDEO_BIT_RATE;
+        } else if (pRtcMediaStreamTrack->kind == MEDIA_STREAM_TRACK_KIND_AUDIO) {
+            DLOGW("Rolling buffer bitrate does not fit range (%lf bps - %lf bps) for audio. Setting to default %lf bps", MIN_EXPECTED_BIT_RATE,
+                  MAX_EXPECTED_BIT_RATE, DEFAULT_EXPECTED_AUDIO_BIT_RATE);
+            pKvsRtpTransceiver->pRollingBufferConfig->rollingBufferBitratebps = DEFAULT_EXPECTED_AUDIO_BIT_RATE;
+        } else {
```
sirknightj, line 86: Since 0 = "use defaults", I would suggest not printing a warn log for behavior that is expected.

src/source/PeerConnection/PeerConnection.c
```diff
@@ -1521,6 +1522,19 @@ STATUS setLocalDescription(PRtcPeerConnection pPeerConnection, PRtcSessionDescri
     return retStatus;
 }
 
+STATUS createRollingBufferConfig(PRtcRtpTransceiver pRtcRtpTransceiver, PRtcMediaStreamTrack pRtcMediaStreamTrack, DOUBLE rollingBufferDurationSec,
```
sirknightj, line 1525: Thinking this method should be called something more along the lines of "configureTransceiver" and we can pass in whatever settings. It makes more sense given the CHK_WARN below.

src/source/PeerConnection/Rtp.c
```diff
@@ -49,12 +49,63 @@ STATUS createKvsRtpTransceiver(RTC_RTP_TRANSCEIVER_DIRECTION direction, PKvsPeer
...
+            DLOGW("Rolling buffer bitrate does not fit range (%lf bps - %lf bps) for unknown codec. Setting to default %lf bps",
+                  MIN_EXPECTED_BIT_RATE, MAX_EXPECTED_BIT_RATE, DEFAULT_EXPECTED_VIDEO_BIT_RATE);
+            pKvsRtpTransceiver->pRollingBufferConfig->rollingBufferBitratebps = DEFAULT_EXPECTED_VIDEO_BIT_RATE;
+        }
+    }
+
```
sirknightj, line 92: Can we add an info log with what parameters the rolling buffer was created with?

src/source/PeerConnection/Rtp.c
```diff
@@ -49,12 +49,63 @@ STATUS createKvsRtpTransceiver(RTC_RTP_TRANSCEIVER_DIRECTION direction, PKvsPeer
...
+CleanUp:
+    return retStatus;
```
sirknightj, line 94: chk log err retstatus

src/source/PeerConnection/Rtp.c
```diff
@@ -76,6 +127,11 @@ STATUS freeKvsRtpTransceiver(PKvsRtpTransceiver* ppKvsRtpTransceiver)
     if (pKvsRtpTransceiver->sender.retransmitter != NULL) {
         freeRetransmitter(&pKvsRtpTransceiver->sender.retransmitter);
     }
+
+    if (pKvsRtpTransceiver->pRollingBufferConfig != NULL) {
```
sirknightj, line 131:
```
if (pKvsRtpTransceiver->pRollingBufferConfig != NULL) {
```
Isn't this check unnecessary since SAFE_MEMFREE of freeRollingBufferConfig checks for null already?

The pre-requisite of this function would be pKvsRtpTransceiver != NULL, which is already checked above on line 114

src/source/PeerConnection/Rtp.h
```diff
@@ -9,13 +9,19 @@ extern "C" {
...
-#define DEFAULT_MTU_SIZE                           1200
-#define DEFAULT_ROLLING_BUFFER_DURATION_IN_SECONDS 3
-#define HIGHEST_EXPECTED_BIT_RATE                  (10 * 1024 * 1024)
-#define DEFAULT_SEQ_NUM_BUFFER_SIZE                1000
-#define DEFAULT_VALID_INDEX_BUFFER_SIZE            1000
-#define DEFAULT_PEER_FRAME_BUFFER_SIZE             (5 * 1024)
-#define SRTP_AUTH_TAG_OVERHEAD                     10
+#define DEFAULT_MTU_SIZE_BYTES                     1200
+#define DEFAULT_ROLLING_BUFFER_DURATION_IN_SECONDS (DOUBLE) 3
+#define DEFAULT_EXPECTED_VIDEO_BIT_RATE            (DOUBLE)(5 * 1024 * 1024)
```
sirknightj, line 14: Suggest leaving the defaults as it is to prevent backwards compat. issues.

tst/SdpApiTest.cpp
stefankiesz:
```
pKvsRtpTransceiver = (PKvsRtpTransceiver) MEMCALLOC(1, SIZEOF(KvsRtpTransceiver));
```
It actually gets zeroed internally in the SDK in `createKvsRtpTransceiver()` [here](https://github.com/awslabs/amazon-kinesis-video-streams-webrtc-sdk-c/blob/30f52d18e647747de1ec29ad6f6d3f62128f5494/src/source/PeerConnection/Rtp.c#L16) which gets called when the application calls `addTransceiver()`.

We only it on the application-side for the tests and not in the samples apps.

src/source/PeerConnection/Rtp.h
sirknightj: Discard the junk values since struct this is constructed internally by the SDK.

src/source/PeerConnection/Rtp.c
sirknightj: Thanks for updating. The log line still makes it seem like the user did something wrong.. suggest to reword for the default case along the lines of: "Using rolling buffer default of %lf bps for <type>".

unicornss (approved):

src/source/PeerConnection/SessionDescription.c
```diff
@@ -315,9 +315,23 @@ STATUS setTransceiverPayloadTypes(PHashTable codecTable, PHashTable rtxTable, PD
...
 
-        CHK_STATUS(createRtpRollingBuffer(DEFAULT_ROLLING_BUFFER_DURATION_IN_SECONDS * HIGHEST_EXPECTED_BIT_RATE / 8 / DEFAULT_MTU_SIZE,
-                                          &pKvsRtpTransceiver->sender.packetBuffer));
-        CHK_STATUS(createRetransmitter(DEFAULT_SEQ_NUM_BUFFER_SIZE, DEFAULT_VALID_INDEX_BUFFER_SIZE, &pKvsRtpTransceiver->sender.retransmitter));
+        if (pKvsRtpTransceiver != NULL) {
+            if (pKvsRtpTransceiver->pRollingBufferConfig == NULL) {
+                // Passing in 0,0. The default values will be set up since application has not set up rolling buffer config with the
+                // configureTransceiverRollingBuffer() call
+                DLOGI("Rolling buffer config not set up for tranceiver. Setting defaults.");
+                CHK_STATUS(setUpRollingBufferConfigInternal(pKvsRtpTransceiver, &pKvsRtpTransceiver->sender.track, 0, 0));
```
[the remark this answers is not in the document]
stefankiesz: No need as we do not dereference it here, and `setUpRollingBufferConfigInternal` checks for the null parameter.

PR #2089, opened 2024-12-04
Samples: Fit the RTP transceiver rolling buffers to fit the set of sample frames

sirknightj:
*Issue #, if available:*
-N/A

*What was changed?*
- Rolling buffer configuration for audio and video streams in the sample Master WebRTC client application was updated to better fit the pre-canned sample frames.
- This uses the new optional configuration API introduced in #2077.

*Why was it changed?*
- Reduce the memory usage of the sample application. The default rolling buffers are too big, causing extra RTP packets to get stored for the retransmitter (nacks) that the other peer would have considered too old to get retransmitted.

*How was it changed?*
- Reduced the bitrate in the rolling buffer parameter for the sample frames.
- Added calls to configure transceiver rolling buffers with the new parameters, for audio and video separately.

- Ran the updated master sample Ingenic T31, M1 Mac and Ubuntu EC2 and checked the playback (past 3 seconds when the rolling buffer fills up) with the [JS](https://awslabs.github.io/amazon-kinesis-video-streams-webrtc-sdk-js/examples/index.html) and to check for regressions.
- Added some print statements when the rolling buffer fills up (not committed) to confirm it takes longer than 3 seconds.

### DIFF
--- a/samples/kvsWebRTCClientMaster.c
+++ b/samples/kvsWebRTCClientMaster.c
@@ -50,6 +50,20 @@ INT32 main(INT32 argc, CHAR* argv[])
     pSampleConfiguration->audioCodec = audioCodec;
     pSampleConfiguration->videoCodec = videoCodec;
 
+    // Configure the RTP rolling buffer sizes for the set of pre-canned sample frames (add a bit extra for padding)
+    if (pSampleConfiguration->videoCodec == RTC_CODEC_H264_PROFILE_42E01F_LEVEL_ASYMMETRY_ALLOWED_PACKETIZATION_MODE) {
+        pSampleConfiguration->videoRollingBufferDurationSec = 3;
+        pSampleConfiguration->videoRollingBufferBitratebps = 1.4 * 1024 * 1024;
+    } else if (pSampleConfiguration->videoCodec == RTC_CODEC_H265) {
+        pSampleConfiguration->videoRollingBufferDurationSec = 3;
+        pSampleConfiguration->videoRollingBufferBitratebps = 462 * 1024;
+    }
+
+    if (pSampleConfiguration->audioCodec == RTC_CODEC_OPUS) {
+        pSampleConfiguration->audioRollingBufferDurationSec = 3;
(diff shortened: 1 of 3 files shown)


stefankiesz (approved):

samples/kvsWebRTCClientMaster.c
```diff
@@ -50,6 +50,20 @@ INT32 main(INT32 argc, CHAR* argv[])
     pSampleConfiguration->audioCodec = audioCodec;
     pSampleConfiguration->videoCodec = videoCodec;
 
+    // Configure the RTP rolling buffer sizes for the set of pre-canned sample frames (add a bit extra for padding)
```
unicornss, line 53: default settings

unicornss: suggest adding note on the bit rate of the sample frames -

unicornss (approved):

samples/kvsWebRTCClientMaster.c
```diff
@@ -50,6 +50,21 @@ INT32 main(INT32 argc, CHAR* argv[])
     pSampleConfiguration->audioCodec = audioCodec;
     pSampleConfiguration->videoCodec = videoCodec;
 
+    // Configure the RTP rolling buffer sizes for the set of sample frames
+    // to be smaller than the default settings (plus some extra for padding).
+    if (pSampleConfiguration->videoCodec == RTC_CODEC_H264_PROFILE_42E01F_LEVEL_ASYMMETRY_ALLOWED_PACKETIZATION_MODE) {
+        pSampleConfiguration->videoRollingBufferDurationSec = 3;
+        pSampleConfiguration->videoRollingBufferBitratebps = 1.4 * 1024 * 1024;
```
sirknightj, line 57: Note: In the other samples this is not set, meaning default values are used. Since CALLOC is used for the sample configuration, these values are 0 everywhere else. Calling `configureTransceiverRollingBuffer` with 0's (or also not calling it at all) will use the default values.

Issue #2094, opened 2024-12-18
[Bug]: Building 1.12 with arm32


thomas-roos:
### Please confirm you have already done the following

- [X] I have searched the repository for related/existing bug reports
- [X] I have all the details the issue requires

### Please answer the following prompt

- [X] This issue is replicable using the unmodified sample application

### Describe the bug

`ROR: amazon-kvs-webrtc-sdk-1.12.0-r0 do_compile: Execution of '/codebuild/output/src783793070/src/actions-runner/_work/meta-aws/meta-aws/build/tmp/work/cortexa15t2hf-neon-poky-linux-gnueabi/amazon-kvs-webrtc-sdk/1.12.0/temp/run.do_compile.13133' failed with exit code 1
ERROR: Logfile of failure stored in: /codebuild/output/src783793070/src/actions-runner/_work/meta-aws/meta-aws/build/tmp/work/cortexa15t2hf-neon-poky-linux-gnueabi/amazon-kvs-webrtc-sdk/1.12.0/temp/log.do_compile.13133
Log data follows:
| DEBUG: Executing shell function do_compile
| NOTE: VERBOSE=1 cmake --build /codebuild/output/src783793070/src/actions-runner/_work/meta-aws/meta-aws/build/tmp/work/cortexa15t2hf-neon-poky-linux-gnueabi/amazon-kvs-webrtc-sdk/1.12.0/build --target all --
| Change Dir: '/codebuild/output/src783793070/src/actions-runner/_work/meta-aws/meta-aws/build/tmp/work/cortexa15t2hf-neon-poky-linux-gnueabi/amazon-kvs-webrtc-sdk/1.12.0/build'
| 
| Run Build Command(s): ninja -v -j 36 all
[...]

PR #2097, opened 2025-01-09
Change onSctpOutboundPacket 3rd parameter type to SIZE_T

stefankiesz:
*Issue #, if available:*
https://github.com/awslabs/amazon-kinesis-video-streams-webrtc-sdk-c/issues/2094

*What was changed?*
The 3rd parameter of the `onSctpOutboundPacket` function.

*Why was it changed?*
To match the expected pointer-function signature of the second parameter of the `usrsctp_init_nothreads` function as defined [here](https://github.com/sctplab/usrsctp/blob/1ade45cbadfd19298d2c47dc538962d4425ad2dd/usrsctplib/user_socket.c#L117) in the _Usrsctp_ library.

*How was it changed?*
The parameter was changed from a ULONG to a SIZE_T.

Will confirm from user that this resolved the issue, and will allow for the CI to pass.

<br>
<br>

### DIFF
--- a/src/source/Sctp/Sctp.c
+++ b/src/source/Sctp/Sctp.c
@@ -269,7 +269,7 @@ CleanUp:
     return retStatus;
 }
 
-INT32 onSctpOutboundPacket(PVOID addr, PVOID data, ULONG length, UINT8 tos, UINT8 set_df)
+INT32 onSctpOutboundPacket(PVOID addr, PVOID data, SIZE_T length, UINT8 tos, UINT8 set_df)
 {
     UNUSED_PARAM(tos);
     UNUSED_PARAM(set_df);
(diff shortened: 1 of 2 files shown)


sirknightj:
The CI confirms that PIC's SIZE_T macro does not match the actual `size_t` in certain cases, and the original authors went with ULONG as a workaround.
https://github.com/awslabs/amazon-kinesis-video-streams-pic/blob/c98c2a256a7bb3dfc4db41ae26d45e28ac7eec56/src/common/include/com/amazonaws/kinesis/video/common/CommonDefs.h#L329-L338

May need to do some more digging into why the original authors didn't use:
```
#include <stddef.h>

typedef size_t SIZE_T;
```

Issue #2100, opened 2025-02-06
[Bug]: updateTwccHashTable() may run in an infinite loop


vincent-alfred:
### Please confirm you have already done the following

- [x] I have searched the repository for related/existing bug reports
- [x] I have all the details the issue requires

### Please answer the following prompt

- [x] This issue is replicable using the unmodified sample application

### Describe the bug

There is a case that could produce an infinite loop in https://github.com/awslabs/amazon-kinesis-video-streams-webrtc-sdk-c/blob/main/src/source/PeerConnection/Rtcp.c#L347. Although `pTwccManager->lastReportedSeqNum` is `UINT16`, `pTwccManager->lastReportedSeqNum + 1` gets promoted to `int`. So if `pTwccManager->lastReportedSeqNum` is 65535, `pTwccManager->lastReportedSeqNum + 1` is equal to 65536 and `seqNum != (pTwccManager->lastReportedSeqNum + 1)` always evaluates to true since `seqNum` is `UINT16` and wraps around within [0, 65535].

The busy loop causes a 100% CPU load in one thread. Additionally, this bug causes a deadlock that the writeFrame() call is blocked by acquiring `twccLock` while the updateTwccHashTable() call holds this lock.

### Expected Behavior

No infinite loop when receiving TWCC packets.

### Current Behavior

An infinite loop may occur when receiving TWCC packets continuously.

### Reproduction Steps

1. Turn on TWCC
2. Start Live streaming
3. Wait for the TWCC sequence number to reach 65535
4. No audio and video received in Viewer side if the above condition is met.

Note that this bug only occurs when the sequence number is exactly 65535. Since the sequence number does not increment by one always, it may take time to satisfy the infinite loop condition. In my test, using a lower audio/video bitrate to make the sequence number always increment by one could reproduce the bug more easily.

### WebRTC C SDK version being used

1.12.0

### If it was working in a previous version, which one?

1.11.0

### Compiler and Version used

mips-linux-gnu-gcc 7.2.0

### Operating System and version

Linux

### Platform being used

MIPS

sirknightj:
Hi @vincent-alfred, thank you for the details and deep dive.

[...]

PR #2101, opened 2025-02-06
Fix: Twcc wraparound uint16 promotion case

sirknightj:
*Issue #, if available:*
- #2100

*What was changed?*
- `PeerConnection.c`: Added type casting to prevent integer promotion issues when handling sequence number wraparound in the twccRollingWindowDeletion function.
- `Rtcp.c`: Added type casting to prevent integer promotion issues when handling sequence number wraparound in the updateTwccHashTable function.
- `RtcpFunctionalityTest.cpp`: Added a new test case to validate the behavior of sequence number handling when integer promotion occurs due to UINT16_MAX.

*Why was it changed?*
- Integer promotion issue was identified in the handling of sequence numbers during RTP processing by the user of the linked issue. Specifically, when the `lastReportedSeqNum` is exactly `UINT16_MAX`, the subsequent increment can cause incorrect behavior due to promotion to a larger integer type.
- This fix ensures that sequence number comparisons are performed correctly, even when the `lastReportedSeqNum` is at the value of `UINT16_MAX`.

*How was it changed?*
- In `PeerConnection.c` and `Rtcp.c`, explicit casting to `UINT16` was added to the sequence number comparisons to prevent unintentional integer promotion to a larger type. 
- Note: Although the expression (`lastReportedSeqNum` + 1) can be promoted to a larger integer type, the explicit casting ensures that the result remains within the bounds of UINT16, effectively voiding the promotion.
- In `RtcpFunctionalityTest.cpp`, a new test was added to simulate the case where `lastReportedSeqNum` is `UINT16_MAX` and ensure that the integer promotion does not cause any issues when updating the hash table.

- A new unit test, updateTwccHashTableIntPromotionCase, was created to verify that the system correctly handles the sequence number wraparound and integer promotion scenario. This test checks the proper clearing of the hash table when sequence numbers wrap around from UINT16_MAX. On my Macbook Pro, the newly added test gets stuck in an infinite loop without the fix. After the fix, the new test exits the loop cleanly.
- Existing functionality was validated to ensure that no regressions were introduced.

### DIFF
--- a/tst/RtcpFunctionalityTest.cpp
+++ b/tst/RtcpFunctionalityTest.cpp
@@ -471,6 +471,44 @@ TEST_F(RtcpFunctionalityTest, updateTwccHashTableTest)
     EXPECT_EQ(STATUS_SUCCESS, freePeerConnection(&pRtcPeerConnection));
 }
 
+TEST_F(RtcpFunctionalityTest, updateTwccHashTableIntPromotionCase) {
+    PRtcPeerConnection pRtcPeerConnection = NULL;
+    PKvsPeerConnection pKvsPeerConnection = NULL;
+    RtcConfiguration config{};
+    EXPECT_EQ(STATUS_SUCCESS, createPeerConnection(&config, &pRtcPeerConnection));
+    pKvsPeerConnection = reinterpret_cast<PKvsPeerConnection>(pRtcPeerConnection);
+    PTwccRtpPacketInfo pTwccRtpPacketInfo = NULL;
+    INT64 duration = 0;
+    UINT64 receivedBytes = 0, receivedPackets = 0, sentBytes = 0, sentPackets = 0;
+    UINT16 i;
+
+    // Grab the hash table
+    PHashTable pTwccRtpPktInfosHashTable = pKvsPeerConnection->pTwccManager->pTwccRtpPktInfosHashTable;
+    UINT16 hashTableInsertionCount = 0;
+
+    // Set up the hash table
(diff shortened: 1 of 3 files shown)


stefankiesz (approved):

tst/RtcpFunctionalityTest.cpp
```diff
@@ -471,6 +471,44 @@ TEST_F(RtcpFunctionalityTest, updateTwccHashTableTest)
...
+    UINT64 receivedBytes = 0, receivedPackets = 0, sentBytes = 0, sentPackets = 0;
+    UINT16 i;
+
+    // Grab the hash table
+    PHashTable pTwccRtpPktInfosHashTable = pKvsPeerConnection->pTwccManager->pTwccRtpPktInfosHashTable;
+    UINT16 hashTableInsertionCount = 0;
+
+    // Set up the hash table
+    pKvsPeerConnection->pTwccManager->prevReportedBaseSeqNum = UINT16_MAX;
+    pKvsPeerConnection->pTwccManager->lastReportedSeqNum = UINT16_MAX;
```
unicornss, line 491: may this be named as
 `#define  TWCC_SEQ_NUM_MAX UINT16_MAX` and
`#define TWCC_SEQ_NUM_ROLLOVER (TWCC_SEQ_NUM_MAX + 1)`

unicornss (approved):

tst/RtcpFunctionalityTest.cpp
sirknightj: Yes we can adjust the naming in a future PR.

Issue #2065, opened 2024-10-22
[Bug]: onSctpInboundPacket arm 32 bit compile error


thomas-roos:
### Please confirm you have already done the following

- [X] I have searched the repository for related/existing bug reports
- [X] I have all the details the issue requires

### Please answer the following prompt

- [X] This issue is replicable using the unmodified sample application

### Describe the bug

[error.txt](https://github.com/user-attachments/files/17475298/error.txt)

### Expected Behavior

Same code (same libs etc) builds just fine with arm64.
 

### Current Behavior

See compile error

### Reproduction Steps

MACHINE=qemuarm bitbake amazon-kvs-webrtc-sdk 

### WebRTC C SDK version being used

1.11.0 / 30f52d18e647747de1ec29ad6f6d3f62128f5494

### Compiler and Version used

arm-poky-linux-gnueabi-gcc (GCC) 14.2.0

### Operating System and version

Ubuntu 22.04 / Yocto

### Platform being used

Yocto

sirknightj:
Could you share more some more details such as:
* What CMake command and flags you're using or passing in (what's the libsrtp host and destination platforms set to?) (are you building with dependencies off?)
* What is the error mentioned in your log causing it to stop?
* Can you write some detailed steps for us to use the same toolchain and reproduce the issue you are seeing from a base Ubuntu image?

thomas-roos:
Had a discussion with Hassan (not bug related),but he gave me that pointers: "As you may know in the underlying PIC layer which webrtc c sdk is built on we have our own macros for everything like INT32, SIZE_T, etc.  it is very likely the logic is unable to detect 32bit and map one of the macros correctly.  This always results in issues like you're saying we call down to the lower level sctp library which expects the correct c types but our macro somehow got messed up, most likely ULONG is mapped to the wrong thing in PIC for your platform.  If not that something similar."

Regards your questions:
1. yes, see here: https://github.com/aws4embeddedlinux/meta-aws/blob/5b36c1a8c4c07a80c904508e77e8280395d2b30b/recipes-sdk/amazon-kvs-webrtc-sdk/amazon-kvs-webrtc-sdk_1.11.0.bb#L54
2.
` ` `
| FAILED: CMakeFiles/kvsWebrtcClient.dir/src/source/Sctp/Sctp.c.o
[...]

PR #2112, opened 2025-04-15
crypto/mbedtls: Add support for mbedtls 3.x

vikramdattu:
- Also clone mbedtls 3.6.x instead of 2.8.x to match with rainmaker's version

*Issue #, if available:*

*What was changed?*

*Why was it changed?*

*How was it changed?*


vikramdattu:
@unicornss @sirknightj please start reviewing this change.

src/source/Crypto/Dtls_mbedtls.c
```diff
@@ -229,6 +234,32 @@ STATUS dtlsTransmissionTimerCallback(UINT32 timerID, UINT64 currentTime, UINT64
...
+#if MBEDTLS_VERSION_NUMBER >= 0x03000000
+VOID dtlsSessionKeyDerivationCallback(PVOID customData, mbedtls_ssl_key_export_type secret_type, const unsigned char* pMasterSecret,
+                                      SIZE_T pMasterSecretLen, const unsigned char clientRandom[MAX_DTLS_RANDOM_BYTES_LEN],
+                                      const unsigned char serverRandom[MAX_DTLS_RANDOM_BYTES_LEN], mbedtls_tls_prf_types tlsProfile)
+{
+    ENTERS();
+
+    /* We're only interested in the TLS 1.2 master secret */
+    if (secret_type != MBEDTLS_SSL_KEY_EXPORT_TLS12_MASTER_SECRET) {
+        DLOGE("Secret type is not matching...");
```
sirknightj, line 246: Just checking that you don't want to return early here?

src/source/Crypto/Dtls_mbedtls.c
```diff
@@ -229,6 +234,32 @@ STATUS dtlsTransmissionTimerCallback(UINT32 timerID, UINT64 currentTime, UINT64
...
+    }
+
+    PDtlsSession pDtlsSession = (PDtlsSession) customData;
```
sirknightj, line 249: As this part is the same as below, wondering if we can just have one of them call the other one or maybe factor it out to a common method.

src/source/Crypto/Dtls_mbedtls.c
```diff
@@ -8,6 +8,11 @@ mbedtls_ssl_srtp_profile DTLS_SRTP_SUPPORTED_PROFILES[] = {
     MBEDTLS_TLS_SRTP_UNSET,
 };
 
+#if MBEDTLS_VERSION_NUMBER >= 0x03000000
+static mbedtls_ctr_drbg_context ctr_drbg;
```
sirknightj, line 12: It seems like these were added since we needed them for `mbedtls_pk_check_pair` in `copyCertificateAndKey`, and even though we have these two fields in `struct __DtlsSession`, the `copyCertificateAndKey` doesn't have access to this struct.

sirknightj: Maybe an optimization you can make would be to add these two pointer arguments to `copyCertificateAndKey` (since it's only used within the SDK we should be OK for backwards compat.), since these two are already initialized in `createDtlsSession` anyways (avoiding an extra initialization)

src/source/Crypto/Dtls_mbedtls.c
```diff
@@ -598,6 +672,29 @@ STATUS copyCertificateAndKey(mbedtls_x509_crt* pCert, mbedtls_pk_context* pKey,
     return retStatus;
 }
 
+#if !(defined(MBEDTLS_BIGNUM_C) && !defined(MBEDTLS_DEPRECATED_REMOVED))
+int mbedtls_x509write_crt_set_serial(mbedtls_x509write_cert* ctx, const mbedtls_mpi* serial)
```
sirknightj, line 676: Just checking, it seems they removed this function: https://github.com/Mbed-TLS/mbedtls/issues/9892, but seems like it wasn't removed in the version specified here, 3.6.0

src/source/Crypto/Dtls.h
```diff
@@ -208,9 +208,17 @@ INT32 dtlsSessionSendCallback(PVOID, const unsigned char*, ULONG);
...
+
+#if MBEDTLS_VERSION_NUMBER >= 0x03000000
+VOID dtlsSessionKeyDerivationCallback(PVOID customData, mbedtls_ssl_key_export_type secret_type, const unsigned char* pMasterSecret,
+                                      SIZE_T pMasterSecretLen, const unsigned char clientRandom[MAX_DTLS_RANDOM_BYTES_LEN],
+                                      const unsigned char serverRandom[MAX_DTLS_RANDOM_BYTES_LEN], mbedtls_tls_prf_types tlsProfile);
+#else
 INT32 dtlsSessionKeyDerivationCallback(PVOID, const unsigned char*, const unsigned char*, ULONG, ULONG, ULONG,
                                        const unsigned char[MAX_DTLS_RANDOM_BYTES_LEN], const unsigned char[MAX_DTLS_RANDOM_BYTES_LEN],
                                        mbedtls_tls_prf_types);
+#endif
```
sirknightj, line 220: Suggest to add an error if MBEDTLS_VERSION_NUMBER is not defined?

```
#ifndef MBEDTLS_VERSION_NUMBER
#error "MBEDTLS_VERSION_NUMBER is not defined"
#endif
```

sirknightj: Maybe another suggestion would be to create a new macro so that we don't have `0x03000000` everywhere?

```
#define MBEDTLS_BEFORE_V3 (MBEDTLS_VERSION_NUMBER < 0x03000000)
```

sirknightj (commented):
Some minor suggestions. Changes look good overall, please take a look at the CI.

src/source/Crypto/Dtls.h
vikramdattu: Makes sense. Will add a common macro in Dtls.h under `KVS_USE_MBEDTLS`.

src/source/Crypto/Dtls_mbedtls.c
vikramdattu: I see, I think this still should hold valid. Maybe we can use `mbedtls_x509write_crt_set_serial_raw` in some other change. But it was lot of code change in the current file so I chose to simply make the deprecated function available for us.

To add the a different discussion to this: I am setting the mbedtls version to `v3.6` instead of `v3.6.0` so we always pull the latest patch version of mbedtls which keep addressing the bug fixes and any other minor issues related to mbedtls.

src/source/Crypto/Dtls.h
```diff
@@ -208,9 +210,17 @@ INT32 dtlsSessionSendCallback(PVOID, const unsigned char*, ULONG);
...
 VOID dtlsSessionSetTimerCallback(PVOID, UINT32, UINT32);
 INT32 dtlsSessionGetTimerCallback(PVOID);
+
+#if MBEDTLS_BEFORE_V3
 INT32 dtlsSessionKeyDerivationCallback(PVOID, const unsigned char*, const unsigned char*, ULONG, ULONG, ULONG,
                                        const unsigned char[MAX_DTLS_RANDOM_BYTES_LEN], const unsigned char[MAX_DTLS_RANDOM_BYTES_LEN],
                                        mbedtls_tls_prf_types);
+#else
+VOID dtlsSessionKeyDerivationCallback(PVOID customData, mbedtls_ssl_key_export_type secret_type, const unsigned char* pMasterSecret,
+                                      size_t pMasterSecretLen, const unsigned char clientRandom[MAX_DTLS_RANDOM_BYTES_LEN],
```
vikramdattu, line 220: Note: `size_t` here is intentional. KVS defines `SIZE_T` to `UINT_PTR` which is not always `size_t` and hence cannot use the same.

On Mac, with following version:
```
Apple clang version 16.0.0 (clang-1600.0.26.6)
Target: arm64-apple-darwin24.3.0
```

I get the error:

```
<...>/Dtls_mbedtls.c:328:59: error: incompatible function pointer types passing 
'void (PVOID, mbedtls_ssl_key_export_type, const unsigned char *, SIZE_T, const unsigned char *, const unsigned char *, mbedtls_tls_prf_types)' (aka 'void (void *, mbedtls_ssl_key_export_type, const unsigned char *, unsigned long long, const unsigned char *, const unsigned char *, mbedtls_tls_prf_types)') to parameter of type 'mbedtls_ssl_export_keys_t *' (aka 'void (*)(void *, mbedtls_ssl_key_export_type, const unsigned char *, unsigned long, const unsigned char *, const unsigned char *, mbedtls_tls_prf_types)') [-Wincompatible-function-pointer-types]
```

src/source/Crypto/Dtls_mbedtls.c
vikramdattu: The signature itself is different in both the mbedtls versions. I found it cleaner to maintain both separate.

src/source/Crypto/Dtls.h
sirknightj: Thanks. We are aware that size_t doesn't always match and have an open pull request to fix it:
- https://github.com/awslabs/amazon-kinesis-video-streams-pic/pull/273

Similar issues:
- https://github.com/awslabs/amazon-kinesis-video-streams-webrtc-sdk-c/issues/2065

src/source/Crypto/Dtls_mbedtls.c
vikramdattu: Added the early return. (Shouldn't have mattered really though.)

CMake/Dependencies/libwebsockets-CMakeLists.txt
```diff
@@ -2,8 +2,6 @@ cmake_minimum_required(VERSION 3.6.3)
 
 project(libwebsocket-download NONE)
 
-SET(PATCH_COMMAND git apply --verbose --ignore-whitespace ${CMAKE_CURRENT_LIST_DIR}/libwebsockets-old-gcc-fix-cast-cmakelists.patch)
```
stefankiesz, line 5: Since we are not longer applying this patch, let's remove the libwebsockets-old-gcc-fix-cast-cmakelists.patch file, and we should remove the already not used libwebsockets-leak-pipe-fix.patch file too.

stefankiesz (approved):
Some unused files to remove, otherwise looks good!

CMake/Dependencies/libwebsockets-CMakeLists.txt
vikramdattu: @stefankiesz removed unused patch files as a part of the `libwebsockets` update commit.

src/source/Ice/TurnConnection.c
```diff
@@ -7,6 +7,48 @@
...
+    urlStart = STRSTR(url, "://");
+    if (urlStart != NULL) {
+        urlStart += 3; // Skip "://"
+    } else {
+        // If no protocol prefix, use the entire URL
+        urlStart = url;
+    }
+
+    // Find port separator if it exists
+    urlEnd = STRCHR(urlStart, ':');
```
unicornss, line 31: Should we check if a '?' parameter exists without a port number? queryStart = STRCHR(urlStart, '?');

src/source/Crypto/Dtls_mbedtls.c
```diff
@@ -598,6 +662,29 @@ STATUS copyCertificateAndKey(mbedtls_x509_crt* pCert, mbedtls_pk_context* pKey,
...
+    /* Ensure that the MPI value fits into the buffer */
+    tmp_len = mbedtls_mpi_size(serial);
+    if (tmp_len > MBEDTLS_X509_RFC5280_MAX_SERIAL_LEN) {
+        return MBEDTLS_ERR_X509_BAD_INPUT_DATA;
+    }
+
+    ctx->MBEDTLS_PRIVATE(serial_len) = tmp_len;
+
+    ret = mbedtls_mpi_write_binary(serial, ctx->MBEDTLS_PRIVATE(serial), tmp_len);
+    if (ret < 0) {
```
unicornss, line 680: can this be != 0 ?
Should we clear (zeroize) the buffer in case of a partial write when the buffer is too small?

src/source/Crypto/Dtls_mbedtls.c
```diff
@@ -247,6 +249,36 @@ INT32 dtlsSessionKeyDerivationCallback(PVOID customData, const unsigned char* pM
...
+    if (secret_type != MBEDTLS_SSL_KEY_EXPORT_TLS12_MASTER_SECRET) {
+        DLOGE("SRTP only supports TLS 1.2 master secret, bailing out for type %d", secret_type);
+        goto CleanUp;
+    }
+
+    PDtlsSession pDtlsSession = (PDtlsSession) customData;
+    PTlsKeys pKeys = &pDtlsSession->tlsKeys;
+
+    if (pMasterSecretLen != SIZEOF(pKeys->masterSecret)) {
+        DLOGE("Length check failed, pMasterSecretLen = %d, sizeof(pKeys->masterSecret) = %d", pMasterSecretLen, SIZEOF(pKeys->masterSecret));
```
unicornss, line 269: Secret length check failed

CMake/Dependencies/libmbedtls-CMakeLists.txt
```diff
@@ -4,6 +4,12 @@ project(libmbedtls-download NONE)
 
 include(ExternalProject)
 
+if (BUILD_OLD_MBEDTLS_VERSION)
+  SET(MBEDTLS_GIT_TAG "v2.28.8")
+else()
+  SET(MBEDTLS_GIT_TAG "v3.6.3")
```
unicornss, line 10: Nice!!

src/source/Crypto/Dtls_mbedtls.c
vikramdattu: Done. Though the connection will anyway fail if the key is not clear, good idea to clear the secret for any failure.

src/source/Ice/TurnConnection.c
vikramdattu: Improved to support URLs without port.

sirknightj (approved):

PR #2116, opened 2025-04-25
Upgrade to Ubuntu 22, refactor CI to use DRY, seperate the jobs

sirknightj:
*Issue #, if available:*
- N/A

*What was changed?*
- GitHub Actions deprecated the native Ubuntu 20.04 runner. Therefore, we are now using a container-based alternative for legacy builds. [Reference](https://github.blog/changelog/2025-01-15-github-actions-ubuntu-20-runner-image-brownout-dates-and-other-breaking-changes/)
- CI was refactored to use DRY (reducing repetition) since updating the Ubuntu 20.04 GitHub actions image to the Ubuntu 20 ECR image needed to be repeated multiple times. Now, we can more quickly and easily perform future updates for any future deprecations.

*Why was it changed?*
- Ubuntu 20 deprecation. Current CI will fail and won't give us confidence that PRs won't break anything.

*How was it changed?*
- Separate the jobs by category into different files
- Pin the Cmake version
- Combine the various Ubuntu unit test jobs into one using matrix for the different build combinations.
- Add `parallel` (-j) flag for faster builds
- Use the ECR containers for Ubuntu 20 jobs instead, and for others, upgrade to Ubuntu 22. The Thread sanitizer reported some new issues on Ubuntu 22, so keeping it at Ubuntu 20.
- Removed extra things that were installed unnecessarily. (e.g. libcurl4-openssl-dev for non-tests)
  - AWS SDK CPP (for tests) requires some extra things to be pre-installed as a peer-dependency. Namely, zlib (zlib1g-dev and libcurl4-openssl-dev).
- Clean up extra commands (ulimit -c unlimited -S)
- **Reduce credentials duration** -- we don't need more than 1 hour. Also change the order so that we ask for credentials only after the build finishes.
- Use the `timeout` property instead of timeout command
- Use `working-directory` property instead of `cd`
- We had 3 of the same job running the same combination of tests. Removed the 2 duplicate ones since the tests all use the same signaling channel, reducing the repetition will decrease the likelihood of the tests interfering with each other.
- Add a condition to run the codecov job only in the main repo (this), not on forks by default, since it requires a codecov token.

- Tested on a fork

### DIFF
--- a/.github/workflows/sanitizers.yaml
+++ b/.github/workflows/sanitizers.yaml
@@ -1,0 +1,160 @@
+name: WebRTC C Sanitizers
+
+on:
+  push:
+    branches:
+      - develop
+      - main
+  pull_request:
+    branches:
+      - develop
+      - main
+
+env:
+  AWS_KVS_LOG_LEVEL: 7
+  DEBIAN_FRONTEND: noninteractive
+
+jobs:
+  address-sanitizer:
+    runs-on: ubuntu-22.04
+    env:
+      ASAN_OPTIONS: detect_odr_violation=0:detect_leaks=1
+      LSAN_OPTIONS: suppressions=../tst/suppressions/LSAN.supp
+      CC: clang
+      CXX: clang++
+      AWS_KVS_LOG_LEVEL: 2
+    permissions:
+      id-token: write
+      contents: read
+    steps:
+      - name: Clone repository
+        uses: actions/checkout@v4
+      - name: Configure AWS Credentials
+        uses: aws-actions/configure-aws-credentials@v4
+        with:
+          role-to-assume: ${{ secrets.AWS_ROLE_TO_ASSUME }}
+          aws-region: ${{ secrets.AWS_REGION }}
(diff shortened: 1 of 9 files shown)


.github/workflows/sanitizers.yaml
```diff
@@ -0,0 +1,179 @@
...
+          cd build
+          cmake .. -DBUILD_TEST=ON -DUNDEFINED_BEHAVIOR_SANITIZER=ON
+          make -j
+
+      - name: Run tests
+        working-directory: ./build
+        run: |
+          ./tst/webrtc_client_test
+
+  # memory-sanitizer:
```
unicornss, line 113: any comments ?

unicornss (approved):

.github/workflows/sanitizers.yaml
sirknightj: It was previously already disabled. Did not look into re-enabling it yet.

stefankiesz (approved):

PR #2117, opened 2025-04-29
Cleaning up stale resources, adjusted logging

sirknightj:
*Issue #, if available:*

*What was changed?*
- Added a script to clean up the old test resources since we hit the channel limit. It appears that the unit tests don't properly clean up their resources. That will be investigated at a later time.
- Set the log level to 7 (silent), and increase the verbosity on retries. This leads to cleaner CI output and quicker to debug what went wrong.
- Add GIT_PROGRESS and GIT_SHALLOW to make it both faster to clone, and to allow to see the cloning progress for larger dependencies like OpenSSL and AWS SDK CPP.

*Why was it changed?*
- Avoid getting billed for resources that are unused
- Tests currently fail and block the CI from passing.
- Improve developer experience by having less unimportant logs to load.

*How was it changed?*
- Run the script added in the demos repository to cleanup the stale resources based on a regex and staleness time.
- On the 2nd run of the tests (retry), set the log level to verbose (1).

- CI is the testing.

Before: 47k lines of logs to load on passing scenario.
image

After: 10k lines of logs to load on passing scenario.
image

Additional logs we can look into disabling is the LWS logs

### DIFF
--- a/CMake/Dependencies/libsrtp-CMakeLists.txt
+++ b/CMake/Dependencies/libsrtp-CMakeLists.txt
@@ -59,6 +59,8 @@ ExternalProject_Add(project_libsrtp
     GIT_REPOSITORY    https://github.com/cisco/libsrtp.git
     GIT_TAG           bd0f27ec0e299ad101a396dde3f7c90d48efc8fc
     PREFIX            ${CMAKE_CURRENT_BINARY_DIR}/build
+    GIT_PROGRESS      TRUE
+    GIT_SHALLOW       TRUE
     CMAKE_ARGS        -DCMAKE_BUILD_TYPE:STRING=${CMAKE_BUILD_TYPE}
                       -DCMAKE_INSTALL_PREFIX:STRING=${OPEN_SRC_INSTALL_PREFIX}
                       -DENABLE_OPENSSL=${LIBSRTP_ENABLE_OPENSSL}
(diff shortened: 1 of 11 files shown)


sirknightj:
Also: Fixed the codecov job silently erroring not being able to find the files. ("No such file or directory") but still passing..

image

sirknightj:
Verified the codecov report is accepted & is being reported again: https://app.codecov.io/github/awslabs/amazon-kinesis-video-streams-webrtc-sdk-c/pull/2117

To check which branches & commits have it running: https://app.codecov.io/github/awslabs/amazon-kinesis-video-streams-webrtc-sdk-c/pulls

CMake/Dependencies/libsrtp-CMakeLists.txt
```diff
@@ -59,6 +59,7 @@ ExternalProject_Add(project_libsrtp
     GIT_REPOSITORY    https://github.com/cisco/libsrtp.git
     GIT_TAG           bd0f27ec0e299ad101a396dde3f7c90d48efc8fc
     PREFIX            ${CMAKE_CURRENT_BINARY_DIR}/build
+    GIT_PROGRESS      TRUE
```
sirknightj, line 62: Note for later:

Can't add GIT_SHALLOW to these since it doesn't reference a branch or tag (specific commit hash). We can use it when upgrading to later.

tst/main.cpp
```diff
@@ -75,6 +75,16 @@ int main(int argc, char** argv)
         if (trial >= MAX_TRIALS) {
             ::testing::GTEST_FLAG(break_on_failure) = breakOnFailure;
         }
+
+        // Increase logging verbosity on retry
+        if (trial >= 1) {
```
stefankiesz, line 80: Since there is a possibility it retries and succeeds, can consider changing this to `if (trial >= MAX_TRIALS)` so it's only verbose on the last attempt.

stefankiesz (approved):

tst/main.cpp
sirknightj: That's a good suggestion.. I think maybe even better would be to buffer the logs and capture all failures.. because then the first failure won't have the verbose logs.

sirknightj: Can do in a follow -up

PR #2127, opened 2025-05-07
Enhance the ice config parser

sirknightj:
*Issue #, if available:*
- N/A

*What was changed?*
- Refactored ICE server configuration parsing into a dedicated function `parseIceConfigResponse`
- Fixed potential buffer overflow in string handling
- Added proper buffer length macros to clearly distinguish between string length and buffer size
- Improved error handling for malformed JSON responses
- Fixed incorrect array indexing in password field null termination

*Why was it changed?*
- Initially investigated to fix a bug where password field wasn't being properly null-terminated:

```
// Before: Incorrectly using userName instead of password
pSignalingClient->iceConfigs[configCount - 1].userName[MAX_ICE_CONFIG_CREDENTIAL_LEN] = '\0';
```

> [!NOTE]
> This fixes a potential non-null terminated password string. This is not currently an issue since the buffer is 257 long (to match [256 limit](https://docs.aws.amazon.com/kinesisvideostreams/latest/dg/API_signaling_IceServer.html#KinesisVideo-Type-signaling_IceServer-Password)). From calling getIceServer config a handful of times, the passwords returned to me were all under 50 characters long. The buffer is memset to 0 beforehand, so the string is always currently null-terminated.

- While fixing this, identified several opportunities for improvement:
  - String handling safety: Previous implementation used STRNCPY which required manual null termination
  - Buffer size clarity: Added explicit BUFFER_LEN macros to distinguish between content length and total buffer size
  - JSON parsing robustness: Made tokenCount signed to properly handle [jsmn_parse error codes](https://github.com/zserge/jsmn/blob/25647e692c7906b96ffd2b05ca54c097948e879c/jsmn.h#L54C1-L61C3)
  - Code maintainability: Extracted parsing logic to separate function for better testing and readability

*How was it changed?*
- Introduced new `BUFFER_LEN` macros (e.g., `MAX_ICE_CONFIG_URI_BUFFER_LEN`) to explicitly include null terminator
- Replaced `STRNCPY` with `SNPRINTF` for safer string handling
- Added comprehensive error handling for malformed JSON
- Made tokenCount signed (INT32) instead of UINT32 to properly handle JSON parser error codes
- Created dedicated unit tests to verify parsing behavior

- Added comprehensive unit test suite (IceConfigParsingTest) covering:
  - Successful parsing scenarios (happy path)
  - Invalid argument handling
  - Buffer overflow prevention
  - Malformed JSON handling
  - Extra field handling
  - Maximum URI/config count and length handling

### DIFF
--- a/tst/IceConfigParsingTest.cpp
+++ b/tst/IceConfigParsingTest.cpp
@@ -1,0 +1,169 @@
+
+#include "WebRTCClientTestFixture.h"
+
+namespace com {
+namespace amazonaws {
+namespace kinesis {
+namespace video {
+namespace webrtcclient {
+
+class IceConfigParsingTest : public WebRtcClientTestBase {};
+
+TEST_F(IceConfigParsingTest, ParseSuccess)
+{
+    IceConfigInfo iceConfigs[MAX_ICE_CONFIG_COUNT];
+    UINT32 iceConfigCount;
+
+    const std::string mockResponse = R"({
+        "IceServerList": [
+            {
+                "Username": "testUser1",
+                "Password": "testPass1",
+                "Ttl": 300,
+                "Uris": [
+                    "turn:example1.com:443?transport=udp",
+                    "turn:example1.com:443?transport=tcp"
+                ]
+            },
+            {
+                "Username": "testUser2",
+                "Password": "testPass2",
+                "Ttl": 300,
+                "Uris": [
+                    "turn:example2.com:443?transport=udp"
+                ]
+            }
+        ]
+    })";
+    UINT32 mockResponseLen = mockResponse.length();
+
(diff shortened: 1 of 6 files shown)


sirknightj:
I did also check:
- Master (C) to JS viewer with ForceTURN and setting the max ice configs to 1 and observed a successful relay-relay connection and playback.

tst/IceConfigParsingTest.cpp
```diff
@@ -0,0 +1,361 @@
...
+    EXPECT_EQ(1, iceConfigs[2].uriCount);
+    EXPECT_STREQ("turn:example3.com:443?transport=udp", iceConfigs[2].uris[0]);
+}
+
+TEST_F(IceConfigParsingTest, InvalidArgs)
+{
+    IceConfigInfo iceConfigs[MAX_ICE_CONFIG_COUNT];
+    UINT32 iceConfigCount;
+
+    const std::string emptyResponse = "";
```
unicornss, line 81: nice

unicornss (approved):

PR #2134, opened 2025-05-14
Increase robustness of LwsApiCall implementation

vikramdattu:
## What changed
Improved robustness of the LWS (libwebsockets) API call implementation for signaling.

## Why
The existing implementation had several reliability issues:
- Partial HTTP body writes were rescheduled but the buffer pointer/size weren't updated, causing duplicate data
- WSS message sends had no retry mechanism — a single CVAR_WAIT timeout would fail the entire send
- No ping/pong keepalive for the WSS connection
- Peer-initiated close always returned error even for normal closure
- No tracking of in-progress message reception, leading to premature termination during multi-fragment receives

## How
- **HTTP partial write fix**: Update `body` pointer and `bodySize` on partial write before rescheduling
- **WSS ping/pong**: Add `LWS_CALLBACK_TIMER` handler for periodic keepalive, schedule via `lws_set_timer_usecs`
- **Close callback**: Distinguish normal vs abnormal closure using `LWS_CLOSE_STATUS_NORMAL` and `receiveMessage` state
- **Receive callback**: Track multi-fragment message reception with `receiveMessage` atomic; reset buffer on final fragment; check shutdown before scheduling writes
- **Writeable callback**: Add send pipe choke detection via `lws_send_pipe_choked`; handle partial writes with `MEMMOVE`; signal `messageResult` on success
- **Send retry logic**: Add `MAX_LWS_SEND_RETRY_COUNT` (3) retries in `writeLwsData` with proper CVAR_WAIT timeout handling
- **HTTP/1.1 forcing**: Force HTTP/1.1 via ALPN to reduce memory footprint on constrained platforms (ESP32)

## Review comments addressed
- Replaced `CHK(FALSE, retValue)` with `STATUS_SIGNALING_LWS_CALL_FAILED`
- Fixed CVAR_WAIT timeout bypassing retry count (store result, reset status, continue loop)
- Added `DLOGD` for retry count on CVAR_WAIT wake-up
- Moved `MAX_RETRY_COUNT` to `#define MAX_LWS_SEND_RETRY_COUNT` in LwsApiCalls.h
- Added comment explaining H2 quirk flag + HTTP/1.1 rationale

## Testing done
- 277/278 non-integration tests pass (1 pre-existing sandbox socket failure)
- Clang-format compliance check passes
- Rebased onto latest `develop` branch

### DIFF
--- a/src/source/Signaling/LwsApiCalls.c
+++ b/src/source/Signaling/LwsApiCalls.c
@@ -517,7 +517,7 @@ INT32 lwsWssCallbackRoutine(PVOID wsi, INT32 reason, PVOID user, PVOID pDataIn, 
                 retValue = (INT32) lws_write(wsi, pLwsCallInfo->sendBuffer + LWS_PRE, remainingSize, LWS_WRITE_TEXT);
                 if (retValue < 0) {
                     DLOGW("Write failed with %d", retValue);
-                    CHK(FALSE, retValue); // Return non-zero value to callback to indicate failure
+                    CHK(FALSE, STATUS_SIGNALING_LWS_CALL_FAILED);
                 } else if ((SIZE_T) retValue < remainingSize) {
                     DLOGW("Partial write occurred: %d of %zu bytes", retValue, remainingSize);
                     // Move remaining data to start of buffer
@@ -612,10 +612,12 @@ STATUS lwsCompleteSync(PLwsCallInfo pCallInfo)
     // Execute the LWS REST call
     MEMSET(&connectInfo, 0x00, SIZEOF(struct lws_client_connect_info));
     connectInfo.context = pContext;
-    connectInfo.ssl_connection = LCCSCF_USE_SSL | LCCSCF_H2_QUIRK_OVERFLOWS_TXCR; // Add flag to handle H2 flow control
+    // H2 quirk flag handles flow control edge cases. HTTP/1.1 is forced via ALPN to reduce memory
+    // footprint and improve reliability on constrained platforms (e.g. ESP32). KVS endpoints support both.
+    connectInfo.ssl_connection = LCCSCF_USE_SSL | LCCSCF_H2_QUIRK_OVERFLOWS_TXCR;
     connectInfo.port = SIGNALING_DEFAULT_SSL_PORT;
-    connectInfo.alpn = "http/1.1";     // Force HTTP/1.1 only
-    connectInfo.protocol = "http/1.1"; // Force HTTP/1.1 protocol
+    connectInfo.alpn = "http/1.1";
+    connectInfo.protocol = "http/1.1";
 
     CHK_STATUS(getRequestHost(pCallInfo->callInfo.pRequestInfo->url, &pHostStart, &pHostEnd));
     CHK(pHostEnd == NULL || *pHostEnd == '/' || *pHostEnd == '?', STATUS_INTERNAL_ERROR);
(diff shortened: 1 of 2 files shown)


src/source/Signaling/LwsApiCalls.c
```diff
@@ -569,8 +607,10 @@ STATUS lwsCompleteSync(PLwsCallInfo pCallInfo)
     // Execute the LWS REST call
     MEMSET(&connectInfo, 0x00, SIZEOF(struct lws_client_connect_info));
     connectInfo.context = pContext;
-    connectInfo.ssl_connection = LCCSCF_USE_SSL;
+    connectInfo.ssl_connection = LCCSCF_USE_SSL | LCCSCF_H2_QUIRK_OVERFLOWS_TXCR; // Add flag to handle H2 flow control
     connectInfo.port = SIGNALING_DEFAULT_SSL_PORT;
+    connectInfo.alpn = "http/1.1";     // Force HTTP/1.1 only
+    connectInfo.protocol = "http/1.1"; // Force HTTP/1.1 protocol
```
sirknightj, line 613: Wondering the purpose of the H2 flag if 1.1 is forced, is it for latency/performance optimization?

It seems that HTTP 2 should from this quick test:
```
curl -sI https://m-xxxxxxxx.kinesisvideo.us-west-2.amazonaws.com -o/dev/null -w '%{http_version}\n'
2
curl -sI https://v-xxxxxxxx.kinesisvideo.us-west-2.amazonaws.com -o/dev/null -w '%{http_version}\n'
2
```

src/source/Signaling/LwsApiCalls.c
```diff
@@ -1907,6 +1948,9 @@ STATUS writeLwsData(PSignalingClient pSignalingClient, BOOL awaitForResponse)
     SIZE_T offset, size;
     SERVICE_CALL_RESULT result;
 
+    UINT32 retryCount = 0;
+    const UINT32 MAX_RETRY_COUNT = 3;
```
sirknightj, line 1952: Thinking this should be a `#define` (very very minor memory reduction)

sirknightj (commented):
These are good improvements..

Wanted to know if you had a method or setup to test these failure recovery scenarios and if we could add something for it? Seems like it would be a lot of additional effort.

src/source/Signaling/LwsApiCalls.c
```diff
@@ -1920,22 +1964,36 @@ STATUS writeLwsData(PSignalingClient pSignalingClient, BOOL awaitForResponse)
...
-    while (iterate) {
+    while (iterate && retryCount < MAX_RETRY_COUNT) {
         offset = ATOMIC_LOAD(&pSignalingClient->pOngoingCallInfo->sendOffset);
         size = ATOMIC_LOAD(&pSignalingClient->pOngoingCallInfo->sendBufferSize);
 
         result = (SERVICE_CALL_RESULT) ATOMIC_LOAD(&pSignalingClient->messageResult);
 
         if (offset != size && result == SERVICE_CALL_RESULT_NOT_SET) {
             CHK_STATUS(CVAR_WAIT(pSignalingClient->sendCvar, pSignalingClient->sendLock, SIGNALING_SEND_TIMEOUT));
+            retryCount++;
```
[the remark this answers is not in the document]
vikramdattu: @sirknightj looks a bit odd, but here is what's happening:
- If `CVAR_WAIT` returns on timeout, we exit anyway.
- Only if it was returned early, we increment the `retryCount` and go again for wait if `offset != size`.
 else we stop iterating and go for the rest of the code path.
 
So, retryCount is basically, making it `retry` if wakeup was within timeout.
Let me know if you have a better suggestion to handle `sendCvar` wake-ups without just giving up entirely.

src/source/Signaling/LwsApiCalls.c
vikramdattu: I added this to make this ws connection work more reliably on ESP platforms. Helps reduce memory consumption a bit and works much reliably. BTW, do we really gain much with http2 for our use case?

Now that, I have lot of optimisations in place, I can test http2 with ESP32 and revert the change.

PR #2108, opened 2025-03-21
Updated .h265 frames and improved the looping logic

Varunaditya1:
*Issue #, if available:*
The Viewer JS pages stops at 09.960 timestamp frame. 

*What was changed?*
All the .h265 frames are modified along with the Macro of `NUMBER_OF_H265_FRAME_FILES`, along with the logic of looping among the frames.

*Why was it changed?*
The Viewer JS pages stops at 09.960 timestamp frame. 

*How was it changed?*
The frames were re-generated using ffmpeg.

*What testing was done for the changes?*

Yes tested the changes multiple times and on multiple devices.

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.

PR #2135, opened 2025-05-14
Adjust the constant used to read the sample frames

sirknightj:
*Issue #, if available:*
- See #2108 for the original details

*What was changed?*
- Changing the incorrect constant used for the sample frames (H.264 constant for H.265 frames).

*Why was it changed?*
- There is currently no issue since both the H.264 and H.265 values are the same (1500).
- This change helps to future-proof the code in case the number changes in the future.

*How was it changed?*
- Instead of using the H.264 constant for both paths, set the value for each path separately.

- This is a trivial change so the CI should be more than sufficient.


stefankiesz (approved):

unicornss (approved):
note: same number of frames have been used in both h264/h65 but as noticed by Aditya's PR, this should be fixed with the right const.

PR #2139, opened 2025-05-15
Handle small MAX_ICE_CONFIG_COUNT gracefully

vikramdattu:
- Since, we only use a single ICE server config anyway. To save space we might want to set the value to say 1
 - The received list may have more than one entries. We gracefully drop the remaining entries and continue.


sirknightj:
Hi @vikramdattu, thanks for the changes.

I have another PR out (#2127) which touches the same code. I've added your changes to there and adjusted the test to match.

Note that there is some ICE config parsing logic in `receiveLwsMessage` which I believe is not relevant since get ice server config is an rest api, not a websocket one. We should remove that at some point in the future.

vikramdattu:
I'm not sure if we indeed get the ICE server config alongside the offer in the KVS case. So, I added the logic in that parser as well.

sirknightj:
Adding this as part of the other PR.

PR #2141, opened 2025-05-22
Refactor Signaling sources to make `Include_i.h` independent of libwebsockets

vikramdattu:
- This way, one could write a new CMakeLists.txt, with ApiCalls using different websocket implementation
 - One should simply remove existing Signaling.c and LwsApiCalls.c from compilation and add their own

*What was changed?*
-  Include_i.h and other SIgnaling files modified to contain libwebsockets dependency to C files

*Why was it changed?*
- To make it easier to swap the signaling implementation keeping headers re-usable

### DIFF
--- a/src/source/Include_i.h
+++ b/src/source/Include_i.h
@@ -54,8 +54,6 @@ extern "C" {
 #define INET6 1
 #include <usrsctp.h>
 
-#include <libwebsockets.h>
-
 #if !defined __WINDOWS_BUILD__
 #include <signal.h>
 #include <sys/types.h>
@@ -67,6 +65,7 @@ extern "C" {
 #include <arpa/inet.h>
 #include <fcntl.h>
 #include <netinet/tcp.h>
+#include <poll.h>
 #endif
 
 // Max uFrag and uPwd length as documented in https://tools.ietf.org/html/rfc5245#section-15.4
--- a/src/source/Signaling/LwsApiCalls.h
+++ b/src/source/Signaling/LwsApiCalls.h
@@ -167,11 +167,14 @@ extern "C" {
 // Encoded max ice server infos string len
 #define MAX_ENCODED_ICE_SERVER_INFOS_STR_LEN (MAX_ICE_SERVER_INFOS_STR_LEN + ICE_SERVER_INFO_TEMPLATE_BLOAT_SIZE)
 
+// Alignment bytes for libwebsockets (Internally, it might end up using lesser than 16)
+#define LWS_ALIGN_BYTES 16
+
 // Scratch buffer size
-#define LWS_SCRATCH_BUFFER_SIZE (MAX_JSON_PARAMETER_STRING_LEN + LWS_PRE)
+#define LWS_SCRATCH_BUFFER_SIZE (MAX_JSON_PARAMETER_STRING_LEN + LWS_ALIGN_BYTES)
 
 // Send and receive buffer size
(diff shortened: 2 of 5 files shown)


src/source/Include_i.h
```diff
@@ -65,6 +63,7 @@ extern "C" {
 #include <arpa/inet.h>
 #include <fcntl.h>
 #include <netinet/tcp.h>
+#include <poll.h>
```
vikramdattu, line 66: Because this was earlier satisfied from `libwebsockets.h`

src/source/Signaling/Signaling.h
```diff
@@ -337,14 +337,14 @@ typedef struct {
...
     ThreadTracker reconnecterTracker;
 
-    // LWS context to use for Restful API
-    struct lws_context* pLwsContext;
+    // Generic websocket context - can be used by any implementation
+    PVOID pWebsocketContext;
 
-    // Signaling protocols - one more for the NULL terminator protocol
-    struct lws_protocols signalingProtocols[LWS_PROTOCOL_COUNT + 1];
+    // Generic websocket protocols array - can be used by any implementation
```
sirknightj, line 343: Can we keep the note about the +1 for the NULL terminator protocol?

src/source/Signaling/Signaling.c
```diff
@@ -223,13 +236,15 @@ STATUS freeSignaling(PSignalingClient* ppSignalingClient)
...
+    if (pSignalingClient->pWebsocketContext != NULL) {
         MUTEX_LOCK(pSignalingClient->lwsServiceLock);
-        lws_context_destroy(pSignalingClient->pLwsContext);
-        pSignalingClient->pLwsContext = NULL;
+        lws_context_destroy((struct lws_context*) pSignalingClient->pWebsocketContext);
+        pSignalingClient->pWebsocketContext = NULL;
         MUTEX_UNLOCK(pSignalingClient->lwsServiceLock);
     }
 
+    SAFE_MEMFREE(pSignalingClient->signalingProtocols[PROTOCOL_INDEX_HTTPS]);
```
sirknightj, line 246: Can we add a comment that this frees the entire struct (HTTPS and WSS)

src/source/Signaling/LwsApiCalls.h
```diff
@@ -167,11 +167,14 @@ extern "C" {
 // Encoded max ice server infos string len
 #define MAX_ENCODED_ICE_SERVER_INFOS_STR_LEN (MAX_ICE_SERVER_INFOS_STR_LEN + ICE_SERVER_INFO_TEMPLATE_BLOAT_SIZE)
 
+// Alignment bytes for libwebsockets (Internally, it might end up using lesser than 16)
```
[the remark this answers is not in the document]
sirknightj: Seems like it would eventually make sense to have an "ApiCalls" interface to abstract away lws (similar to how Tls.h is setup).

src/source/Include_i.h
vikramdattu: To make this platform independent, I think we should pass it as a compiler flag and only include if true.

cmake:
```cmake
include(CheckIncludeFile)
check_include_file("poll.h" HAVE_POLL_H)

if(HAVE_POLL_H)
    add_definitions(-DHAVE_POLL_H=1)
endif()
```

C:
```c
#ifdef HAVE_POLL_H
#include <poll.h>
#endif
```

vikramdattu: Maybe similar for `ifaddrs.h`

stefankiesz (approved):
LGTM once we get CI passing. Thank you for this work!

sirknightj (approved):

PR #2147, opened 2025-06-06
Miscellaneous build fixes

vikramdattu:
- Check if socketpair available on the platform in CMakeList instead of assuming for non-Windows platforms
- bugfix(kvspicUtils): Implicit declaration error on pthread_getname_np
- Move ifaddrs.h and poll.h under macro checks


unicornss (approved):
Notes for PIC update: 
https://github.com/awslabs/amazon-kinesis-video-streams-pic/blame/master/src/utils/src/Thread.c 

https://github.com/awslabs/amazon-kinesis-video-streams-pic/blob/master/src/utils/CMakeLists.txt

sirknightj:
Related to pthread_getname_np: https://github.com/awslabs/amazon-kinesis-video-streams-pic/pull/272

I'm thinking we should actually remove it since it's not used anywhere except the test.

PR #2153, opened 2025-06-17
Bugfix: unused label build error observed

vikramdattu:
Build issue was observed with strict compiler flags for unused label

- Fixed by removing unused `CleanUp` label in `LwsApiCalls.c`


sirknightj:
Thanks for this. I would like to eventually get a strict compiler check added to the CI at some point to check for these as they don't show up in the current runs.

PR #2155, opened 2025-07-08
Bugfix: Only operate on `pSenderTranceiver` if it was found

vikramdattu:
- The code path traces and tries to take lock on `pSenderTranceiver` even if it is NULL
 - This makes the code crash if pSenderTranceiver is NULL

FIX: Check if pSenderTranceiver is NULL before dereferencing

### DIFF
--- a/src/source/PeerConnection/Retransmitter.c
+++ b/src/source/PeerConnection/Retransmitter.c
@@ -121,13 +121,17 @@ STATUS resendPacketOnNack(PRtcpPacket pRtcpPacket, PKvsPeerConnection pKvsPeerCo
             pRtpPacket = NULL;
         }
     }
-CleanUp:
 
-    MUTEX_LOCK(pSenderTranceiver->statsLock);
-    pSenderTranceiver->outboundStats.nackCount += nackCount;
-    pSenderTranceiver->outboundStats.retransmittedPacketsSent += retransmittedPacketsSent;
-    pSenderTranceiver->outboundStats.retransmittedBytesSent += retransmittedBytesSent;
-    MUTEX_UNLOCK(pSenderTranceiver->statsLock);
+CleanUp:
+    if (pSenderTranceiver != NULL) {
+        MUTEX_LOCK(pSenderTranceiver->statsLock);
+        pSenderTranceiver->outboundStats.nackCount += nackCount;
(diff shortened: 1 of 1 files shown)


src/source/PeerConnection/Retransmitter.c
```diff
@@ -121,13 +121,17 @@ STATUS resendPacketOnNack(PRtcpPacket pRtcpPacket, PKvsPeerConnection pKvsPeerCo
...
     }
-CleanUp:
 
-    MUTEX_LOCK(pSenderTranceiver->statsLock);
-    pSenderTranceiver->outboundStats.nackCount += nackCount;
-    pSenderTranceiver->outboundStats.retransmittedPacketsSent += retransmittedPacketsSent;
-    pSenderTranceiver->outboundStats.retransmittedBytesSent += retransmittedBytesSent;
-    MUTEX_UNLOCK(pSenderTranceiver->statsLock);
+CleanUp:
+    if (pSenderTranceiver != NULL) {
```
sirknightj, line 126: Unrelated: but the variable should be spelt: `transceiver `

src/source/PeerConnection/Retransmitter.c
```diff
@@ -121,13 +121,17 @@ STATUS resendPacketOnNack(PRtcpPacket pRtcpPacket, PKvsPeerConnection pKvsPeerCo
...
+        MUTEX_LOCK(pSenderTranceiver->statsLock);
+        pSenderTranceiver->outboundStats.nackCount += nackCount;
+        pSenderTranceiver->outboundStats.retransmittedPacketsSent += retransmittedPacketsSent;
+        pSenderTranceiver->outboundStats.retransmittedBytesSent += retransmittedBytesSent;
+        MUTEX_UNLOCK(pSenderTranceiver->statsLock);
+    } else {
+        DLOGD("Retransmit pSenderTranceiver is NULL");
```
sirknightj, line 133: Should be WARN right?

sirknightj (approved):
The function calls `findTransceiverBySsrc()` which can return `STATUS_NOT_FOUND`. In this case, it jumps to the `CleanUp`. In the `CleanUp`, `pSenderTranceiver` is still `NULL`, but `pSenderTranceiver->statsLock` is accessed.

stefankiesz (approved):

PR #2174, opened 2025-10-14
Further features and improvements for ESP Port

vikramdattu:
- Fix network coprocessor for incorrect length of wifi_connected event
- esp_hosted: fixes including wifi_started event miss
- bugfix(media_stream): Fixed frame memory leak in video_capture
- webrtc_split: Added deep-sleep and wake-up logic
- data_channel: Added config in app_webrtc_config_t
- Dependency cleanup from app_webrtc and examples
- Remove requirement for ets_sys_define patch on esp-idf
- bugfix: Use timezone aware conversions to get utc time
- simplification: Remove dependency of lwip patch from slave and host side


vikramdattu:
cc @sirknightj @unicornss

PR #2181, opened 2025-11-13
Dual-Stack Endpoints and IPv6 STUN Candidates

stefankiesz:
*Issue #, if available:*
N/A

*What was changed?*
- Added support for dual-stack KVS signaling channel endpoints.
- Added support for dual-stack KVS STUN servers and IPv6 srflx candidates.
- Added support for dual-stack KVS TURN servers (still only using IPv4 relay allocations).
- Added an environment variable to switch between dual-stack and legacy (default) modes.

*Why was it changed?*
- To provide dual-stack AWS KVS endpoint support.

*How was it changed?*
- ICE server structs now maintain two IP addresses per server: IPv4 and IPv6.
  - If the ICE server is not dual-stack, then only the IPv4 address will be set.
- Dual-stack peers using dual-stack STUN servers gather both IPv4 and IPv6 srflx candidates.
- 2 turn connections (IPv4 and IPv6) will be made if using dual-stack TURN service; currently, the IPv6 allocation request will include a `REQUESTED-ADDRESS-FAMILY` attribute set to IPv4 until there is full support for IPv6 allocations, so both allocation relay addresses will be IPv4.

*What testing was done for the changes?*
- CI to pass.
- New unit tests added.

This PR includes changes from https://github.com/awslabs/amazon-kinesis-video-streams-webrtc-sdk-c/pull/2115.

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.

PR #2182, opened 2025-11-24
isIpAddr to not Return True when Extra Characters Follow IP Address

stefankiesz:
*Issue #, if available:*
- N/A

*What was changed?*
- `isIpAddr` now considers the length of the inputted string compared to the IP-format-matched string.

*Why was it changed?*
- To prevent false positives from the `isIpAddr` function.

*How was it changed?*
- If the length of the inputted string does not match the length of the formatted string, then the function returns false, even if it was able to fill the template to make an IP address.

- Ran p2p test locally using a hardcoded TURN hostname in the format "turn:44.251.170.58?transport=udp" which resulted in isIpAddr returning false.
-  Ran p2p test locally using a hardcoded TURN hostname in the format "turn:44.251.170.58" which resulted in isIpAddr returning true (the SDK removes the 'turn:' prefix before calling `isIpaddr`), and resulted in a successful relayed WebRTC streaming session.

### DIFF
--- a/src/source/Ice/Network.c
+++ b/src/source/Ice/Network.c
@@ -343,7 +343,7 @@ BOOL isIpAddr(PCHAR hostname, UINT16 length)
         DLOGW("Provided NULL hostname.");
         return FALSE;
     }
-    if (length >= MAX_ICE_CONFIG_URI_LEN) {
+    if (length > MAX_ICE_CONFIG_URI_LEN) {
         DLOGW("Provided invalid hostname length: %u.", length);
         return FALSE;
     }


src/source/Ice/Network.c
```diff
@@ -336,17 +336,32 @@ STATUS socketWrite(INT32 sockfd, const void* pBuffer, SIZE_T length)
...
+
+    // Check if IPv4 address.
+    if (sscanf(hostname, "%u.%u.%u.%u%n", &ip_1, &ip_2, &ip_3, &ip_4, &offset) == 4 && ip_1 <= 255 && ip_2 <= 255 && ip_3 <= 255 && ip_4 <= 255 &&
+        offset == strlen(hostname)) {
+        return TRUE;
+    }
+
+    // Check if IPv6 address.
+    offset = 0;
+    if (sscanf(hostname, "%x:%x:%x:%x:%x:%x:%x:%x%n", &ip_1, &ip_2, &ip_3, &ip_4, &ip_5, &ip_6, &ip_7, &ip_8, &offset) == 8 &&
```
Ahhorse1: Should we do checks here as similar to the `<= 255` done for IPv4?

src/source/Ice/Network.c
```diff
@@ -336,17 +336,32 @@ STATUS socketWrite(INT32 sockfd, const void* pBuffer, SIZE_T length)
...
+        return FALSE;
+    }
+    if (length > MAX_ICE_CONFIG_URI_LEN) {
+        DLOGW("Provided invalid hostname length: %u.", length);
+        return FALSE;
     }
-    return status;
+
+    // Check if IPv4 address.
+    if (sscanf(hostname, "%u.%u.%u.%u%n", &ip_1, &ip_2, &ip_3, &ip_4, &offset) == 4 && ip_1 <= 255 && ip_2 <= 255 && ip_3 <= 255 && ip_4 <= 255 &&
```
sirknightj: Is the IPV4/IPV6 template still used?
Also suggest to use the macro version SSCANF and STRLEN

stefankiesz:
Closing as these changes will be included in https://github.com/awslabs/amazon-kinesis-video-streams-webrtc-sdk-c/pull/2181

Comments have been addressed there, and unit test will be added too.

PR #2188, opened 2025-12-16
Update `createTurnConnection` parameter and `TURN_CONNECTION_DATA_TRANSFER_MODE` enum member name

stefankiesz:
*Issue #, if available:*
- N/A

*What was changed?*
- The `TURN_CONNECTION_DATA_TRANSFER_MODE` parameter passed into the `createTurnConnection` call in `iceAgentInitRelayCandidate`.
- The name of one of the members of the `TURN_CONNECTION_DATA_TRANSFER_MODE` enum.

*Why was it changed?*
- The passed parameter was changed because not only is it not used, the enum value that parameter is to set gets hardcoded in the same function to be a _different_ enum value from the one passed in making it very confusing.
- The enum member name had a typo.

*How was it changed?*
- The passed parameter was changed to the one we support, and a check was added to `createTurnConnection` .
- The typo was corrected.

- Allowing for the CI tests to pass.


sirknightj (approved):

Ahhorse1 (approved):

src/source/Ice/TurnConnection.h
```diff
@@ -68,7 +68,7 @@ typedef enum {
 } TURN_PEER_CONNECTION_STATE;
 
 typedef enum {
-    TURN_CONNECTION_DATA_TRANSFER_MODE_SEND_INDIDATION,
+    TURN_CONNECTION_DATA_TRANSFER_MODE_SEND_INDICATION,
```
unicornss: //   not used  currently

PR #2194, opened 2025-12-23
Support IPv6 TURN Allocations

stefankiesz:
*Issue #, if available:*
- N/A

*What was changed?*
1. TURN Allocation Requests will now contain the `REQUESTED-ADDRESS-FAMILY` attribute which will be set to either IPv4 or IPv6 depending on the IP family of the server-client connection. When in dual-stack mode (`KVS_DUALSTACK_ENDPOINTS`), both an IPv4 and an IPv6 connection will be attempted with the TURN server.
2. Initialization for the IPv6 server-client connection was moved to come prior to that of the IPv4 connection.

*Why was it changed?*
1. To support IPv6 allocated relay addresses.
2. To prioritize IPv6 over IPv4 TURN connections (when in dual-stack mode).

*How was it changed?*
1. By appending the `REQUESTED-ADDRESS-FAMILY` attribute to allocation requests and setting it to equal the same IP family type as the server-client connection.

- Tested successful e2e streaming with the following combinations:
  - C-Macbook-master <-> JS-Chrome-viewer
    - IPv4 relay <-> IPv4 srflx 
    - IPv6 relay <-> IPv6 host
- Letting the CI tests to pass. 

### DIFF
--- a/src/source/Ice/IceAgent.c
+++ b/src/source/Ice/IceAgent.c
@@ -1868,28 +1868,6 @@ STATUS iceAgentInitRelayCandidates(PIceAgent pIceAgent)
             DLOGD("Initializing TURN relay candidates for ICE server %u with IPv4 family %u and IPv6 family (if available) %u", j,
                   pIceAgent->iceServers[j].ipAddresses.ipv4Address.family, pIceAgent->iceServers[j].ipAddresses.ipv6Address.family);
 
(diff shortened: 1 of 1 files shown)


unicornss (commented):
Does the C viewer work with this setup - IPv6 relay <-> IPv6 relay?

stefankiesz:
@unicornss, yes, I tested IPv6-relay<->IPv6-relay both ends C SDK. I set log level to 1 and saw frame logs coming in on viewer end.

PR #2199, opened 2026-01-14
Signaling Cache Improvements

stefankiesz:
*Issue #, if available:*
- N/A

*What was changed?*
- Added control-plane-url and dual-stack-mode signaling cache entries.
- Added check for consistent control-plane-url and dual-stack-mode entries when reading from the signaling cache.

*Why was it changed?*
- To be able to verify whether the Control Plane URL or Dual-Stack Mode parameters were changed between runs.

*How was it changed?*
- Updated the signaling cache entries struct, added the corresponding checks, and incremented the cache file name version.

- Added checks for the new entries to the existing signaling cache tests.
- Manually tested by changing between `KVS_DUALSTACK_ENDPOINTS` modes. The SDK no longer attempts to use the previously cached endpoints when the dual-stack mode in not the same. The result is now a successful p2p connection in such a case. Also confirmed that the signaling cache _is_ used when the necessary parameters are consistent.
- Letting CI to pass.

### DIFF
--- a/tst/SignalingApiFunctionalityTest.cpp
+++ b/tst/SignalingApiFunctionalityTest.cpp
@@ -3367,8 +3367,8 @@ TEST_F(SignalingApiFunctionalityTest, fileCachingUpdateCache)
     STRCPY(testEntry2.region, "testRegion");
     STRCPY(testEntry2.channelArn, "testChannelArn2");
     STRCPY(testEntry2.channelName, "testChannel2");
-    STRCPY(testEntry.controlPlaneUrl, "testControlPlaneUrl");
-    STRCPY(testEntry.useDualStackEndpoints, testUseDualStack ? "1" : "0");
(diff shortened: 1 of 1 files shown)


src/source/Signaling/FileCache.h
```diff
@@ -12,7 +12,7 @@ extern "C" {
 
 /* If SignalingFileCacheEntry layout is changed, change the version in cache file name so we wont read from older
  * cache file. */
-#define DEFAULT_CACHE_FILE_PATH                     (PCHAR) "./.SignalingCache_v0"
+#define DEFAULT_CACHE_FILE_PATH                     (PCHAR) "./.SignalingCache_v1"
```
sirknightj: Wondering if we should cleanup older cache files.

If someone upgrades the SDK version then the older cache file will be stuck there forever until manually deleted

YPandas (approved):

PR #2200, opened 2026-01-22
Improve error handling in the master sample pipeline

sirknightj:
*Issue #, if available:*
- N/A

*What was changed?*
- Improved error handling in the GStreamer master sample pipeline when the pipeline fails to initialize (prevent silent failures)

*Why was it changed?*
- **Helps users who are tinkering with the out-of-the-box GStreamer pipeline in the sample**
- The previous implementation ignored GStreamer pipeline parsing errors and state change failures, meaning that there won't be any media getting sent. It also made that thread get stuck (can't exit with CTRL+C), while having no errors shown in the logs.
- Initialized `bus` and `msg` to NULL to prevent undefined behavior with uninitialized memory

*How was it changed?*
- Added explicit error checking after `gst_parse_launch()` calls to detect and handle recoverable parsing errors
- Validated return values from `gst_element_set_state()` to catch state transition failures
- Improved resource cleanup by ensuring all GStreamer objects are properly checked, freed, and set to NULL
- Renamed `error` variable to `gError` to avoid confusion with status codes
- Added logging for state change failures and thread exit

- Manually modified the pipeline to an incorrect one, by replacing `videotestsrc pattern=ball` with `autovideosrc`, which will cause a pipeline error since `autovideosrc` doesn't have `is-live` parameter.

| Before | After |
|-------|-------|
|  image | image |
| CTRL+C does not work | CTRL+C works |
| No errors shown | Pipeline error printed |

Additional note:
- When I set the pipeline state of that incorrect `autovideosrc` pipeline from `NULL` to `PLAYING`, it returned `GST_STATE_CHANGE_ASYNC (2)` but didn't change to playing after a few seconds, indicating the pipeline wasn't usable.

```c
if (stateChangeStatus == GST_STATE_CHANGE_ASYNC) {
    stateChangeStatus = gst_element_get_state(senderPipeline, &state, NULL, 5 * GST_SECOND);
    if (stateChangeStatus != GST_STATE_CHANGE_SUCCESS) {
        DLOGE("Pipeline failed to reach PLAYING after 5 seconds");
        retStatus = STATUS_INVALID_OPERATION;
        goto CleanUp;
    }
}
```

### DIFF
--- a/samples/kvsWebrtcClientMasterGstSample.c
+++ b/samples/kvsWebrtcClientMasterGstSample.c
@@ -154,6 +154,7 @@ PVOID sendGstreamerAudioVideo(PVOID args)
     GstBus* bus = NULL;
     GstMessage* msg = NULL;
     GError* gError = NULL;
+    gchar* gDebug = NULL;
     GstStateChangeReturn stateChangeStatus = GST_STATE_CHANGE_SUCCESS;
     PSampleConfiguration pSampleConfiguration = (PSampleConfiguration) args;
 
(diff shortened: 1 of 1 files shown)


samples/kvsWebrtcClientMasterGstSample.c
```diff
@@ -318,37 +328,48 @@ PVOID sendGstreamerAudioVideo(PVOID args)
     if (appsinkAudio != NULL) {
         g_signal_connect(appsinkAudio, "new-sample", G_CALLBACK(on_new_sample_audio), (gpointer) pSampleConfiguration);
     }
-    gst_element_set_state(senderPipeline, GST_STATE_PLAYING);
+    stateChangeStatus = gst_element_set_state(senderPipeline, GST_STATE_PLAYING);
```
unicornss: DLOGD("State change to PLAYING returned status: %d", stateChangeStatus);

unicornss (approved):

PR #2211, opened 2026-02-04
Speeding up the build: libwebsockets

sirknightj:
*Issue #, if available:*
- N/A

*What was changed?*
- Don't build the sample applications

*Why was it changed?*
- Saw lines in the build log (see below) that indicated that samples were being built

*How was it changed?*
- Turned off the samples

- Checked the new build log which doesn't the above log lines anymore

| Before | After |
|-------|-------|
| image | image |


sirknightj:
According to CI, the build phase went from 4:40~4:50 down to about 4:25-4:40

PR #2214, opened 2026-02-09
Added SDP re-negotiation support

vikramdattu:
*Issue #, if available:*
N/A

*What was changed?*
- SDP re-negotiation on an active P2P connection: a second offer from the same client id updates the existing `PeerConnection` instead of being rejected. `findTransceiversByRemoteDescription()` intersects the remote m-line direction against the direction the application configured at `addTransceiver()` time, so a narrowed direction can widen again when the peer re-enables a track; unmatched m-lines are answered inactive.
- Answer m-lines now mirror the offer's m-line count and order, including the case where media follows the datachannel m-line.
- `setTransceiverPayloadTypes()` reuses an already-allocated rolling buffer and retransmitter across a re-negotiation instead of freeing and recreating them under the media sender thread, and warns if the codec changed under a reused buffer.
- Thread-safety on the re-negotiation path: `writeFrame()` and the retransmitter read the transceiver direction and the sender payload types from atomic snapshots, `dtlsIsServer` and the remote certificate fingerprint are frozen once the DTLS handshake completes, and two pre-existing SDK races were fixed (`onIceCandidate` checked outside `peerConnectionObjLock`, `pSrtpSession` read without `pSrtpSessionLock`).
- `samples/`: the master handles a re-offer on a live session, and the session-replacement path now unlinks from the streaming session list and the client-id hash table in one critical section before freeing, matching `sessionCleanupWait()`. Locking rules for `sampleStreamingSessionList` are documented at the field.
- `samples/`: `createSampleStreamingSession()` no longer clobbers `pSampleConfiguration->enableIceStats`, and `signalingMessageReceived()` snapshots that flag under the lock. Pre-existing on `develop`; surfaced while testing this work.
- Tests: re-negotiation coverage in `PeerConnectionFunctionalityTest.cpp` and answer-shape coverage in `SdpApiTest.cpp`, including two concurrency tests.

*Why was it changed?*
Viewers routinely re-offer on an established session -- pausing and resuming a track, adding audio to a video-only call, or restarting ICE. Rejecting the second offer with `STATUS_INVALID_OPERATION` forced the peer to tear the session down and reconnect, which costs a full ICE/DTLS handshake and drops media. The companion JS SDK change (awslabs/amazon-kinesis-video-streams-webrtc-sdk-js#826) drives exactly this flow from the browser.

The `enableIceStats` change is a bug fix rather than cleanup: the assignment always landed before `initializePeerConnection()` read the field, so the flag was pinned to FALSE and `ENABLE_STATS_CALCULATION_CONTROL` had no effect in the samples once a session existed. The same write raced the read in `signalingMessageReceived()`, which happens after the lock is released.

*How was it changed?*
Five commits: the re-negotiation support, answer m-line ordering, test coverage, then the CI/thread-safety follow-ups and the two sample-layer review follow-ups. Re-negotiation is applied to the existing `PeerConnection`; ICE restarts only when the remote ICE credentials change. Fields read on the media hot path are published as atomics rather than read plainly, since re-negotiation rewrites them from the signaling thread while `writeFrame()` runs on the application's media thread. No public API change, and the single-offer path is unchanged.

- `webrtc_client_test`: 69/69 across the re-negotiation and SDP suites, and the 7 re-negotiation/concurrency tests looped 3x back-to-back for flake resistance (21/21) -- these races are intermittent, so a single clean run proves little.
- ThreadSanitizer: full binary run clean; the two concurrency tests are what surfaced the direction/payload-type and `dtlsIsServer` races fixed here.
- `scripts/run-static-sample.sh` green for both `policy=all` and `KVS_ICE_TRANSPORT_POLICY=relay` (relay/relay pair verified).
- Interop: driven with an aiortc viewer on a fixed client id, since the C viewer sample randomises its client id and so never reaches the master's re-offer path. The master takes the re-offer branch, media keeps flowing across the renegotiation (video packets 1617 -> 4405 with the connection staying up), and the run is free of TSan data races.
- gcc-4.4 / C89 and clang-format checks pass; cross-compilation matrix exercised via CI.
- Not covered: the sample session-replacement branch. It is a lock race between the disconnect handler setting `terminateFlag` and `sessionCleanupWait()` acquiring the object lock, and I could not win that race on demand on a desktop, nor via ingestion mode (a storage-mode master never receives viewer-originated offers). That change is reasoned and reviewed rather than exercised.


vikramdattu:
@sirknightj @unicornss 

This does below things:
1. Makes sure, the correct answer is sent to the new offer. (As we correctly parse and set new remote SDP)
2. New received tracks are added if we already have added transceivers
3. If the ongoing sender threads are active, keep using for the updated session

Track removal is not supported though and it is kept until the session is killed.

src/source/PeerConnection/PeerConnection.c
```diff
@@ -1483,6 +1483,23 @@ STATUS setRemoteDescription(PRtcPeerConnection pPeerConnection, PRtcSessionDescr
     CHK_STATUS(setTransceiverPayloadTypes(pKvsPeerConnection->pCodecTable, pKvsPeerConnection->pRtxTable, pKvsPeerConnection->pTransceivers));
     CHK_STATUS(setReceiversSsrc(pSessionDescription, pKvsPeerConnection->pTransceivers));
 
+    /* Apply remote offer to transceiver directions (e.g. mark removed tracks INACTIVE) so the app sees updated state before createAnswer. */
+    if (!pKvsPeerConnection->isOffer) {
+        PHashTable pUnknownCodecPayloadTypesTable = NULL;
```
unicornss: // stores payload types for codecs that aren't recognized

unicornss (commented):
Thank you for your contribution to the renegotiation support.

samples/Common.c
```diff
@@ -1487,9 +1487,47 @@ STATUS signalingMessageReceived(UINT64 customData, PReceivedSignalingMessage pRe
...
+                // Remove old session from hash table
+                CHK_STATUS(hashTableRemove(pSampleConfiguration->pRtcPeerConnectionForRemoteClient, clientIdHash));
+
+                // Remove old session from session list and free it
+                MUTEX_LOCK(pSampleConfiguration->streamingSessionListReadLock);
+                for (UINT32 idx = 0; idx < pSampleConfiguration->streamingSessionCount; ++idx) {
+                    if (pSampleConfiguration->sampleStreamingSessionList[idx] == pSampleStreamingSession) {
+                        pSampleConfiguration->streamingSessionCount--;
+                        pSampleConfiguration->sampleStreamingSessionList[idx] =
+                            pSampleConfiguration->sampleStreamingSessionList[pSampleConfiguration->streamingSessionCount];
```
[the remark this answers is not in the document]
sirknightj: Sanitizers seem to be failing as well:

https://github.com/awslabs/amazon-kinesis-video-streams-webrtc-sdk-c/actions/runs/21876594728

sirknightj:
Thanks @vikramdattu, there's one compile issue on the `mbedtls-ubuntu-gcc-4-4` https://github.com/awslabs/amazon-kinesis-video-streams-webrtc-sdk-c/actions/runs/22110791424/job/63936735562?pr=2214.

```
/__w/amazon-kinesis-video-streams-webrtc-sdk-c/amazon-kinesis-video-streams-webrtc-sdk-c/samples/Common.c: In function 'respondWithAnswer':
/__w/amazon-kinesis-video-streams-webrtc-sdk-c/amazon-kinesis-video-streams-webrtc-sdk-c/samples/Common.c:290: warning: format '%llu' expects type 'long long unsigned int', but argument 4 has type 'UINT64'
/__w/amazon-kinesis-video-streams-webrtc-sdk-c/amazon-kinesis-video-streams-webrtc-sdk-c/samples/Common.c:290: warning: format '%llu' expects type 'long long unsigned int', but argument 5 has type 'SIZE_T'
/__w/amazon-kinesis-video-streams-webrtc-sdk-c/amazon-kinesis-video-streams-webrtc-sdk-c/samples/Common.c:290: warning: format '%llu' expects type 'long long unsigned int', but argument 4 has type 'UINT64'
/__w/amazon-kinesis-video-streams-webrtc-sdk-c/amazon-kinesis-video-streams-webrtc-sdk-c/samples/Common.c:290: warning: format '%llu' expects type 'long long unsigned int', but argument 5 has type 'SIZE_T'
/__w/amazon-kinesis-video-streams-webrtc-sdk-c/amazon-kinesis-video-streams-webrtc-sdk-c/samples/Common.c: In function 'signalingMessageReceived':
/__w/amazon-kinesis-video-streams-webrtc-sdk-c/amazon-kinesis-video-streams-webrtc-sdk-c/samples/Common.c:1516: error: 'for' loop initial declarations are only allowed in C99 mode
/__w/amazon-kinesis-video-streams-webrtc-sdk-c/amazon-kinesis-video-streams-webrtc-sdk-c/samples/Common.c:1516: note: use option -std=c99 or -std=gnu99 to compile your code
/__w/amazon-kinesis-video-streams-webrtc-sdk-c/amazon-kinesis-video-streams-webrtc-sdk-c/samples/Common.c: In function 'onDataChannelMessage':
/__w/amazon-kinesis-video-streams-webrtc-sdk-c/amazon-kinesis-video-streams-webrtc-sdk-c/samples/Common.c:1848: warning: format '%llu' expects type 'long long unsigned int', but argument 4 has type 'UINT64'
/__w/amazon-kinesis-video-streams-webrtc-sdk-c/amazon-kinesis-video-streams-webrtc-sdk-c/samples/Common.c:1848: warning: format '%llu' expects type 'long long unsigned int', but argument 4 has type 'UINT64'
/__w/amazon-kinesis-video-streams-webrtc-sdk-c/amazon-kinesis-video-streams-webrtc-sdk-c/samples/Common.c:1864: warning: format '%llu' expects type 'long long unsigned int', but argument 4 has type 'UINT64'
/__w/amazon-kinesis-video-streams-webrtc-sdk-c/amazon-kinesis-video-streams-webrtc-sdk-c/samples/Common.c:1864: warning: format '%llu' expects type 'long long unsigned int', but argument 4 has type 'UINT64'
/__w/amazon-kinesis-video-streams-webrtc-sdk-c/amazon-kinesis-video-streams-webrtc-sdk-c/samples/Common.c:1902: warning: format '%llu' expects type 'long long unsigned int', but argument 4 has type 'UINT64'
/__w/amazon-kinesis-video-streams-webrtc-sdk-c/amazon-kinesis-video-streams-webrtc-sdk-c/samples/Common.c:1902: warning: format '%llu' expects type 'long long unsigned int', but argument 5 has type 'UINT64'
/__w/amazon-kinesis-video-streams-webrtc-sdk-c/amazon-kinesis-video-streams-webrtc-sdk-c/samples/Common.c:1902: warning: format '%llu' expects type 'long long unsigned int', but argument 6 has type 'UINT64'
/__w/amazon-kinesis-video-streams-webrtc-sdk-c/amazon-kinesis-video-streams-webrtc-sdk-c/samples/Common.c:1902: warning: format '%llu' expects type 'long long unsigned int', but argument 7 has type 'UINT64'
/__w/amazon-kinesis-video-streams-webrtc-sdk-c/amazon-kinesis-video-streams-webrtc-sdk-c/samples/Common.c:1902: warning: format '%llu' expects type 'long long unsigned int', but argument 8 has type 'UINT64'
/__w/amazon-kinesis-video-streams-webrtc-sdk-c/amazon-kinesis-video-streams-webrtc-sdk-c/samples/Common.c:1902: warning: format '%llu' expects type 'long long unsigned int', but argument 9 has type 'UINT64'
/__w/amazon-kinesis-video-streams-webrtc-sdk-c/amazon-kinesis-video-streams-webrtc-sdk-c/samples/Common.c:1902: warning: format '%llu' expects type 'long long unsigned int', but argument 10 has type 'UINT64'
/__w/amazon-kinesis-video-streams-webrtc-sdk-c/amazon-kinesis-video-streams-webrtc-sdk-c/samples/Common.c:1902: warning: format '%llu' expects type 'long long unsigned int', but argument 11 has type 'UINT64'
/__w/amazon-kinesis-video-streams-webrtc-sdk-c/amazon-kinesis-video-streams-webrtc-sdk-c/samples/Common.c:1902: warning: format '%llu' expects type 'long long unsigned int', but argument 12 has type 'UINT64'
/__w/amazon-kinesis-video-streams-webrtc-sdk-c/amazon-kinesis-video-streams-webrtc-sdk-c/samples/Common.c:1902: warning: format '%llu' expects type 'long long unsigned int', but argument 13 has type 'UINT64'
/__w/amazon-kinesis-video-streams-webrtc-sdk-c/amazon-kinesis-video-streams-webrtc-sdk-c/samples/Common.c:1902: warning: format '%llu' expects type 'long long unsigned int', but argument 9 has type 'UINT64'
/__w/amazon-kinesis-video-streams-webrtc-sdk-c/amazon-kinesis-video-streams-webrtc-sdk-c/samples/Common.c:1902: warning: format '%llu' expects type 'long long unsigned int', but argument 10 has type 'UINT64'
/__w/amazon-kinesis-video-streams-webrtc-sdk-c/amazon-kinesis-video-streams-webrtc-sdk-c/samples/Common.c:1902: warning: format '%llu' expects type 'long long unsigned int', but argument 11 has type 'UINT64'
/__w/amazon-kinesis-video-streams-webrtc-sdk-c/amazon-kinesis-video-streams-webrtc-sdk-c/samples/Common.c:1902: warning: format '%llu' expects type 'long long unsigned int', but argument 12 has type 'UINT64'
/__w/amazon-kinesis-video-streams-webrtc-sdk-c/amazon-kinesis-video-streams-webrtc-sdk-c/samples/Common.c:1902: warning: format '%llu' expects type 'long long unsigned int', but argument 13 has type 'UINT64'
/__w/amazon-kinesis-video-streams-webrtc-sdk-c/amazon-kinesis-video-streams-webrtc-sdk-c/samples/Common.c:1902: warning: format '%llu' expects type 'long long unsigned int', but argument 14 has type 'UINT64'
/__w/amazon-kinesis-video-streams-webrtc-sdk-c/amazon-kinesis-video-streams-webrtc-sdk-c/samples/Common.c:1902: warning: format '%llu' expects type 'long long unsigned int', but argument 15 has type 'UINT64'
/__w/amazon-kinesis-video-streams-webrtc-sdk-c/amazon-kinesis-video-streams-webrtc-sdk-c/samples/Common.c:1902: warning: format '%llu' expects type 'long long unsigned int', but argument 16 has type 'UINT64'
/__w/amazon-kinesis-video-streams-webrtc-sdk-c/amazon-kinesis-video-streams-webrtc-sdk-c/samples/Common.c:1902: warning: format '%llu' expects type 'long long unsigned int', but argument 17 has type 'UINT64'
/__w/amazon-kinesis-video-streams-webrtc-sdk-c/amazon-kinesis-video-streams-webrtc-sdk-c/samples/Common.c:1902: warning: format '%llu' expects type 'long long unsigned int', but argument 18 has type 'UINT64'
/__w/amazon-kinesis-video-streams-webrtc-sdk-c/amazon-kinesis-video-streams-webrtc-sdk-c/samples/Common.c:1902: warning: format '%llu' expects type 'long long unsigned int', but argument 19 has type 'UINT64'
/__w/amazon-kinesis-video-streams-webrtc-sdk-c/amazon-kinesis-video-streams-webrtc-sdk-c/samples/Common.c:1902: warning: format '%llu' expects type 'long long unsigned int', but argument 4 has type 'UINT64'
/__w/amazon-kinesis-video-streams-webrtc-sdk-c/amazon-kinesis-video-streams-webrtc-sdk-c/samples/Common.c:1902: warning: format '%llu' expects type 'long long unsigned int', but argument 5 has type 'UINT64'
/__w/amazon-kinesis-video-streams-webrtc-sdk-c/amazon-kinesis-video-streams-webrtc-sdk-c/samples/Common.c:1902: warning: format '%llu' expects type 'long long unsigned int', but argument 6 has type 'UINT64'
/__w/amazon-kinesis-video-streams-webrtc-sdk-c/amazon-kinesis-video-streams-webrtc-sdk-c/samples/Common.c:1902: warning: format '%llu' expects type 'long long unsigned int', but argument 7 has type 'UINT64'
/__w/amazon-kinesis-video-streams-webrtc-sdk-c/amazon-kinesis-video-streams-webrtc-sdk-c/samples/Common.c:1902: warning: format '%llu' expects type 'long long unsigned int', but argument 8 has type 'UINT64'
/__w/amazon-kinesis-video-streams-webrtc-sdk-c/amazon-kinesis-video-streams-webrtc-sdk-c/samples/Common.c:1902: warning: format '%llu' expects type 'long long unsigned int', but argument 9 has type 'UINT64'
/__w/amazon-kinesis-video-streams-webrtc-sdk-c/amazon-kinesis-video-streams-webrtc-sdk-c/samples/Common.c:1902: warning: format '%llu' expects type 'long long unsigned int', but argument 10 has type 'UINT64'
/__w/amazon-kinesis-video-streams-webrtc-sdk-c/amazon-kinesis-video-streams-webrtc-sdk-c/samples/Common.c:1902: warning: format '%llu' expects type 'long long unsigned int', but argument 11 has type 'UINT64'
/__w/amazon-kinesis-video-streams-webrtc-sdk-c/amazon-kinesis-video-streams-webrtc-sdk-c/samples/Common.c:1902: warning: format '%llu' expects type 'long long unsigned int', but argument 12 has type 'UINT64'
/__w/amazon-kinesis-video-streams-webrtc-sdk-c/amazon-kinesis-video-streams-webrtc-sdk-c/samples/Common.c:1902: warning: format '%llu' expects type 'long long unsigned int', but argument 13 has type 'UINT64'
/__w/amazon-kinesis-video-streams-webrtc-sdk-c/amazon-kinesis-video-streams-webrtc-sdk-c/samples/Common.c:1902: warning: format '%llu' expects type 'long long unsigned int', but argument 14 has type 'UINT64'
/__w/amazon-kinesis-video-streams-webrtc-sdk-c/amazon-kinesis-video-streams-webrtc-sdk-c/samples/Common.c:1902: warning: format '%llu' expects type 'long long unsigned int', but argument 15 has type 'UINT64'
/__w/amazon-kinesis-video-streams-webrtc-sdk-c/amazon-kinesis-video-streams-webrtc-sdk-c/samples/Common.c:1902: warning: format '%llu' expects type 'long long unsigned int', but argument 16 has type 'UINT64'
/__w/amazon-kinesis-video-streams-webrtc-sdk-c/amazon-kinesis-video-streams-webrtc-sdk-c/samples/Common.c:1902: warning: format '%llu' expects type 'long long unsigned int', but argument 17 has type 'UINT64'
/__w/amazon-kinesis-video-streams-webrtc-sdk-c/amazon-kinesis-video-streams-webrtc-sdk-c/samples/Common.c:1902: warning: format '%llu' expects type 'long long unsigned int', but argument 18 has type 'UINT64'
/__w/amazon-kinesis-video-streams-webrtc-sdk-c/amazon-kinesis-video-streams-webrtc-sdk-c/samples/Common.c:1902: warning: format '%llu' expects type 'long long unsigned int', but argument 19 has type 'UINT64'
/__w/amazon-kinesis-video-streams-webrtc-sdk-c/amazon-kinesis-video-streams-webrtc-sdk-c/samples/Common.c:1909: warning: format '%llu' expects type 'long long unsigned int', but argument 4 has type 'UINT64'
/__w/amazon-kinesis-video-streams-webrtc-sdk-c/amazon-kinesis-video-streams-webrtc-sdk-c/samples/Common.c:1909: warning: format '%llu' expects type 'long long unsigned int', but argument 5 has type 'UINT64'
/__w/amazon-kinesis-video-streams-webrtc-sdk-c/amazon-kinesis-video-streams-webrtc-sdk-c/samples/Common.c:1909: warning: format '%llu' expects type 'long long unsigned int', but argument 4 has type 'UINT64'
/__w/amazon-kinesis-video-streams-webrtc-sdk-c/amazon-kinesis-video-streams-webrtc-sdk-c/samples/Common.c:1909: warning: format '%llu' expects type 'long long unsigned int', but argument 5 has type 'UINT64'
/__w/amazon-kinesis-video-streams-webrtc-sdk-c/amazon-kinesis-video-streams-webrtc-sdk-c/samples/Common.c:1915: warning: format '%llu' expects type 'long long unsigned int', but argument 4 has type 'UINT64'
/__w/amazon-kinesis-video-streams-webrtc-sdk-c/amazon-kinesis-video-streams-webrtc-sdk-c/samples/Common.c:1915: warning: format '%llu' expects type 'long long unsigned int', but argument 5 has type 'UINT64'
/__w/amazon-kinesis-video-streams-webrtc-sdk-c/amazon-kinesis-video-streams-webrtc-sdk-c/samples/Common.c:1915: warning: format '%llu' expects type 'long long unsigned int', but argument 4 has type 'UINT64'
/__w/amazon-kinesis-video-streams-webrtc-sdk-c/amazon-kinesis-video-streams-webrtc-sdk-c/samples/Common.c:1915: warning: format '%llu' expects type 'long long unsigned int', but argument 5 has type 'UINT64'
make[2]: *** [samples/CMakeFiles/kvsWebrtcClientMaster.dir/build.make:79: samples/CMakeFiles/kvsWebrtcClientMaster.dir/Common.c.o] Error 1
```

Would you also be able to rebase to the latest `develop` branch? We plan to get this in as part of the next release. Thank you

sirknightj (dismissed):

vikramdattu:
@sirknightj @unicornss — rebased onto latest `develop` (PR base updated from `release-v1.18.0` to `develop`). The 1.18.x release commits that were already in develop dropped automatically; the 4 PR commits apply cleanly with no conflicts.

Locally on macOS arm64, all 5 renegotiation tests pass:
* `renegotiateWithSameIceCredentials`
* `renegotiateWithIceRestart`
* `renegotiateWithNewTransceiver`
* `renegotiateWithTrackRemoved` (covers the audio+video → audio-only re-offer path with the dropped media line marked INACTIVE)
* `renegotiateDirectionChangeOnly`

If everything looks ok, can we get this merged?

samples/common/Common.c
```diff
@@ -1510,9 +1510,47 @@ STATUS signalingMessageReceived(UINT64 customData, PReceivedSignalingMessage pRe
 
     switch (pReceivedSignalingMessage->signalingMessage.messageType) {
         case SIGNALING_MESSAGE_TYPE_OFFER:
-            // Check if we already have an ongoing master session with the same peer
```
sirknightj, line 1739: Webrtc ingestion mode since it doesn't support re-negotiation (retry: new offer with new ice candidates when joinstoragesession is called).

Can we adjust the condition to only do renegotiation in p2p mode for now?

src/source/PeerConnection/Rtp.c
```diff
@@ -288,6 +288,10 @@ STATUS writeFrame(PRtcRtpTransceiver pRtcRtpTransceiver, PFrame pFrame)
 
     CHK(pKvsRtpTransceiver != NULL && pFrame != NULL, STATUS_NULL_ARG);
     pKvsPeerConnection = pKvsRtpTransceiver->pKvsPeerConnection;
+    if (pKvsRtpTransceiver->transceiver.direction == RTC_RTP_TRANSCEIVER_DIRECTION_INACTIVE ||
```
sirknightj: Suggest adding a null check just in case

src/source/PeerConnection/PeerConnection.c
```diff
@@ -1537,6 +1540,27 @@ STATUS setRemoteDescription(PRtcPeerConnection pPeerConnection, PRtcSessionDescr
...
+                                              &pUnknownCodecPayloadTypesTable);
+        if (STATUS_SUCCEEDED(retStatus)) {
+            retStatus = hashTableCreateWithParams(unknownHashTableBucketCount, CODEC_RTPMAP_PAYLOAD_TYPES_HASH_TABLE_BUCKET_LENGTH,
+                                                  &pUnknownCodecRtpmapTable);
+        }
+        if (STATUS_SUCCEEDED(retStatus)) {
+            retStatus = findTransceiversByRemoteDescription(pKvsPeerConnection, pSessionDescription, pUnknownCodecPayloadTypesTable,
+                                                            pUnknownCodecRtpmapTable);
+        }
+        (void) hashTableFree(pUnknownCodecPayloadTypesTable);
```
sirknightj: Can use CHK_LOG_ERR (won't update the retStatus) for additional safety/debug

samples/common/Common.c
vikramdattu: Done — added `&& !pSampleConfiguration->channelInfo.useMediaStorage` to the gate. WebRTC ingestion mode falls through to the existing session-replacement path, preserving the JoinStorageSession retry-with-new-ICE behaviour.

src/source/PeerConnection/Rtp.c
vikramdattu: Done — tightened the existing `CHK` to use the public-API arg `pRtcRtpTransceiver` instead of the post-cast `pKvsRtpTransceiver` (semantically equivalent for the cast, but explicit at the API boundary).

src/source/PeerConnection/PeerConnection.c
vikramdattu: Done — swapped both `(void) hashTableFree(...)` calls to `CHK_LOG_ERR(hashTableFree(...))`.

vikramdattu:
@sirknightj — addressed all three review comments and rebased onto the latest `develop` (which now also includes the `Handle NULL arguments to writeFrame` change from #2271, so my `writeFrame` change sits on top of that cleanly). Force-pushed; new head is 90b0540d3a.

Fixes folded into the existing PR commits via `git commit --fixup` + `git rebase --autosquash` (no extra commits added):

- **`samples/common/Common.c`** — re-negotiation gate now `&& !pSampleConfiguration->channelInfo.useMediaStorage`. Ingestion mode falls through to the existing session-replacement path (preserves the JoinStorageSession retry-with-new-ICE behaviour).
- **`src/source/PeerConnection/Rtp.c`** — `CHK` now references the public-API arg `pRtcRtpTransceiver` instead of the post-cast `pKvsRtpTransceiver`.
- **`src/source/PeerConnection/PeerConnection.c`** — `(void) hashTableFree(...)` → `CHK_LOG_ERR(hashTableFree(...))` for both calls.

Locally verified: all 5 renegotiation tests + the 16 `RtpFunctionalityTest` cases (including the new `writeFrameNullArgs`) pass. Could you take another look?

vikramdattu:
ping @sirknightj @unicornss for re-review/approval and merge.

vikramdattu:
@sirknightj @unicornss can we fast-track this PR please? Rebasing it on latest develop

src/source/PeerConnection/SessionDescription.c
```diff
@@ -1536,6 +1628,68 @@ STATUS findTransceiversByRemoteDescription(PKvsPeerConnection pKvsPeerConnection
...
+        } else {
+            tokenLen = (UINT32) STRLEN(attributeValue);
+        }
+        if (STRNCMP(MEDIA_SECTION_AUDIO_VALUE, attributeValue, tokenLen) == 0 || STRNCMP(MEDIA_SECTION_VIDEO_VALUE, attributeValue, tokenLen) == 0) {
+            avMlineCount++;
+        }
+    }
+    CHK_STATUS(doubleListGetNodeCount(pKvsPeerConnection->pAnswerTransceivers, &answerTransceiverCount));
+    if (avMlineCount != answerTransceiverCount) {
+        DLOGW("Audio/video m-line count (%u) does not match answer transceiver count (%u)", avMlineCount, answerTransceiverCount);
```
sirknightj: Wondering if this is needed in the main path. Good to have additional checks (safety) but if the tests show transceiver mismatch won't happen, and all this does is print a warning (no user action)

src/source/PeerConnection/SessionDescription.c
```diff
@@ -1346,6 +1355,73 @@ STATUS writeTransceiverDirection(PCHAR buf, UINT32 len, RTC_RTP_TRANSCEIVER_DIRE
...
+    }
+    return RTC_RTP_TRANSCEIVER_DIRECTION_INACTIVE;
+}
+
+// Intersect the application-configured local direction with what the remote offer permits.
+// invertedRemote is the remote m-line direction already mapped to our perspective
+// (e.g. remote `a=sendonly` -> RECVONLY here). The result is the direction we should
+// place on the transceiver before createAnswer fires, so we don't clobber an explicit
+// app intent (e.g. SENDONLY) with a permissive remote (e.g. SENDRECV).
+static RTC_RTP_TRANSCEIVER_DIRECTION intersectLocalDirectionWithRemote(RTC_RTP_TRANSCEIVER_DIRECTION local,
```
sirknightj: There should be a recent function addition that adds this intersection logic - wondering if you can use that one?

sirknightj (commented):
Minor comments, LGTM otherwise. Could you pl. update to the latest develop branch?

src/source/PeerConnection/PeerConnection.c
vikramdattu: Done — added this comment.

src/source/PeerConnection/PeerConnection.c
```diff
@@ -1483,6 +1483,23 @@ STATUS setRemoteDescription(PRtcPeerConnection pPeerConnection, PRtcSessionDescr
     CHK_STATUS(setTransceiverPayloadTypes(pKvsPeerConnection->pCodecTable, pKvsPeerConnection->pRtxTable, pKvsPeerConnection->pTransceivers));
     CHK_STATUS(setReceiversSsrc(pSessionDescription, pKvsPeerConnection->pTransceivers));
 
+    /* Apply remote offer to transceiver directions (e.g. mark removed tracks INACTIVE) so the app sees updated state before createAnswer. */
+    if (!pKvsPeerConnection->isOffer) {
+        PHashTable pUnknownCodecPayloadTypesTable = NULL;
+        PHashTable pUnknownCodecRtpmapTable = NULL;
```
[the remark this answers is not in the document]
vikramdattu: Done — added.

src/source/PeerConnection/PeerConnection.c
```diff
@@ -1483,6 +1483,23 @@ STATUS setRemoteDescription(PRtcPeerConnection pPeerConnection, PRtcSessionDescr
...
+        UINT32 unknownHashTableBucketCount =
```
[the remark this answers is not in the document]
vikramdattu: Done — added (bucket count = number of m-lines, or MIN_HASH_BUCKET_COUNT, whichever is larger).

src/source/PeerConnection/PeerConnection.c
```diff
@@ -1483,6 +1483,23 @@ STATUS setRemoteDescription(PRtcPeerConnection pPeerConnection, PRtcSessionDescr
...
+            pSessionDescription->mediaCount < MIN_HASH_BUCKET_COUNT ? MIN_HASH_BUCKET_COUNT : pSessionDescription->mediaCount;
+        CHK_STATUS(hashTableCreateWithParams(unknownHashTableBucketCount, CODEC_RTPMAP_PAYLOAD_TYPES_HASH_TABLE_BUCKET_LENGTH,
+                                             &pUnknownCodecPayloadTypesTable));
+        CHK_STATUS(
+            hashTableCreateWithParams(unknownHashTableBucketCount, CODEC_RTPMAP_PAYLOAD_TYPES_HASH_TABLE_BUCKET_LENGTH, &pUnknownCodecRtpmapTable));
+        retStatus = findTransceiversByRemoteDescription(pKvsPeerConnection, pSessionDescription, pUnknownCodecPayloadTypesTable,
```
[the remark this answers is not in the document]
vikramdattu: Done — added a doc comment on findTransceiversByRemoteDescription describing this: it maps remote m-lines to local transceivers (reusing a matching user transceiver or creating a fake RTC_CODEC_UNKNOWN one) and records unknown-codec payload-type/rtpmap into the two parallel hashtables.

src/source/PeerConnection/SessionDescription.c
```diff
@@ -1231,6 +1231,32 @@ STATUS findCodecInTransceivers(PKvsPeerConnection pKvsPeerConnection, RTC_CODEC
...
+// Used when applying remote offer to update transceiver directions for re-negotiation.
+static RTC_RTP_TRANSCEIVER_DIRECTION getLocalDirectionFromRemoteMediaDescription(PSdpMediaDescription pMediaDescription)
+{
+    UINT32 i;
+    if (pMediaDescription == NULL) {
+        return RTC_RTP_TRANSCEIVER_DIRECTION_INACTIVE;
+    }
+    for (i = 0; i < pMediaDescription->mediaAttributesCount; i++) {
+        PCHAR name = pMediaDescription->sdpAttributes[i].attributeName;
+        if (STRCMP(name, "sendonly") == 0) {
```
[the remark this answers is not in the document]
vikramdattu: getRemoteDirectionFromMediaDescription() returns the remote m-line's raw direction (not inverted) on purpose — the send<->recv crossover happens in intersectTransceiverDirection(), so a 'remote sendonly -> local recvonly' table on this function would be misleading. I documented that split in the function header. Happy to add the explicit mapping table next to intersectTransceiverDirection() if you'd prefer it spelled out there.

src/source/PeerConnection/SessionDescription.c
vikramdattu: Yes, good catch — refactored to reuse the existing intersectTransceiverDirection() and removed my duplicate helper. The re-negotiation hook now feeds it the raw remote direction and it does the send<->recv crossover. Verified with the renegotiate* tests.

src/source/PeerConnection/SessionDescription.c
vikramdattu: It is now DLOGV (verbose-only), purely informational with no user action, and the tests confirm the counts match on the normal path. Kept it only as a lightweight diagnostic for the abnormal case — happy to drop it entirely if you'd prefer no log.

vikramdattu:
@sirknightj — rebased onto latest `develop` as requested.

While putting together an end-to-end renegotiation demo (viewer test page with buttons to add/remove the local track and pause/resume the incoming one — awslabs/amazon-kinesis-video-streams-webrtc-sdk-js#826), we found and fixed a direction bug in this flow: re-offers intersected the remote direction against the *previously negotiated* value and stored the result back, so any narrowing was permanent — after a peer paused a track with a `sendonly` re-offer, a later `sendrecv` re-offer could never widen it back and the track stayed paused. The new commit (`SessionDescription: intersect re-offers against the configured direction`) intersects against the direction the app configured at `addTransceiver()` time instead.

Verified live with the JS page against this branch's master sample: pause stops the outbound track, resume restarts it, repeatable for audio and video, on a single session.

One coordination note: the test-fixture segfault commit on this branch is the same change as #2365 — happy to drop either copy once one of them merges.

src/source/PeerConnection/SessionDescription.c
```diff
@@ -355,6 +355,15 @@ STATUS setTransceiverPayloadTypes(PHashTable codecTable, PHashTable rtxTable, PD
         }
 
         if (pKvsRtpTransceiver != NULL) {
+            // Skip transceivers that already have rolling buffer and retransmitter allocated
```
unicornss, line 410: Codec changes mid-session are not typical, but if the codec did change between negotiations (testing) -  worth adding a DLOGW and skip, or at minimum a code comment documenting the assumption...

src/source/PeerConnection/SessionDescription.c
```diff
@@ -1370,6 +1379,35 @@ STATUS writeTransceiverDirection(PCHAR buf, UINT32 len, RTC_RTP_TRANSCEIVER_DIRE
...
+    return direction == RTC_RTP_TRANSCEIVER_DIRECTION_SENDRECV || direction == RTC_RTP_TRANSCEIVER_DIRECTION_SENDONLY ||
+        direction == RTC_RTP_TRANSCEIVER_DIRECTION_RECVONLY;
+}
+
+// Parse remote offer media direction and return our perspective (e.g. remote sendonly -> we recvonly).
+// Used when applying remote offer to update transceiver directions for re-negotiation.
+// Returns the remote m-line's declared direction (raw, from the remote's perspective);
+// INACTIVE if the m-line has no direction attribute. No inversion here on purpose:
+// intersectTransceiverDirection() does the send<->recv crossover between local and remote.
+static RTC_RTP_TRANSCEIVER_DIRECTION getRemoteDirectionFromMediaDescription(PSdpMediaDescription pMediaDescription)
```
unicornss: /nit parseRemoteDirectionFromMediaDescription to match the doc string

samples/common/Common.c
```diff
@@ -1570,9 +1570,52 @@ STATUS signalingMessageReceived(UINT64 customData, PReceivedSignalingMessage pRe
...
+                    CHK_STATUS(handleOffer(pSampleConfiguration, pSampleStreamingSession, &pReceivedSignalingMessage->signalingMessage));
+                    startStats = pSampleConfiguration->iceCandidatePairStatsTimerId == MAX_UINT32;
+                    break;
+                }
+
+                /* Session replacement: the old connection is terminated (DISCONNECTED/FAILED/CLOSED).
+                 * Clean up the old session and fall through to create a new one. */
+                DLOGI("Session replacement: replacing terminated session for peer %s", pReceivedSignalingMessage->signalingMessage.peerClientId);
+
+                // Remove old session from hash table
```
unicornss: is the ordering is safe? could another thread race ?

tst/PeerConnectionFunctionalityTest.cpp
```diff
@@ -1549,6 +1553,431 @@ TEST_F(PeerConnectionFunctionalityTest, parseExtIdHandlesNull)
...
+    this->threads.clear();
+    this->noNewThreads = TRUE;
+    this->lock.unlock();
+
+    closePeerConnection(offerPc);
+    closePeerConnection(answerPc);
+
+    freePeerConnection(&offerPc);
+    freePeerConnection(&answerPc);
+}
```
unicornss, line 1979: Good coverage on sequential re-negotiations. Should we also consider:

Same Session, Concurrent Threads: Run a writeFrame loop on one thread while triggering setRemoteDescription (direction → INACTIVE) on another thread for the same session. Confirm there are no crashes or unexpected access (UAF), and verify that writeFrame eventually becomes a no-op.

Multiple Sessions, Concurrent Re-offers: Simulate two viewers sending re-offers to the same master simultaneously. Confirm safe mutations of the session list/hash table in the signaling callback under contention.

unicornss (commented):
Thanks for your contributions!
Suggest considering the PR description as - Handle SDP re-offers on **active** P2P connections (track add/remove/direction change)

src/source/PeerConnection/SessionDescription.c
vikramdattu: We now warn on code change

src/source/PeerConnection/SessionDescription.c
vikramdattu: Renamed.

samples/common/Common.c
vikramdattu: Good question! I went ahead and tested it rather than reasoning about it, and it turned up more than I expected.

The ordering is safe. Both the session list and the hash table are only mutated while
`sampleConfigurationObjLock` is held, and the session is taken out of the list — under
`streamingSessionListReadLock`, which the `writeFrame` loops hold for their whole walk — before it
is freed, so no media thread can still reach it. But it was safe more by accident than by design,
and it didn't follow the order `sessionCleanupWait()` already uses. Fixed in 49bb7ac: both
structures are now unlinked in one critical section, the free happens after that, a failed
`hashTableRemove` no longer bails out and leaks the session, `freeSampleStreamingSession()`'s
status is no longer discarded, and the locking rule for the list is written down at the field
instead of being something you infer from the neighbouring samples.

Two things came out of the testing.

**There was a real race in this callback, just not on the list — fixed in 8879a16.**
`enableIceStats` was written in `createSampleStreamingSession()` under the lock and read in
`signalingMessageReceived()` after the unlock; since each signaling message is dispatched on its own
thread, two offers can be in the callback at once, and ThreadSanitizer flagged the pair on a
re-offer run. The value is now snapshotted while the lock is held. The same write was also a
functional bug: it always landed before `initializePeerConnection()` read the field, so
`enableIceStats` was pinned to FALSE and `ENABLE_STATS_CALCULATION_CONTROL` did nothing in the
samples once a session existed. The field is already zeroed by `MEMCALLOC`, so dropping the
assignment keeps the default and lets an application actually enable it. Both are pre-existing on
`develop` rather than from this PR, but they surfaced here so I fixed them in their own commit.

**I could not get the session-replacement branch to execute, but it is reachable and worth being
correct.** On a desktop the teardown wins easily — `onConnectionStateChange()` sets `terminateFlag`
and broadcasts the cvar, and `sessionCleanupWait()` frees the session in the same millisecond as the
disconnect:

```
14:44:43.344 INFO  onConnectionStateChange(): New connection state 6
14:44:43.344 DEBUG onConnectionStateChange(): p2p connection disconnected
14:44:43.344 DEBUG freeSampleStreamingSession(): Freeing streaming session with peer id: aiortcViewer
```

That is a race, though, not an ordering guarantee. `onConnectionStateChange()` does not hold
`sampleConfigurationObjLock` — it only sets the flag and broadcasts — so `sessionCleanupWait()` still
has to wake and acquire that lock, while an incoming offer takes the same lock at the top of
`signalingMessageReceived()`. Whichever gets there first decides, and if the offer wins it finds the
session still in the hash with `terminateFlag` set, which is exactly this branch. The window widens
on a constrained target: one core, the cleanup task likely lower priority than the signaling task,
and the teardown itself running under the lock — `closePeerConnection()` plus `freePeerConnection()`
measured ~200 ms even on my desktop — against a viewer that can re-offer over an already-open
signaling socket in milliseconds. So it stays, and the fixes above matter for it.

For completeness I also tried the ingestion side on a storage-enabled channel: there the master never
sees a viewer-originated offer at all, and the backend's next offer only arrives after
`sessionCleanupWait()` has already freed the session and reconnected.

So to be straight about it: the fix to this block is reasoned rather than exercised. I could not get
the branch to run on demand in any setup I could build, and nothing in `webrtc_client_test` reaches
it, so it is reviewed-and-argued, not test-covered. If you know a setup that does hit it, I'll run it.

On coverage, for the record: `samples/` isn't compiled into `webrtc_client_test`, and the C viewer
randomises its client id, so no sample run reaches the re-offer path — the master just creates two
independent sessions. I ended up driving it with a small aiortc viewer on a fixed client id. With
that, the re-offer path runs clean under ThreadSanitizer and media keeps flowing across the
renegotiation (video packets went 1382 -> 3934 with the connection staying up). The only TSan
warnings left on the master are pre-existing and unrelated: a `peerConnectionObjLock` / DTLS lock
order inversion between `createAnswer()` and `closePeerConnection()`, the same family as the
`deadlock:dtlsSessionStart` entry already in `tst/suppressions/TSAN.supp`.

vikramdattu:
@sirknightj @unicornss I have addressed the review comments. And the other piece of the puzzle, the re-neogiation option on JS Test Page it available to try as well, request you to take a look again.

sirknightj (approved):

YPandas (approved):

PR #2218, opened 2026-02-14
Configurability for pre-generated certs

sirknightj:
*Issue #, if available:*
- N/A

*What was changed?*
- Add new environment variables: `KVS_PRE_GENERATE_CERT`, `KVS_PRE_GENERATE_CERT_PERIOD_MS`, and `KVS_PRE_GENERATE_CERT_MAX` to configure certificate pre-generation (sample application)
- Changed the defaults
- Updated README

*Why was it changed?*
- Certificate pre-generation settings were hardcoded, requiring code changes to adjust behavior for different use cases or testing
- README is outdated

- Changed the defaults to reduce CPU load on startup since it's done in a timerQueue thread (background thread)

| Var | Old |  New |
| ------ | ------- | ------ |
| KVS_PRE_GENERATE_CERT_PERIOD_MS | 1s | 3s 
| KVS_PRE_GENERATE_CERT_MAX | 3 | 2 |

*How was it changed?*
- Use environment variables to configure

- CI/CD
- Run it locally and checked the logs to verify the behavior


samples/Samples.h
```diff
@@ -41,7 +41,12 @@ extern "C" {
 #define SAMPLE_VIDEO_FRAME_DURATION (HUNDREDS_OF_NANOS_IN_A_SECOND / DEFAULT_FPS_VALUE)
 
 #define SAMPLE_PRE_GENERATE_CERT        TRUE
-#define SAMPLE_PRE_GENERATE_CERT_PERIOD (1000 * HUNDREDS_OF_NANOS_IN_A_MILLISECOND)
+#define SAMPLE_PRE_GENERATE_CERT_PERIOD (3000 * HUNDREDS_OF_NANOS_IN_A_MILLISECOND)
+#define SAMPLE_PRE_GENERATE_CERT_MAX    2
+
+#define PRE_GENERATE_CERT_ENV_VAR        ((PCHAR) "KVS_PRE_GENERATE_CERT")
+#define PRE_GENERATE_CERT_PERIOD_ENV_VAR ((PCHAR) "KVS_PRE_GENERATE_CERT_PERIOD_MS")
+#define PRE_GENERATE_CERT_MAX_ENV_VAR    ((PCHAR) "KVS_PRE_GENERATE_CERT_MAX")
```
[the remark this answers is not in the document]
sirknightj: Don't think so

```c
/**
 * Max certificates an RtcConfiguration can accept
 */
#define MAX_RTCCONFIGURATION_CERTIFICATES 3
```

The sample app only passes in 1 pre-generated cert to each peerconnection. This variable controls the pending queue size (burst of peerconnections).

From the stackQueue (singleList/linked list), the max elements count would be uint32 which no one will realistically set (you'd run out of memory first).

I added the log:

```c
    DLOGD("Pre-generate cert max standby: %d", preGenerateCertMax);
```

samples/common/Common.c
```diff
@@ -1147,7 +1191,7 @@ STATUS pregenerateCertTimerCallback(UINT32 timerId, UINT64 currentTime, UINT64 c
 
     // Quick check if there is anything that needs to be done.
     CHK_STATUS(stackQueueGetCount(pSampleConfiguration->pregeneratedCertificates, &certCount));
-    CHK(certCount != MAX_RTCCONFIGURATION_CERTIFICATES, retStatus);
+    CHK(certCount != pSampleConfiguration->pregenerateCertificatesMax, retStatus);
```
sirknightj: Note: This fn is called every "interval" seconds. This check says "if the queue is full, return early"

Ahhorse1 (dismissed):

Ahhorse1 (approved):

samples/common/Common.c
vabhasin: Is there a possibility we could exceed this and generate more certs than max? Not sure if this is the only place for cert generation.
Basically suggesting if we should change this to `certCount < pSampleConfiguration->pregenerateCertificatesMax`

sirknightj: It's not possible since the timer queue thread (this) is single-threaded.

The suggestion would help to make the code more readable, so will adjust it.

vabhasin (approved):

PR #2225, opened 2026-02-23
Use parameterized test for the transceiver direction matrix and default for invalid args

sirknightj:
*Issue #, if available:*
- N/A

*What was changed?*
- Change the unit test format for the IntersectTransceiverDirectionTest
- Matrix contains 16 entries (4x4)
- Also handle the case when the enum input is invalid (default to sendrecv), just in case. Redundant safety check since it should be sanitized by #2223 PR.

*Why was it changed?*
- Making it easier to verify the results of the function

*How was it changed?*
- Use parameterized test
- Check the enum value range

- Ran the new test locally

```
./tst/webrtc_client_test --gtest_filter="*SdpFunctionalityTest*"
...
[       OK ] SdpFunctionalityTest/IntersectTransceiverDirectionTest.AllScenarios/15 (408 ms)
[----------] 16 tests from SdpFunctionalityTest/IntersectTransceiverDirectionTest (6525 ms total)

[----------] Global test environment tear-down
[==========] 25 tests from 2 test suites ran. (10202 ms total)
[  PASSED  ] 25 tests.
```


tst/SdpFunctionalityTest.cpp
```diff
@@ -60,62 +60,66 @@ TEST_F(SdpFunctionalityTest, writeTransceiverDirection_bufferTooSmall)
...
-    EXPECT_EQ(intersectTransceiverDirection(RTC_RTP_TRANSCEIVER_DIRECTION_SENDONLY, RTC_RTP_TRANSCEIVER_DIRECTION_RECVONLY),
-              RTC_RTP_TRANSCEIVER_DIRECTION_SENDONLY);
-    EXPECT_EQ(intersectTransceiverDirection(RTC_RTP_TRANSCEIVER_DIRECTION_RECVONLY, RTC_RTP_TRANSCEIVER_DIRECTION_SENDONLY),
-              RTC_RTP_TRANSCEIVER_DIRECTION_RECVONLY);
+    RTC_RTP_TRANSCEIVER_DIRECTION transceiver, offer, expected;
+    std::tie(transceiver, offer, expected) = GetParam();
+    EXPECT_EQ(intersectTransceiverDirection(transceiver, offer), expected);
 }
 
+// tuple: [transceiver (local), offer (remote), generated answer]
```
unicornss: // Transceiver direction intersection follows RFC 8829 offer/answer semantics: 
// https://datatracker.ietf.org/doc/html/rfc8829#name-setdirection
// - The answer direction is the intersection of local capabilities and remote offer
// - "send" in local matches "recv" in remote (and vice versa)
// - "inactive" always results in "inactive"
// - Incompatible directions (e.g., both sendonly) result in "inactive"

unicornss (approved):

PR #2251, opened 2026-04-06
add hostname verification for stuns

jgooge:
*Issue #, if available:*
- N/A

*What was changed?*
- Added strict CA + hostname validation for the `stuns:` DTLS path while preserving the existing relaxed peer-DTLS/fingerprint flow.
- Added a GovCloud `stuns:` integration test for srflx gathering and a DTLS test that verifies strict validation rejects an untrusted server certificate.
- Made the OpenSSL hostname-verification path version-aware so older OpenSSL support is preserved.

*Why was it changed?*
- We wanted `stuns:` to fail closed on untrusted or misconfigured certificates and to add coverage that proves both the success and rejection paths.
- The OpenSSL follow-up keeps the new validation behavior without breaking older supported OpenSSL versions at compile time.

*How was it changed?*
- Added an internal DTLS validation mode/options path and used it only for `stuns:` sockets.
- Plumbed the parsed STUNS hostname through ICE/socket setup so DTLS can load the CA bundle, set SNI, and verify the expected hostname.
- Added explicit validation logging and version-gated OpenSSL hostname verification:
  - `>= 1.1.0`: `SSL_set1_host(...)`
  - `>= 1.0.2` and `< 1.1.0`: `X509_VERIFY_PARAM_set1_host(...)`
  - `< 1.0.2`: fail closed with `STATUS_NOT_IMPLEMENTED`

- Added `IceFunctionalityTest.IceAgentGovCloudStunsCandidateGatheringTest`
- Added `DtlsFunctionalityTest.strictServerValidationRejectsUntrustedServerCertificate`
- Verified local formatting with `clang-format` / `git diff --check`
- Verified local OpenSSL build with `cmake --build build-test-openssl --target kvsWebrtcClient`
- GitHub Actions on the fork were used to validate the new integration/unit coverage and follow-up fixes
- Full local `webrtc_client_test` execution is still blocked on this machine by the pre-existing `arm64` vs `x86_64` `libgtest.a` link mismatch


src/source/Crypto/Dtls_mbedtls.c
```diff
@@ -8,8 +8,123 @@ mbedtls_ssl_srtp_profile DTLS_SRTP_SUPPORTED_PROFILES[] = {
     MBEDTLS_TLS_SRTP_UNSET,
 };
 
+static STATUS dtlsReadAndParseCACertificate(PDtlsSession pDtlsSession)
+{
+    ENTERS();
+    STATUS retStatus = STATUS_SUCCESS;
+    UINT64 certLen = 0;
+    PBYTE certBuf = NULL;
+    CHAR errBuf[128];
```
YPandas: Can we extract this 128 to a #define?

src/source/Crypto/Dtls.h
```diff
@@ -65,6 +65,16 @@ typedef struct {
     DtlsSessionOnStateChange stateChangeFn;
 } DtlsSessionCallbacks, *PDtlsSessionCallbacks;
 
+typedef enum {
+    DTLS_SESSION_VALIDATION_MODE_RELAXED,
```
YPandas: Can we comment out the meaning of these mode?

src/source/Crypto/Dtls_mbedtls.c
```diff
@@ -8,8 +8,123 @@ mbedtls_ssl_srtp_profile DTLS_SRTP_SUPPORTED_PROFILES[] = {
     MBEDTLS_TLS_SRTP_UNSET,
 };
 
+static STATUS dtlsReadAndParseCACertificate(PDtlsSession pDtlsSession)
```
YPandas, line 14: Is there a reason why we want this to be static?

src/source/Crypto/Dtls_mbedtls.c
```diff
@@ -8,8 +8,123 @@ mbedtls_ssl_srtp_profile DTLS_SRTP_SUPPORTED_PROFILES[] = {
...
+    CHK(sslRet == 0, STATUS_INVALID_CA_CERT_PATH);
+
+CleanUp:
+    CHK_LOG_ERR(retStatus);
+    SAFE_MEMFREE(certBuf);
+    LEAVES();
+    return retStatus;
+}
+
+STATUS dtlsSessionCopyOptions(PDtlsSession pDtlsSession, PDtlsSessionOptions pDtlsSessionOptions)
```
YPandas: recommend to put in the .h file with documentation

src/source/Crypto/Dtls_mbedtls.c
```diff
@@ -8,8 +8,123 @@ mbedtls_ssl_srtp_profile DTLS_SRTP_SUPPORTED_PROFILES[] = {
...
+    pDtlsSession->pExpectedServerHostname = NULL;
+
+    if (pDtlsSessionOptions == NULL) {
+        CHK(FALSE, retStatus);
+    }
+
+    pDtlsSession->validationMode = pDtlsSessionOptions->validationMode;
+    if (pDtlsSession->validationMode == DTLS_SESSION_VALIDATION_MODE_STRICT_SERVER) {
+        CHK(pDtlsSessionOptions->pExpectedServerHostname != NULL && pDtlsSessionOptions->pExpectedServerHostname[0] != '\0', STATUS_INVALID_ARG);
+        hostnameLen = (UINT32) STRLEN(pDtlsSessionOptions->pExpectedServerHostname);
```
YPandas: Can we validate if hostnameLen doesn't exceed MBEDTLS_SSL_MAX_HOST_NAME_LEN?

src/source/Crypto/Dtls_mbedtls.c
```diff
@@ -8,8 +8,123 @@ mbedtls_ssl_srtp_profile DTLS_SRTP_SUPPORTED_PROFILES[] = {
...
+}
+
+STATUS dtlsSessionCheckRemoteCertificateVerification(PDtlsSession pDtlsSession)
+{
+    ENTERS();
+    STATUS retStatus = STATUS_SUCCESS;
+    UINT32 verifyResult = 0;
+
+    CHK(pDtlsSession != NULL, STATUS_NULL_ARG);
+    CHK(pDtlsSession->validationMode == DTLS_SESSION_VALIDATION_MODE_STRICT_SERVER, retStatus);
```
YPandas, line 99: Can we add a warning here that we don't recommend to use relaxed mode in production?

src/source/Crypto/Dtls_mbedtls.c
```diff
@@ -8,8 +8,123 @@ mbedtls_ssl_srtp_profile DTLS_SRTP_SUPPORTED_PROFILES[] = {
...
+        DLOGW("DTLS strict server certificate verification failed for %s. mbedTLS verify flags: 0x%08x",
+              pDtlsSession->pExpectedServerHostname != NULL ? pDtlsSession->pExpectedServerHostname : "(null)", verifyResult);
+    }
+
+    CHK_LOG_ERR(retStatus);
+    LEAVES();
+    return retStatus;
+}
+
+STATUS dtlsSessionGetCertificateVerificationFailureStatus(PDtlsSession pDtlsSession, INT32 sslRet)
```
YPandas, line 119: Can we add comment on why we need to verify the result again?
Is there an occasion we only need to verify the result without getting peer cert?

src/source/Crypto/Dtls_mbedtls.c
```diff
@@ -36,6 +152,12 @@ STATUS createDtlsSession(PDtlsSessionCallbacks pDtlsSessionCallbacks, TIMER_QUEU
     pDtlsSession->timerId = MAX_UINT32;
     pDtlsSession->sslLock = MUTEX_CREATE(TRUE);
     pDtlsSession->dtlsSessionCallbacks = *pDtlsSessionCallbacks;
+    pDtlsSession->validationMode = DTLS_SESSION_VALIDATION_MODE_RELAXED;
+    pDtlsSession->pExpectedServerHostname = NULL;
```
YPandas: It seems we will set these two variables again in dtlsSessionCopyOptions. Consider removing duplicate code.

src/source/Crypto/Dtls_mbedtls.c
```diff
@@ -36,6 +152,12 @@ STATUS createDtlsSession(PDtlsSessionCallbacks pDtlsSessionCallbacks, TIMER_QUEU
...
+    CHK_STATUS(dtlsSessionCopyOptions(pDtlsSession, pDtlsSessionOptions));
+    if (pDtlsSession->validationMode == DTLS_SESSION_VALIDATION_MODE_STRICT_SERVER) {
```
YPandas: Can you make it follow the consistent style, where the function is always called regardless of the mode and do an early return if the mode is not RELAXED?

src/source/Crypto/Dtls_mbedtls.c
```diff
@@ -8,8 +8,123 @@ mbedtls_ssl_srtp_profile DTLS_SRTP_SUPPORTED_PROFILES[] = {
...
+        retStatus = STATUS_SSL_REMOTE_CERTIFICATE_VERIFICATION_FAILED;
+    }
+
+CleanUp:
+    if (retStatus == STATUS_SSL_REMOTE_CERTIFICATE_VERIFICATION_FAILED && pDtlsSession != NULL &&
+        pDtlsSession->validationMode == DTLS_SESSION_VALIDATION_MODE_STRICT_SERVER) {
+        DLOGW("DTLS strict server certificate verification failed for %s during handshake processing. mbedTLS verify flags: 0x%08x, sslRet: %d",
+              pDtlsSession->pExpectedServerHostname != NULL ? pDtlsSession->pExpectedServerHostname : "(null)", verifyResult, sslRet);
+    }
+
```
sirknightj, line 141: Let's make sure we have the CHK log err and enters/leaves

YPandas (commented):
We have reviewed changes to embedded tls. Will come back to openssl later

src/source/Crypto/Dtls_mbedtls.c
jgooge: Resolved. Extracted the mbedTLS error string buffer size into `DTLS_MBEDTLS_ERROR_STRING_BUFFER_SIZE`

src/source/Crypto/Dtls.h
jgooge: Resolved. Added comments for both modes in Dtls.h

src/source/Crypto/Dtls_mbedtls.c
jgooge: Kept this static because it is a backend-local mbedTLS helper and should not be exposed through the shared DTLS interface. Added a short comment explaining that.

src/source/Crypto/Dtls_mbedtls.c
jgooge: Resolved by moving it into shared Dtls.c and adding its documented declaration to Dtls.h

src/source/Crypto/Dtls_mbedtls.c
jgooge: Resolved. Added an mbedTLS-only check against `MBEDTLS_SSL_MAX_HOST_NAME_LEN` before copying the strict server hostname.

src/source/Crypto/Dtls_mbedtls.c
jgooge: Resolved.

src/source/Crypto/Dtls_mbedtls.c
jgooge: Resolved. Added comments explaining that strict post-handshake validation checks both the peer certificate and verify flags, while handshake/read/write error paths can fail before the normal peer-cert check runs, so we re-read verify flags to preserve the certificate failure reason.

src/source/Crypto/Dtls_mbedtls.c
jgooge: Resolved.

src/source/Crypto/Dtls_mbedtls.c
jgooge: Resolved. Refactored the mbedTLS validation setup so helper functions are called consistently and return early when the current mode/server role does not require strict validation work.

src/source/Crypto/Dtls_mbedtls.c
jgooge: Resolved. Added `ENTERS`, `LEAVES`, and `CHK_LOG_ERR(retStatus)` to the mbedTLS certificate verification failure helper

Ahhorse1 (dismissed):

src/source/Crypto/Dtls_openssl.c
```diff
@@ -1,15 +1,108 @@
...
+
+    // Allow all certificates in the default peer DTLS path since identity is verified later via SDP fingerprint.
     return 1;
 }
 
+STATUS dtlsSessionConfigureStrictHostnameValidation(PDtlsSession pDtlsSession)
+{
+    ENTERS();
+    STATUS retStatus = STATUS_SUCCESS;
+    X509_VERIFY_PARAM* pVerifyParam = NULL;
```
sirknightj, line 33: Suggest to move this into the elif with the scoping:

```diff
 #elif (OPENSSL_VERSION_NUMBER >= 0x10002000L)
     {
+        X509_VERIFY_PARAM* pVerifyParam = SSL_get0_param(pDtlsSession->pSsl);
         CHK(pVerifyParam != NULL, STATUS_SSL_CTX_CREATION_FAILED);
         CHK(X509_VERIFY_PARAM_set1_host(pVerifyParam, pDtlsSession->pExpectedServerHostname, 0) == 1, STATUS_SSL_CTX_CREATION_FAILED);
     }
 #else
```

To avoid any unused variable compiler warnings when using the other path.
Or adding `UNUSED_PARAM(pVerifyParam)` in the other two paths.

sirknightj (approved):

Ahhorse1 (approved):

PR #2252, opened 2026-04-16
esp_port: Expose API for creating data channels and skip video media lines in offer if capture callback is not provided

shreyash-b:
- If an application wants to use the WebRTC SDK for application involving only audio streaming, the SDK is still including the video lines in the SDP offer.
- Check if video_capture callbacks are provided. If not, don't include the video lines in the SDP offer
- Earlier there was no provision to create data channels using the SDK. Data channels could be used only if the other peer created them.
- This PR exposes a function which creates a data channel with the specified label, peerId, and reliability.
- Additionally it includes a bugfix wherein the audio/video frame callbacks were not being set if we are the ones sending offer.


vikramdattu:
LGTM.

@sirknightj @unicornss PTAL.

Note: Since the new repo is not ready yet to be used and hence, the important PRs will need to go to this branch till then.

PR #2255, opened 2026-04-23
enforce AES-256-GCM SRTP encryption in GovCloud/ADC regions

jgooge:
*Issue #, if available:*
  - N/A

  *What was changed?*
  - Added SRTP_AEAD_AES_256_GCM support for SRTP media encryption per CNSA 1.0 (RFC 9151 (https://datatracker.ietf.org/doc/html/rfc9151#section-4)). In GovCloud/ADC regions, only AES-256-GCM is offered. In commercial regions, both AES-256-GCM and AES-128-CM are supported for backward compatibility.

  *Why was it changed?*
  - CNSA 1.0 requires a minimum 256-bit encryption key size for all data in transit in ADC regions. The SDK previously only offered `SRTP_AES128_CM_SHA1_80` and `SRTP_AES128_CM_SHA1_32`, which does not meet this requirement. See [RFC 9151 Section 4](https://datatracker.ietf.org/doc/html/rfc9151#section-4).

  *How was it changed?*
  - **`Crypto.h`**: Added `KVS_SRTP_PROFILE_AEAD_AES_256_GCM` to both OpenSSL and mbedTLS profile enums, with a fallback `#define` for `SRTP_AEAD_AES_256_GCM` (`0x0008`) when not provided by the crypto backend
  - **`Dtls.h`**: Increased `MAX_SRTP_MASTER_KEY_LEN` from 16 to 32 to accommodate AES-256 key size
  - **`Dtls_openssl.c`**: Region-based SRTP profile selection — GCM-only for gov/ADC (`us-gov-*`, `us-iso-*`, `us-isob-*`), GCM-preferred with AES-128-CM fallback for commercial. Updated `dtlsSessionPopulateKeyingMaterial` to dynamically determine key/salt lengths per profile (32/12 for GCM vs 16/14 for AES-128-CM)
  - **`Dtls_mbedtls.c`**: Same dynamic key/salt length handling in `dtlsSessionPopulateKeyingMaterial` for mbedTLS backend
  - **`SrtpSession.c`**: Added GCM policy using `srtp_crypto_policy_set_aes_gcm_256_16_auth` for both RTP and RTCP. Set `window_size = 1024` for replay protection on all SRTP policies
  - **`Rtp.h`**: Increased `SRTP_AUTH_TAG_OVERHEAD` from 10 to 16 for GCM's 128-bit authentication tag (per [RFC 7714](https://datatracker.ietf.org/doc/html/rfc7714))
  - **`PeerConnectionFunctionalityTest.cpp`**: Updated `bytesSent` assertion to accept both AES-128-CM and AES-256-GCM values since the auth tag size differs by profile

  - Verified AES-256-GCM negotiation via `chrome://webrtc-internals` (`srtpCipher: AEAD_AES_256_GCM`) in both master and viewer roles
  - Verified DTLS-SRTP ClientHello offers `SRTP_AEAD_AES_256_GCM` (`0x0008`) via Wireshark pcap analysis
  - Tested C SDK master with JS viewer and C SDK viewer with JS master — both directions negotiate AES-256-GCM successfully
  - Confirmed backward compatibility in commercial regions with AES-128-CM fallback
  - Passed integration tests on my forked repository


src/source/Crypto/Dtls_mbedtls.c
sirknightj, line 323: I noticed DTLS_SRTP_SUPPORTED_PROFILES array still doesn't include the new one. Could you double check?

```c
mbedtls_ssl_srtp_profile DTLS_SRTP_SUPPORTED_PROFILES[] = {
    MBEDTLS_TLS_SRTP_AES128_CM_HMAC_SHA1_80,
    MBEDTLS_TLS_SRTP_AES128_CM_HMAC_SHA1_32,
    MBEDTLS_TLS_SRTP_UNSET,
};
```

src/source/Crypto/Dtls_openssl.c
```diff
@@ -202,7 +202,18 @@ STATUS createSslCtx(PDtlsSessionCertificateInfo pCertificates, UINT32 certCount,
 #endif
 
     SSL_CTX_set_verify(pSslCtx, SSL_VERIFY_PEER | SSL_VERIFY_FAIL_IF_NO_PEER_CERT, dtlsCertificateVerifyCallback);
-    CHK(SSL_CTX_set_tlsext_use_srtp(pSslCtx, "SRTP_AES128_CM_SHA1_32:SRTP_AES128_CM_SHA1_80") == 0, STATUS_SSL_CTX_CREATION_FAILED);
+
+    // Use AES-256-GCM only for CNSA 1.0 compliance in gov/ADC regions
+    PCHAR pRegion = GETENV(DEFAULT_REGION_ENV_VAR);
```
sirknightj: The region is read from the `AWS_DEFAULT_REGION` env var here, but applications can set the region programmatically via `channelInfo.pRegion` without setting the env var. In that case, GCM-only enforcement won't trigger (silently).

Consider: should the region or desired SRTP profile list be passed down from the caller so it works regardless of how the region is configured?

tst/PeerConnectionFunctionalityTest.cpp
```diff
@@ -794,10 +794,11 @@ TEST_F(PeerConnectionFunctionalityTest, exchangeMedia)
     RtcOutboundRtpStreamStats stats{};
     EXPECT_EQ(STATUS_SUCCESS, getRtpOutboundStats(offerPc, offerVideoTransceiver, &stats));
     EXPECT_EQ(206, stats.sent.packetsSent);
+    // bytesSent varies by SRTP profile: AES-128-CM (10-byte tag) vs AES-256-GCM (16-byte tag)
 #ifdef KVS_USE_MBEDTLS
     EXPECT_EQ(248026, stats.sent.bytesSent);
 #else
-    EXPECT_EQ(246790, stats.sent.bytesSent);
+    EXPECT_TRUE(stats.sent.bytesSent == 246790 || stats.sent.bytesSent == 249262);
```
sirknightj: Suggest to put this in an if/else for commercial and fips.

src/source/Crypto/Dtls_openssl.c
```diff
@@ -202,7 +210,20 @@ STATUS createSslCtx(PDtlsSessionCertificateInfo pCertificates, UINT32 certCount,
 #endif
 
     SSL_CTX_set_verify(pSslCtx, SSL_VERIFY_PEER | SSL_VERIFY_FAIL_IF_NO_PEER_CERT, dtlsCertificateVerifyCallback);
-    CHK(SSL_CTX_set_tlsext_use_srtp(pSslCtx, "SRTP_AES128_CM_SHA1_32:SRTP_AES128_CM_SHA1_80") == 0, STATUS_SSL_CTX_CREATION_FAILED);
+
+    // Use AES-256-GCM only for CNSA 1.0 compliance in gov/ADC regions
+    if (IS_NULL_OR_EMPTY_STRING(pRegion)) {
+        pRegion = GETENV(DEFAULT_REGION_ENV_VAR);
```
sirknightj, line 216: Would you be able to add the debug lines for the region selection?

Especially for default region when unset

And which path for "dtlsShouldUseGcmOnlySrtpProfile" it went in

src/include/com/amazonaws/kinesis/video/webrtcclient/Include.h
```diff
@@ -1437,6 +1437,8 @@ typedef struct {
     BOOL generateRSACertificate; //!< GenerateRSACertificate controls if an ECDSA or RSA certificate is generated.
                                  //!< By default we generate an ECDSA certificate but some platforms may not support them.
 
+    PCHAR pRegion; //!< Optional AWS region used for region-specific media security policy.
```
sirknightj, line 1440: I think we should call out in the documentation:

- The default
- A bit more specific on what it's used for (why someone should set it). We can mention currently affects DTLS cipher suites in commercial/FIPS regions. Also note lowercase restriction and an example

sirknightj: One other consideration:

If a user haven't initialized this new member (upgrading from older version and didn't memset this value), we'll potentially be reading uninitialized memory. We typically get around this by using versioned structs, or using a constructor that zero-initializes this struct + setters. However, that's not currently possible while preserving backwards compatibility.

Would suggest doing something similar to `ENABLE_STATS_CALCULATION_CONTROL` member below

src/source/Crypto/Dtls_openssl.c
jgooge: I pushed [743a9b6](https://github.com/awslabs/amazon-kinesis-video-streams-webrtc-sdk-c/pull/2255/commits/743a9b649c82bb1fc2baa1d63380e566e3697fc4) which passes the resolved region into DTLS and lets DTLS pick the SRTP profiles. Locally, I’ve since refactored it so the caller resolves the SRTP profile list and passes that into DTLS instead. That feels a bit cleaner since DTLS no longer needs to know about AWS regions, and it would probably be easier to extend later if we add something like a FIPS config flag. Would you prefer I push that version, or keep the current region-passing approach?

PR #2257, opened 2026-04-27
Samples: Defer cert pre-generation when there are active PeerConnections

sirknightj:
*Issue #, if available:*
- N/A

*What was changed?*
- Added an early return in `pregenerateCertTimerCallback` that defers certificate pre-generation when there are active streaming sessions

*Why was it changed?*
- When a new peer connection is created, a pre-generated certificate is dequeued from the pre-generated certificate queue. The timer callback then attempts to refill the queue (almost) immediately. Certificate generation involves CPU-intensive key pair creation that can impact the performance of active peer connections (e.g., increased latency or jitter during media streaming), more noticeable in lower-powered devices.
- This change defers refilling the queue until after all streaming sessions have ended.

*How was it changed?*
-  Added a check for `pSampleConfiguration->streamingSessionCount == 0` before proceeding with cert generation in `samples/common/Common.c`. If sessions are active, the callback returns early and defers generation to a future timer invocation.

- CI/CD
- Ran locally with an immediate viewer peerconnection after master start. Observed the "New certificate has been pre-generated and added to the queue" log only after the viewer left (not during).

### DIFF
--- a/samples/common/Common.c
+++ b/samples/common/Common.c
@@ -1150,6 +1150,9 @@ STATUS pregenerateCertTimerCallback(UINT32 timerId, UINT64 currentTime, UINT64 c
         locked = TRUE;
     }
 
+    // Don't pre-generate certs when there are active streaming sessions to avoid impacting existing connections
+    CHK(pSampleConfiguration->streamingSessionCount == 0, retStatus);
+
     // Quick check if there is anything that needs to be done.
     CHK_STATUS(stackQueueGetCount(pSampleConfiguration->pregeneratedCertificates, &certCount));
     CHK(certCount != MAX_RTCCONFIGURATION_CERTIFICATES, retStatus);


Ahhorse1 (approved):

unicornss (commented):
helps in low power devices

unicornss (approved):

samples/common/Common.c
```diff
@@ -1150,6 +1150,9 @@ STATUS pregenerateCertTimerCallback(UINT32 timerId, UINT64 currentTime, UINT64 c
         locked = TRUE;
     }
 
+    // Don't pre-generate certs when there are active streaming sessions to avoid impacting existing connections
+    CHK(pSampleConfiguration->streamingSessionCount == 0, retStatus);
```
vikramdattu, line 1154: Good change but, can lead to problem if active connections reach max  pre-gen certs count. Recommend  defering it only when there is a peer connection process happening.

sirknightj: In the case all the pre-generated certs are used up (pool empty), no cert will be passed into the constructor. In that case, the SDK will generate the cert during connection phase. New connections won't fail, instead they will be slightly slower since the cert generation is happening during the connection phase instead of outside.

I added sessionStreamingCount=0 instead of just "any PeerConnection in connecting state" to avoid jitter on the streaming path (minimize impact to existing connections). In #2218 we have the ability to configure the pool size so users can tune it according to their requirements.

sirknightj:
CI/CD measured:

Before: 2026-04-20 03:07:13.035
After: 2026-04-27 21:27:09.035

Unit: ms 
Cand pair: host-host

| Metric  | Avg    | StdDev   | Min   | P50    | P90    | Max    | Count   |
|-----|-----|-----|-----|-----|-----|-----|-----|
| Time to first frame  ([before](https://github.com/awslabs/amazon-kinesis-video-streams-webrtc-sdk-c/actions/runs/24646394910/job/72059822957))  | 348.8  | 40.1     | 308.7 | 343.8  | 403.9  | 458.9  | 50      |
| Time to first frame   ([after](https://github.com/awslabs/amazon-kinesis-video-streams-webrtc-sdk-c/actions/runs/25020232322/job/73278620065))   | 280.2 | 33.7     | 257.1 | 259.0 | 323.3 | 407.7 | 50      |

vabhasin (approved):

PR #2264, opened 2026-05-04
fix(RtpOpusPayloader): handle empty input cleanly

vikramdattu:
*Issue #, if available:*
- N/A

*What was changed?*
- `createPayloadForOpus` now early-returns with `payloadLength=0` and `payloadSubLenSize=0` when `opusFrameLength==0`.

*Why was it changed?*
- Previously `payloadSubLenSize=1` was set unconditionally, so on an empty input `writeFrame` would `MEMALLOC(pPacketList)` for one packet, but `constructRtpPackets` would early-return on `payloadLength > 0` without filling it — the for-loop then read uninitialized memory (observed as `MEMALLOC` failing for multi-GB sizes via garbage `RTP_GET_RAW_PACKET_SIZE`, or `srtp_protect → replay_fail` from corrupted seq numbers).
- The H264 payloader already handles empty input via `CHK(remainNalusLength != 0, retStatus)`. This change brings Opus to the same shape.

*How was it changed?*
- One-line `CHK(opusFrameLength > 0, retStatus)` in `createPayloadForOpus`. CleanUp leaves both out-params at their zero initialisers, so writeFrame's loop iterates zero times.

- New unit test `RtpFunctionalityTest.packingEmptyOpusFrameReturnsZeroSubLenSize` verifies `payloadLength=0` and `payloadSubLenSize=0` for empty input.
- Existing `packingUnpackingVerifySameOpusFrame` continues to pass.
- `scripts/check-clang.sh` clean.

### DIFF
--- a/src/source/Rtp/Codecs/RtpOpusPayloader.c
+++ b/src/source/Rtp/Codecs/RtpOpusPayloader.c
@@ -14,6 +14,7 @@ STATUS createPayloadForOpus(UINT32 mtu, PBYTE opusFrame, UINT32 opusFrameLength,
 
     CHK(opusFrame != NULL && pPayloadSubLenSize != NULL && pPayloadLength != NULL && (sizeCalculationOnly || pPayloadSubLength != NULL),
         STATUS_NULL_ARG);
+    CHK(opusFrameLength > 0, retStatus);
 
     payloadLength = opusFrameLength;
     payloadSubLenSize = 1;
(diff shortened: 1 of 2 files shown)


vikramdattu:
@sirknightj @unicornss

sirknightj (approved):
From rfc6716: https://datatracker.ietf.org/doc/html/rfc6716
> A well-formed Opus packet MUST contain at least one byte [[R1](https://datatracker.ietf.org/doc/html/rfc6716#ref-R1)].

Would suggest returning a warning log for observability for invalid frames as the application shouldn't try submitting invalid frames.

unicornss (approved):

Issue #2263, opened 2026-04-30
[Bug]: Leaking File Descriptor every 2 hours


tross-skybell:
### Please confirm you have already done the following

- [x] I have searched the repository for related/existing bug reports
- [x] I have all the details the issue requires

### Please answer the following prompt

- [x] This issue is replicable using the unmodified sample application

### Describe the bug

If you let the sample code run so that it does the 2 hour TURN/ICE credential TTL cycle it will leak 1 FD each time. This is especially bad for embedded systems where the code runs for days (weeks?) at a time before restarting.

_It also seems to leak FDs if it has a network re-connect event. Still working on reproducible steps for this._

### Expected Behavior

The number of open File Descriptors should not be continually increasing as the program runs.

### Current Behavior

Every 2 hours I see the following (expected) logs

`[2026/04/29 20:13:24:3937] N: __lws_lc_untag:  -- [wsi|0|pipe] (0) 255.541ms
[2026/04/29 20:13:24:3949] N: __lws_lc_untag:  -- [vh|1|default||-1] (1) 256.366ms
[2026/04/29 20:13:24:3950] N: __lws_lc_untag:  -- [wsicli|0|GET/h1/default/c1duopey9fjeq4.credentials.iot.us-eas] (0) 240.609ms
[2026/04/29 20:13:24:3951] N: __lws_lc_untag:  -- [vh|0|netlink] (0) 256.895ms
2026-04-30 00:13:24.396 PROFILE executeGetTokenSignalingState(): [Get token call] Time taken: 258 ms
2026-04-30 00:13:24.397 PROFILE executeDescribeSignalingState(): [Describe signaling call] Time taken: 0 ms
2026-04-30 00:13:24.397 PROFILE executeGetEndpointSignalingState(): [Get endpoint signaling call] Time taken: 0 ms
[2026/04/29 20:13:24:3984] N: __lws_lc_tag:  ++ [wsicli|4|POST/h1/default/r-0606824a.kinesisvideo.us-east-1.am] (1)
2026-04-30 00:13:24.582 PROFILE executeGetIceConfigSignalingState(): [Get ICE config signaling call] Time taken: 185 ms
2026-04-30 00:13:24.783 PROFILE executeConnectSignalingState(): [Connect signaling call] Time taken: 200 ms`

Each time this happens the number of open FDs increases

//Before
sudo ls -al /proc/2820404/fd | wc -l
19

//After
[...]

PR #2269, opened 2026-05-05
Update libwebsockets to fix FD leak

tross-skybell:
*Issue #, if available:*
#2263 

*What was changed?*
Update libwebsockets to latest stable (4.3.10)

*Why was it changed?*
Fix file descriptor leak by incorporating fix in libwebsockets
Specifically this commit:
https://github.com/warmcat/libwebsockets/commit/b9c3ed4cd1032e9b14cda733a753aad868b5382b

*How was it changed?*
Updated libwebsockets-CMakeLists.txt with appropriate GIT_TAG

Run default kvsWebrtcClientMaster sample. Without the change it leaks 1 file descriptor every 2 hours (the lws ICE TTL reconnect). Can be verified with:
> ls -l /proc/<proc>/fd | wc -l.
With the updated libwebsockets change leak stops.

### DIFF
--- a/CMake/Dependencies/libwebsockets-CMakeLists.txt
+++ b/CMake/Dependencies/libwebsockets-CMakeLists.txt
@@ -28,7 +28,7 @@ endif()
 
 ExternalProject_Add(project_libwebsockets
     GIT_REPOSITORY    https://github.com/warmcat/libwebsockets.git
-    GIT_TAG           v4.3.5
+    GIT_TAG           v4.3.10
     GIT_PROGRESS      TRUE
     GIT_SHALLOW       TRUE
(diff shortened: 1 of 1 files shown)


vabhasin:
Thanks for providing a fix, we are currently running some additional validations based on this change.

vabhasin:
We ran some tests and can confirm the behavior where fd count increases by 1 when a reconnect event occurs after the credential duration has expired. The same test does not result in fd leak with updated `libwebsockets` version `4.3.10`.

@tross-skybell thank you for reporting the issue and creating the PR for fix. I'm changing the base branch of this PR to `develop`.

sirknightj (approved):

unicornss (approved):

vabhasin (approved):

PR #2274, opened 2026-05-07
fix(esp_port): mempool static fix + signaling cache v2 + esp_video bump

vikramdattu:
## Summary

Three independent fixes in `esp_port/`:

### 1. `network_coprocessor/`: mempool / NimBLE link conflict (`dc83686500`)

Backport of upstream [esp-hosted-mcu c8e59587](https://github.com/espressif/esp-hosted-mcu/commit/c8e59587).

- Mark all `os_mempool_*` / `os_memblock_*` functions in `mempool_ll.c` as `static` so they no longer collide with NimBLE's identical symbols when BT NimBLE is enabled (e.g. RainMaker BLE provisioning on the C6 in `split_mode/rmaker_split_camera`).
- Add a `struct mempool_ops_t` ops table and an `os_mempool_get_ops()` accessor; route `hosted_mempool_*` calls through the ops table so the implementation can be swapped without re-exporting the LL symbols.
- Move per-pool metadata struct allocations to plain `calloc()` (default heap, SPIRAM-eligible where available); data buffers continue to use `heap_caps_malloc` with `INTERNAL | DMA | 8BIT`.

Without this fix, building any image that links *both* `network_coprocessor` and IDF's `bt` (NimBLE) fails at link time with multiple-definition errors on `os_memblock_put`, `os_memblock_put_from_cb`, `os_mempool_info_get_next`. The same fix is already in `espressif/esp-hosted-mcu` upstream (commit c8e59587 + follow-up 4b420272).

### 2. `kvs_signaling/SignalingESP.c`: bump `SIGNALING_CACHE_VERSION` v1 → v2 (`2eb8bda92c`)

Keeps the ESP-port NVS cache version in sync with `src/source/Signaling/FileCache.h`'s `DEFAULT_CACHE_FILE_PATH` bump.

`SignalingFileCacheEntry`'s binary layout gained two new fields between `region` and `httpsEndpoint` upstream — `controlPlaneUrl` and `useDualStackEndpoints` — in commit [`795a60d4f7`](https://github.com/awslabs/amazon-kinesis-video-streams-webrtc-sdk-c/commit/795a60d4f7) ("Signaling Cache Improvements"), landed on `beta-reference-esp-port` via `b6396850a4` ("Release v1.16.0", 2026-02-17). The upstream file-cache path bumped its filename suffix at the same time (`./.SignalingCache_v0` → `./.SignalingCache_v1`), but the ESP-port equivalent `SIGNALING_CACHE_VERSION` constant — which forms the prefix of the NVS key — was left at `"v1"`.

Devices flashed with both layouts under the same `v1` key read stale-shape blobs into the new struct: every field after `region` is shifted by `sizeof(controlPlaneUrl) + sizeof(useDualStackEndpoints)`, so `channelEndpointWss` / `channelEndpointHttps` / `storageStreamArn` / `channelEndpointWebrtc` come out as garbage. `channelArn` still reads correctly because it sits before the inserted fields.

**Symptom**: `signAwsRequestInfoQueryParam` returns `STATUS_INVALID_ARG (0x2)`, bubbled up from `getRequestHost()` failing because the corrupted WSS URL has no `://`.

Bumping the cache version orphans old blobs (they remain under the v1 NVS key but are simply never read again) and forces fresh entries under v2, written with the current layout.

### 3. `media_stream`: bump `esp_video` constraint `~1.4.0` → `~2.0.0` (`cb4244f583`)

Pulls upstream [`f5cd4bd1`](https://github.com/espressif/esp-video-components/commit/f5cd4bd1) ("Fix the WBG stop issue after starting the WBG module twice", first in `esp_video` v2.0.0). Without it, repeated camera start/stop cycles leave WBG in an inconsistent state — `isp_stop_wbg` skips the actual `esp_isp_wbg_disable()` call when red/blue balance gain ≠ 1.0 — and the next `VIDIOC_STREAMON` returns `EBUSY` (errno 16). Compatible with `esp32_p4_function_ev_board` ^5.1.0 (resolves to 5.2.3, which itself requires `esp_video ~2.0`).

## Test plan

- [x] `webrtc_classic` example builds clean on ESP32-P4 + C6 transport (default `CONFIG_ESP_CACHE_MALLOC=y`)
- [x] `rmaker_split_camera` (esp-rainmaker) builds for ESP32-C6 with BT NimBLE enabled — link errors gone
- [x] On a device previously running pre-`v1.16.0` firmware, signaling progresses past `CONNECTING` (state 8) without `signAwsRequestInfoQueryParam` failing 0x2
- [x] Camera survives multiple consecutive start/stop cycles — no `wbg isn't enabled yet` errors, no `VIDIOC_STREAMON … EBUSY`
- [x] `network_adapter` example regression check

### DIFF
--- a/esp_port/components/network_coprocessor/mempool_ll.c
+++ b/esp_port/components/network_coprocessor/mempool_ll.c
@@ -1,4 +1,10 @@
-// SPDX-License-Identifier: Apache-2.0
+/*
+ * SPDX-FileCopyrightText: 2015-2022 The Apache Software Foundation (ASF)
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ *
+ * SPDX-FileContributor: 2019-2026 Espressif Systems (Shanghai) CO LTD
+ */
 /*
  * Licensed to the Apache Software Foundation (ASF) under one
  * or more contributor license agreements.  See the NOTICE file
@@ -16,6 +22,10 @@
  * KIND, either express or implied.  See the License for the
  * specific language governing permissions and limitations
  * under the License.
+ *
+ */
+/*
+ * NOTICE: File has been changed from original implementation.
  */
 
 #include <string.h>
@@ -66,7 +76,7 @@ os_mempool_poison_check(void *start, int sz)
 #define os_mempool_poison_check(start, sz)
 #endif
 
-os_error_t
+static os_error_t
 os_mempool_init(struct os_mempool *mp, uint16_t blocks, uint32_t block_size,
                 void *membuf, const char *name)
 {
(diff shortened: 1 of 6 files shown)


vikramdattu:
cc @mantriyogesh for hosted change

esp_port/components/network_coprocessor/mempool_ll.c
```diff
@@ -331,7 +237,118 @@ os_memblock_put(struct os_mempool *mp, void *block_addr)
 	return os_memblock_put_from_cb(mp, block_addr);
 }
 
-struct os_mempool *
+#if 0
```
sirknightj: Intentional?

Looks like os_mempool_clear is being removed

vikramdattu: Good catch — `os_mempool_clear` and the other functions in that block (`os_mempool_ext_init`, `os_mempool_ext_clear`, `os_memblock_from`, `os_mempool_is_sane`, `os_mempool_info_get_next`) are intentionally compiled out: they aren't exposed via `mempool_ops_t` and have no callers on the hosted-networking path. The `#if 0` here mirrors the same block in upstream esp-hosted-mcu — see [`common/mempool/mempool_ll.c` after commit `c8e5958`](https://github.com/espressif/esp-hosted-mcu/blob/c8e5958/common/mempool/mempool_ll.c#L204-L222) for the corresponding upstream guards.

That said, you're right that a bare `#if 0` reads as silently deleted code. I just pushed a follow-up amend that:

- Replaces the `#if 0` with two semantically-named gates so the *why* is visible from source:
  - `#if MYNEWT_VAL(OS_MEMPOOL_CHECK)` wraps `os_memblock_from` + `os_mempool_is_sane` (sanity-check helpers — same macro that already guards the `assert(os_memblock_from(...))` inside `os_memblock_put`).
  - `#if MYNEWT_VAL(OS_MEMPOOL_EXTENDED_API)` wraps `os_mempool_clear`, `os_mempool_ext_init`, `os_mempool_ext_clear`, `os_mempool_info_get_next` (clear / extended-pool / info-walk surface).
- Adds a comment block explaining the gating and crediting the upstream esp-hosted-mcu source.

Neither flag is defined in this build, so the effective object code is unchanged. Side benefit: `os_memblock_from` now sits under the same macro as its sole caller, so the latent link error if anyone ever flips `OS_MEMPOOL_CHECK=1` is gone.

PR #2279, opened 2026-05-09
PeerConnection: always use DEFAULT_H264_FMTP for our SDP fmtp

vikramdattu:
## Problem

The override of `currentFmtp` to `DEFAULT_H264_FMTP` was gated on `isOffer`. When **answering**, the SDK echoed back whatever H.264 fmtp the peer offered — including the peer's `profile-level-id` — even though the SDK only ever encodes Baseline.

For platforms whose decoder cannot handle every profile (e.g. the tinyh264 software decoder on ESP32-P4 / ESP32-S3, which is Baseline-only), this meant a browser offering High or Main profile would negotiate successfully but then send a stream the device could not decode, producing endless `profile_idc is error` / ACCESS UNIT BOUNDARY CHECK failures.

There is a long-standing TODO in `PeerConnection.c` flagging exactly this case.

## Fix

Always advertise our own fmtp on both offer and answer, defaulting to `DEFAULT_H264_FMTP`. Add a `transceiverSetFmtp(PRtcRtpTransceiver, PCHAR)` public API so applications whose codec needs a different `profile-level-id` (or different fmtp parameters) can pin one per transceiver; the override falls back to `DEFAULT_H264_FMTP` when unset or cleared.

Per RFC 6184 §8.2.2 `profile-level-id` and `packetization-mode` are symmetric parameters — the fmtp we advertise should reflect what we will actually send, not what the peer happened to support. Echoing the peer's High/Main fmtp from the answer path was misadvertising the SDK's real capability. Codec *selection* still uses the matched remote fmtp via `getH264FmtpScore`; only the answer-side advertised fmtp changes.

## Behavior change

- **Peers that already offer `42e01f` (Baseline)**: no change.
- **Peers that offered only High / Main**: now receive a Baseline answer. If they cannot produce Baseline, they will drop video rather than send an undecodable stream — which is the correct WebRTC behavior on a profile mismatch.
- **Tests**: `TestPayloadNoFmtp` now expects DEFAULT fmtp in the answer; `SdpMatch` fixtures for Chrome_Android (was `42e015`) and Firefox Linux/Mac (different attr ordering) updated to the canonical DEFAULT string.

## Validated on

ESP32-P4 with tinyh264 software decoder, against Chrome / Firefox initiating a WebRTC call with High / Main profile in the offer.

### DIFF
--- a/src/source/PeerConnection/SessionDescription.c
+++ b/src/source/PeerConnection/SessionDescription.c
@@ -724,9 +724,16 @@ STATUS populateSingleMediaSection(PKvsPeerConnection pKvsPeerConnection, PKvsRtp
 
     if (pRtcMediaStreamTrack->codec == RTC_CODEC_H264_PROFILE_42E01F_LEVEL_ASYMMETRY_ALLOWED_PACKETIZATION_MODE) {
         // TODO: Need additional condition for a signaling channel with an ENABLED media storage configuration
-        if (pKvsPeerConnection->isOffer) {
-            currentFmtp = DEFAULT_H264_FMTP;
-        }
+        //
+        // Advertise our own fmtp on both offer and answer. The previous
+        // answer-path behavior echoed the peer's fmtp, which on Baseline-only
+        // decoders (e.g. tinyh264 on ESP32-P4/S3) caused the SDK to accept
+        // High/Main offers and then fail to decode the resulting stream.
+        // Per RFC 6184 §8.2.2 profile-level-id and packetization-mode are
+        // symmetric parameters; advertising them ourselves correctly reflects
+        // what we will actually send. transceiverSetFmtp() lets callers pin
+        // a different fmtp when their codec needs it.
+        currentFmtp = pKvsRtpTransceiver->fmtpOverride[0] != '\0' ? pKvsRtpTransceiver->fmtpOverride : DEFAULT_H264_FMTP;
         STRCPY(pSdpMediaDescription->sdpAttributes[attributeCount].attributeName, "rtpmap");
         amountWritten = SNPRINTF(pSdpMediaDescription->sdpAttributes[attributeCount].attributeValue,
                                  SIZEOF(pSdpMediaDescription->sdpAttributes[attributeCount].attributeValue), "%" PRId64 " H264/90000", payloadType);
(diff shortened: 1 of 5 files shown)


src/source/PeerConnection/SessionDescription.c
```diff
@@ -719,9 +719,18 @@ STATUS populateSingleMediaSection(PKvsPeerConnection pKvsPeerConnection, PKvsRtp
...
+        //
+        // Force DEFAULT_H264_FMTP for both offers and answers. Previously we
+        // only overrode currentFmtp when offering; answering inherited
+        // whatever the peer's fmtp was, which meant when a browser offered
+        // High/Main profile we agreed and then received a stream we couldn't
+        // decode on Baseline-only decoders (e.g., tinyh264 on ESP32-P4/S3).
+        //
+        // level-asymmetry-allowed=1 is included in DEFAULT_H264_FMTP and in
+        // practice in every peer's offer; it explicitly permits the answerer
+        // to pick different profile/level than the offer. This resolves the
```
sirknightj: One note: 

`level-asymmetry-allowed=1` means the two directions can use different **levels** (levels = resolution/bitrate caps like 3.0, 3.1, 4.0), not a different **profile**. Profile is always symmetric according to RFC 6184

https://datatracker.ietf.org/doc/html/rfc6184#section-8.2.2
> The parameters identifying a media format configuration for H.264
      are profile-level-id and packetization-mode.  These media format
      configuration parameters (except for the level part of profile-
      level-id) MUST be used symmetrically

src/source/PeerConnection/SessionDescription.c
```diff
@@ -719,9 +719,18 @@ STATUS populateSingleMediaSection(PKvsPeerConnection pKvsPeerConnection, PKvsRtp
...
+        // long-standing TODO on the fmtp echo below.
+        currentFmtp = DEFAULT_H264_FMTP;
```
sirknightj: Wondering on your thoughts on if we add an override to this to make it more configurable for users:

```c
CHAR fmtpOverride[MAX_SDP_ATTRIBUTE_VALUE_LENGTH]; // new member to KvsRtpTransceiver:
```

```c
PUBLIC_API STATUS transceiverSetFmtp(PRtcRtpTransceiver, PCHAR);

STATUS transceiverSetFmtp(PRtcRtpTransceiver pTransceiver, PCHAR fmtp) {
    PKvsRtpTransceiver pKvsRtpTransceiver = (PKvsRtpTransceiver) pTransceiver;
    CHK(pKvsRtpTransceiver != NULL && fmtp != NULL, STATUS_NULL_ARG);
    STRNCPY(pKvsRtpTransceiver->fmtpOverride, fmtp, MAX_SDP_ATTRIBUTE_VALUE_LENGTH - 1);
CleanUp:
    CHK_LOG_ERR(retStatus);
    return retStatus;
}
```

```c
currentFmtp = pKvsRtpTransceiver->fmtpOverride[0] != '\0' 
    ? pKvsRtpTransceiver->fmtpOverride 
    : DEFAULT_H264_FMTP;
```

sirknightj:
Hi @vikramdattu looks some of the tests are failing, please take a look!

src/source/PeerConnection/SessionDescription.c
vikramdattu: Good idea from the configurability point of view. I think, I can incorporate this as is.

sirknightj (approved):

YPandas (approved):

PR #2295, opened 2026-05-21
Espressif Port: JPEG snapshot pipeline + binary cmd channel on webrtc_bridge

vikramdattu:
*Issue #, if available:*
- N/A

*What was changed?*
- JPEG snapshot capture in `media_stream`, plus a binary command channel
  on `webrtc_bridge` so a signaling-side CLI can request/receive snapshot
  bytes end-to-end.
- streaming_only / webrtc_classic (P4): switch to esp-rainmaker's 8 MB
  partition layout (dual OTA + secure_cert + fctry).

*Why was it changed?*
- Snapshot is the building block needed for **Matter Camera** support
  (Camera cluster snapshot endpoint) and for RainMaker-style
  image-on-event flows.
- JSON-over-signaling isn't the right channel for binary payloads.
  The bridge already moves raw bytes; the examples just needed an
  explicit non-JSON path.
- IDF v5.5's bootloader on P4 outgrew the 0x6000 window left by the
  default 4 MB layout. The rmaker layout fixes that and brings OTA +
  secure_cert in for free.

*How was it changed?*
6 commits — preprocess hook for raw frames, `video_capture_get_snapshot()`,
P4 rev<3 RGB565 conversion, `webrtc_bridge` binary channel + CLI demo,
`signaling_client_if` const fix, partition layout. Per-commit detail in
the commit messages.

- gitlab CI `build_idf_v5.4` + `build_idf_v5.5` on the equivalent internal MR.
- Bench: P4-EYE running `streaming_only`; ESP32-C5 running `signaling_only`
  issues `query-snapshot [quality]` over the binary cmd channel; JPEG bytes
  verified on the requesting side.


vikramdattu:
@unicornss @sirknightj PTAL
cc @sayondeep @dhrishi

esp_port/components/media_stream/src/esp32p4_frame_grabber.c
```diff
@@ -571,4 +626,182 @@ esp_err_t esp32p4_frame_grabber_deinit(void)
...
+        }
+    }
+
+    /* Prepare the snapshot request */
+    s_snapshot.buffer = buf;
+    s_snapshot.buffer_size = buf_size;
+    s_snapshot.frame_len = 0;
+    s_snapshot.width = 0;
+    s_snapshot.height = 0;
+    s_snapshot.requested = true;  /* Trigger the interceptor in encoder task */
```
sirknightj, line 692: Minor: two concurrent callers could race here. The second would overwrite buffer/buffer_size before the first completes. A quick early-out like `if (s_snapshot.requested) return ESP_ERR_INVALID_STATE;` would be cheap insurance, even if the current usage is single-threaded.

Issue #2296, opened 2026-05-21
[Feature request]: mbedtls 4 support


vikramdattu:
### Please confirm you have already done the following

- [x] I have searched the repository for related/existing feature requests
- [x] I have all the details the issue requires

### Describe the feature

The current latest version of mbedtls supported by the SDK is 3.6.x (old: 2.x).
mbedtls version 4 is already out(for about a year and already at 4.1.0 release) and other SDKs have already adopted the same (e.g., latest ESP-IDF releases from Espressif already ship with the mbedtls 4).

mbedtls 4 support should be added.

### Use Case

mbedtls latest version upgrade. Security and maintainance update

### Proposed Solution

Add the support for mbedtls 4.x along with existing 2.x and 3.x

vikramdattu:
Dear @unicornss @sirknightj let us know if aws wants to take this up.

sirknightj:
It has been merged and will be part of the 1.20.0 release.

Thank you for your contribution!

vikramdattu:
@sirknightj thanks and kudos to you guys persistently pushing for the polished, high-standard version!

PR #2299, opened 2026-05-23
feat(srtp): opt-in libsrtp 3.x support via -DUSE_LIBSRTP3=ON

vikramdattu:
*Issue #, if available:*
- #2296 (mbedTLS 4 support — feature request)

*What was changed?*
- Added a `USE_LIBSRTP3` CMake option (default `OFF`). When `ON`, the SDK builds and links against libsrtp 3.x (the `libsrtp3` line on cisco/libsrtp `main`) instead of the long-standing 2.x default.
- `src/source/Srtp/SrtpSession.c` now carries two parallel implementations under `#ifdef USE_LIBSRTP3`. `freeSrtpSession()` is shared.
- Build scaffolding: `CMake/Utilities.cmake` flips the lib filename (`srtp2` ↔ `srtp3`); `CMake/Dependencies/libsrtp-CMakeLists.txt` picks the `GIT_TAG` (2.x stays at `bd0f27ec`, 3.x at cisco/libsrtp main tip `6e23ad8d` — no `v3.0.0` release tag exists yet), skips the autotools `configure` step on 3.x (CMake is canonical now), and passes `-DLIBSRTP_TEST_APPS=OFF` so libsrtp's `cipher_driver` test binary doesn't fail to link in an SDK-dep build; top-level `CMakeLists.txt` flips `find_library` / `pkg_check_modules` between `srtp2` / `srtp3` and emits `-DUSE_LIBSRTP3` to the SDK target.
- `src/source/Include_i.h`: `<srtp3/srtp.h>` vs `<srtp2/srtp.h>` under the same guard.

*Why was it changed?*
- libsrtp 3.x adds mbedTLS 4 / TF-PSA-Crypto support (cisco/libsrtp#782). The 2.x line on `2_x_dev` still includes the legacy `<mbedtls/aes.h>` / `<mbedtls/gcm.h>` paths and therefore cannot build against mbedTLS 4. That blocks embedded targets shipping mbedTLS 4 — most concretely **ESP-IDF v6**, which is in pre-release and will be the default for new Espressif chip bring-ups. The KVS SDK is otherwise the bottleneck in the dependency chain for those consumers (cisco/libsrtp#812 tracks whether 2.x will catch up; no upstream commitment yet).
- The 3.x line changes the public API in non-trivial ways (opaque policy handle, separate src/dst buffers with `size_t` lengths on protect/unprotect, profile enum that folds RTP/RTCP), and cisco renamed the project on `main` to `libsrtp3` — a deliberate parallel line. So this is a real fork, not a transparent bump, and forcing every consumer to move would break those still on 2.x. Hence opt-in via compile flag rather than a hard cutover. Default `OFF` preserves the existing 2.x build, pin, behaviour, tests, and CI exactly.

*How was it changed?*
- **API break #1 — `srtp_policy_t` is opaque.** Allocate via `srtp_policy_create()`, configure via `srtp_policy_set_profile()` + `srtp_policy_set_ssrc()` + `srtp_policy_add_key()`. The 2.x pattern of writing into struct fields directly is gone. Encapsulated in a small `buildPolicy()` helper used for both transmit and receive sessions.
- **API break #2 — `srtp_protect` / `srtp_unprotect` signature.** Now `(ctx, src, src_len, dst, &dst_len, mki_index)` with `size_t` lengths. In-place is still supported when `src == dst`, which KVS keeps using to preserve its existing buffer-ownership contract (no extra allocation, no memcpy). `mki_index` is hardcoded to 0 (MKI not used in this SDK).
- **API break #3 — master key + master salt split.** KVS today concatenates master key + master salt into `policy.key`; the 3.x path splits via `srtp_profile_get_master_key_length()` / `srtp_profile_get_master_salt_length()` so `srtp_policy_add_key()` gets each separately.
- **KVS profile → libsrtp profile mapping.** `KVS_SRTP_PROFILE_AES128_CM_HMAC_SHA1_{32,80}` mapped to `srtp_profile_aes128_cm_sha1_{32,80}`. In 2.x KVS used `srtp_crypto_policy_set_aes_cm_128_hmac_sha1_32` for RTP and `srtp_crypto_policy_set_rtp_default` for RTCP; libsrtp 3.x's profile enum follows DTLS-SRTP RFC 5764 §4.1.2 which already folds RTP+RTCP into a single profile per session, so the mapping is direct.

- Local build, macOS arm64, `USE_OPENSSL=OFF -DUSE_MBEDTLS=ON -DUSE_LIBSRTP3=ON -DBUILD_TEST=ON -DBUILD_DEPENDENCIES=ON`: full SDK (`libkvsWebrtcClient`, `libkvsWebrtcSignalingClient`), all samples (`kvsWebrtcClient{Master,Viewer}`, `kvsWebrtcStorage*`, `*GstSample`), and `webrtc_client_test` build clean — no warnings, no errors.
- `tst/SrtpApiTest` — the closest unit-test signal — runs 3/3 green:
  - `encryptDecryptRtpPacketSuccess` — end-to-end in-place `srtp_protect` → `srtp_unprotect` roundtrip; encrypted length grows by the 10-byte HMAC-SHA1-80 tag and shrinks back on decrypt as expected.
  - `encryptDecryptKeyMisMatchFails` — receive key intentionally differs from transmit key; decrypt correctly returns `srtp_err_status_auth_fail` (7).
  - `noSrtpKeyReturnsFailure` — null-arg validation intact.
- Existing 2.x matrix (default `USE_LIBSRTP3=OFF`) is unchanged and exercised by CI as before. Happy to add a parallel `USE_LIBSRTP3=ON` matrix entry if maintainers want it.


vikramdattu:
cc @unicornss @sirknightj with the srtp upgrade from v2 to v3 (mbedtls 4 is supported with srtp3), it should unblock hurdle for the SDK mbedtls 4 support as well. [Feature request here.](https://github.com/awslabs/amazon-kinesis-video-streams-webrtc-sdk-c/issues/2296)

CMake/Dependencies/libsrtp-CMakeLists.txt
```diff
@@ -2,14 +2,28 @@ cmake_minimum_required(VERSION 3.6.3)
 
 project(libsrtp-download LANGUAGES C)
 
+# Pick the libsrtp tag based on USE_LIBSRTP3. The 2.x default tracks what the
```
sirknightj, line 5: Wondering if you'd be able to add at least 1 combination to the CI that verifies the new path?

CMakeLists.txt
```diff
@@ -21,6 +21,7 @@ option(BUILD_STATIC_LIBS "Build all libraries statically. (This includes third-p
 
 option(BUILD_WEBSOCKETS "Build libwebsockets from source" ${BUILD_DEPENDENCIES})
 option(BUILD_SRTP "Build libsrtp from source" ${BUILD_DEPENDENCIES})
+option(USE_LIBSRTP3 "Build/link against libsrtp 3.x (default: 2.x). Required for mbedTLS 4." OFF)
```
sirknightj, line 24: Can you add the info to the README as well?

src/source/Srtp/SrtpSession.c
```diff
@@ -1,6 +1,231 @@
...
+    CHK(pSrtpSession != NULL && message != NULL && len != NULL, STATUS_NULL_ARG);
+
+    /* Caller is required to allocate the buffer with at least SRTP_MAX_TRAILER_LEN
+     * bytes of headroom (matches the libsrtp 2.x in-place contract that KVS
+     * already follows). dst_len starts as the available capacity and is
+     * overwritten with the protected packet length. */
+    srcLen = (size_t) *len;
+    dstLen = srcLen + SRTP_MAX_TRAILER_LEN;
+    errStatus = srtp_protect(pSrtpSession->srtp_transmit_session, (const uint8_t*) message, srcLen, (uint8_t*) message, &dstLen, 0);
+    CHK_ERR(errStatus == srtp_err_status_ok, STATUS_SRTP_ENCRYPT_FAILED, "srtp_protect returned %lu on srtp session %" PRIu64, errStatus,
```
sirknightj: The encrypt functions use `%lu` for `srtp_err_status_t` while the decrypt functions use `%u`. This is inherited from the existing 2.x code, so not blocking, but since you're writing the 3.x path fresh, you could use `%u` consistently here (lines 197 and 218).

src/source/Srtp/SrtpSession.c
```diff
@@ -153,3 +349,35 @@ STATUS encryptRtcpPacket(PSrtpSession pSrtpSession, PVOID message, PINT32 len)
...
+
+    PSrtpSession pSrtpSession = NULL;
+
+    CHK(ppSrtpSession != NULL, STATUS_NULL_ARG);
+    CHK(*ppSrtpSession != NULL, retStatus);
+
+    pSrtpSession = *ppSrtpSession;
+
+    if ((pSrtpSession->srtp_transmit_session != NULL) && (errStatus = srtp_dealloc(pSrtpSession->srtp_transmit_session)) != srtp_err_status_ok) {
+        DLOGW("Dealloc of transmit session failed with error code %d\n", errStatus);
```
sirknightj: Could clean up the `\n` here (also copied over from old path) since the logging functions already add newlines.

src/source/Srtp/SrtpSession.c
```diff
@@ -1,6 +1,231 @@
...
+    if (receivePolicy != NULL) {
+        srtp_policy_destroy(receivePolicy);
+    }
+    if (transmitPolicy != NULL) {
+        srtp_policy_destroy(transmitPolicy);
+    }
+    if (STATUS_FAILED(retStatus) && pSrtpSession != NULL) {
+        freeSrtpSession(&pSrtpSession);
+    }
+
```
sirknightj, line 141: Could we add a CHK_LOG_ERR here?

sirknightj (commented):
Action items: README, clang-format, and new CI to verify

Functionally, looks good. Thank you for the contribution

src/source/Srtp/SrtpSession.c
vikramdattu: Used %u

src/source/Srtp/SrtpSession.c
```diff
@@ -1,6 +1,232 @@
...
+    size_t srcLen, dstLen;
+
+    CHK(pSrtpSession != NULL && encryptedMessage != NULL && len != NULL, STATUS_NULL_ARG);
+
+    /* libsrtp 3.x: in-place is supported when src == dst. dst_len starts as
+     * the buffer capacity and is overwritten with the decrypted packet length. */
+    srcLen = (size_t) *len;
+    dstLen = srcLen;
+    errStatus = srtp_unprotect(pSrtpSession->srtp_receive_session, (const uint8_t*) encryptedMessage, srcLen, (uint8_t*) encryptedMessage, &dstLen);
+    CHK_ERR(errStatus == srtp_err_status_ok, STATUS_SRTP_DECRYPT_FAILED, "Decrypting rtp packet failed with error code %u on srtp session %" PRIu64,
```
unicornss: do we need the pointer pSrtpSession->srtp_receive_session in the log?

src/source/Srtp/SrtpSession.c
```diff
@@ -1,6 +1,232 @@
...
+{
+    switch (kvsProfile) {
+        case KVS_SRTP_PROFILE_AES128_CM_HMAC_SHA1_32:
+            *pProfile = srtp_profile_aes128_cm_sha1_32;
+            return STATUS_SUCCESS;
+        case KVS_SRTP_PROFILE_AES128_CM_HMAC_SHA1_80:
+            *pProfile = srtp_profile_aes128_cm_sha1_80;
+            return STATUS_SUCCESS;
+        default:
+            return STATUS_SSL_UNKNOWN_SRTP_PROFILE;
```
unicornss: DLOGE on unknown profile?

src/source/Srtp/SrtpSession.c
```diff
@@ -1,6 +1,232 @@
...
+static STATUS kvsToLibsrtpProfile(KVS_SRTP_PROFILE kvsProfile, srtp_profile_t* pProfile)
+{
+    switch (kvsProfile) {
+        case KVS_SRTP_PROFILE_AES128_CM_HMAC_SHA1_32:
+            *pProfile = srtp_profile_aes128_cm_sha1_32;
+            return STATUS_SUCCESS;
+        case KVS_SRTP_PROFILE_AES128_CM_HMAC_SHA1_80:
+            *pProfile = srtp_profile_aes128_cm_sha1_80;
+            return STATUS_SUCCESS;
+        default:
```
unicornss, line 47: Null check on pProfile ?

unicornss (approved):
Thanks for your contribution!

src/source/Srtp/SrtpSession.c
vikramdattu: Maybe not and it may just end up cluttering the error log. Will remove!

vikramdattu: Removed the session pointer print from the log

sirknightj (approved):

PR #2300, opened 2026-05-27
Ice: release agent lock around the packet send in iceAgentSendPacket

vikramdattu:
*Issue #, if available:*
- N/A

*What was changed?*
- iceAgentSendPacket now releases pIceAgent->lock around the actual packet send (iceUtilsSendData) instead of holding it across the send.
- Adds a bidirectionalConcurrentMediaSend functional test.

*Why was it changed?*
- pIceAgent->lock is one coarse lock for the whole ICE subsystem — RX data handling, connectivity checks and timers all take it. Holding it across the send serialized every TX packet behind that work, so under bidirectional A/V load each outbound packet waited on RX/ICE activity. This showed up as large per-packet send stalls and collapsed outbound video fps on an ESP32-P4 + WiFi-coprocessor deployment.

*How was it changed?*
- Cache the selected pair's socketConnection, destination address, turn connection and relay flag under the lock; release the lock; send; re-acquire to update stats. socketConnection/turnConnection carry their own locks and connection-closed guards, so the send is safe outside the agent lock.
- The selected pair is re-read and NULL-checked after re-acquiring the lock, since it can change while unlocked.

- Existing PeerConnection loopback tests pass (exchangeMedia, noLostFramesAfterConnected, connect/disconnect, close).
- New bidirectionalConcurrentMediaSend test: both agents send concurrently while each receives.
- ThreadSanitizer: 0 data races across the concurrency and disconnect tests.


vikramdattu:
@sirknightj @unicornss PTAL

PR #2305, opened 2026-05-28
Add minimum length validation for stun ERROR_CODE parsing

vabhasin:
*Issue #, if available:*
- N/A

*What was changed?*
- Added minimum length validation for STUN attribute ERROR_CODE in `deserializeStunPacket` (Stun.c)

*Why was it changed?*
- The ERROR_CODE attribute parser only validated an upper bound (length <= 128) but not a lower bound (length >= 4). When `attribute.length = 0`, the computation `paddedLength - 4` underflows (UINT16 wraps to ~65532), causing a massive MEMCPY that crashes the process. Per RFC 5389, ERROR_CODE must contain at least 4 bytes (reserved + error class + error number).

*How was it changed?*
- Added `stunAttributeHeader.length >= STUN_ERROR_CODE_PACKET_ERROR_PHRASE_OFFSET` to the existing validation check, rejecting malformed attributes with length more than 4.
- Added an if guard around the `MEMCPY` to prevent the underflow even if a zero-length attribute somehow reaches the copy stage.

- Added unit test for cases where attribute length is less than 4.
- All tests are now passing

### DIFF
--- a/src/source/Stun/Stun.c
+++ b/src/source/Stun/Stun.c
@@ -700,8 +700,10 @@ STATUS deserializeStunPacket(PBYTE pStunBuffer, UINT32 bufferSize, PBYTE passwor
             case STUN_ATTRIBUTE_TYPE_ERROR_CODE:
                 attributeSize = SIZEOF(StunAttributeErrorCode);
 
-                // Validate the size of the length against the max value of error phrase
-                CHK(stunAttributeHeader.length <= STUN_MAX_ERROR_PHRASE_LEN, STATUS_STUN_INVALID_ERROR_CODE_ATTRIBUTE_LENGTH);
+                // Validate the size of the length against the min and max value of error phrase
+                CHK(stunAttributeHeader.length >= STUN_ERROR_CODE_PACKET_ERROR_PHRASE_OFFSET &&
+                        stunAttributeHeader.length <= STUN_MAX_ERROR_PHRASE_LEN,
+                    STATUS_STUN_INVALID_ERROR_CODE_ATTRIBUTE_LENGTH);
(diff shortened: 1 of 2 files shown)


sirknightj (approved):

src/source/Stun/Stun.c
```diff
@@ -700,8 +700,10 @@ STATUS deserializeStunPacket(PBYTE pStunBuffer, UINT32 bufferSize, PBYTE passwor
             case STUN_ATTRIBUTE_TYPE_ERROR_CODE:
                 attributeSize = SIZEOF(StunAttributeErrorCode);
 
-                // Validate the size of the length against the max value of error phrase
-                CHK(stunAttributeHeader.length <= STUN_MAX_ERROR_PHRASE_LEN, STATUS_STUN_INVALID_ERROR_CODE_ATTRIBUTE_LENGTH);
+                // Validate the size of the length against the min and max value of error phrase
+                CHK(stunAttributeHeader.length >= STUN_ERROR_CODE_PACKET_ERROR_PHRASE_OFFSET &&
+                        stunAttributeHeader.length <= STUN_MAX_ERROR_PHRASE_LEN,
+                    STATUS_STUN_INVALID_ERROR_CODE_ATTRIBUTE_LENGTH);
```
unicornss, line 706: should we need to have two separate error codes (too_short, too_long)?

unicornss (approved):

src/source/Stun/Stun.c
vabhasin: I think INVALID should be ok to cover both, we are following this same pattern across SDK.

PR #2312, opened 2026-05-29
fix packet length validation in handleDcepPacket

vabhasin:
*Issue #, if available:*
- N/A

*What was changed?*
- Fixed the packet length validation in `handleDcepPacket` (Sctp.c:322) and added unit tests for the bounds check.

*Why was it changed?*
- The bounds check used `>=` instead of `<=`, which inverted the validation logic. It was intended to reject packets where the declared `labelLength + protocolLength` exceeds the actual packet size, but instead it rejected packets where the declared content was smaller than the packet. For example, A DCEP DataChannelOpen message with `labelLength=0xFFFF` but only a few bytes of actual data would pass the check and  uncapped length would be passed to `STRNCPY` in `onSctpSessionDataChannelOpen`, overflowing the 256-byte `dataChannel.name` buffer.

*How was it changed?*
- Changed `>=` to `<=` on the comparison at Sctp.c:322 so the check correctly rejects packets where declared content exceeds actual packet length.
- Added a test (`handleDcepPacket_RejectsOversizedLabelLength`) that constructs a 13-byte DCEP packet with `labelLength=0xFFFF` and verifies it is rejected with `STATUS_SCTP_INVALID_DCEP_PACKET`.

- Ran `webrtc_client_test --gtest_filter="DataChannelApiTest.*"` new and existing DataChannel tests are passing.
- Verified the malformed packet test fails without the fix and passes with the fix (rejected at the bounds check).
- Verified a well-formed packet with matching `labelLength` passes the bounds check but fails in the next step`usrsctp_sendv` which requires a real SCTP socket.
- Verified data channel path through the JS SDK Test page. A message sent through viewer is received on master side and a response message is sent back

### DIFF
--- a/tst/DataChannelApiTest.cpp
+++ b/tst/DataChannelApiTest.cpp
@@ -32,8 +32,6 @@ TEST_F(DataChannelApiTest, createDataChannel_Disconnected)
     freePeerConnection(&pPeerConnection);
 }
 
-extern "C" STATUS handleDcepPacket(PSctpSession pSctpSession, UINT32 streamId, PBYTE data, SIZE_T length);
-
 TEST_F(DataChannelApiTest, handleDcepPacket_RejectsOversizedLabelLength)
 {
     /** Construct a minimal DCEP DataChannelOpen packet (13 bytes total):
--- a/tst/DataChannelApiTest.cpp
+++ b/tst/DataChannelApiTest.cpp
@@ -7,8 +7,7 @@ namespace kinesis {
 namespace video {
 namespace webrtcclient {
 
-class DataChannelApiTest : public WebRtcClientTestBase {
-};
+class DataChannelApiTest : public WebRtcClientTestBase {};
 
 TEST_F(DataChannelApiTest, createDataChannel_Disconnected)
 {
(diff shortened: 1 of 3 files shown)


tst/DataChannelApiTest.cpp
```diff
@@ -33,6 +32,30 @@ TEST_F(DataChannelApiTest, createDataChannel_Disconnected)
     freePeerConnection(&pPeerConnection);
 }
 
+extern "C" STATUS handleDcepPacket(PSctpSession pSctpSession, UINT32 streamId, PBYTE data, SIZE_T length);
```
sirknightj: Would recommend adding it to the Sctp.h:

```c
STATUS sctpSessionWriteDcep(PSctpSession, UINT32, PCHAR, UINT32, PRtcDataChannelInit);
STATUS handleDcepPacket(PSctpSession, UINT32, PBYTE, SIZE_T);
```

The test file already includes WebRTCClientTestFixture.h which pulls in the SDK headers, and Sctp.h is an internal header that's part of the build. As long as the test's include chain reaches Sctp.h, the prototype is available through the header's existing extern "C" { ... } block, no per-test-file declaration needed.

vabhasin: Thanks, updated

YPandas (approved):

sirknightj (approved):

Ahhorse1 (approved):

PR #2317, opened 2026-06-01
add packet length checks for h264 and h65 RTP packet

vabhasin:
*Issue #, if available:*
- N/A

*What was changed?*
- Added minimum packet length validation in H.264 FU-A, FU-B, and H.265 FU depayloading paths to prevent integer underflow when processing malformed packets shorter than the header size.
- Fix header constant used for calculating `naluLength` in case of FU-B packet.

*Why was it changed?*
- The H.264 and H.265 depayloaders subtract a header-size constant from `packetLength` without first verifying that the packet is large enough. If `packetLength <= headerSize`, the subtraction underflows and the subsequent MEMCPY attempts a massive copy that crashes the process. The SDK's own payloaders always produce packets larger than the header, so this only affects malformed input from non-SDK sources.

*How was it changed?*
- Added `CHK(packetLength > FU_A_HEADER_SIZE, STATUS_RTP_INPUT_PACKET_TOO_SMALL)` in the H.264 FU-A case before any header byte access or arithmetic.
- Added `CHK(packetLength > FU_B_HEADER_SIZE, STATUS_RTP_INPUT_PACKET_TOO_SMALL)` in the H.264 FU-B case.
- Added `CHK(packetLength > H265_FU_HEADER_SIZE, STATUS_RTP_INPUT_PACKET_TOO_SMALL)` in the H.265 FU case.
- All checks use the existing `STATUS_RTP_INPUT_PACKET_TOO_SMALL` error code.
- Use header constant `FU_B_HEADER_SIZE` instead of `FU_B_HEADER_SIZE` in `naluLength` calculation

- `tst/RtpFunctionalityTest.cpp`: added `depayH264FuARejectsTooSmallPacket` and `depayH265FuRejectsTooSmallPacket` tests that construct packets at and below the header size and verify they are rejected. Both tests fail without the fix (function returns SUCCESS with garbage naluLength) and pass with the fix.

### DIFF
--- a/tst/RtpFunctionalityTest.cpp
+++ b/tst/RtpFunctionalityTest.cpp
@@ -6,13 +6,12 @@ namespace kinesis {
 namespace video {
 namespace webrtcclient {
 
-#define NUMBER_OF_FRAME_FILES 403
+#define NUMBER_OF_FRAME_FILES      403
 #define NUMBER_OF_H265_FRAME_FILES 1500
-#define DEFAULT_FPS_VALUE     25
+#define DEFAULT_FPS_VALUE          25
 BYTE start4ByteCode[] = {0x00, 0x00, 0x00, 0x01};
 
-class RtpFunctionalityTest : public WebRtcClientTestBase {
-};
+class RtpFunctionalityTest : public WebRtcClientTestBase {};
 
 TEST_F(RtpFunctionalityTest, packetUnderflow)
 {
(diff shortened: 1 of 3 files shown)


tst/RtpFunctionalityTest.cpp
vabhasin, line 1: Also contains some clang-format changes

sirknightj (approved):

Ahhorse1 (approved):

PR #2318, opened 2026-06-01
Add bounds check before reading recvDelta in TWCC packet

vabhasin:
*Issue #, if available:*

- N/A

*What was changed?*

- Added minimum payload length validation and bounds checking in `parseRtcpTwccPacket` to prevent out-of-bounds reads when parsing recv-delta bytes

*Why was it changed?*

- The TWCC parser could read past `payloadLength` when consuming recv-delta bytes, resulting in undefined behavior. Additionally, payloads shorter than the minimum TWCC feedback header (16 bytes) were not rejected upfront.

*How was it changed?*

- Added `CHK(pRtcpPacket->payloadLength >= TWCC_FB_PAYLOAD_MIN_LEN, STATUS_RTCP_INPUT_PACKET_TOO_SMALL)` at function entry to reject payloads that are too short to contain a valid TWCC header
- Added condition `(recvOffset >= payloadLength)` guards before each delta read (small and large) in both the run-length and status vector loops, so the parser gracefully stops processing when the payload is exhausted without returning an error or logging false failures
- Fixed the `parseTwcc` test helper to use the full 256-byte buffer size as `payloadLength`

- `tst/RtcpFunctionalityTest.cpp`: Added `parseRtcpTwccPacketRejectsTruncatedDeltas` (validates graceful stop on truncated delta section) and `parseRtcpTwccPacketRejectsShortPayload` (validates `STATUS_RTCP_INPUT_PACKET_TOO_SMALL` for lengths 0, 10, 15 and `STATUS_SUCCESS` for length 16)


YPandas (approved):

YPandas (commented):
Seems some CIs are failing

Ahhorse1 (approved):

Ahhorse1 (approved):

PR #2319, opened 2026-06-02
fix data channel stalls from missing sctp timer

kevincgall-stackav:
*Issue #, if available:*
- N/A

*What was changed?*
- Add configuration options (currently hard-coded) to the SCTP header for configuring retry and path liveness detection for data channels over sctp.
- Periodically call the usrsctp state machine step function while data channels are active
- Unit test covering the new code

*Why was it changed?*
- Observed issues in using this library where, under certain network congestion conditions, data channels would stall and never recover even though ICE transport stayed alive and media channels recovered. Issue was root-caused to a bug in the usage of the usrsctp library: since there are no periodic invocations of the timer handler, the advancement of the SCTP state machine relied on a steady stream of data being published by the data channel, which is not a guarantee. In particular, if the data channel's send buffers fill up, there is no way to clear the stall without receiving a packet from the peer.

*How was it changed?*
- Implemented a timer callback to run periodically which invokes `usrsctp_handle_timers()`. By doing so, we advance the sctp state machine, even when there is no incoming data, which allows for retry or failure detection as applicable.

- Unit test included in this PR. I did verify that without the fix, the new unit test fails.
- Tested with proprietary application. With this fix, data channels that are stalled during a period of low network bandwidth are able to fully recover alongside the media channels.

### DIFF
--- a/src/source/Sctp/Sctp.c
+++ b/src/source/Sctp/Sctp.c
@@ -1,6 +1,87 @@
 #define LOG_CLASS "SCTP"
 #include "../Include_i.h"
 
+PSctpContext acquireSctpContext()
+{
+    ENTERS();
+    static SctpContext s = {.lastTickTime = 0, .isSctpInitialized = FALSE, .contextRefCnt = 0, .sctpContextLock = INVALID_MUTEX_VALUE};
+    ATOMIC_INCREMENT(&s.contextRefCnt);
+    LEAVES();
+    return &s;
+}
+
+VOID releaseSctpContext(PSctpContext pSctpContext)
+{
+    ENTERS();
+    ATOMIC_DECREMENT(&pSctpContext->contextRefCnt);
+    LEAVES();
+}
+
+// Initializes SCTP context, in particular the lastTickTime used for timer handling
+STATUS createSctpContext()
+{
+    ENTERS();
+    PSctpContext pSctpContext = acquireSctpContext();
+    STATUS retStatus = STATUS_SUCCESS;
+    BOOL locked = FALSE;
+
+    CHK_WARN(!ATOMIC_LOAD_BOOL(&pSctpContext->isSctpInitialized), retStatus, "SCTP context already initialized, nothing to do");
+    CHK_ERR(!IS_VALID_MUTEX_VALUE(pSctpContext->sctpContextLock), retStatus, "Mutex seems to have been created already");
+
+    pSctpContext->sctpContextLock = MUTEX_CREATE(TRUE);
+    CHK_ERR(IS_VALID_MUTEX_VALUE(pSctpContext->sctpContextLock), STATUS_NULL_ARG, "Mutex creation failed");
+    MUTEX_LOCK(pSctpContext->sctpContextLock);
+    locked = TRUE;
+    pSctpContext->lastTickTime = GETTIME();
+    ATOMIC_STORE_BOOL(&pSctpContext->isSctpInitialized, TRUE);
+    DLOGI("Initialized SCTP context instance");
+
+CleanUp:
+    if (locked) {
+        MUTEX_UNLOCK(pSctpContext->sctpContextLock);
+    }
+    releaseSctpContext(pSctpContext);
+    CHK_LOG_ERR(retStatus);
+
+    LEAVES();
+    return retStatus;
+}
+
+STATUS cleanupSctpContext()
+{
+    ENTERS();
+    UINT64 shutdownTimeout;
+    STATUS retStatus = STATUS_SUCCESS;
+
+    PSctpContext pSctpContext = acquireSctpContext();
+
+    DLOGD("Releasing SCTP context instance from cleanupSctpContext");
+    releaseSctpContext(pSctpContext);
+
+    CHK_WARN(ATOMIC_LOAD_BOOL(&pSctpContext->isSctpInitialized), STATUS_INVALID_OPERATION, "SCTP context not initialized, nothing to clean up");
+
+    ATOMIC_STORE_BOOL(&pSctpContext->isSctpInitialized, FALSE);
+
+    shutdownTimeout = GETTIME() + SCTP_CONTEXT_REFERENCE_WAIT_TIMEOUT;
+    while (ATOMIC_LOAD(&pSctpContext->contextRefCnt) > 0 && GETTIME() < shutdownTimeout) {
+        DLOGV("Waiting on all references to be returned...%d", pSctpContext->contextRefCnt);
+        THREAD_SLEEP(100 * HUNDREDS_OF_NANOS_IN_A_MILLISECOND);
+    }
+
+    if (IS_VALID_MUTEX_VALUE(pSctpContext->sctpContextLock)) {
+        MUTEX_FREE(pSctpContext->sctpContextLock);
+        pSctpContext->sctpContextLock = INVALID_MUTEX_VALUE;
+    }
+
+    DLOGI("Destroyed SCTP context");
+
+CleanUp:
+    CHK_LOG_ERR(retStatus);
+
+    LEAVES();
+    return retStatus;
+}
+
 STATUS initSctpAddrConn(PSctpSession pSctpSession, struct sockaddr_conn* sconn)
 {
     ENTERS();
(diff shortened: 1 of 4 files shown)


src/source/Sctp/Sctp.c
```diff
@@ -300,6 +322,24 @@ STATUS putSctpPacket(PSctpSession pSctpSession, PBYTE buf, UINT32 bufLen)
...
+
+    CHK(pSctpSession != NULL, STATUS_NULL_ARG);
+    CHK(ATOMIC_LOAD(&pSctpSession->shutdownStatus) == SCTP_SESSION_ACTIVE, retStatus);
+
+    elapsedMs = (currentTime - pSctpSession->lastTickTime) / HUNDREDS_OF_NANOS_IN_A_MILLISECOND;
+    pSctpSession->lastTickTime = currentTime;
+    usrsctp_handle_timers((UINT32) elapsedMs);
+
+CleanUp:
+    return retStatus;
```
sirknightj, line 444: Can we add CHK_LOG_ERR?

src/source/Sctp/Sctp.c
```diff
@@ -300,6 +322,24 @@ STATUS putSctpPacket(PSctpSession pSctpSession, PBYTE buf, UINT32 bufLen)
...
+    STATUS retStatus = STATUS_SUCCESS;
+    PSctpSession pSctpSession = (PSctpSession) customData;
+    UINT64 elapsedMs;
+
+    CHK(pSctpSession != NULL, STATUS_NULL_ARG);
+    CHK(ATOMIC_LOAD(&pSctpSession->shutdownStatus) == SCTP_SESSION_ACTIVE, retStatus);
+
+    elapsedMs = (currentTime - pSctpSession->lastTickTime) / HUNDREDS_OF_NANOS_IN_A_MILLISECOND;
+    pSctpSession->lastTickTime = currentTime;
+    usrsctp_handle_timers((UINT32) elapsedMs);
```
sirknightj, line 435: Is this a global function? Is there implication if multiple peerconnections are created?

sirknightj, line 435: Noting: `usrsctp_init_nothreads` mode has no internal locking, so `usrsctp_handle_timers` running on the timer queue thread can race with `usrsctp_conninput` (DTLS receive thread) and `usrsctp_sendv` (application thread). The frequent timerQueue invocations adds to the risk.

If data channel stability issues resurface under high concurrency, serializing all `usrsctp_*` calls behind a global mutex would be the fix and reduce timerQueue frequency.

sirknightj (commented):
Good bug fix. The root cause is clear - `usrsctp_init_nothreads(0, ...)` disables the internal timer thread, so usrsctp_handle_timers() was never being called, causing SCTP retransmissions to never fire and data channels to stall on packet loss. The fix adds a periodic timer via the SDK's timer queue. CI is all green.

Could you please target the `develop` branch? `develop` gets periodically merged to `main` as part of SDK release cycle.

src/source/Sctp/Sctp.c
kevincgall-stackav: Did a little tracing through `usrsctp` to see what the practical implications are for race conditions:

- `usrsctp_handle_timers` delegates directly to [sctp_handle_tick](https://github.com/sctplab/usrsctp/blob/fd070e05a7474f38c7fecdf4d4b6005d2547ee00/usrsctplib/user_socket.c#L3338)
- `sctp_handle_tick` acquires a [timer queue lock](https://github.com/sctplab/usrsctp/blob/fd070e05a7474f38c7fecdf4d4b6005d2547ee00/usrsctplib/netinet/sctp_callout.c#L160) while reading the call queue of timer functions
- The only place the callqueue is written to is `sctp_os_timer_start` ([here](https://github.com/sctplab/usrsctp/blob/fd070e05a7474f38c7fecdf4d4b6005d2547ee00/usrsctplib/netinet/sctp_callout.c#L127)), also while holding the same lock.

`sctp_os_timer_start` is called in many places, but I think by tracing the lock, we can conclude that it is safe to call code that adds timer callbacks from multiple threads, even while the callback queue is being read.

kevincgall-stackav: Good call! Yes, this is a global function, so if 2 peer connections are created, the practical effect will be that the sctp state machine will be advanced approx. twice as fast as intended. In my application, I only ever have 1 peer connection, so that never came up for me!

I'll try to get a commit with a proposal for deconflicting global calls between peer connections and post the commit hash here. Let me know what you're thinking.

src/source/Sctp/Sctp.c
kevincgall-stackav: 95506cf32bda699bb25a3b87dff91ef581072dab

src/source/Sctp/Sctp.c
kevincgall-stackav: Proposed fix: 88694ef7b93a1f08c9f10f640a51a91335d120e8

I replicated the `WebRtcClientContext` idea to create a singleton context that stores the global last tick. I kept the per-peer connection timer though:

- Don't need to create a global always-running timer queue
- If there is no active peer connection, don't run any timers

The cost is that if there are N peer connections, we'll still invoke the timer Nx as many times, though the actual timer behavior should be correct now. It also requires a mutex now. Let me know what you think.

src/source/Sctp/Sctp.c
```diff
@@ -1,6 +1,85 @@
 #define LOG_CLASS "SCTP"
 #include "../Include_i.h"
 
+PSctpContext getSctpContext()
```
Ahhorse1: Could we rename these functions to `acquireSctpContext` and `releaseSctpContext`? `getSctpContext` could be misinterpeted as just a getter function and renaming makes the lifecycle clear that every acquire should be followed with a release.

src/source/Sctp/Sctp.c
```diff
@@ -1,6 +1,85 @@
...
+    PSctpContext pSctpContext = getSctpContext();
+
+    DLOGD("Releasing SCTP context instance from cleanupSctpContext");
+    releaseHoldOnSctpContext(pSctpContext);
+
+    CHK_WARN(ATOMIC_LOAD_BOOL(&pSctpContext->isSctpInitialized), STATUS_INVALID_OPERATION, "SCTP context not initialized, nothing to clean up");
+
+    ATOMIC_STORE_BOOL(&pSctpContext->isSctpInitialized, FALSE);
+
+    while (ATOMIC_LOAD(&pSctpContext->contextRefCnt) > 0) {
```
Ahhorse1: Could we add a timeout for this loop?

src/source/Sctp/Sctp.c
Ahhorse1: Given we are setting the last tick time during the `initSctpSession`, there could be an extended amount on time before the first `createSctpSession` is called for the connection. Would it be better to set the last tick time when the first session is created?

kevincgall-stackav: When a new peer connection is created, if we reset the last tick time there, it will throw off the timing for the original peer connection's queued callbacks.
We could coordinate that with additional fields on the global context, but another option would be to always invoke `usrsctp_handle_timers` on SCTP session creation and reset the last tick time then and there. That would end up being most correct in my estimation - we'd effectively reset the clock w.r.t. elapsed time calculations for the new peer connection and correctly advance any in-progress handlers for the existing peer connection.

I'll do that and post the commit hash

src/source/Sctp/Sctp.c
kevincgall-stackav: c69be3b2e4a024e226192051ea70cea34e3fbd6d
Resolving

src/source/Sctp/Sctp.c
kevincgall-stackav: Done in c69be3b2e4a024e226192051ea70cea34e3fbd6d

src/source/Sctp/Sctp.c
kevincgall-stackav: Proposed change in c69be3b2e4a024e226192051ea70cea34e3fbd6d

Let me know if this satisfies your concerns. Thanks!

Ahhorse1: Invoking the timer here makes sense to update the last tick time. Should it be placed prior to `usrsctp_socket` so that none of the new session's SCTP timers (e.g., the INIT retransmit timer scheduled by `usrsctp_connect`) are scheduled against a stale lastTickTime and fire immediately?

kevincgall-stackav: Sounds good.
6e547b4965ed9b9d6eb35d6b5ea6b8b507524bd7

Ahhorse1:
Thank you for the contribution! We are currently performing some additional validations on our end, and we will update this PR when those are finished.

kevincgall-stackav:
> Thank you for the contribution! We are currently performing some additional validations on our end, and we will update this PR when those are finished.

Hey folks! Any word on the validations? Are we expecting this to merge any time soon?
Thanks!

sirknightj:
Hi @kevincgall-stackav, apologies for the delay. Could you please change the target branch to `develop`? We have performed the additional validations (manual fault-inject) and the changes look good on our side.

sirknightj (approved):

sirknightj (approved):

YPandas (approved):

PR #2322, opened 2026-06-03
Optimization: don't call the TWCC feedback callback for empty packets

sirknightj:
*Issue #, if available:*
- N/A

*What was changed?*
- Skip TWCC parsing and callback invocation when `packetStatusCount == 0` (empty report)
- Cap feedback list allocation size to packetStatusCount in the custom callback path

*Why was it changed?*
- There is no useful work to do for an empty TWCC report. No packet statuses to parse, no delay trends to compute, and no feedback to deliver to the application.
- Previously, an empty report would still enter the full parsing and callback path. Because `parseRtcpTwccPacket` doesn't update `lastReportedSeqNum` when there are zero packets to iterate, the downstream `reportLen` calculation in `onRtcpTwccPacket` operates on a stale value, producing an unnecessary heap allocation and invoking the callback with zero useful entries. Skipping early avoids this wasted work and the associated allocation overhead.

*How was it changed?*
- parseRtcpTwccPacket: Early-return STATUS_SUCCESS when packetStatusCount == 0. No chunks or deltas exist to parse in this case.
- onRtcpTwccPacket: In the custom `onTwccFeedbackReceived` callback path, skip if `packetStatusCount == 0` (nothing to report). Additionally, cap `reportLen` to `packetStatusCount` to tightly bound the allocation to the actual report size rather than the full sequence number range.

  - Added three new tests in RtcpFunctionalityTest:
    - `twccZeroPacketStatusCountDoesNotInvokeCallback`: Verifies the callback is not invoked for a zero-count TWCC packet.
    - `twccZeroStatusCountAfterValidReportNoSpuriousCallback`: Verifies that after a valid report, a subsequent empty report does not spuriously invoke the callback.
    - `twccSeqNumWraparoundInvokesCallbackCorrectly`: Verifies that a legitimate report spanning the UINT16 wraparound boundary (seqNums 65534..1) correctly invokes the callback with all 4 reported packets, confirming the `reportLen` cap does not break valid wraparound behavior.
  - All 17 TWCC-related tests pass with zero regressions.

### DIFF
--- a/src/source/PeerConnection/Rtcp.c
+++ b/src/source/PeerConnection/Rtcp.c
@@ -524,6 +524,7 @@ STATUS onRtcpTwccPacket(PRtcpPacket pRtcpPacket, PKvsPeerConnection pKvsPeerConn
         // Also cap reportLen to packetStatusCount to bound the allocation size
         CHK(packetStatusCount > 0, STATUS_SUCCESS);
         if (reportLen > packetStatusCount) {
+            DLOGD("Malformed TWCC packet: reportLen %u exceeds packetStatusCount %u", reportLen, packetStatusCount);
             reportLen = packetStatusCount;
         }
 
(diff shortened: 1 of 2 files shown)


src/source/PeerConnection/Rtcp.c
```diff
@@ -514,6 +518,14 @@ STATUS onRtcpTwccPacket(PRtcpPacket pRtcpPacket, PKvsPeerConnection pKvsPeerConn
         UINT64 twccPktVal = 0;
         PTwccRtpPacketInfo pPktInfo = NULL;
         UINT16 reportLen = (UINT16) (pTwccManager->lastReportedSeqNum - pTwccManager->prevReportedBaseSeqNum + 1);
+        UINT16 packetStatusCount = TWCC_PACKET_STATUS_COUNT(pRtcpPacket->payload);
+
+        // Skip callback for empty reports
+        // Also cap reportLen to packetStatusCount to bound the allocation size
+        CHK(packetStatusCount > 0, STATUS_SUCCESS);
```
unicornss: does it help to have verbose or debug?

unicornss (approved):

src/source/PeerConnection/Rtcp.c
sirknightj: Sure we can add that

Ahhorse1: Is there a reason we want the CHK inside the if statement? Do we not want to skip empty packets for the default trendline

sirknightj: Added

sirknightj: Hoisted it above the if statement to skip both callbacks

sirknightj: Added a test to confirm all of them are skipped as well

Ahhorse1 (approved):

PR #2323, opened 2026-06-04
Handle null parameter to parseExtId

sirknightj:
*Issue #, if available:*
- N/A

*What was changed?*
- Fixed a logic bug in `parseExtId` where `&&` should have been `||`, causing a NULL pointer dereference. Note: this code path is unreachable since the input string is guaranteed to not be null.

*Why was it changed?*
  - The guard condition (`extmapValue == NULL && STRCHR(extmapValue, ' ') == NULL`) evaluates the second operand even when `extmapValue` is `NULL` because C short-circuits `&&` only when the left operand is false. Since the left operand is true when `NULL`, `STRCHR(NULL, ' ')` executes, which is undefined behavior.
  - The call site currently guarantees non-NULL since it only call `parseExtId` after a successful `STRSTR` match), so this is a latent defect, but any future caller passing `NULL` would crash.

*How was it changed?*
  - Changed `&&` to `||` in the `NULL`/space guard in `parseExtId`. Now if `extmapValue` is `NULL`, the function returns 0 immediately without evaluating `STRCHR`.

- Added unit test to exercise this path
- It fails before the change, and passes after the change

### DIFF
--- a/src/source/PeerConnection/PeerConnection.c
+++ b/src/source/PeerConnection/PeerConnection.c
@@ -1372,14 +1372,19 @@ CleanUp:
 UINT32 parseExtId(PCHAR extmapValue)
 {
     ENTERS();
-    UINT32 extid = 0;
-    if (extmapValue == NULL || STRCHR(extmapValue, ' ') == NULL) {
-        LEAVES();
-        return 0;
-    }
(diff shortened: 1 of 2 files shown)


Ahhorse1 (approved):

unicornss (approved):
DLOGW?

Ahhorse1 (approved):

PR #2324, opened 2026-06-04
esp_video_if: avoid memory fragmentation by re-using buffers

vikramdattu:
*Issue #, if available:*
- N/A

*What was changed?*
- esp_video_if: switch v4l2 frame buffers to USERPTR I/O and allocate once.

*Why was it changed?*
- v4l2 reallocated raw frame buffers on every init/deinit, fragmenting PSRAM. After a few sessions the camera couldn't secure contiguous frames.

*How was it changed?*
- Allocate the raw frame buffers up-front and use USERPTR I/O so the same buffers are queued/dequeued across sessions instead of being freed and re-malloc'd.

- ESP32-P4-EYE: repeated camera start/stop + WebRTC reconnect cycles; the previous fragmentation path no longer triggers.

### DIFF
--- a/esp_port/components/media_stream/src/esp_video_if.c
+++ b/esp_port/components/media_stream/src/esp_video_if.c
@@ -205,6 +205,14 @@ static esp_err_t queue_all_buffers(v4l2_src_t *v4l2)
 
         if (ioctl(v4l2->cap_fd, VIDIOC_QBUF, &v4l2->v4l2_buf[i]) < 0) {
             ESP_LOGE(TAG, "Failed to requeue buffer %d, errno: %d", i, errno);
+            /* Buffers 0..i-1 are already queued to the driver. Flush them with
+             * STREAMOFF (legal before STREAMON; returns every queued buffer to
+             * the dequeued state) so the driver is not left half-queued and a
+             * later start attempt begins from a clean queue. */
+            int type = V4L2_BUF_TYPE_VIDEO_CAPTURE;
+            if (i > 0 && ioctl(v4l2->cap_fd, VIDIOC_STREAMOFF, &type) < 0) {
+                ESP_LOGE(TAG, "STREAMOFF after partial queue failed, errno: %d (full re-init required)", errno);
+            }
             return ESP_FAIL;
         }
     }
@@ -226,6 +234,15 @@ static video_fb_t *video_fb_get_cb(void *cb_ctx)
             int ret = ioctl(v4l2->cap_fd, VIDIOC_DQBUF, &buf);
             if (ret != 0) {
                 ESP_LOGE(TAG, "failed to receive video frame ret %d", ret);
+                return NULL;
+            }
+
+            /* Guard the driver-supplied index before it touches fb_used[] /
+             * cap_buffer[] / v4l2_buf[] (all sized BUFFER_COUNT): a buggy or
+             * racing driver returning an unexpected index must not become an
+             * out-of-bounds write. */
+            if (buf.index >= BUFFER_COUNT) {
+                ESP_LOGE(TAG, "DQBUF returned out-of-range buffer index %u", (unsigned) buf.index);
                 return NULL;
             }
 
--- a/esp_port/components/media_stream/src/esp_video_if.c
+++ b/esp_port/components/media_stream/src/esp_video_if.c
@@ -639,7 +639,24 @@ static esp_err_t configure_camera_format(v4l2_src_t *v4l2, uint32_t pixelformat)
         }
     }
 
-    ESP_LOGE(TAG, "Failed to set any supported format. Check the camera resolution in menuconfig");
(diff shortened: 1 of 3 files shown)


esp_port/components/media_stream/src/esp_video_if.c
```diff
@@ -155,28 +162,80 @@ static void video_stop_cb(void *cb_ctx)
...
+            v4l2->v4l2_buf[i].length = v4l2->buffer_size[i];  /* Restore from saved size */
+            if (v4l2->buffer_size[i] == 0) {
+                ESP_LOGE(TAG, "USERPTR buffer size not set for buffer %d", i);
+                return ESP_FAIL;
+            }
+            v4l2->v4l2_buf[i].bytesused = 0;
+        }
+
+        if (ioctl(v4l2->cap_fd, VIDIOC_QBUF, &v4l2->v4l2_buf[i]) < 0) {
+            ESP_LOGE(TAG, "Failed to requeue buffer %d, errno: %d", i, errno);
```
sirknightj, line 219: If VIDIOC_QBUF fails for buffer 2 (of 0,1,2), buffers 0 and 1 are already queued to the driver but the function returns ESP_FAIL. The caller (restart_streaming_with_existing_buffers) then returns FAIL without calling STREAMOFF, leaving the driver in an inconsistent state where some buffers are queued but streaming never started.
  
Consider:
  1. Calling STREAMOFF in the error path to release queued buffers back, or
  2. Noting in a comment that partial queue failure is non-recoverable and requires a full cleanup (call esp_video_if_cleanup())

esp_port/components/media_stream/src/esp_video_if.c
```diff
@@ -155,28 +162,80 @@ static void video_stop_cb(void *cb_ctx)
...
                 return NULL;
             }
-            v4l2->fb_used[i] = true;
-            v4l2->fb.buf = v4l2->cap_buffer[buf->index];
-            // printf("Captured buffer num: %d\n", (int) buf->index);
-            v4l2->fb.len = buf->bytesused;
-            // v4l2->fb.width = format.fmt.pix.width;
-            // v4l2->fb.height = format.fmt.pix.height;
+
+            v4l2->fb_used[buf.index] = true;
```
sirknightj, line 261: After `VIDIOC_DQBUF`, `buf.index` is used directly to index into `fb_used[]`, `cap_buffer[]`, and `v4l2_buf[]` arrays (all sized BUFFER_COUNT=3) without a bounds check. If the driver ever returns an unexpected index (e.g., due to a driver bug or race condition), this would be an out-of-bounds array write - a memory corruption that's extremely hard to debug on embedded targets.

The original code had the same latent issue but was partially protected because it used the loop index `i` for `fb_used[i]` (so only `cap_buffer[buf->index]` was unbounded). The new code uses `buf.index` for all three arrays, increasing the blast radius.

Suggest adding a bounds check after DQBUF:

  ```c
  int ret = ioctl(v4l2->cap_fd, VIDIOC_DQBUF, &buf);
  if (ret != 0) {
      ESP_LOGE(TAG, "failed to receive video frame ret %d", ret);
      return NULL;
  }
  if (buf.index >= BUFFER_COUNT) {
      ESP_LOGE(TAG, "DQBUF returned invalid index %d (max %d)", buf.index, BUFFER_COUNT - 1);
      return NULL;
  }

sirknightj (commented):
Well-structured refactoring to avoid memory fragmentation on ESP32 by reusing V4L2 buffers across start/stop cycles. Includes decomposition into helper functions.

esp_port/components/media_stream/src/esp_video_if.c
vikramdattu: Fixed in `ecb43c8` — on a mid-loop QBUF failure the queued buffers are now flushed with STREAMOFF (legal before STREAMON; returns them to the dequeued state) so the driver isn't left half-queued for the next start attempt.

esp_port/components/media_stream/src/esp_video_if.c
vikramdattu: Fixed in `ecb43c8` — `buf.index` is validated against `BUFFER_COUNT` right after DQBUF, before it touches any of the three arrays; an out-of-range index logs and returns NULL instead of corrupting memory.

Also pushed `eaa16d5`: it syncs this PR with a fix we made downstream after this change originally shipped — re-QBUFing descriptors that still carry the previous session's DQBUF state desyncs the driver queue (STREAMON succeeds, DQBUF never delivers → black video on session 2+), and the close/reopen cycle in deinit is the fragile transition. The PR now rebuilds descriptors from scratch and keeps the fd open (STREAMOFF-only stop) — verified on hardware across 3+ consecutive sessions. Without this, the buffer-reuse change would introduce that regression here.

PR #2325, opened 2026-06-04
Add opt-in mbedTLS 4 support (-DUSE_MBEDTLS4=ON)

### LINKED ISSUE #2296: [Feature request]: mbedtls 4 support

vikramdattu:
*Issue #, if available:*
- Closes #2296 (mbedTLS 4 support feature request).
- Follow-up to #2299 (libsrtp 3.x opt-in, merged 2026-06-04), which was the precondition.

*What was changed?*
- New CMake option `-DUSE_MBEDTLS4=ON` selects mbedTLS 4.1.0 instead of 3.6.6 for `BUILD_DEPENDENCIES`. Implies `USE_LIBSRTP3=ON` (the only libsrtp line that supports mbedTLS 4).
- Bumped `libwebsockets` pin (mbedTLS 4 arm only) to the **v5.0.0 release tag** — the first release carrying the mbedtls4 migration. Earlier revisions of this PR pinned a `main` SHA out of necessity; that's gone, along with the build-time parallel-connect patch (warmcat/libwebsockets#3633 is included in v5.0.0). The default (mbedTLS 3.x / OpenSSL) builds keep v4.3.10.
- Source-side compatibility layer for mbedTLS 4: PSA-backed cert/key generation, `mbedtls_ssl_export_keying_material` for SRTP keying, `mbedtls_ssl_is_handshake_over` helper, and the `MBEDTLS_DECLARE_PRIVATE_IDENTIFIERS` shortcut so legacy headers still expose the symbols our hot paths use.
- One CI matrix entry — `Shared MbedTLS libsrtp3 mbedTLS4` — gating the opt-in path.

*Why was it changed?*
- Consumers on platforms shipping mbedTLS 4 (ESP-IDF v6.0 is the first) currently have no path forward. v3.6.6 stays the default; mbedTLS 4 is one CMake flag away. Avoids forcing a downstream downgrade.

*How was it changed?*
- 4 commits: build wiring (USE_MBEDTLS4 + lws v5.0.0 pin), source crypto migration, signaling fixes, CI + README.
- Defaults untouched — `-DUSE_MBEDTLS4=ON` is opt-in. mbedTLS 2/3 paths are not modified beyond the small shared helpers (handshake-over gate).
- libwebsockets `LWS_ROLE_QUIC=OFF` + `LWS_WITH_HTTP3=OFF` — unused by this SDK; kept off to minimise the dependency build.

- Local Mac: `cmake -DUSE_MBEDTLS=ON -DUSE_LIBSRTP3=ON -DUSE_MBEDTLS4=ON -DBUILD_TEST=ON && make -j` → clean build; libtfpsacrypto links in.
- `webrtc_client_test --gtest_filter="DtlsApiTest.*:DtlsFunctionalityTest.*:SrtpApiTest.*"` → **9/9 PASS** (cert gen, DTLS handshake both directions, SRTP encrypt/decrypt).
- v3 regression build (`-DUSE_MBEDTLS=ON` without `USE_MBEDTLS4`) passes — existing path unchanged.
- `bash scripts/check-clang.sh` → zero diff.
- After the v5.0.0 repin (2026-07-27): full Linux build with `-DUSE_OPENSSL=OFF -DUSE_MBEDTLS=ON -DUSE_MBEDTLS4=ON` → clean; installed lws reports 5.0.0.


vikramdattu:
@sirknightj and @unicornss PTAL

CMake/Dependencies/libwebsockets-CMakeLists.txt
```diff
@@ -28,9 +28,34 @@ endif()
...
+    #   - GIT_SUBMODULES "README.md" sidesteps broken qpack/qifs gitlink that
+    #     has no .gitmodules URL entry (tracked: warmcat upstream PR pending,
+    #     see esp-rainmaker task #130)
+    #   - LWS_ROLE_QUIC=OFF / LWS_WITH_HTTP3=OFF skip in-flight features that
+    #     don't compile cleanly (tracked: warmcat upstream PR pending,
+    #     see esp-rainmaker task #131)
+    # v4.5.8 (latest release tag) predates the mbedtls4 migration. Revert to
+    # GIT_TAG vX.Y.Z + GIT_SHALLOW TRUE + drop GIT_SUBMODULES + drop the
+    # LWS_ROLE_QUIC/LWS_WITH_HTTP3 lines once the upstream release ships.
+    GIT_TAG           7726361cf453ec7bb83eb33ec6d5f468219c1c4d
```
sirknightj: This version bump from v4.3.10 to a mainline SHA is unconditional - it applies to all builds (OpenSSL, mbedTLS 3.x, etc.), not just mbedTLS 4 builds. 

Could this SHA pin be made conditional on `USE_MBEDTLS4`, keeping `v4.3.10` for the default path? Something like:

  ```cmake
  if (USE_MBEDTLS4)
    set(LWS_GIT_TAG 7726361cf453ec7bb83eb33ec6d5f468219c1c4d)
    set(LWS_GIT_SHALLOW FALSE)
    set(LWS_GIT_SUBMODULES "README.md")
  else()
    set(LWS_GIT_TAG v4.3.10)
    set(LWS_GIT_SHALLOW TRUE)
    set(LWS_GIT_SUBMODULES "")
  endif()
```

File: .github/workflows/ci.yml [hunk 2]
```diff
@@ -86,6 +86,8 @@ jobs:
             cmake_flags: "-DBUILD_TEST=ON -DBUILD_STATIC_LIBS=ON -DUSE_OPENSSL=OFF -DUSE_MBEDTLS=ON -DPARALLEL_BUILD=ON"
           - name: Shared MbedTLS libsrtp3
             cmake_flags: "-DBUILD_TEST=ON -DUSE_OPENSSL=OFF -DUSE_MBEDTLS=ON -DUSE_LIBSRTP3=ON -DPARALLEL_BUILD=ON"
+          - name: Shared MbedTLS libsrtp3 mbedTLS4
+            cmake_flags: "-DBUILD_TEST=ON -DUSE_OPENSSL=OFF -DUSE_MBEDTLS=ON -DUSE_LIBSRTP3=ON -DUSE_MBEDTLS4=ON -DPARALLEL_BUILD=ON"
```
sirknightj, line 92: The new mbedTLS4 CI jobs are failing with:

  `ModuleNotFoundError: No module named 'jsonschema'`

Please take a look.

src/source/Crypto/Dtls_mbedtls.c
```diff
@@ -783,6 +809,37 @@ STATUS copyCertificateAndKey(mbedtls_x509_crt* pCert, mbedtls_pk_context* pKey,
...
+        psa_status_t psaStatus;
+        psa_key_attributes_t attr = PSA_KEY_ATTRIBUTES_INIT;
+        mbedtls_svc_key_id_t newKeyId = MBEDTLS_SVC_KEY_ID_INIT;
+        BOOL keyWrapped = FALSE;
+
+        CHK(psa_crypto_init() == PSA_SUCCESS, STATUS_CERTIFICATE_GENERATION_FAILED);
+        /* Match createCertificateAndKey's policy: both SIGN_HASH and SIGN_MESSAGE.
+         * The source key must have PSA_KEY_USAGE_EXPORT for mbedtls_pk_import_into_psa
+         * to succeed — a hardened caller-provided key without EXPORT will fail here. */
+        if (mbedtls_pk_get_psa_attributes(pKey, PSA_KEY_USAGE_SIGN_HASH | PSA_KEY_USAGE_SIGN_MESSAGE, &attr) != 0) {
```
sirknightj: The comment above notes that "the source key must have PSA_KEY_USAGE_EXPORT for mbedtls_pk_import_into_psa to succeed." However, createCertificateAndKey does not set EXPORT/COPY on its generated keys (line 951: "EXPORT/COPY are intentionally not set on this ephemeral key").

I believe this is fine because copyCertificateAndKey is only called with user-provided certificates (not SDK-generated ones), but it's a new restriction in v4 that didn't exist in v3.x. Users who provide their own certs will get an opaque STATUS_CERTIFICATE_GENERATION_FAILED if their PSA key lacks EXPORT permission.

Worth considering adding action item for the user (configuration issue):
  1. A more specific error log when this path fails (maybe log "source key requires PSA_KEY_USAGE_EXPORT for copyCertificateAndKey to succeed on mbedTLS 4")
  2. Documenting this requirement in the README's mbedTLS 4 notes

src/source/Crypto/Dtls_mbedtls.c
```diff
@@ -851,6 +909,103 @@ int mbedtls_x509write_crt_set_serial(mbedtls_x509write_cert* ctx, const mbedtls_
...
+        psa_set_key_type(&keyAttr, PSA_KEY_TYPE_ECC_KEY_PAIR(PSA_ECC_FAMILY_SECP_R1));
+        psa_set_key_bits(&keyAttr, 256);
+        psa_set_key_usage_flags(&keyAttr, PSA_KEY_USAGE_SIGN_HASH | PSA_KEY_USAGE_SIGN_MESSAGE);
+        /* MBEDTLS_PK_ALG_ECDSA resolves to deterministic ECDSA when available
+         * (the v4 default for TLS); set the enrollment alg to plain ECDSA so
+         * the key matches both flavours during ssl_pick_cert. */
+        psa_set_key_algorithm(&keyAttr, MBEDTLS_PK_ALG_ECDSA(PSA_ALG_ANY_HASH));
+        psa_set_key_enrollment_algorithm(&keyAttr, PSA_ALG_ECDSA(PSA_ALG_ANY_HASH));
+    }
+    CHK(psa_generate_key(&keyAttr, &keyId) == PSA_SUCCESS, STATUS_CERTIFICATE_GENERATION_FAILED);
```
sirknightj: Suggestion: the v4 path for `createCertificateAndKey` has multiple PSA failure points (psa_crypto_init, psa_generate_key, mbedtls_pk_wrap_psa, mbedtls_x509write_crt_der) that all map to the same `STATUS_CERTIFICATE_GENERATION_FAILED`. If any of these fail, a user will only see "certificate generation failed" with no indication of which step broke.

The `copyCertificateAndKey` v4 path does a nice job of this - it logs specific messages on `mbedtls_pk_get_psa_attributes` and `mbedtls_pk_import_into_psa` failures. Consider adding similar `DLOGE` calls before the CHK in this function, e.g.:

  ```c
  psaStatus = psa_generate_key(&keyAttr, &keyId);
  if (psaStatus != PSA_SUCCESS) {
      DLOGE("psa_generate_key failed (status %d)", (int) psaStatus);
  }
  CHK(psaStatus == PSA_SUCCESS, STATUS_CERTIFICATE_GENERATION_FAILED);
```

Same for mbedtls_pk_wrap_psa and mbedtls_x509write_crt_der.

File: src/source/Crypto/Dtls_mbedtls.c [hunk 5]
```diff
@@ -681,19 +693,29 @@ STATUS dtlsSessionPopulateKeyingMaterial(PDtlsSession pDtlsSession, PDtlsKeyingM
...
 
     MUTEX_LOCK(pDtlsSession->sslLock);
     locked = TRUE;
 
+#if MBEDTLS_V4_OR_LATER
+    /* mbedTLS 4 removes mbedtls_ssl_tls_prf(). Pull the SRTP keying material
+     * directly via the RFC 5705 exporter (same shape we use for OpenSSL). */
+    CHK(mbedtls_ssl_export_keying_material(&pDtlsSession->sslCtx, keyingMaterialBuffer, ARRAY_SIZE(keyingMaterialBuffer), KEYING_EXTRACTOR_LABEL,
+                                           ARRAY_SIZE(KEYING_EXTRACTOR_LABEL) - 1, NULL, 0, 0) == 0,
+        STATUS_INTERNAL_ERROR);
```
sirknightj: if mbedtls_ssl_export_keying_material fails, the caller only sees STATUS_INTERNAL_ERROR with no log. Since this is the new keying-material derivation path (replacing the v3 PRF-based approach), and it can fail if the handshake didn't complete properly or the SSL context is in an unexpected state, a LOG_MBEDTLS_ERROR before or after the CHK would help troubleshooting:

```c
int ret = mbedtls_ssl_export_keying_material(&pDtlsSession->sslCtx, keyingMaterialBuffer, ...);
if (ret != 0) {
    LOG_MBEDTLS_ERROR("mbedtls_ssl_export_keying_material", ret);
}
CHK(ret == 0, STATUS_INTERNAL_ERROR);
```

This matches the pattern used elsewhere in the file (e.g. LOG_MBEDTLS_ERROR("mbedtls_ssl_handshake", ...) on line 368).

CMake/Dependencies/libwebsockets-CMakeLists.txt
vikramdattu: Done. lws revision is now gated as you suggested.

.github/workflows/ci.yml
vikramdattu: Fixed — the mbedTLS-4 job pulls a mainline libwebsockets whose codegen imports jsonschema (the v4.3.10 path doesn't). Added a setup-python + pip install jsonschema step gated to the USE_MBEDTLS4 config.

src/source/Crypto/Dtls_mbedtls.c
vikramdattu: Good catch! Testing against a live viewer, the assumption that copyCertificateAndKey only sees user-provided certs doesn't hold: apps that pre-generate a cert via createRtcCertificate and pass it back in RtcConfiguration hit this path too, and it failed until the generated key set PSA_KEY_USAGE_EXPORT.

Added that, documented the requirement for user-supplied certs in the README's mbedTLS-4 note, and also fixed mbedtls_pk_get_psa_attributes being called with OR'd usage flags (its switch rejects that with PK_TYPE_MISMATCH), now derives from SIGN_HASH and adds SIGN_MESSAGE.

src/source/Crypto/Dtls_mbedtls.c
vikramdattu: Done

src/source/Crypto/Dtls_mbedtls.c
vikramdattu: Done, added LOGE on failure paths

sirknightj:
Hi @vikramdattu, some of the CI is failing, please take a look!

.github/workflows/ci.yml
```diff
@@ -131,6 +133,18 @@ jobs:
...
+      # codegen step imports jsonschema; the v4.3.10 release path does not.
+      - name: Set up Python (libwebsockets mainline codegen)
+        if: ${{ contains(matrix.config.cmake_flags, 'USE_MBEDTLS4') }}
+        uses: actions/setup-python@v5
+        with:
+          python-version: "3.12"
+
+      - name: Install jsonschema + jinja2
+        if: ${{ contains(matrix.config.cmake_flags, 'USE_MBEDTLS4') }}
+        run: pip install jsonschema jinja2
```
vikramdattu, line 146: Heads-up @sirknightj — before this change the mbedTLS-4 jobs failed at `ModuleNotFoundError: No module named 'jinja2'`. The lws-mainline / tf-psa-crypto codegen imports **jinja2** in addition to `jsonschema`, so I added it here. Since it applies to the mbedTLS-4 build generally (not just this PR), it may be worth folding into the base CI setup.

(The other red job, `mbedtls-ubuntu-gcc-4-4`, was unrelated: GCC 4.4 predates `__has_include`, so the one-line `defined(__has_include) && __has_include(...)` was a parse error — fixed in `Include_i.h` with a nested `#ifdef __has_include` guard.)

sirknightj: Thanks for sharing the details - Would you be able to add the MbedTLS4 path install prereq's to the README? (Python 3.12/Pip modules)

sirknightj: Also: the updated CI is now failing with:

```
fatal: unable to read tree (7316933aefbc10369251c3b6717b166c3f7a86ca)
CMake Error at /Users/runner/work/amazon-kinesis-video-streams-webrtc-sdk-c/amazon-kinesis-video-streams-webrtc-sdk-c/open-source/libwebsockets/build/tmp/project_libwebsockets-gitclone.cmake:61 (message):
  Failed to checkout tag: '7316933aefbc10369251c3b6717b166c3f7a86ca'
```

Could you please check?

vikramdattu: Root-caused: the pinned SHA `7316933a` was warmcat main's *tip* when I set it, but warmcat force-pushed main on 2026-07-09 (`7316933a` → `4cc6ce14`, histories diverged — confirmed via the GitHub events log), which orphaned the commit and broke the fetch.

Re-pinned to the current main tip in `ba65ca5`. This is inherently fragile though — warmcat main gets rewritten routinely — so the durable fix is to **wait for Andy to tag an lws release carrying the mbedTLS 4 migration** and pin that tag instead. I'll follow up with upstream on the tag; until then the SHA may need occasional re-bumps if main is rewritten again. (The CMake comment now documents this.)

vikramdattu: Added to the README (`92ddf4a`): mbedTLS 4 build-prereqs note in the build matrix section — Python 3 with `jinja2` + `jsonschema` for the TF-PSA-Crypto codegen, with the exact `ModuleNotFoundError` you'd see without them.

On the lws pin instability more broadly: I've asked upstream for a release on warmcat/libwebsockets#3607, which would let this pin move to a durable tag.

vikramdattu:
@sirknightj The CI fixes are in: re-pinned the lws SHA (root cause: warmcat force-pushed main, orphaning the old tip — details in the thread above), fixed a second gcc-4.4 `__has_include` instance in Tls.h, and added the mbedTLS-4 build prereqs to the README. The workflow runs for the latest push are sitting in **"action required"** — could you approve the runs so we get a CI verdict?

sirknightj:
Thanks @vikramdattu. I pulled in the latest. It looks like the CI is running into an issue with the data channel:
```
[----------] 9 tests from DataChannelFunctionalityTest
[ RUN      ] DataChannelFunctionalityTest.createDataChannel_Disconnected
/Users/runner/work/_temp/43c3b0e0-8e86-4ccb-884d-f91c6c2f713d.sh: line 1: 56006 Segmentation fault: 11  ./tst/webrtc_client_test
```

vikramdattu:
Repinned the mbedTLS 4 arm's libwebsockets to the **v5.0.0 release tag** (08b13da57c) — warmcat's first release carrying the mbedtls4 migration. This drops the two workarounds this PR had to carry: the force-push-prone `main` SHA pin and the build-time parallel-connect fds cleanup patch (warmcat/libwebsockets#3633 is included in v5.0.0). Verified with a full `-DUSE_MBEDTLS4=ON` build; installed lws reports 5.0.0. PR description updated to match.

vikramdattu:
@sirknightj @unicornss please approve the workflow, should be good to go.

src/source/Signaling/LwsApiCalls.c
```diff
@@ -607,6 +607,10 @@ STATUS lwsCompleteSync(PLwsCallInfo pCallInfo)
     connectInfo.path = path;
     connectInfo.host = connectInfo.address;
     connectInfo.method = pVerb;
+    // The REST/WSS plumbing here is HTTP/1.1-only. lws v5 offers "h2,http/1.1" by
+    // default and the per-channel endpoints negotiate h2, then reject the request
+    // (HTML 400 on JoinStorageSession/GetIceServerConfig), so pin ALPN to http/1.1.
```
sirknightj, line 612: Wondering if this is an issue with LWS v5?

I observed h2 is negotiated fine with GetIceServerConfig and JoinStorageSession APIs using Curl:

```sh
curl -v --http2 \
  "https://r-xxx.kinesisvideo.us-west-2.amazonaws.com/v1/get-ice-server-config" \
  --aws-sigv4 "aws:amz:us-west-2:kinesisvideo" \
  --user "$AWS_ACCESS_KEY_ID:$AWS_SECRET_ACCESS_KEY" \
  -H "Content-Type: application/json" \
  -d '{"ChannelARN":"arn:aws:kinesisvideo:us-west-2:xxx:channel/xxx/xxx"}'
```
```
* ALPN: curl offers h2,http/1.1
...
* SSL connection using TLSv1.3 / AEAD-AES128-GCM-SHA256 / [blank] / UNDEF
* ALPN: server accepted h2
...
< HTTP/2 200 
< date: Wed, 12 Aug 2026 15:18:46 GMT
< content-type: application/json
< content-length: 1042
<  ...
{
  "IceServerList" : [ {
    "Password" : "...",
    "Ttl" : 300,
```

```sh
curl -v --http2 \
    "$STORAGE_ENDPOINT/joinStorageSession" \
    --aws-sigv4 "aws:amz:us-west-2:kinesisvideo" \
    --user "$AWS_ACCESS_KEY_ID:$AWS_SECRET_ACCESS_KEY" \
    -H "Content-Type: application/json" \
    -d "{\"channelArn\":\"$CHANNEL_ARN\"}"
```

Both return 200 OK

vikramdattu: @sirknightj  you are right. The endpoints accept h2.

1. **Why h2 suddenly appears**: lws 4.3.10 unconditionally took the connection's ALPN from connect_info.alpn (alpn_comma = wsi->stash->cis[CIS_ALPN] in openssl-client.c/mbedtls-client.c), so with it unset (as in this SDK) no ALPN extension was sent at all and servers defaulted to HTTP/1.1. <br>v5.0.0 only overrides the default when the app actually set one, so unset now falls back to the role default h2,http/1.1. (The v4 mbedTLS wrapper also had ALPN compiled out, so every handshake logs mbedtls_ssl_conf_alpn_protocols absent.)

2. **Why our h2 requests get the HTML 400**: our LWS_CALLBACK_CLIENT_HTTP_WRITEABLE handler writes the entire body on every callback with plain LWS_WRITE_HTTP and never uses LWS_WRITE_HTTP_FINAL. lws's minimal-http-client-post example documents FINAL as "necessary to support H2" and warns about extra writeable callbacks. Under h1 that's benign (single callback) but under h2 the stream is never half-closed, a second writeable arrives, and the body is sent twice. DATA overruns content-length and the front-end rejects the stream. Curl command works, as it frames the h2 frame correctly.
 
So: not an lws v5 defect (actually v5 fixed v4's "unset ALPN disables ALPN" open end). The pin keeps the existing h1-shaped plumbing on the protocol it implements.

Suggestion: Making the handler h2-clean (track sent state + LWS_WRITE_HTTP_FINAL) is a small, self-contained follow-up. Happy to file it separately.

vikramdattu: @sirknightj the CI is passed with the fix and the one failure seems some CI side issue. I will rephrase the commit description for pinning alpn to http 1.1 if you want, or can leave as is if it's anyway going to be squashed

sirknightj (approved):

vabhasin (approved):

src/source/Signaling/LwsApiCalls.c
vabhasin: Sounds good, this can be tackled in a separate change.

sirknightj: Should be fine as it is, thanks for the deep dive!

PR #2327, opened 2026-06-08
Add bounds check for data channel name/label copy in onSctpSessionDat…

vabhasin:
*Issue #, if available:*
- N/A

*What was changed?*
- Added bounds check in `onSctpSessionDataChannelOpen` to reject data channel names exceeding `MAX_DATA_CHANNEL_NAME_LEN` (255).
- Fixed memory leak of `pKvsDataChannel` on `hashTablePut` failure path.

*Why was it changed?*
- The `nameLen` parameter originates from the remote peer's DCEP packet `labelLength` field (a UINT16, max 65535). Without validation, `STRNCPY` could write past the 256-byte destination buffers and cause overflow.

*How was it changed?*
- Added `CHK(nameLen <= MAX_DATA_CHANNEL_NAME_LEN, STATUS_INVALID_ARG)` before any allocation or copy. Oversized names are rejected entirely. Data channel is not created when `onDataChannel` callback fires and the error is logged via `CHK_LOG_ERR`
- Added `SAFE_MEMFREE(pKvsDataChannel)` in the `CleanUp` block when `STATUS_FAILED(retStatus)` to prevent a leak if `hashTablePut` fails after allocation.

- Added two unit tests in `DataChannelFunctionalityTest`:
  - `dataChannelOpen_OversizedNameIsRejected`: Calls `onSctpSessionDataChannelOpen` with `nameLen = 355`, verifies `onDataChannel` callback is NOT invoked, no entry exists in the hash table and no memory leak.
  - `dataChannelOpen_NormalNameIsPreserved`: Calls with a normal 15-byte name, verifies the name and label are preserved exactly and callback fires.
- Both tests pass with zero memory leaks under the SDK's instrumented allocator.

### DIFF
--- a/src/source/PeerConnection/PeerConnection.c
+++ b/src/source/PeerConnection/PeerConnection.c
@@ -641,27 +641,21 @@ VOID onSctpSessionDataChannelOpen(UINT64 customData, UINT32 channelId, PBYTE pNa
     STATUS retStatus = STATUS_SUCCESS;
     PKvsPeerConnection pKvsPeerConnection = (PKvsPeerConnection) customData;
     PKvsDataChannel pKvsDataChannel = NULL;
-    UINT32 copyLen = 0;
 
     CHK(pKvsPeerConnection != NULL && pKvsPeerConnection->onDataChannel != NULL, STATUS_NULL_ARG);
+    CHK(nameLen <= MAX_DATA_CHANNEL_NAME_LEN, STATUS_INVALID_ARG);
 
     pKvsDataChannel = (PKvsDataChannel) MEMCALLOC(1, SIZEOF(KvsDataChannel));
     CHK(pKvsDataChannel != NULL, STATUS_NOT_ENOUGH_MEMORY);
 
-    // Cap copy length to destination buffer size to prevent heap overflow
-    copyLen = MIN(nameLen, MAX_DATA_CHANNEL_NAME_LEN);
-    STRNCPY(pKvsDataChannel->dataChannel.name, (PCHAR) pName, copyLen);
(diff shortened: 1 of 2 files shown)


sirknightj (approved):

Ahhorse1 (approved):

src/source/PeerConnection/PeerConnection.c
```diff
@@ -641,25 +641,34 @@ VOID onSctpSessionDataChannelOpen(UINT64 customData, UINT32 channelId, PBYTE pNa
...
+    UINT32 copyLen = 0;
 
     CHK(pKvsPeerConnection != NULL && pKvsPeerConnection->onDataChannel != NULL, STATUS_NULL_ARG);
 
     pKvsDataChannel = (PKvsDataChannel) MEMCALLOC(1, SIZEOF(KvsDataChannel));
     CHK(pKvsDataChannel != NULL, STATUS_NOT_ENOUGH_MEMORY);
 
-    STRNCPY(pKvsDataChannel->dataChannel.name, (PCHAR) pName, nameLen);
+    // Cap copy length to destination buffer size to prevent heap overflow
+    copyLen = MIN(nameLen, MAX_DATA_CHANNEL_NAME_LEN);
```
sirknightj: Minor: You can use SNPRINTF %s and check for truncation based on the return value (length).

The upside to that approach is that we can have observability/debug when the truncation occurs

vabhasin: Actually, since we know theres no other possibility that we would see a long name other than that it is from a malformed input, I have changed it to fully deny.

sirknightj (approved):

Ahhorse1 (approved):

PR #2363, opened 2026-07-08
PeerConnection/Ice: portable format specifiers + explicit fallthrough for -Wformat/-Wextra

vikramdattu:
These surface building the SDK on ESP-IDF (Xtensa/RISC-V GCC), where UINT32 is `unsigned long` rather than `unsigned int`, so `SNPRINTF`/`SSCANF` calls trip `-Wformat`, and `-Wextra`'s implicit-fallthrough level does not recognize the existing comment.

- **SessionDescription.c** — `SNPRINTF`/`SSCANF` pass `UINT32`/`PUINT32` to `%u`/`%d`/`%x` (and `UINT32` to the `%*.*` field width). Use the fixed-width `PRIu32`/`SCNx32` macros (the file already uses `PRId64`), and cast the field-width/precision args to `int`. The `SSCANF("%x", PUINT32)` case is also a genuine type mismatch, fixed with `SCNx32`.
- **IceAgent.c** — `iceCandidateSerialize()` passes `UINT32` foundation/priority to `%u`; use `PRIu32`. The `ICE_CANDIDATE_PAIR_STATE_WAITING` fallthrough used a `// NOTE: Explicit fall-through` comment; the `NOTE: Explicit` prefix is not matched by `-Wimplicit-fallthrough` comment recognition (enabled via `-Wextra`), so it warns while the SDK's bare `// fallthrough` comments do not. Drop the prefix to the recognized bare form.

No behavior change on platforms where `uint32_t == unsigned int`. Built clean on GCC (Xtensa, `-Wextra`) and clang.

### DIFF
--- a/src/source/Ice/IceAgent.c
+++ b/src/source/Ice/IceAgent.c
@@ -1735,7 +1735,7 @@ STATUS iceAgentCheckCandidatePairConnection(PIceAgent pIceAgent)
             switch (pIceCandidatePair->state) {
                 case ICE_CANDIDATE_PAIR_STATE_WAITING:
                     pIceCandidatePair->state = ICE_CANDIDATE_PAIR_STATE_IN_PROGRESS;
-                    // NOTE: Explicit fall-through
+                    // fallthrough
                 case ICE_CANDIDATE_PAIR_STATE_IN_PROGRESS:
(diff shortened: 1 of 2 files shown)


sirknightj (approved):
Thank you for the contribution!

sirknightj:
CI is passing except the `clang-format`. Could you take a look? Thanks @vikramdattu!

YPandas (approved):

PR #2356, opened 2026-06-29
Configurable SDP Track Count

sirknightj:
*Issue #, if available:*
- N/A

*What was changed?*
- Made `MAX_SDP_SESSION_MEDIA_COUNT` a CMake-configurable parameter (default: 5/unchanged, range: 1-6) instead of a hardcoded value.
- Replaced `CHK` with `CHK_ERR` at all media count limit checks for observability; to log the max and current count on failure.
- Added `docs/SDP_MEDIA_COUNT.md` documenting what consumes slots, stack size implications, and troubleshooting.

*Why was it changed?*
- Adding support for 4 video track + 1 audio track + data channel use case, which consumes 6 slots.
- Keeping it backwards compatible for existing users in terms of memory consumption.

*How was it changed?*
- `Sdp.h`: Replaced `#ifdef ENABLE_DATA_CHANNEL` (5 vs 6) with a single `#ifndef MAX_SDP_SESSION_MEDIA_COUNT` guard defaulting to 5.
- `CMakeLists.txt`: Added `MAX_SDP_SESSION_MEDIA_COUNT` cache variable with range validation (1-6) and conditional `add_definitions`.
- `README.md`: Added the new `-DMAX_SDP_SESSION_MEDIA_COUNT=N` flag entry linking to the doc.
- `Deserialize.c`: Replaced `CHK` with `CHK_ERR` to log max and current count on failure.
- Added `docs/SDP_MEDIA_COUNT.md` with configuration, slot breakdown, stack sizing, and troubleshooting.

*What testing was done for the changes?*
- Tested with `MAX_SDP_SESSION_MEDIA_COUNT=6` on macOS: all 53 SDP tests pass.
- Tested with `MAX_SDP_SESSION_MEDIA_COUNT=6` end-to-end against the JS test page via an internal demo.
- Tested with `MAX_SDP_SESSION_MEDIA_COUNT=1` on macOS: the SDK errors gracefully with `STATUS_SESSION_DESCRIPTION_MAX_MEDIA_COUNT` (0x5500000A) and does not crash.
- Verified CMake rejects out-of-range values (0 and 7).
- Unit tests cover deserialization and offer generation boundary conditions at the limit.
- Data channel test validates `MAX_SDP_SESSION_MEDIA_COUNT - 1` transceivers + data channel succeeds, and one more fails.

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.

PR #2367, opened 2026-07-27
New Feature: revise RTCP receiver report parsing to handle multiple report blocks

YPandas:
*Issue #, if available:*
- N/A

*What was changed?*
- `src/source/PeerConnection/Rtcp.c` — `onRtcpReceiverReport()` now parses **all** report blocks in an RTCP Receiver Report instead of discarding any RR that doesn't contain exactly one block. Each block updates the remote-inbound stats (`reportsReceived`, `fractionLost`, `roundTripTime`, `totalRoundTripTime`, `roundTripTimeMeasurements`) of the transceiver matching that block's SSRC.
- `onRtcpReceiverReport()` now validates the payload length against the header's `receptionReportCount` and logs a warning for malformed reports instead of silently ignoring them.
- Added inbound RTCP debug logging: one line per compound RTCP packet (`onRtcpPacket`) and one line per RR block (SSRC, fractionLost, cumulativeLost, extended highest seq, jitter, LSR, DLSR).
- `samples/webrtcstorage/kvsWebRTCStorageAudioVideoMasterSample.c` — added a periodic (10s) debug thread that dumps remote-inbound RTP stats for the video and audio transceivers via `rtcPeerConnectionGetMetrics(RTC_STATS_TYPE_REMOTE_INBOUND_RTP)` to verify the stats are populated during ingestion.

*Why was it changed?*
- In typical P2P sessions, a remote peer sends a Receiver Report carrying a single report block for the one stream it receives, which is the case the existing RR handling was built around.
- The media storage backend reports differently: it aggregates reception feedback for **all** ingested streams (audio + video) into a single compound Receiver Report carrying multiple report blocks, as permitted by RFC 3550 §6.4.2.
- The SDK's RR handling did not yet support this multi-block reporting style, so Receiver Reports from the storage backend were not consumed into per-transceiver stats — leaving `getRtpRemoteInboundStats` empty (`fractionLost`, RTCP RTT unavailable) during ingestion sessions.
- This change extends RR parsing to the general multi-block form so remote-inbound stats are populated regardless of how the reporting peer packs its report blocks.

*How was it changed?*
- Read `receptionReportCount` from the RTCP header and validate `payloadLength >= SIZEOF(UINT32) + count * RTCP_PACKET_RECEIVER_REPORT_BLOCK_LEN` (warn and skip on malformed input).
- Iterate over the report blocks at their per-block offsets. For each block:
  - Look up the transceiver by the block's SSRC (`findTransceiverBySsrc`); skip unknown SSRCs with a warning instead of aborting the whole report.
  - Parse `fractionLost`, `cumulativeLost`, extended highest sequence number, interarrival jitter, LSR, and DLSR from the block.
  - Compute round-trip propagation delay per RFC 3550 §6.4.1 (`A - LSR - DLSR`) when LSR is non-zero, resetting the value per block.
  - Update that transceiver's `remoteInboundStats` under `statsLock`.
- Verified end-to-end on a storage (ingestion) master: RR blocks for both audio and video SSRCs are now parsed, and `rtcPeerConnectionGetMetrics(RTC_STATS_TYPE_REMOTE_INBOUND_RTP)` returns non-zero `reportsReceived` with populated `fractionLost`/`roundTripTime`.

- Tested locally to check if receiver report show up in the log
image

- Tested with SDK canary setup
  
BEFORE:
image

  AFTER:
image


src/source/PeerConnection/Rtcp.c
```diff
@@ -84,65 +84,68 @@ static STATUS onRtcpReceiverReport(PRtcpPacket pRtcpPacket, PKvsPeerConnection p
...
-        // TODO: handle multiple receiver report blocks
+    // A receiver report contains the sender SSRC followed by receptionReportCount report blocks. The media
+    // storage backend reports on all streams (audio + video) in a single RR, so all blocks must be processed.
+    reportBlockCount = pRtcpPacket->header.receptionReportCount;
+    if (pRtcpPacket->payloadLength < SIZEOF(UINT32) + (UINT32) reportBlockCount * RTCP_PACKET_RECEIVER_REPORT_BLOCK_LEN) {
+        DLOGW("Malformed receiver report: payloadLength %u too small for %u report block(s)", pRtcpPacket->payloadLength, reportBlockCount);
         return STATUS_SUCCESS;
     }
 
-#ifdef LOG_STREAMING
```
unicornss, line 105: any reason to disable ?

src/source/PeerConnection/Rtcp.c
```diff
@@ -84,65 +84,68 @@ static STATUS onRtcpReceiverReport(PRtcpPacket pRtcpPacket, PKvsPeerConnection p
...
-    UNUSED_PARAM(rttPropDelayMsec);
-    UNUSED_PARAM(rttPropDelay);
-    UNUSED_PARAM(delaySinceLastSR);
-    UNUSED_PARAM(lastSR);
+    UINT32 reportBlockOffset;
+    UINT8 reportBlockCount, i;
 
     CHK(pKvsPeerConnection != NULL && pRtcpPacket != NULL, STATUS_NULL_ARG);
     // https://tools.ietf.org/html/rfc3550#section-6.4.2
-    if (pRtcpPacket->payloadLength != RTCP_PACKET_RECEIVER_REPORT_MINLEN) {
```
unicornss, line 100: Should we  consider capping reportBlockCount to a reasonable max (e.g., 2–5) ?

src/source/PeerConnection/Rtcp.c
```diff
@@ -84,65 +84,68 @@ static STATUS onRtcpReceiverReport(PRtcpPacket pRtcpPacket, PKvsPeerConnection p
...
+        }
+
+        fractionLost = pRtcpPacket->payload[reportBlockOffset + 4] / 255.0;
+        cumulativeLost = ((UINT32) getUnalignedInt32BigEndian(pRtcpPacket->payload + reportBlockOffset + 4)) & 0x00ffffffu;
+        extHiSeqNumReceived = getUnalignedInt32BigEndian(pRtcpPacket->payload + reportBlockOffset + 8);
+        interarrivalJitter = getUnalignedInt32BigEndian(pRtcpPacket->payload + reportBlockOffset + 12);
+        lastSR = getUnalignedInt32BigEndian(pRtcpPacket->payload + reportBlockOffset + 16);
+        delaySinceLastSR = getUnalignedInt32BigEndian(pRtcpPacket->payload + reportBlockOffset + 20);
+
+        DLOGI("RTCP inbound: RECEIVER_REPORT block %u/%u - senderSSRC=%u ssrc=%u fractionLost=%.3f cumulativeLost=%u extHiSeq=%u jitter=%u "
```
unicornss: Should we reduce the level, as this could impact lossy network scenarios?

src/source/PeerConnection/Rtcp.c
```diff
@@ -84,65 +84,68 @@ static STATUS onRtcpReceiverReport(PRtcpPacket pRtcpPacket, PKvsPeerConnection p
...
+            //      SSRC_r by recording the time A when this reception report block is
+            //      received.  It calculates the total round-trip time A-LSR using the
+            //      last SR timestamp (LSR) field, and then subtracting this field to
+            //      leave the round-trip propagation delay as (A - LSR - DLSR).
+            rttPropDelay = MID_NTP(currentTimeNTP) - lastSR - delaySinceLastSR;
+            rttPropDelayMsec = KVS_CONVERT_TIMESCALE(rttPropDelay, DLSR_TIMESCALE, 1000);
+            DLOGS("RTCP_PACKET_TYPE_RECEIVER_REPORT rttPropDelay %u msec", rttPropDelayMsec);
+        }
+
+        MUTEX_LOCK(pTransceiver->statsLock);
```
unicornss: nice!

unicornss (commented):
Are there unit tests planned (follow-up or in this PR) for the multi-block RR parsing? Given this touches security-sensitive protocol parsing, tests covering RC=0, RC=2, and malformed payloads would help prevent regressions.

src/source/PeerConnection/Rtcp.c
```diff
@@ -662,6 +665,10 @@ STATUS onRtcpPacket(PKvsPeerConnection pKvsPeerConnection, PBYTE pBuff, UINT32 b
     while (currentOffset < buffLen) {
         CHK_STATUS(setRtcpPacketFromBytes(pBuff + currentOffset, buffLen - currentOffset, &rtcpPacket));
 
+        // Debug: log every RTCP packet received from the remote peer (200=SR, 201=RR, 202=SDES, 205=RTPFB/NACK/TWCC, 206=PSFB/PLI/REMB)
+        DLOGI("RTCP inbound: packetType=%u fmt/rc=%u payloadLength=%u (compound offset %u/%u)", rtcpPacket.header.packetType,
```
sirknightj: Logging level: this should not be DLOGI.

This fires for every compound RTCP packet, including high-frequency types like TWCC/NACK/REMB. At INFO level (compiled in and on by default) this will flood production logs. The comment even labels it "Debug" — please downgrade to DLOGD/DLOGV, or gate it behind LOG_STREAMING like the other streaming logs in this file. Same applies if this line is only needed for local verification and can be dropped.

src/source/PeerConnection/Rtcp.c
sirknightj: The previous version logged this at DLOGS (verbose, gated behind LOG_STREAMING). This promotes it to INFO and multiplies it per report block, so a multi-stream RR emits several INFO lines per packet. Suggest keeping it at DLOGS/DLOGV to match the surrounding convention.

src/source/PeerConnection/Rtcp.c
```diff
@@ -84,65 +84,68 @@ static STATUS onRtcpReceiverReport(PRtcpPacket pRtcpPacket, PKvsPeerConnection p
...
+            rttPropDelayMsec = KVS_CONVERT_TIMESCALE(rttPropDelay, DLSR_TIMESCALE, 1000);
+            DLOGS("RTCP_PACKET_TYPE_RECEIVER_REPORT rttPropDelay %u msec", rttPropDelayMsec);
+        }
+
+        MUTEX_LOCK(pTransceiver->statsLock);
+        pTransceiver->remoteInboundStats.reportsReceived++;
+        if (fractionLost > -1.0) {
+            pTransceiver->remoteInboundStats.fractionLost = fractionLost;
+        }
+        pTransceiver->remoteInboundStats.roundTripTimeMeasurements++;
```
sirknightj: roundTripTimeMeasurements is incremented unconditionally, but Stats.h:422 documents it as the count of RR blocks "that contain a valid round trip time." When lastSR == 0 there's no valid RTT (rttPropDelayMsec stays 0), yet we still bump the measurement count and add 0 to totalRoundTripTime. This was pre-existing but is now more visible per-block. 

Consider moving the three RTT stat updates inside the if (lastSR != 0) block:
```c
if (lastSR != 0) {
    ...
    rttPropDelayMsec = KVS_CONVERT_TIMESCALE(...);
    pTransceiver->remoteInboundStats.roundTripTimeMeasurements++;
    pTransceiver->remoteInboundStats.totalRoundTripTime += rttPropDelayMsec;
    pTransceiver->remoteInboundStats.roundTripTime = rttPropDelayMsec;
}
```
(reportsReceived and fractionLost should stay unconditional.)

samples/webrtcstorage/kvsWebRTCStorageAudioVideoMasterSample.c
```diff
@@ -7,10 +7,62 @@ extern PSampleConfiguration gSampleConfiguration;
 
 STATUS addSendOnlyVideoSendrecvAudioTransceivers(PSampleConfiguration, PSampleStreamingSession);
 
+// Debug: periodically dump remote-inbound RTP stats (populated from RTCP receiver reports sent by the media service)
```
sirknightj: It also sent by normal p2p peers too?

src/source/PeerConnection/Rtcp.c
```diff
@@ -84,65 +84,68 @@ static STATUS onRtcpReceiverReport(PRtcpPacket pRtcpPacket, PKvsPeerConnection p
...
 
     CHK(pKvsPeerConnection != NULL && pRtcpPacket != NULL, STATUS_NULL_ARG);
     // https://tools.ietf.org/html/rfc3550#section-6.4.2
-    if (pRtcpPacket->payloadLength != RTCP_PACKET_RECEIVER_REPORT_MINLEN) {
-        // TODO: handle multiple receiver report blocks
+    // A receiver report contains the sender SSRC followed by receptionReportCount report blocks. The media
+    // storage backend reports on all streams (audio + video) in a single RR, so all blocks must be processed.
+    reportBlockCount = pRtcpPacket->header.receptionReportCount;
+    if (pRtcpPacket->payloadLength < SIZEOF(UINT32) + (UINT32) reportBlockCount * RTCP_PACKET_RECEIVER_REPORT_BLOCK_LEN) {
+        DLOGW("Malformed receiver report: payloadLength %u too small for %u report block(s)", pRtcpPacket->payloadLength, reportBlockCount);
```
sirknightj, line 102: Could we add a unit test in tst/RtcpFunctionalityTest.cpp feeding some RRs with 2 blocks (audio + video SSRCs) and asserting both transceivers' reportsReceived/fractionLost get populated? That directly covers the new multi-block path with the tools (Asan/Valgrind) to verify memory. Include malformed packets.

Suggest looking into: dump the raw packet and use that in your unit test

samples/webrtcstorage/kvsWebRTCStorageAudioVideoMasterSample.c
```diff
@@ -7,10 +7,62 @@ extern PSampleConfiguration gSampleConfiguration;
...
+}
+
+static PVOID remoteInboundStatsThread(PVOID args)
+{
+    PSampleConfiguration pSampleConfiguration = (PSampleConfiguration) args;
+    PSampleStreamingSession pSampleStreamingSession;
+    UINT32 i, sleepCounter = 0;
+
+    while (!ATOMIC_LOAD_BOOL(&pSampleConfiguration->appTerminateFlag)) {
+        THREAD_SLEEP(HUNDREDS_OF_NANOS_IN_A_SECOND);
```
sirknightj: What's up with the sleep logic? Can you add a comment

samples/webrtcstorage/kvsWebRTCStorageAudioVideoMasterSample.c
```diff
@@ -55,6 +107,9 @@ INT32 main(INT32 argc, CHAR* argv[])
 
     DLOGI("[%s] Channel %s set up done ", SAMPLE_NAME, pChannelName);
 
+    // Debug: start periodic remote-inbound stats dump
+    THREAD_CREATE(&statsTid, remoteInboundStatsThread, (PVOID) pSampleConfiguration);
```
sirknightj: Extra threads use more memory. Can you gate the thread creation around the set log level, given the thread just prints logs?

sirknightj: Also: suggest moving to Common.C so it can be helpful/used for all samples

src/source/PeerConnection/Rtcp.c
YPandas: I will keep this at DLOGV

samples/webrtcstorage/kvsWebRTCStorageAudioVideoMasterSample.c
YPandas: Modified to more accurate comment

samples/webrtcstorage/kvsWebRTCStorageAudioVideoMasterSample.c
YPandas: This is because I want to log the stats every 10 seconds, but also make sure we can notice shutdown within one second. Comment added

src/source/PeerConnection/Rtcp.c
YPandas: Sure, maybe we can just cap it to 2

src/source/PeerConnection/Rtcp.c
YPandas: The #ifdef LOG_STREAMING guards existed only to suppress unused-variable warnings: those fields were consumed solely by DLOGS, which compiles out unless LOG_STREAMING is defined. In this change all of those fields are used unconditionally (per-block log + stats/RTT update), so the guards are no longer needed.

src/source/Rtcp/RtcpPacket.h
```diff
@@ -29,6 +29,9 @@ extern "C" {
 #define RTCP_PACKET_SENDER_REPORT_MINLEN      24
 #define RTCP_PACKET_RECEIVER_REPORT_BLOCK_LEN 24
 #define RTCP_PACKET_RECEIVER_REPORT_MINLEN    4 + RTCP_PACKET_RECEIVER_REPORT_BLOCK_LEN
+// Maximum receiver report blocks processed per RR. The SDK negotiates at most one audio and one
+// video stream per peer connection, so report blocks beyond 2 reference no known transceiver.
+#define RTCP_PACKET_RECEIVER_REPORT_MAX_BLOCKS 2
```
sirknightj: The cap of 2 and its justifying comment are stale after #2356 ("Configurable SDP Track Count"), which landed one commit before this branch. MAX_SDP_SESSION_MEDIA_COUNT now defaults to 5 (range 1–6), and transceivers live in an unbounded linked list (findTransceiverBySsrc just walks pTransceivers), so a peer connection can have more than 2 transceivers. 

Suggest keying the cap off MAX_SDP_SESSION_MEDIA_COUNT rather than a literal 2, or dropping the cap entirely and relying on the natural bounds: receptionReportCount is a 5-bit field (max 31).

samples/webrtcstorage/kvsWebRTCStorageVideoOnlyMasterSample.c
```diff
@@ -53,6 +54,12 @@ INT32 main(INT32 argc, CHAR* argv[])
 
     DLOGI("[%s] Channel %s set up done ", SAMPLE_NAME, pChannelName);
 
+    // Periodically log remote-inbound RTP stats. The thread only produces VERBOSE-level logs, so skip
+    // creating it (and its memory overhead) unless the configured log level is VERBOSE.
+    if (logLevel == LOG_LEVEL_VERBOSE) {
+        THREAD_CREATE(&statsTid, getPeriodicRemoteInboundStats, (PVOID) pSampleConfiguration);
```
sirknightj: It should be for all samples including the GSt storage samples and p2p samples too?
Let's add it if is useful

src/source/PeerConnection/Rtcp.c
```diff
@@ -84,65 +84,74 @@ static STATUS onRtcpReceiverReport(PRtcpPacket pRtcpPacket, PKvsPeerConnection p
...
-    if (pRtcpPacket->payloadLength != RTCP_PACKET_RECEIVER_REPORT_MINLEN) {
-        // TODO: handle multiple receiver report blocks
+    // A receiver report contains the sender SSRC followed by receptionReportCount report blocks. The media
+    // storage backend reports on all streams (audio + video) in a single RR, so all blocks must be processed.
+    reportBlockCount = pRtcpPacket->header.receptionReportCount;
+    if (reportBlockCount > RTCP_PACKET_RECEIVER_REPORT_MAX_BLOCKS) {
+        DLOGW("Receiver report contains %u report blocks, processing only the first %u", reportBlockCount, RTCP_PACKET_RECEIVER_REPORT_MAX_BLOCKS);
+        reportBlockCount = RTCP_PACKET_RECEIVER_REPORT_MAX_BLOCKS;
+    }
+    if (pRtcpPacket->payloadLength < SIZEOF(UINT32) + (UINT32) reportBlockCount * RTCP_PACKET_RECEIVER_REPORT_BLOCK_LEN) {
```
sirknightj, line 101: Looks like the previous constant: RTCP_PACKET_RECEIVER_REPORT_MINLEN is now replaced with a dynamic one. Suggest to add a comment for the constant definition that it's no longer used

sirknightj (approved):

samples/common/Common.c
```diff
@@ -1132,6 +1133,86 @@ STATUS logSignalingClientStats(PSignalingClientMetrics pSignalingClientMetrics)
...
+static VOID logRemoteInboundStatsForTransceiver(PRtcPeerConnection pPeerConnection, PRtcRtpTransceiver pTransceiver, PCHAR streamName)
+{
+    RtcStats rtcStats;
+
+    if (pTransceiver == NULL) {
+        return;
+    }
+
+    rtcStats.requestedTypeOfStats = RTC_STATS_TYPE_REMOTE_INBOUND_RTP;
+    if (STATUS_SUCCEEDED(rtcPeerConnectionGetMetrics(pPeerConnection, pTransceiver, &rtcStats))) {
```
unicornss: do we need error logging on failure ?

samples/common/Common.c
```diff
@@ -1132,6 +1133,86 @@ STATUS logSignalingClientStats(PSignalingClientMetrics pSignalingClientMetrics)
     return retStatus;
 }
 
+// Log remote-inbound RTP stats for one transceiver. These stats are populated from RTCP receiver
+// reports sent by the remote peer (the media service in a storage session, or a viewer in a P2P session).
+static VOID logRemoteInboundStatsForTransceiver(PRtcPeerConnection pPeerConnection, PRtcRtpTransceiver pTransceiver, PCHAR streamName)
+{
+    RtcStats rtcStats;
+
+    if (pTransceiver == NULL) {
```
unicornss: do we have this check already pPeerConnection == NULL ||   ?

samples/common/Common.c
```diff
@@ -1132,6 +1133,86 @@ STATUS logSignalingClientStats(PSignalingClientMetrics pSignalingClientMetrics)
     return retStatus;
 }
 
+// Log remote-inbound RTP stats for one transceiver. These stats are populated from RTCP receiver
+// reports sent by the remote peer (the media service in a storage session, or a viewer in a P2P session).
+static VOID logRemoteInboundStatsForTransceiver(PRtcPeerConnection pPeerConnection, PRtcRtpTransceiver pTransceiver, PCHAR streamName)
+{
+    RtcStats rtcStats;
```
unicornss, line 1199: zero initialize?

samples/common/Common.c
```diff
@@ -1132,6 +1133,86 @@ STATUS logSignalingClientStats(PSignalingClientMetrics pSignalingClientMetrics)
...
+              " rttMeasurements=%" PRIu64,
+              streamName, rtcStats.rtcStatsObject.remoteInboundRtpStreamStats.reportsReceived,
+              rtcStats.rtcStatsObject.remoteInboundRtpStreamStats.fractionLost, rtcStats.rtcStatsObject.remoteInboundRtpStreamStats.roundTripTime,
+              rtcStats.rtcStatsObject.remoteInboundRtpStreamStats.totalRoundTripTime,
+              rtcStats.rtcStatsObject.remoteInboundRtpStreamStats.roundTripTimeMeasurements);
+    }
+}
+
+// Thread routine that logs remote-inbound RTP stats for all streaming sessions every 10 seconds.
+// The stats are logged at VERBOSE level; callers should skip creating this thread when the configured
```
unicornss, line 1221: can we add this check?
```
   if (GET_LOGGER_LOG_LEVEL() < LOG_LEVEL_VERBOSE) {
        return NULL;
    }
```

unicornss: #if LOG_LEVEL_ENABLED(LOG_LEVEL_VERBOSE)

samples/common/Common.c
```diff
@@ -1132,6 +1133,86 @@ STATUS logSignalingClientStats(PSignalingClientMetrics pSignalingClientMetrics)
...
+        if (++sleepCounter < 10) {
+            continue;
+        }
+        sleepCounter = 0;
+
+        // Use trylock to avoid contending with session setup/teardown; on failure we simply skip
+        // this cycle. Stats are cumulative SDK counters, so nothing is lost by skipping a read.
+        if (!MUTEX_TRYLOCK(pSampleConfiguration->sampleConfigurationObjLock)) {
+            continue;
+        }
```
unicornss, line 1255: should we add a check ?  
```
MIN(pSampleConfiguration->streamingSessionCount, MAX_STREAMING_SESSION_COUNT);
```

samples/common/Common.c
```diff
@@ -1132,6 +1133,86 @@ STATUS logSignalingClientStats(PSignalingClientMetrics pSignalingClientMetrics)
...
+        for (i = 0; i < pSampleConfiguration->streamingSessionCount; ++i) {
```
unicornss: can we add a guard?
```
pSampleStreamingSession = pSampleConfiguration->sampleStreamingSessionList[i];

    //  Avoid crashes if the session is torn down during array access
    if (pSampleStreamingSession == NULL) {
        continue;
    }
```

tst/RtcpFunctionalityTest.cpp
```diff
@@ -229,6 +229,212 @@ TEST_F(RtcpFunctionalityTest, onRtcpPacketCompoundSenderReport)
...
+
+    initTransceiver(0x11111111);
+
+    EXPECT_EQ(STATUS_SUCCESS, onRtcpPacket(pKvsPeerConnection, rawpacket, rawpacketSize));
+
+    RtcRemoteInboundRtpStreamStats stats{};
+    EXPECT_EQ(STATUS_SUCCESS, getRtpRemoteInboundStats(pRtcPeerConnection, pRtcRtpTransceiver, &stats));
+    EXPECT_EQ(0, stats.reportsReceived);
+
+    freePeerConnection(&pRtcPeerConnection);
```
unicornss, line 435: Good set of scenarios!

In the follow-up, we might want to note that we’ve confirmed the following:

- fractionLost boundary values are correctly calculated using N/255.0.
- RTT values are accurate.
- The count of cumulative reportsReceived across multiple RRs is properly asserted.

unicornss (approved):

PR #2370, opened 2026-08-05
Add support for RTX for H.265 and VP8 codecs

sirknightj:
*Issue #, if available:*
- Internal. High level summary: reported RTX advertised but retransmissions sent as plain resends for H.265; Chrome `retransmittedPacketsReceived` inbound-rtp stat was empty for H.265).

*What was changed?*
- Added RTX (RFC 4588) retransmission support for **H.265 and VP8**. H.264 RTX is already implemented.
  - Previously: on NACK, the SDK used plain resend (original SSRC / payload type / sequence number) instead of sending RFC 4588 RTX packets for H.265 and VP8.
- Added additional observability: a negotiation-time log (WARN) when RTX does not resolve for a codec, plus a verbose per-packet log when the plain-resend fallback is taken.
- Added a new status code: STATUS_SESSION_DESCRIPTION_CODEC_NOT_MAPPED_TO_RTX
- Added unit + end-to-end regression tests, and fixed an existing H.265 test that was asserting incorrect behavior.
- Added `docs/NACK_RTX.md` documenting the RFC 4588 flow (SDP negotiation, NACK parsing, RTX construction) and the key log lines to look for (troubleshooting).

*Why was it changed?*
- Adding support for H.265 and VP8 RTX
- Why it used plain-resend previously for those two codecs:`setTransceiverPayloadTypes()` looked up the RTX payload-type table using the **media-codec** enum value (`RTC_CODEC_*`), but the table is keyed by the **RTX-codec** enum (`RTC_RTX_CODEC_*`). These two enums have different numbering:

  | Codec | `RTC_CODEC_*` | `RTC_RTX_CODEC_*` |
  |-------|---------------|-------------------|
  | H264  | 1             | 1                 |
  | VP8   | 3             | 2                 |
  | H265  | 7             | 3                 |

  The lookup only matched for H.264 (both enums = 1). For H.265 and VP8 it missed, so `rtxPayloadType` stayed equal to `payloadType`, and `resendPacketOnNack()` took the `payloadType == rtxPayloadType` branch - a plain resend. The SDP-emission path used the correct key, so RTX was still **advertised**, producing "advertised but not used" behavior for H.265/VP8.
- **This is not a multi-track issue.** It is purely codec-dependent - an oversight in codec→RTX-table mapping. H.264 (single- or multi-track) was compliant; H.265/VP8 failed identically in single- and multi-track.
- Impact: Retransmissions were invisible in the Chrome's inbound-rtp receiver stats (`rtxSsrc` / `retransmittedPacketsReceived`), and it poses an interop risk with stricter receivers.

*How was it changed?*
- `SessionDescription.c`: added `getRtxCodecKeyForCodec()` mapping `RTC_CODEC_* -> RTC_RTX_CODEC_*` (H264->H264, VP8->VP8, H265->H265). It returns `STATUS` (new code `STATUS_SESSION_DESCRIPTION_CODEC_NOT_MAPPED_TO_RTX`, defined in `Include.h`), following the codebase's ENTERS/CHK/LEAVES convention, and `CHK_ERR`s with the offending codec value if no mapping exists. This mirrors the mapping the SDP-emission path (`populateSingleMediaSection()`) already performs.
- `setTransceiverPayloadTypes()`: the RTX-table lookup now maps the codec via this helper. RTX is video-only, so the lookup is guarded by `track.kind == MEDIA_STREAM_TRACK_KIND_VIDEO` (for readability) - audio transceivers never call the helper (avoiding a spurious error log), and the helper's `CHK_ERR` only fires for the genuinely unexpected case of a video codec with no RTX mapping.
- `SessionDescription.c`: after the lookup, log once per transceiver whether RTX resolved (`DLOGD`) or fell back to plain resend (`DLOGW`).
- `Retransmitter.c`: in the plain-resend branch of `resendPacketOnNack()`, added a `DLOGV` marking the fallback (ssrc / seq / pt).
- The change adds support for H.265 and VP8 in one place; H.264 behavior is unchanged.

- `tst/SdpApiTest.cpp`:
  - Fixed `setTransceiverPayloadTypes_HasRtxType_H265` - it previously seeded the RTX table with `RTC_CODEC_H265` (the same wrong key as the bug), which is why the bug shipped despite a passing test. It now seeds `RTC_RTX_CODEC_H265`, matching production.
  - Added `setTransceiverPayloadTypes_HasRtxType_VP8` to cover the new code.
  - Added `setTransceiverPayloadTypes_H265RtxResolvedEndToEnd` - drives the full negotiation (`createPeerConnection` → `setRemoteDescription` → `createAnswer`) so `setPayloadTypesFromOffer()` and `setTransceiverPayloadTypes()` must agree on the key, then asserts the sender's runtime `rtxPayloadType != payloadType` (52 vs 51).
- Verified the new/updated tests **fail on the pre-fix code** (H.265/VP8 resolve `rtxPayloadType == payloadType == media PT`) and **pass with the fix** - confirming they are real regression guards. All targeted `SdpApiTest` RTX/payload-type tests pass (8/8).
- Manually verified end-to-end (C SDK master → Chrome viewer, H.265 multi-track): pre-fix, Wireshark showed zero RTX (PT 52) packets and duplicate-sequence plain resends on the original PT 51, and Chrome never created `retransmittedPacketsReceived`; post-fix, RTX packets flow on the RTX SSRCs and Chrome's `retransmittedPacketsReceived` populates on all streams. H.264 multi-track remained working throughout.

| Before | After | 
|-----|-----|
|  image | image  |


src/source/PeerConnection/SessionDescription.c
```diff
@@ -348,10 +379,24 @@ STATUS setTransceiverPayloadTypes(PHashTable codecTable, PHashTable rtxTable, PD
...
+                getRtxCodecKeyForCodec(pKvsRtpTransceiver->sender.track.codec, &rtxCodecKey) == STATUS_SUCCESS &&
+                hashTableGet(rtxTable, rtxCodecKey, &data) == STATUS_SUCCESS) {
                 pKvsRtpTransceiver->sender.rtxPayloadType = (UINT8) data;
             }
+
+            // Emit once per negotiation, whether RTX was resolved for this codec. When rtxPayloadType
+            // stays equal to payloadType, retransmissions on NACK fall back to a plain resend on the original SSRC/PT
+            // instead of being sent as RTX packets.
+            if (pKvsRtpTransceiver->sender.rtxPayloadType != pKvsRtpTransceiver->sender.payloadType) {
+                DLOGD("RTX negotiated for codec %u: payloadType %u, rtxPayloadType %u", pKvsRtpTransceiver->sender.track.codec,
```
YPandas: The doc(NACK_RTX.md) mentioned this log is INFO level. Should we modify this log level accordingly?

sirknightj: Thank you for catching it. The neighboring "success" SDP-parsing logs in this same file are low-level, so I'll adjust this log to match the doc (debug).

YPandas (approved):

vabhasin (approved):

PR #2371, opened 2026-08-06
Add multi-track H265 sample and troubleshooting H265 docs

sirknightj:
*Issue #, if available:*
- N/A

*What was changed?*
- Added an H.265 option to the multi-track sample (`kvsWebrtcClientMaster4Video1Audio`) and its frame-generation script (`scripts/generate_multi_track_frames.sh`).
- Updated the H.264/H.265 `gst-launch` commands in the README so generated keyframes are self-contained.
- Added `docs/TROUBLESHOOTING_H265_FRAMES.md` and linked it from the README.
- Added output-path logging to the generation script.

*Why was it changed?*
- The previous H.265 generation command omitted `h265parse config-interval=-1`, so only the first keyframe carried VPS/SPS/PPS. Later keyframes were bare slices, which decode locally in ffmpeg but freeze in a WebRTC viewer after the first frame (audio keeps playing) even with good network. The multi-track sample needed matching H.265 support so its per-track frames were valid too.

*How was it changed?*
- The script and README commands now run the encoder through `h264parse`/`h265parse config-interval=-1`, re-inserting parameter sets before every keyframe, plus `key-int-max=25` and B-frames disabled.
- The troubleshooting doc explains the freeze symptom, how to inspect NAL units / find keyframes (Python scanner + `ffprobe`), receiver-side debugging in `chrome://webrtc-internals`, and how to regenerate frames.
- The script now logs the absolute output directory for each track, since the sample resolves frame paths relative to its own working directory.

- Generated both H.264 and H.265 frames with the updated commands and verified via a raw NAL scan that every keyframe (every 25 frames) carries its parameter sets (VPS/SPS/PPS for H.265, SPS/PPS for H.264).
- Used the commands to generate frames locally into the master sample and confirmed continuous playback in Chrome for both codecs and checked the playback.
- Verified `ffprobe` keyframe detection against a raw byte-scan ground truth.


samples/p2p/kvsWebRTCClientMaster4Video1Audio.c
YPandas: Do we want to add the codec to this warning message? In case users keep regenerating h264(default) when they actually want to run h265.

scripts/generate_multi_track_frames.sh
```diff
@@ -1,24 +1,69 @@
...
+        exit 1
+        ;;
+esac
+
 PATTERNS=("ball" "smpte" "snow" "circular")
 NUM_BUFFERS=1500
 WIDTH=1280
 HEIGHT=720
 FPS=25
 BITRATE=512
```
YPandas: Should we move this line above the case block? ${BITRATE:-512} in the ENCODER string is expanded before this assignment. Also, the h265 branch's ENCODER doesn't set a bitrate ==> is that intentional?

samples/p2p/kvsWebRTCClientMaster4Video1Audio.c
sirknightj: Good suggestion, added

scripts/generate_multi_track_frames.sh
sirknightj: Added the bitrate to both cases and moved it up to a common section

YPandas (approved):

docs/TROUBLESHOOTING_H265_FRAMES.md
vabhasin, line 1: Seems like some issue in doc formatting (mid sentence line breaks), is it intentional?

unicornss (commented):
Should we move this into a separate category — for example, samples/p2p/multi-track/ — so we can add master_4video_1audio.c and master_2video_1audio.c?

docs/TROUBLESHOOTING_H265_FRAMES.md
sirknightj: Due to fixed-width setting. It doesn't appear in the final rendered markdown, but I can reformat it

vabhasin (approved):

Issue #2373, opened 2026-08-08
[Bug]: Media attribute overflow silently corrupts the parsed SDP - the MAX_SDP_ATTRIBUTES_COUNT guard is unreachable


xxxxxxjun:
### Please confirm you have already done the following

- [x] I have searched the repository for related/existing bug reports
- [x] I have all the details the issue requires

### Please answer the following prompt

- [ ] This issue is replicable using the unmodified sample application

### Describe the bug

`SdpMediaDescription` stores up to `MAX_SDP_ATTRIBUTES_COUNT` (256) attributes but counts them in a `UINT8` ([Sdp.h#L188-L190](https://github.com/awslabs/amazon-kinesis-video-streams-webrtc-sdk-c/blob/b0959ff4c1c704c6dbff20ececa2dc3662b42813/src/source/Sdp/Sdp.h#L188-L190)), so the guard in `parseMediaAttributes()` can never fail ([Deserialize.c#L51-L66](https://github.com/awslabs/amazon-kinesis-video-streams-webrtc-sdk-c/blob/b0959ff4c1c704c6dbff20ececa2dc3662b42813/src/source/Sdp/Deserialize.c#L51-L66)). The count is read out of that `UINT8`, so `currentMediaAttributesCount < MAX_SDP_ATTRIBUTES_COUNT` is always true, and the `mediaAttributesCount++` at the end of the function wraps 255 back to 0.

After the 256th attribute of a media section is stored the counter wraps to 0, so the 257th attribute lands in `sdpAttributes[0]` and later ones keep overwriting from there. Browser offers put `a=ice-ufrag`, `a=ice-pwd`, `a=fingerprint`, `a=setup` and `a=mid` at the start of a media section, so the ICE and DTLS parameters are the first thing destroyed. `deserializeSessionDescription()` still returns `STATUS_SUCCESS` and logs nothing, so the caller has no way to notice. It is a struct field width issue in the parser, so it is platform independent.

The v1.7.1 fix for #1389 looks incomplete: it raised the limit from 128 to 256 and widened `sessionAttributesCount` to `UINT16`, but left `mediaAttributesCount` at `UINT8`. At a limit of 128 a `UINT8` counter could still represent the overflow, which is why #1394 and #1396 saw a clear `STATUS_SDP_ATTRIBUTE_MAX_EXCEEDED` instead of silent corruption.

[...]

Issue #1389, opened 2022-02-07
[QUESTION] How to Configure the SDK to adapt to Chrome 98?


suggestedfixes:
We are having issues with the SDK with the recent deprecation of SDES on chrome 98. Do we wonder what would be the steps to disable SDES? We understand this could be a chrome-specific config, js config, or c config. Our version is based off of v1.3.0.
https://developer.chrome.com/blog/deps-rems-98/

Edit:
As determined by hassantech. The root cause was Chrome 98 had exceeded the max allowed lines of SDP configured by the SDK.

disa6302:
AFAIK, we do not use SDES for security implementation in the SDK. We use `DTLS-SRTP`. Can you highlight what the issues you are facing are and include SDK logs?

FYI: You are using a very old version of the SDK. Please upgrade to `v1.7.0`.

suggestedfixes:
Here is an example of other people reporting on the js sdk https://github.com/awslabs/amazon-kinesis-video-streams-webrtc-sdk-js/issues/160

disa6302:
It is possible the offer from the borwser side has changed. In previous version, `sdes` parameter is a part of the `extmap` sdp attribute. While we try to reproduce this, can you also attach SDK logs with `export DEBUG_LOG_SDP=TRUE`? It would be super helpful for us. Thanks!

suggestedfixes:
Good find, my log confirms this as well!
![image](https://user-images.githubusercontent.com/61390265/152883918-e1492600-d890-4750-88d1-116b440e9864.png)

suggestedfixes:
```
#define STATUS_WEBRTC_BASE 0x55000000
#define STATUS_SDP_BASE                         STATUS_WEBRTC_BASE + 0x01000000
#define STATUS_SDP_ATTRIBUTE_MAX_EXCEEDED       STATUS_SDP_BASE + 0x0000000F
```
Those 3 sum up to 0x56.....F

hassanctech:
It does not also these are static with size determined at compile time.  These SDKs are targeted to C89.

hassanctech:
@suggestedfixes can you confirm that this change resolved your issue?

hassanctech:
@suggestedfixes Also perhaps we should change the title of the question, as per your logs there is zero indication of a SDES related issue.

suggestedfixes:
[...]

PR #2374, opened 2026-08-08
Fix silent SDP media attribute counter overflow

xxxxxxjun:
*Issue #, if available:*
- #2373

*What was changed?*
- `SdpMediaDescription.mediaAttributesCount` is now `UINT16` instead of `UINT8`.
- The `currentAttribute` loop counter in `setPayloadTypesFromOffer()` is now `UINT16` too. Every other consumer of `mediaAttributesCount` already used `UINT32`.
- Added two regression tests to `tst/SdpApiTest.cpp`.

*Why was it changed?*
- `MAX_SDP_ATTRIBUTES_COUNT` is 256, but `mediaAttributesCount` could only represent 0..255, so `CHK(currentMediaAttributesCount < MAX_SDP_ATTRIBUTES_COUNT, ...)` in `parseMediaAttributes()` could never fail. Once 256 attributes had been stored the counter wrapped to 0 and the next attribute overwrote `sdpAttributes[0]`, where browser offers keep the ICE and DTLS parameters. `deserializeSessionDescription()` still returned `STATUS_SUCCESS`.
- The guard worked while the limit was 128, so this looks like something the v1.7.1 fix for #1389 missed when it raised the limit to 256.

*How was it changed?*
- Only the counter widths changed; the 256-attribute capacity is untouched, so `SdpMediaDescription` grows by at most 2 bytes after alignment.
- `Sdp.h` is internal, not under `src/include/`, so the public API is unchanged. Consumers linking a prebuilt library only need to rebuild.

- Built and ran the unit tests on macOS 15.6.1 (arm64, Apple clang 17.0.0). `SdpApiTest`, `SdpFunctionalityTest` and `SdpBufferSizeConfigTest` pass 90 tests, including the two new ones.
- With the source change reverted and only the tests applied, both fail: `mediaAttributesCount` reads `0` instead of `256`, and the over-limit case returns `STATUS_SUCCESS` with `sdpAttributes[0].attributeName` set to `rtcp-fb` instead of `mid`.

### DIFF
--- a/src/source/PeerConnection/SessionDescription.c
+++ b/src/source/PeerConnection/SessionDescription.c
@@ -193,7 +193,7 @@ STATUS setPayloadTypesFromOffer(PHashTable codecTable, PHashTable rtxTable, PSes
     ENTERS();
     STATUS retStatus = STATUS_SUCCESS;
     PSdpMediaDescription pMediaDescription = NULL;
-    UINT8 currentAttribute;
+    UINT16 currentAttribute;
     UINT16 currentMedia;
     PCHAR attributeValue, end;
     UINT64 parsedPayloadType, hashmapPayloadType, fmtpVal, aptVal;
--- a/src/source/Sdp/Sdp.h
+++ b/src/source/Sdp/Sdp.h
@@ -187,7 +187,9 @@ typedef struct {
 
     SdpAttributes sdpAttributes[MAX_SDP_ATTRIBUTES_COUNT];
 
-    UINT8 mediaAttributesCount;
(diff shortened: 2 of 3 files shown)


vabhasin:
@xxxxxxjun Thank you for your contribution. We were able to reproduce the behavior on our end and verify the fix. Your root-cause analysis tracing it back to the v1.7.1 limit increase is accurate and the unit tests clearly demonstrate the before/after behavior.

vabhasin (approved):

sirknightj (approved):

PR #2377, opened 2026-08-11
C SDK as viewer/recvonly: Fix RTP jitter buffer receive-path double-free

vikramdattu:
*Issue #, if available:*
- N/A

*What was changed?*
- `jitterBufferPush()` no longer returns an error once it has taken ownership of the packet: a failure from `jitterBufferInternalParse()` is now logged instead of propagated.
- Added a regression test, `pushErrorPathDoesNotDoubleFreePacket`.

*Why was it changed?*
- `jitterBufferPush()` inserts the packet into the hash table and then calls `jitterBufferInternalParse()`. If the parse fails — for example the depayloader rejects a corrupt packet and `CHK_STATUS` aborts — push returned that error while the packet was still referenced in the buffer. In `sendPacketToRtpReceiver()` the caller sets `ownedByJitterBuffer = TRUE` only after a successful push, so on the error path its `CleanUp` frees the packet. A later `jitterBufferInternalParse()` / `jitterBufferDropBufferData()` then frees the same packet again. On a device this surfaces as a heap double-free (`tlsf_free: block already marked as free`) and a crash, triggered by inbound video with occasional malformed payloads.

*How was it changed?*
- After the packet is inserted, the internal parse is treated as best-effort: on failure it logs and continues instead of returning the status. Push reports success because it has taken the packet, so the caller never frees a buffered packet. The offending packet is removed later as it ages out of the buffer via `maxLatency`.

- Added `pushErrorPathDoesNotDoubleFreePacket`, which drives the ownership sequence: two good frames, then a packet whose depay fails. Built with AddressSanitizer, the test reports a heap-use-after-free in `freeRtpPacket` before the fix and passes after it. The full `JitterBuffer*` suite passes (mbedTLS build).

### DIFF
--- a/src/source/PeerConnection/JitterBuffer.c
+++ b/src/source/PeerConnection/JitterBuffer.c
@@ -454,7 +454,12 @@ STATUS jitterBufferPush(PJitterBuffer pJitterBuffer, PRtpPacket pRtpPacket, PBOO
         }
     }
 
-    CHK_STATUS(jitterBufferInternalParse(pJitterBuffer, FALSE));
+    // Best-effort: the packet is now owned by the jitter buffer, so a parse failure
+    // must not be returned (the caller would free it -> double free). Log and continue.
+    status = jitterBufferInternalParse(pJitterBuffer, FALSE);
+    if (STATUS_FAILED(status)) {
+        DLOGW("jitterBufferInternalParse failed with 0x%08x", status);
+    }
 
 CleanUp:
 
--- a/tst/JitterBufferFunctionalityTest.cpp
+++ b/tst/JitterBufferFunctionalityTest.cpp
@@ -9,6 +9,95 @@ namespace webrtcclient {
 class JitterBufferFunctionalityTest : public WebRtcClientTestBase {
 };
 
+static STATUS reproFrameReadyFunc(UINT64, UINT16, UINT16, UINT32)
+{
+    return STATUS_SUCCESS;
+}
+
+static STATUS reproFrameDroppedFunc(UINT64, UINT16, UINT16, UINT32)
+{
+    return STATUS_SUCCESS;
+}
+
+// Fails the depay for a "corrupt" packet (first byte 0xEE) to force the parse-failure
+// path inside jitterBufferPush().
+static STATUS reproDepayRtpFunc(PBYTE payload, UINT32 payloadLength, PBYTE outBuffer, PUINT32 pBufferSize, PBOOL pIsStart)
+{
+    STATUS retStatus = STATUS_SUCCESS;
+    UINT32 bufferSize = 0;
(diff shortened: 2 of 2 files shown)


sirknightj (approved):
Thanks for the contribution!

The ownership analysis looks correct, the fix and the regression test look good to me. I ran the new test with the fix reverted and it failed as expected.

Wondering if you have additional details on the packet that triggered this? In particular, we'd like to understand whether the packet was genuinely malformed/corrupted, or whether it was a valid packet the depayloader doesn't yet handle (feature request)?

vikramdattu:
> Wondering if you have additional details on the packet that triggered this? In particular, we'd like to understand whether the packet was genuinely malformed/corrupted, or whether it was a valid packet the depayloader doesn't yet handle (feature request)?

@sirknightj there is not feature gap as such, but a packet with smaller size OR corruption making the packet lengthier. I would add some instrumentation for exact which one, and share the exact packet dump failing the parse. The fix is good neverthless as parse failure should not lead to double failure.

vikramdattu:
@sirknightj instrumented this and found the packet to be of below dump:

`78 00 10 | 67 42 c0 0c da dd e3 dc cc db a1 10 1a e4 3c 98 | 58 7c | 09 ac b1 …`

- 0x78 : NAL header, type 0x78 & 0x1F = 24 = STAP-A (aggregation packet), matches indicator=24.
- 00 10 : first aggregation unit length = 16 bytes.
- next 16 bytes start 0x67: NAL type 7 = SPS, profile 0x42/level 0x0c = H.264 Baseline 1.2. A perfectly valid SPS.
- at offset 19, 58 7c: second aggregation unit length = 0x587c = 22652 bytes. But only ~4 bytes of packet remain (total len 25).

That 22652-byte length prefix inside a 25-byte packet is garbage. The depayloader walks the STAP-A aggregation, hits the second unit, and correctly bails with STATUS_RTP_INPUT_PACKET_TOO_SMALL.

1. It's a corrupt packet, not an unsupported feature. The packet is truncated/corrupted mid-aggregation on the lossy inbound link (a valid second unit would be a tiny PPS like 00 04 …, not 22652 bytes). Error is STATUS_RTP_INPUT_PACKET_TOO_SMALL, not STATUS_RTP_INVALID_NALU or a "mode not supported"
2. The fix behaves exactly as intended, verified live. Pre-fix, jitterBufferPush inserted this packet then CHK_STATUS'd out on the parse failure after the jitter buffer already owned it --> caller double-freed. With the fix, the log shows jitterBufferPush(): jitterBufferInternalParse failed with 0x5c000001; packet retained in jitter buffer --> parse is best-effort, ownership stays with the buffer, no crash across the entire run (the SPS/first-NALU that did parse is still usable; only the corrupt tail is dropped).

vabhasin (approved):

PR #2380, opened 2026-08-14
New Feature: parse reception report blocks embedded in RTCP sender reports(follow-up to #2367)

YPandas:
*Issue #, if available:*
- N/A (follow-up to #2367)

*What was changed?*
- `src/source/PeerConnection/Rtcp.c` — `onRtcpSenderReport()` now parses the reception report blocks embedded in RTCP Sender Reports and updates each matching transceiver's `remoteInboundStats`, instead of silently discarding any SR whose payload is larger than the 24-byte sender-info section.
- The report-block parsing loop is extracted into a shared helper, `parseRtcpReceptionReportBlocks()`, used by both `onRtcpReceiverReport()` and `onRtcpSenderReport()`. The block-count cap (`RTCP_PACKET_RECEIVER_REPORT_MAX_BLOCKS`, keyed to `MAX_SDP_SESSION_MEDIA_COUNT`) and payload-length validation from #2367 now apply to both packet types.
- `tst/RtcpFunctionalityTest.cpp` — new unit tests covering SRs with embedded report blocks.

*Why was it changed?*
- Per [RFC 3550 §6.4.1](https://tools.ietf.org/html/rfc3550#section-6.4.1), a peer that both sends and receives media embeds its reception report blocks in its Sender Reports rather than emitting separate Receiver Reports. The previous `onRtcpSenderReport()` bailed out on any SR that was not exactly `RTCP_PACKET_SENDER_REPORT_MINLEN` (24 bytes), so every embedded report block was dropped (the code carried a `TODO: handle sender report containing receiver report blocks`).
- The practical impact: `remoteInboundStats` (`reportsReceived`, `fractionLost`, `roundTripTime`, …) freezes as soon as the remote peer becomes a sender. Observed live against the media storage backend — it switches from RRs to SRs with embedded blocks the moment it starts sending audio on the sendrecv audio transceiver, after which the master receives no remote-inbound updates. #2367 fixed multi-block parsing for standalone RRs; this closes the remaining gap for SRs.

*How was it changed?*
- The RR and SR block formats are identical; only the offset of the first block differs (4 bytes after the sender SSRC for RRs, 24 bytes for SRs due to the sender-info section). `parseRtcpReceptionReportBlocks()` takes that offset as a parameter, and both packet handlers delegate to it.
- `onRtcpSenderReport()` keeps its existing sender-info handling (NTP/RTP timestamp, packet/octet counts) and processes embedded blocks afterwards; malformed packets that claim more blocks than the payload carries are rejected with a warning, mirroring the RR path.

- New unit tests in `RtcpFunctionalityTest`: SR with one embedded report block (verifying stats and RTT computation), SR with two embedded blocks for different SSRCs (audio + video transceivers both updated), and a malformed SR claiming more blocks than its payload carries (rejected without stats corruption).
- Full `RtcpFunctionalityTest` suite passes locally: 49/49 tests.
- Verified live against the media storage backend: with this change, the master's remote-inbound stats continue updating after the backend transitions from RRs to SRs (previously they froze at that point).

### DIFF
--- a/src/source/PeerConnection/Rtcp.c
+++ b/src/source/PeerConnection/Rtcp.c
@@ -49,37 +49,13 @@ CleanUp:
     return retStatus;
 }
 
-// TODO better sender report handling https://tools.ietf.org/html/rfc3550#section-6.4.1
-static STATUS onRtcpSenderReport(PRtcpPacket pRtcpPacket, PKvsPeerConnection pKvsPeerConnection)
-{
-    STATUS retStatus = STATUS_SUCCESS;
-    UINT32 senderSSRC;
-    PKvsRtpTransceiver pTransceiver = NULL;
-
-    CHK(pKvsPeerConnection != NULL && pRtcpPacket != NULL, STATUS_NULL_ARG);
-
-    if (pRtcpPacket->payloadLength != RTCP_PACKET_SENDER_REPORT_MINLEN) {
-        // TODO: handle sender report containing receiver report blocks
-        return STATUS_SUCCESS;
-    }
-
-    senderSSRC = getUnalignedInt32BigEndian(pRtcpPacket->payload);
-    if (STATUS_SUCCEEDED(findTransceiverBySsrc(pKvsPeerConnection, &pTransceiver, senderSSRC))) {
-        UINT64 ntpTime = getUnalignedInt64BigEndian(pRtcpPacket->payload + 4);
-        UINT32 rtpTs = getUnalignedInt32BigEndian(pRtcpPacket->payload + 12);
-        UINT32 packetCnt = getUnalignedInt32BigEndian(pRtcpPacket->payload + 16);
-        UINT32 octetCnt = getUnalignedInt32BigEndian(pRtcpPacket->payload + 20);
-        DLOGV("RTCP_PACKET_TYPE_SENDER_REPORT %d %" PRIu64 " rtpTs: %u %u pkts %u bytes", senderSSRC, ntpTime, rtpTs, packetCnt, octetCnt);
-    } else {
-        DLOGW("Received sender report for non existing ssrc: %u", senderSSRC);
-    }
-
-CleanUp:
-
-    return retStatus;
-}
-
-static STATUS onRtcpReceiverReport(PRtcpPacket pRtcpPacket, PKvsPeerConnection pKvsPeerConnection)
+// Parses the reception report blocks carried in a receiver report or a sender report and updates
+// remoteInboundStats on the transceiver matching each block's source SSRC.
+// https://tools.ietf.org/html/rfc3550#section-6.4.1 (SR), https://tools.ietf.org/html/rfc3550#section-6.4.2 (RR)
+// The block format is identical in both packet types; only the offset of the first block differs:
+//   RR: 4 bytes  (after the sender SSRC)
+//   SR: 24 bytes (after the sender SSRC and the 20-byte sender info section)
+static STATUS parseRtcpReceptionReportBlocks(PRtcpPacket pRtcpPacket, PKvsPeerConnection pKvsPeerConnection, UINT32 firstReportBlockOffset)
 {
     STATUS retStatus = STATUS_SUCCESS;
     PKvsRtpTransceiver pTransceiver = NULL;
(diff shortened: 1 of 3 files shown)


src/source/PeerConnection/Rtcp.c
```diff
@@ -158,6 +135,47 @@ static STATUS onRtcpReceiverReport(PRtcpPacket pRtcpPacket, PKvsPeerConnection p
...
+{
+    STATUS retStatus = STATUS_SUCCESS;
+    UINT32 senderSSRC;
+    PKvsRtpTransceiver pTransceiver = NULL;
+
+    CHK(pKvsPeerConnection != NULL && pRtcpPacket != NULL, STATUS_NULL_ARG);
+
+    // https://tools.ietf.org/html/rfc3550#section-6.4.1
+    if (pRtcpPacket->payloadLength < RTCP_PACKET_SENDER_REPORT_MINLEN) {
+        DLOGW("Malformed sender report: payloadLength %u too small", pRtcpPacket->payloadLength);
```
unicornss, line 156: do we need a sanity check ?
```
RTCP_PACKET_SENDER_REPORT_MINLEN is at least 24 bytes (4 SSRC + 8 NTP + 4 RTP TS + 4 pkt count + 4 octet count).
```

unicornss (approved):

src/source/PeerConnection/Rtcp.c
```diff
@@ -158,6 +135,47 @@ static STATUS onRtcpReceiverReport(PRtcpPacket pRtcpPacket, PKvsPeerConnection p
     return retStatus;
 }
 
+static STATUS onRtcpSenderReport(PRtcpPacket pRtcpPacket, PKvsPeerConnection pKvsPeerConnection)
+{
+    STATUS retStatus = STATUS_SUCCESS;
+    UINT32 senderSSRC;
+    PKvsRtpTransceiver pTransceiver = NULL;
+
+    CHK(pKvsPeerConnection != NULL && pRtcpPacket != NULL, STATUS_NULL_ARG);
```
sirknightj: Suggest adding `payload != NULL` to the NULL guards.

Both parseRtcpReceptionReportBlocks and onRtcpSenderReport check `pRtcpPacket != NULL` but then dereference `pRtcpPacket->payload` directly. The only current caller path always sets payload non-NULL, but a future manually-constructed RtcpPacket, new caller, or unit test would segfault instead of returning cleanly:

`CHK(pKvsPeerConnection != NULL && pRtcpPacket != NULL && pRtcpPacket->payload != NULL, STATUS_NULL_ARG);`

src/source/PeerConnection/Rtcp.c
```diff
@@ -158,6 +135,47 @@ static STATUS onRtcpReceiverReport(PRtcpPacket pRtcpPacket, PKvsPeerConnection p
...
+
+    // https://tools.ietf.org/html/rfc3550#section-6.4.1
+    if (pRtcpPacket->payloadLength < RTCP_PACKET_SENDER_REPORT_MINLEN) {
+        DLOGW("Malformed sender report: payloadLength %u too small", pRtcpPacket->payloadLength);
+        return STATUS_SUCCESS;
```
sirknightj: Route early bail-outs through CleanUp instead of bare return. Three spots do return STATUS_SUCCESS; mid-body (the malformed-length branch in the helper, and the two guards in onRtcpSenderReport). Correct only because CleanUp: is a no-op today; the moment cleanup logic (an allocation, or a MUTEX_LOCK needing release) is added, these bare returns silently skip it. Use the KVS idiom so all exits converge:

```c
if (pRtcpPacket->payloadLength < firstReportBlockOffset + (UINT32) reportBlockCount * RTCP_PACKET_RECEIVER_REPORT_BLOCK_LEN) {
    DLOGW("Malformed report: ...", ...);
    CHK(FALSE, STATUS_SUCCESS);   // -> CleanUp, returns SUCCESS
}
```

And add a CHK_LOG_ERR in the cleanUp

src/source/PeerConnection/Rtcp.c
```diff
@@ -90,28 +66,29 @@ static STATUS onRtcpReceiverReport(PRtcpPacket pRtcpPacket, PKvsPeerConnection p
     UINT8 reportBlockCount, i;
 
     CHK(pKvsPeerConnection != NULL && pRtcpPacket != NULL, STATUS_NULL_ARG);
```
sirknightj: Recommend to validate the firstReportBlockOffset

```
// firstReportBlockOffset must be the byte offset of the first report block within the payload:
// SIZEOF(UINT32) for RR, RTCP_PACKET_SENDER_REPORT_MINLEN for SR.
CHK(firstReportBlockOffset >= SIZEOF(UINT32) && firstReportBlockOffset <= RTCP_PACKET_SENDER_REPORT_MINLEN, STATUS_INVALID_ARG);
```

sirknightj (approved):

sirknightj (approved):

PR #2385, opened 2026-08-18
Crypto: take RNG from PSA when mbedTLS is built without the entropy module

vikramdattu:
*Issue #, if available:*
- N/A

*What was changed?*
- `Include_i.h` derives a new `MBEDTLS_HAS_ENTROPY` from the mbedTLS configuration, and the DTLS/TLS session paths plus the pre-v4 `createCertificateAndKey()` no longer assume the entropy module exists.
- Builds without it source randomness from PSA. Where entropy is present, behaviour is unchanged.

*Why was it changed?*
- A build that supplies randomness through a PSA driver — `MBEDTLS_PSA_CRYPTO_EXTERNAL_RNG`, the default on ESP-IDF v6 targets — compiles mbedTLS's entropy module out entirely, and is permitted to drop CTR-DRBG with it. `MBEDTLS_ENTROPY_C` is then undefined and the symbols do not exist, so the SDK fails to link:
  ```
  undefined reference to `mbedtls_entropy_init'
  undefined reference to `mbedtls_entropy_func'
  undefined reference to `mbedtls_entropy_free'
  ```
- This is a build-*configuration* question rather than a version one, which is why the guard is `MBEDTLS_ENTROPY_C` and not `MBEDTLS_VERSION_MAJOR`: stock mbedTLS 4 still ships the entropy module, so CI does not encounter this.
- Beyond the link error, two paths would have gone wrong silently once entropy was absent. `copyCertificateAndKey()` passes the session DRBG to `mbedtls_pk_check_pair()` on mbedTLS 3.2–3.6; an `init()`-only DRBG has `reseed_counter == -1`, so it never reseeds and drives an unkeyed cipher operation. And the pre-v4 `createCertificateAndKey()` — the default path whenever no `RtcCertificate` is supplied — seeds its own local DRBG from entropy, so guarding only the session paths would have left mbedTLS 3.x unbuildable anyway.

*How was it changed?*
- `Include_i.h`: `MBEDTLS_HAS_ENTROPY` is `1` when both `MBEDTLS_ENTROPY_C` and `MBEDTLS_CTR_DRBG_C` are defined, `0` when PSA Crypto is available instead (`MBEDTLS_PSA_CRYPTO_CLIENT` / `MBEDTLS_PSA_CRYPTO_C`, which is also where `<mbedtls/psa_util.h>` is pulled in), and `#error` when neither can provide randomness. It sits with the other mbedTLS includes so it is defined before the structs that use it.
- `Dtls.h` / `Tls.h`: the `entropy` and `ctrDrbg` members are compiled out when absent, so any future use in such a build is a compile error rather than a silently weak RNG. `KVS_MBEDTLS_DRBG()` yields the DRBG or `NULL`.
- `Dtls_mbedtls.c` / `Tls_mbedtls.c`: entropy/DRBG init, seeding and teardown are guarded. `mbedtls_ssl_conf_rng()` and `mbedtls_pk_check_pair()` receive `mbedtls_psa_get_random`/`MBEDTLS_PSA_RANDOM_STATE` instead of an unseeded DRBG, and the pre-v4 `createCertificateAndKey()` drives RSA/ECP key generation and `mbedtls_x509write_crt_der()` through the same pair.
- `Crypto.h`: `KVS_CRYPTO_INIT()` now also covers entropy-less pre-v4 builds, so `psa_crypto_init()` still happens once from `initKvsWebRtc()`. Calling it per session would have reintroduced the race on builds without `MBEDTLS_THREADING_C` that the existing v4 comment warns about.

- Desktop mbedTLS build (`-DUSE_MBEDTLS=ON -DBUILD_TEST=ON`), entropy present: `webrtc_client_test` passes, 18/18 across `DtlsApiTest`, `DtlsFunctionalityTest`, `PeerConnectionApiTest` — the path this change must leave untouched.
- ESP32-P4 with ESP-IDF v6.0.2 (mbedTLS 4.1.1, external RNG, no entropy module): previously failed to link; now links, the DTLS handshake completes in ~1.1 s, and SRTP video/audio plus a data channel flow to a browser viewer over a KVS signaling channel.
- ESP-IDF v5.5 (mbedTLS 3.x, entropy present): builds and streams unchanged.

### DIFF
--- a/src/source/Crypto/Dtls_mbedtls.c
+++ b/src/source/Crypto/Dtls_mbedtls.c
@@ -1047,6 +1047,8 @@ STATUS createCertificateAndKey(INT32 certificateBits, BOOL generateRSACertificat
 #if MBEDTLS_HAS_ENTROPY
     mbedtls_entropy_context* pEntropy = NULL;
     mbedtls_ctr_drbg_context* pCtrDrbg = NULL;
+#else
+    psa_status_t psaStatus;
 #endif
     /* RNG handed to key generation and certificate signing below. */
     int (*fRng)(void*, unsigned char*, size_t);
@@ -1080,7 +1082,12 @@ STATUS createCertificateAndKey(INT32 certificateBits, BOOL generateRSACertificat
     fRng = mbedtls_ctr_drbg_random;
     pRng = pCtrDrbg;
 #else
-    /* PSA is the RNG; KVS_CRYPTO_INIT() in initKvsWebRtc() has initialised it. */
+    /* PSA is the RNG. Idempotent; guards against callers reaching here before
+     * initKvsWebRtc() has run KVS_CRYPTO_INIT(). */
+    if ((psaStatus = psa_crypto_init()) != PSA_SUCCESS) {
+        DLOGE("psa_crypto_init failed (status %d)", (int) psaStatus);
+    }
+    CHK(psaStatus == PSA_SUCCESS, STATUS_CERTIFICATE_GENERATION_FAILED);
     fRng = mbedtls_psa_get_random;
     pRng = MBEDTLS_PSA_RANDOM_STATE;
 #endif
--- a/src/source/Crypto/Dtls_mbedtls.c
+++ b/src/source/Crypto/Dtls_mbedtls.c
@@ -168,13 +168,19 @@ STATUS createDtlsSessionWithOptions(PDtlsSessionCallbacks pDtlsSessionCallbacks,
     CHK(pDtlsSession != NULL, STATUS_NOT_ENOUGH_MEMORY);
 
     // initialize mbedtls stuff with sane values
+#if MBEDTLS_HAS_ENTROPY
     mbedtls_entropy_init(&pDtlsSession->entropy);
     mbedtls_ctr_drbg_init(&pDtlsSession->ctrDrbg);
+#endif
     mbedtls_ssl_config_init(&pDtlsSession->sslCtxConfig);
     mbedtls_ssl_init(&pDtlsSession->sslCtx);
     mbedtls_x509_crt_init(&pDtlsSession->trustedCaCert);
+#if MBEDTLS_HAS_ENTROPY
     mbedtls_ctr_drbg_set_prediction_resistance(&pDtlsSession->ctrDrbg, MBEDTLS_CTR_DRBG_PR_ON);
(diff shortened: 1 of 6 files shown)


vikramdattu:
cc @sirknightj @unicornss

src/source/Crypto/Dtls_mbedtls.c
```diff
@@ -1027,38 +1044,55 @@ STATUS createCertificateAndKey(INT32 certificateBits, BOOL generateRSACertificat
...
     mbedtls_x509write_crt_init(pWriteCert);
     mbedtls_x509_crt_init(pCert);
     mbedtls_pk_init(pKey);
     initialized = TRUE;
+#if MBEDTLS_HAS_ENTROPY
     CHK(mbedtls_ctr_drbg_seed(pCtrDrbg, mbedtls_entropy_func, pEntropy, NULL, 0) == 0, STATUS_CERTIFICATE_GENERATION_FAILED);
+    fRng = mbedtls_ctr_drbg_random;
+    pRng = pCtrDrbg;
+#else
+    /* PSA is the RNG; KVS_CRYPTO_INIT() in initKvsWebRtc() has initialised it. */
```
sirknightj: It is possible to call `createRtcCertificate` without calling `initKvsWebrtc` in an incorrectly implemented app.

Suggestion to prevent invalid usage:

```diff
  #else      
-     /* PSA is the RNG; KVS_CRYPTO_INIT() in initKvsWebRtc() has initialised it. */                                                                                                                                                                        
+     /* PSA is the RNG. Idempotent; guards against callers reaching here before                                                                                          
+      * initKvsWebRtc() has run KVS_CRYPTO_INIT(). */                                                                                                                               
+     if ((psaStatus = psa_crypto_init()) != PSA_SUCCESS) {                                                                                                                          
+         DLOGE("psa_crypto_init failed (status %d)", (int) psaStatus);                                                                                                              
+     }                                                                                                                                                                              
+     CHK(psaStatus == PSA_SUCCESS, STATUS_CERTIFICATE_GENERATION_FAILED);                                                                                                           
      fRng = mbedtls_psa_get_random;                                                                                                                                                 
      pRng = MBEDTLS_PSA_RANDOM_STATE;                                                                                                                                               
  #endif    
```

sirknightj (approved):

src/source/Crypto/Dtls_mbedtls.c
vikramdattu: @sirknightj done, suggestion incorporated. We now do `psa_crypto_init` whenever the `createRtcCertificate` is called. Totally fine since the `psa_crypto_init' is idempotent/singleton

PR #2387, opened 2026-08-19
Ice/TURN: don't wait the per-state timeout for peers rejected by the TURN server

### LINKED ISSUE #2386: [BUG] TURN state machine waits the full CreatePermission/BindChannel timeout for peers rejected with 403 Forbidden IP — delays ICE gathering completion (breaks non-trickle answers)
### Which SDK are you using
C WebRTC SDK, **v1.15.0** (the same code is present on `master`).

### What did you expect to happen
When gathering local ICE candidates, once all reachable STUN/TURN candidates are collected, `iceAgentGatherCandidateTimerCallback()` should report `Candidate gathering completed` immediately via the early-completion path (`pendingCandidateCount == 0 && addedRelayCandidate`), so the local description is ready right away.

### What actually happens
Gathering does not complete until the `iceLocalCandidateGatheringTimeout` (~10s), even though every usable host/srflx/relay candidate has already been collected within a few hundred ms.

### Root cause
When a *remote* candidate carries a non-routable address (RFC1918 private or RFC3927 link-local), the TURN server rejects `CreatePermission` for it with `403 Forbidden IP`:

```
Received STUN error response. Error type: 0x118, Error Code: 403. Error detail: Forbidden IP.
```

In `turnConnectionHandleStunError()` (`src/source/Ice/TurnConnection.c`) that peer is set to `TURN_PEER_CONN_STATE_FAILED` **but kept in `turnPeerList`** (`turnPeerCount` is not decremented).

The state-advance conditions in `src/source/Ice/TurnConnectionStateMachine.c` only leave a state early when **every** peer is on the ready side:

```c
// fromCreatePermissionTurnState
if (currentTime > stateTimeoutTime || channelWithPermissionCount == turnPeerCount) { ... }
// fromBindChannelTurnState
if (currentTime > stateTimeoutTime || readyPeerCount == turnPeerCount) { ... }
```

`FAILED` peers are never counted in `channelWithPermissionCount` / `readyPeerCount`, so `== turnPeerCount` can never become true once any peer is rejected. The state machine therefore always falls through to the full per-state timeout (`DEFAULT_TURN_CREATE_PERMISSION_TIMEOUT` = 5s, then `DEFAULT_TURN_BIND_CHANNEL_TIMEOUT` = 5s) → ~10s before `TURN_STATE_READY`, even though the permission/channel for the *reachable* peer succeeded instantly.

Observed on device (one TURN connection):

```
CREATE_PERMISSION entered             t=0.548
Create permission succeeded (peer P)  t=0.586   # reachable peer OK immediately
Channel bind succeeded    (peer P)    t=0.637
... 4x "403 Forbidden IP" for the peer's private/link-local candidates ...
CREATE_PERMISSION -> BIND_CHANNEL      t=5.562   # +5s (timeout, not on success)
BIND_CHANNEL     -> READY              t=10.575  # +5s (timeout)
ICE candidate gathering complete       t=10.198  # only at the gathering timeout
```

While the TURN connection is not `READY`, its relay candidate stays pending, so `pendingCandidateCount != 0` and gathering cannot complete early.

### Why this matters — non-trickle peers
With **trickle ICE** the impact is hidden: the answer is sent immediately and candidates trickle out, so the late gathering completion is harmless.

With a **non-trickle** peer the answer is only serialized/sent at gathering-complete, so **every session is delayed by ~10s** before the answer is sent — a large call-setup latency and, depending on the remote's handshake timeout, an outright connection failure.

Our concrete case is **Amazon Alexa Smart Home** (`Alexa.RTCSessionController` `InitiateSessionWithOffer`): the offer is **non-trickle** and includes the device's private/link-local candidates, which the KVS TURN server rejects with `403 Forbidden IP` — so every Alexa live-view session pays the full ~10s gathering timeout before we can answer.

### Proposed fix
Treat `TURN_PEER_CONN_STATE_FAILED` as a terminal state and advance as soon as every peer is either ready or failed (with at least one ready). Failed peers can never be relayed to anyway, and peers added later still get their permission/channel created in `TURN_STATE_READY`. PR to follow.

alitofresko:
### Issue #, if available
Fixes #2386

### Description of changes
`fromCreatePermissionTurnState()` and `fromBindChannelTurnState()` advance the TURN connection early only when **all** peers reached the ready side (`channelWithPermissionCount == turnPeerCount` / `readyPeerCount == turnPeerCount`).

When a peer's `CreatePermission` is rejected by the TURN server (e.g. `403 Forbidden IP` for a non-routable remote candidate), the peer is set to `TURN_PEER_CONN_STATE_FAILED` in `turnConnectionHandleStunError()` but kept in `turnPeerList` (`turnPeerCount` is not decremented). Because a failed peer is never counted as ready, `== turnPeerCount` can never be reached, and the state machine always waits the full `DEFAULT_TURN_CREATE_PERMISSION_TIMEOUT` (5s) then `DEFAULT_TURN_BIND_CHANNEL_TIMEOUT` (5s) — ~10s to reach `TURN_STATE_READY` — even when the reachable peer's permission and channel bind succeeded immediately.

While the TURN connection is not `READY`, its relay candidate stays pending, so `pendingCandidateCount != 0` prevents `iceAgentGatherCandidateTimerCallback()` from completing gathering early; gathering only finishes at `iceLocalCandidateGatheringTimeout`. For **non-trickle** sessions (where the answer is serialized/sent at gathering-complete) this delays every answer by ~10s. Reproduced with **Amazon Alexa Smart Home**, whose `InitiateSessionWithOffer` offer is non-trickle and includes the device's private/link-local candidates that the TURN server rejects with `403 Forbidden IP`. See #2386 for a full trace.

This change counts `TURN_PEER_CONN_STATE_FAILED` peers as terminal and advances as soon as every peer is either ready or failed, keeping the existing guard that at least one peer is ready. Failed peers can never be relayed to, and peers added later still get their permission/channel created while in `TURN_STATE_READY`, so nothing is lost. When no peer fails (`failedPeerCount == 0`) the condition is identical to before.

### How was this change tested
- On device (aarch64 / Yocto, KVS 1.15.0): a non-trickle Alexa live-view session previously reported `ICE candidate gathering complete` at ~10s; with the patch the TURN connections reach `READY` right after the first channel bind and gathering completes early, so the answer is sent without the timeout wait.
- Verified no behavior change for the all-peers-succeed and single-peer cases (`failedPeerCount == 0`).

### License
By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

### DIFF
--- a/src/source/Ice/TurnConnectionStateMachine.c
+++ b/src/source/Ice/TurnConnectionStateMachine.c
@@ -376,7 +376,7 @@ STATUS fromCreatePermissionTurnState(UINT64 customData, PUINT64 pState)
     STATUS retStatus = STATUS_SUCCESS;
     PTurnConnection pTurnConnection = (PTurnConnection) customData;
     UINT64 state = TURN_STATE_CREATE_PERMISSION, currentTime;
-    UINT32 channelWithPermissionCount = 0, i = 0;
+    UINT32 channelWithPermissionCount = 0, failedPeerCount = 0, i = 0;
     BOOL locked = FALSE;
 
     CHK(pTurnConnection != NULL && pState != NULL, STATUS_NULL_ARG);
@@ -399,6 +399,8 @@ STATUS fromCreatePermissionTurnState(UINT64 customData, PUINT64 pState)
         if (pTurnConnection->turnPeerList[i].connectionState == TURN_PEER_CONN_STATE_BIND_CHANNEL ||
             pTurnConnection->turnPeerList[i].connectionState == TURN_PEER_CONN_STATE_READY) {
             channelWithPermissionCount++;
+        } else if (pTurnConnection->turnPeerList[i].connectionState == TURN_PEER_CONN_STATE_FAILED) {
+            failedPeerCount++;
         }
     }
 
@@ -408,7 +410,7 @@ STATUS fromCreatePermissionTurnState(UINT64 customData, PUINT64 pState)
         CHK(FALSE, retStatus);
     }
 
-    if (currentTime > pTurnConnection->stateTimeoutTime || channelWithPermissionCount == pTurnConnection->turnPeerCount) {
+    if (currentTime > pTurnConnection->stateTimeoutTime || channelWithPermissionCount + failedPeerCount == pTurnConnection->turnPeerCount) {
         CHK(channelWithPermissionCount > 0, STATUS_TURN_CONNECTION_FAILED_TO_CREATE_PERMISSION);
 
         // go to next state if we have at least one ready peer
@@ -501,7 +503,7 @@ STATUS fromBindChannelTurnState(UINT64 customData, PUINT64 pState)
     UINT64 state = TURN_STATE_BIND_CHANNEL;
     UINT64 currentTime;
     BOOL locked = FALSE;
-    UINT32 readyPeerCount = 0, i = 0;
+    UINT32 readyPeerCount = 0, failedPeerCount = 0, i = 0;
 
     CHK(pTurnConnection != NULL && pState != NULL, STATUS_NULL_ARG);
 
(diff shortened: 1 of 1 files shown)


YPandas (approved):

vabhasin:
Thank you @alitofresko for your contribution, we were able to reproduce the behavior and verify the fix manually. Can we add/modify any unit/functionality test for this change?

vabhasin (approved):

YPandas:
Will have a follow up PR to address unit tests.

PR #2388, opened 2026-08-24
Fix issue 2386 - #2388  (Adding filter for Relay candidates)

YPandas:
*Issue #, if available:*
- N/A (builds on / complements the issue raised in #2387)

*What was changed?*
- Added a proactive filter that skips creating TURN permissions for **non-routable peer addresses** that a public TURN server (e.g. KVS) rejects with `403 Forbidden IP`.
- New `IS_NON_ROUTABLE_ADDR` classifier in `Include_i.h` covering all not-globally-reachable IPv4 and IPv6 ranges (IANA Special-Purpose registries / RFC 6890).
- New opt-out config flag `KvsRtcConfiguration.disableTurnPeerAddressFilter` (default off = filter on).
- Added a `nonRoutablePeersFiltered` diagnostic counter, logged at INFO on each skip (no peer IP logged).

*Why was it changed?*
- #2387 fixes the symptom reactively: a `403`-failed peer stalls the TURN state machine for the per-state timeouts (~5s + ~5s), delaying ICE gathering and breaking non-trickle answerers.
- This PR removes the root cause for the common case: the `403` is predictable for private/link-local peers, so we never send the doomed `CreatePermission` — no error, no failed peer, no stall.
- Complementary to #2387: this avoids the wasted round-trips; #2387 remains the safety net for any other failure.

*How was it changed?*
- `turnConnectionAddPeer` skips a peer up front when `IS_NON_ROUTABLE_ADDR` matches and the filter isn't disabled; the candidate is untouched, so it can still form direct host/srflx pairs.
- Only the TURN permission is skipped — no change to candidate gathering or ICE pairing.
- The filter can be disabled via config for TURN servers that legitimately relay to private/ULA peers (on-prem/LAN).

- Unit tests for `IS_NON_ROUTABLE_ADDR` covering every IPv4/IPv6 range, boundary values, routable negatives, and NULL/unset-family (all passing).
- Integration test asserting `turnConnectionAddPeer` skips a non-routable peer, adds a routable one, and adds a non-routable one when the filter is disabled.
- Manual end-to-end run (force-relay master + LAN viewer): `403`s eliminated and TURN reaches `READY` in ~110ms vs the ~4s+ stall without the fix.
BEFORE: 
image
AFTER: 
image

### DIFF
--- a/src/include/com/amazonaws/kinesis/video/webrtcclient/Include.h
+++ b/src/include/com/amazonaws/kinesis/video/webrtcclient/Include.h
@@ -1472,6 +1472,11 @@ typedef struct {
     BOOL enableIceStats; //!< Control whether ICE agent stats are to be calculated. ENABLE_STATS_CALCULATION_CONTROL compiler flag must be defined
                          //!< to use this member, else stats are enabled by default.
 #endif
+    BOOL disableTurnPeerAddressFilter; //!< By default the SDK does not create TURN permissions for non-routable peer addresses
+                                       //!< (RFC 1918 private, RFC 6598 CGNAT, loopback, RFC 3927 link-local, etc.), because a public
+                                       //!< TURN server such as the KVS TURN service rejects them with 403 Forbidden IP. Set this to TRUE
+                                       //!< if you use a TURN server that legitimately relays to such addresses (e.g. an on-prem/LAN TURN),
+                                       //!< so those peers are not filtered. Filtering is enabled by default (this flag defaults to FALSE).
 } KvsRtcConfiguration, *PKvsRtcConfiguration;
 
 /**
--- a/src/include/com/amazonaws/kinesis/video/webrtcclient/Include.h
+++ b/src/include/com/amazonaws/kinesis/video/webrtcclient/Include.h
@@ -1472,11 +1472,11 @@ typedef struct {
     BOOL enableIceStats; //!< Control whether ICE agent stats are to be calculated. ENABLE_STATS_CALCULATION_CONTROL compiler flag must be defined
                          //!< to use this member, else stats are enabled by default.
 #endif
-    BOOL disableTurnPeerAddressFilter; //!< By default the SDK does not create TURN permissions for non-routable peer addresses
-                                       //!< (RFC 1918 private, RFC 6598 CGNAT, loopback, RFC 3927 link-local, etc.), because a public
-                                       //!< TURN server such as the KVS TURN service rejects them with 403 Forbidden IP. Set this to TRUE
-                                       //!< if you use a TURN server that legitimately relays to such addresses (e.g. an on-prem/LAN TURN),
-                                       //!< so those peers are not filtered. Filtering is enabled by default (this flag defaults to FALSE).
+    BOOL disableNonRoutablePeersFilterForRelay; //!< By default the SDK does not create TURN permissions for non-routable peer addresses
+                                                //!< (RFC 1918 private, RFC 6598 CGNAT, loopback, RFC 3927 link-local, etc.), because a public
+                                                //!< TURN server such as the KVS TURN service rejects them with 403 Forbidden IP. Set this to TRUE
+                                                //!< if you use a TURN server that legitimately relays to such addresses (e.g. an on-prem/LAN TURN),
+                                                //!< so those peers are not filtered. Filtering is enabled by default (this flag defaults to FALSE).
 } KvsRtcConfiguration, *PKvsRtcConfiguration;
 
 /**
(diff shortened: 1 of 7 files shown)


src/source/Ice/TurnConnection.h
```diff
@@ -197,6 +197,15 @@ struct __TurnConnection {
...
     PStateMachine pStateMachine;
 
+    // When FALSE (default), turnConnectionAddPeer skips non-routable peer addresses that a public
+    // TURN server would reject with 403 Forbidden IP. Set TRUE (via KvsRtcConfiguration) for TURN
+    // servers that legitimately relay to private/link-local peers (e.g. on-prem/LAN deployments).
+    BOOL disableTurnPeerAddressFilter;
+
+    // Diagnostic counter: number of peers skipped by the non-routable address filter on this
+    // TURN connection. Surfaced in logs so an operator can tell when a relay path was removed.
+    UINT32 nonRoutablePeersFiltered;
```
unicornss: nonRoutablePeersFilteredCount

src/source/Ice/TurnConnection.h
```diff
@@ -197,6 +197,15 @@ struct __TurnConnection {
     TurnProfileDiagnostics turnProfileDiagnostics;
     PStateMachine pStateMachine;
 
+    // When FALSE (default), turnConnectionAddPeer skips non-routable peer addresses that a public
+    // TURN server would reject with 403 Forbidden IP. Set TRUE (via KvsRtcConfiguration) for TURN
+    // servers that legitimately relay to private/link-local peers (e.g. on-prem/LAN deployments).
+    BOOL disableTurnPeerAddressFilter;
```
unicornss: disableNonRoutablePeersFilterForRelay

tst/NetworkApiTest.cpp
```diff
@@ -234,6 +234,112 @@ TEST_F(NetworkApiTest, GetIpAddrPortStrIpv6BufferTooSmallByOne)
...
+    setKvsV4Addr(&addr, 100, 128, 0, 0);     EXPECT_FALSE(IS_NON_ROUTABLE_ADDR(&addr)); // just above 100.64/10
+    setKvsV4Addr(&addr, 172, 15, 255, 255);  EXPECT_FALSE(IS_NON_ROUTABLE_ADDR(&addr)); // just below 172.16/12
+    setKvsV4Addr(&addr, 172, 32, 0, 0);      EXPECT_FALSE(IS_NON_ROUTABLE_ADDR(&addr)); // just above 172.16/12
+    setKvsV4Addr(&addr, 198, 17, 255, 255);  EXPECT_FALSE(IS_NON_ROUTABLE_ADDR(&addr)); // just below 198.18/15
+    setKvsV4Addr(&addr, 198, 20, 0, 0);      EXPECT_FALSE(IS_NON_ROUTABLE_ADDR(&addr)); // just above 198.18/15
+    setKvsV4Addr(&addr, 239, 255, 255, 255); EXPECT_FALSE(IS_NON_ROUTABLE_ADDR(&addr)); // just below 240/4 (multicast, not filtered)
+    setKvsV4Addr(&addr, 192, 1, 0, 1);       EXPECT_FALSE(IS_NON_ROUTABLE_ADDR(&addr)); // 192.x but not a reserved /24 nor 192.168
+    setKvsV4Addr(&addr, 203, 0, 114, 1);     EXPECT_FALSE(IS_NON_ROUTABLE_ADDR(&addr)); // adjacent to TEST-NET-3
+}
+
```
unicornss, line 318: test for  // ::2 not loopback

unicornss (approved):
Let's confirm that the logs are not printed twice.

YPandas:
> Let's confirm that the logs are not printed twice.
image

@unicornss  I had the same concern earlier. I confirmed the two lines come from different TURN connections by printing the memory pointer of the TurnConnection object — the duplicated lines carry different pointers, so they're two separate allocations, not a repeated log. The same two pointers also show up in the SDK's existing TURN Allocation succeeded logs, each with a different relay port, which confirms it.

vabhasin (approved):

src/include/com/amazonaws/kinesis/video/webrtcclient/Include.h
```diff
@@ -1472,6 +1472,11 @@ typedef struct {
     BOOL enableIceStats; //!< Control whether ICE agent stats are to be calculated. ENABLE_STATS_CALCULATION_CONTROL compiler flag must be defined
                          //!< to use this member, else stats are enabled by default.
 #endif
+    BOOL disableNonRoutablePeersFilterForRelay; //!< By default the SDK does not create TURN permissions for non-routable peer addresses
```
sirknightj, line 1475: Would it make more sense to have a function callback to make it customizable?

We can provide the default function as you've implemented below. But we can make it more flexible in case someone wants to filter it differently

PR #2392, opened 2026-08-31
Crypto: hash DTLS certificate via mbedtls_md instead of mbedtls_sha256

ak1932:
*Issue #, if available:*
- N/A

*What was changed?*
- Use of public mbedtls_md() API instead of private mbedtls_sha256()

*Why was it changed?*
- mbedTLS 4 moved mbedtls_sha256() into a private header and only builds it when the builtin SHA-256 driver is used, so a build with a PSA accelerator (e.g. ESP-IDF hardware SHA) fails to link. mbedtls_md() is public in mbedTLS 2/3/4 and dispatches to the active driver, dropping the version fork.

*How was it changed?*
- By using public mbedtls_md() API

- Tested on IDF examples

### DIFF
--- a/src/source/Crypto/Dtls_mbedtls.c
+++ b/src/source/Crypto/Dtls_mbedtls.c
@@ -1147,11 +1147,9 @@ STATUS dtlsCertificateFingerprint(mbedtls_x509_crt* pCert, PCHAR pBuff)
     pMdInfo = mbedtls_md_info_from_type(MBEDTLS_MD_SHA256);
     CHK(pMdInfo != NULL, STATUS_INTERNAL_ERROR);
 
-#if MBEDTLS_BEFORE_V3
-    sslRet = mbedtls_sha256_ret(pCert->raw.p, pCert->raw.len, fingerprint, 0);
-#else
-    sslRet = mbedtls_sha256(pCert->raw.p, pCert->raw.len, fingerprint, 0);
-#endif
+    /* Go through the generic message-digest layer rather than the SHA-256 module
+     * directly. */
+    sslRet = mbedtls_md(pMdInfo, pCert->raw.p, pCert->raw.len, fingerprint);
     CHK(sslRet == 0, STATUS_INTERNAL_ERROR);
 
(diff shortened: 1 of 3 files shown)


ak1932:
This is a follow up PR of https://github.com/awslabs/amazon-kinesis-video-streams-webrtc-sdk-c/pull/2385

src/source/Crypto/Dtls_mbedtls.c
```diff
@@ -1147,11 +1147,9 @@ STATUS dtlsCertificateFingerprint(mbedtls_x509_crt* pCert, PCHAR pBuff)
...
     CHK(pMdInfo != NULL, STATUS_INTERNAL_ERROR);
 
-#if MBEDTLS_BEFORE_V3
-    sslRet = mbedtls_sha256_ret(pCert->raw.p, pCert->raw.len, fingerprint, 0);
-#else
-    sslRet = mbedtls_sha256(pCert->raw.p, pCert->raw.len, fingerprint, 0);
-#endif
+    /* Go through the generic message-digest layer rather than the SHA-256 module
+     * directly. */
+    sslRet = mbedtls_md(pMdInfo, pCert->raw.p, pCert->raw.len, fingerprint);
```
sirknightj, line 1152: Note: One other side effect -
  - Before: the hash came from a hardcoded mbedtls_sha256() call
  - After: both the algorithm and the length come from the same pMdInfo = mbedtls_md_info_from_type(MBEDTLS_MD_SHA256). They can't drift apart (helps make it easier for future support of additional algorithms by reducing the number of places to change)

sirknightj (commented):
Couple notes:

1. Please update the target branch to `develop` branch
2. Do we need to update the include headers?  <mbedtls/private/sha256.h> (v4) and <mbedtls/sha256.h> (v2/v3)
3. Do we need the same change for mbedtls_md5 (TURN long-term key in Crypto.h:91)?

ak1932:
I have rebased to the develop target branch and updated KVS_MD5_DIGEST macro as well to use the mbedtls_md API, removing the need for the private sha256 and md5 includes

PR #2394, opened 2026-09-01
test(ice): stress-test non-routable peer filter under flood (#2386)

YPandas:
*Issue #, if available:*

- #2386 (TURN `403 Forbidden IP` stalls ICE gathering). Adds stress/scale coverage for the non-routable peer filter delivered in #2388.

*What was changed?*

- Added two injection/stress unit tests to `tst/TurnConnectionFunctionalityTest.cpp` for the non-routable TURN peer filter:
  - `turnConnectionFiltersDistinctNonRoutableFloodFast`
  - `turnConnectionFloodDoesNotDelayRoutablePeers`
- Added shared `allocStressTestTurnConnection` / `freeStressTestTurnConnection` test helpers.
- Test-only change; no production code modified.

*Why was it changed?*

- The filter's whole point is that a large number of local/non-routable peer candidates can be rejected up front so they never trigger a `CreatePermission` the TURN server would answer with `403 Forbidden IP` (which otherwise stalls the state machine until the ~5s-per-state timeout). Prior tests confirmed the single-peer behavior but not that the filter holds up under scale.
- These lock in two properties a regression could silently break: (1) a huge flood of distinct non-routable peers is absorbed cheaply and never enters the peer list, and (2) a flood does not crowd out or delay genuinely routable peers — the state machine still advances immediately once the real peers succeed.
- Per-range coverage of the classifier itself already lives in `NetworkApiTest.IsNonRoutableAddr*`, so these tests deliberately focus on scale and flood behavior rather than duplicating range checks.

*How was it changed?*

- Both tests build a minimal heap `TurnConnection` (live mutex, `stateTimeoutTime` pushed far into the future so results reflect peer-state counting rather than a timeout fallthrough) and drive `turnConnectionAddPeer` directly — no network or credentials, so they are not gated on `mAccessKeyIdSet` and always run.
- `turnConnectionFiltersDistinctNonRoutableFloodFast` injects 500,000 **distinct** non-routable IPv4 addresses (packed across `10.0.0.0/8`) and asserts every call is a successful no-op, `turnPeerCount == 0`, `nonRoutablePeersFilteredCount == 500000`, and the flood completes well under a single per-state timeout.
- `turnConnectionFloodDoesNotDelayRoutablePeers` interleaves 100,000 non-routable peers with 3 genuinely routable ones, then asserts only the routable peers are added (`turnPeerCount == 3`, the rest filtered) and that `fromCreatePermissionTurnState` advances to `TURN_STATE_BIND_CHANNEL` once they succeed.
- The per-skip `INFO` log is silenced during the flood (`SET_LOGGER_LOG_LEVEL(LOG_LEVEL_WARN)`, restored to `mLogLevel`) so timing reflects filter cost, not log I/O; teardown frees each added peer's transaction-id store to keep the CI sanitizer jobs clean.

- Built `webrtc_client_test` locally (`-DBUILD_TEST=ON`) and ran both new cases:
  `./tst/webrtc_client_test --gtest_filter='TurnConnectionFunctionalityTest.turnConnectionFiltersDistinctNonRoutableFloodFast:TurnConnectionFunctionalityTest.turnConnectionFloodDoesNotDelayRoutablePeers'` — 2/2 passed.
- Confirmed the functional assertions are the real regression guards (if the filter were removed, adds would fill the 32-peer cap then return errors, `turnPeerCount` would be non-zero, and `nonRoutablePeersFilteredCount` would be 0 — all caught independently of timing).
- Tests are pure logic (no network/credentials), so they run identically in CI across the OS/TLS/compiler build matrix, including the sanitizer jobs.

### DIFF
--- a/tst/TurnConnectionFunctionalityTest.cpp
+++ b/tst/TurnConnectionFunctionalityTest.cpp
@@ -1177,6 +1177,159 @@ TEST_F(TurnConnectionFunctionalityTest, turnConnectionCallMultipleTurnSendDataIn
     freeTestTurnConnection();
 }
 
+// ---------------------------------------------------------------------------
+// Injection / stress tests for the non-routable peer filter (issue #2386, PR #2388).
+//
+// A flood of local / non-routable peer candidates must be absorbed cheaply. Each is rejected by
+// turnConnectionAddPeer at O(1) -- before the lock, the duplicate scan and the transaction-id-store
+// allocation -- so it never enters turnPeerList and never produces a CreatePermission that the TURN
+// server would reject with 403 Forbidden IP, and so cannot stall the state machine until the
+// ~5s-per-state timeout. These are pure logic (no network / credentials), so they are not gated on
+// mAccessKeyIdSet and always run. Exhaustive per-range coverage of the classifier itself lives in
+// NetworkApiTest.IsNonRoutableAddr*; here the focus is scale and behaviour under a flood.
+
+// Allocates a minimal TurnConnection usable by turnConnectionAddPeer and the state-machine
+// predicates. stateTimeoutTime is pushed far into the future so advance decisions reflect
+// peer-state counting, not a timeout fallthrough.
+static PTurnConnection allocStressTestTurnConnection()
+{
+    PTurnConnection pTurnConnection = (PTurnConnection) MEMCALLOC(1, SIZEOF(TurnConnection));
+    EXPECT_TRUE(pTurnConnection != NULL);
+    if (pTurnConnection != NULL) {
+        pTurnConnection->lock = MUTEX_CREATE(FALSE);
+        pTurnConnection->state = TURN_STATE_CREATE_PERMISSION;
+        pTurnConnection->stateTimeoutTime = GETTIME() + 100 * HUNDREDS_OF_NANOS_IN_A_SECOND;
+        pTurnConnection->ipFamilyType = KVS_IP_FAMILY_TYPE_IPV4;
+        pTurnConnection->disableNonRoutablePeersFilterForRelay = FALSE; // filter on (default)
+    }
+    return pTurnConnection;
+}
+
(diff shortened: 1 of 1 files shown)


tst/TurnConnectionFunctionalityTest.cpp
```diff
@@ -1177,6 +1177,159 @@ TEST_F(TurnConnectionFunctionalityTest, turnConnectionCallMultipleTurnSendDataIn
...
+        // timed region reflects only the filter cost.
+        allFiltered = (BOOL) (allFiltered && (turnConnectionAddPeer(pTurnConnection, &peer) == STATUS_SUCCESS));
+    }
+    UINT64 floodElapsed = GETTIME() - startTime;
+
+    SET_LOGGER_LOG_LEVEL(mLogLevel);
+
+    EXPECT_TRUE(allFiltered);
+    // None of the flood entered the peer list; every one was counted as filtered.
+    EXPECT_EQ((UINT32) 0, pTurnConnection->turnPeerCount);
```
sirknightj, line 1260: Is there a difference between what this test is testing vs. the tests added in #2388? Both check that the filter filters the invalid ones.

Is there a need to test 500k? Seems like 1 (since we're checking non-zero) would suffice. In case of a slow environment, it may not finish in time.

The delays in the issue comes from the network retries/delay, but there are no network calls in these tests, so the 3s timeout is mostly testing the CPU speed.

YPandas: I originally wanted to sanity-check that throwing a very large number of non-routable candidates at `turnConnectionAddPeer` gets absorbed cleanly (all filtered, none added) rather than accumulating. But you're right that peer #500k exercises the same code as peer #1 and the big count probably only adds CI flakiness risk on slow/sanitizer runners without covering anything new.

I'll close this PR since the same tests has already been done.

PR #2397, opened 2026-09-09
Fix signaling receive worker use-after-free

### LINKED ISSUE #2240: [Bug]: Potential use-after-free race between freeSignaling() and async receiveLwsMessageWrapper() callback thread
### Please confirm you have already done the following

- [ ] I have searched the repository for related/existing bug reports
- [ ] I have all the details the issue requires

### Please answer the following prompt

- [ ] This issue is replicable using the unmodified sample application

### Describe the bug

In our project, we actively perform signaling server reconnection, so we believe we are more likely than usual to hit this issue.

I believe there is a race condition in the signaling client shutdown path that can lead to a use-after-free (UAF).
receiveLwsMessage() dispatches message handling asynchronously (detached thread or threadpool) via receiveLwsMessageWrapper(), and that wrapper accesses pSignalingClient fields and user callbacks.

At the same time, freeSignaling() can free pSignalingClient after stopping the listener thread, but without synchronizing with all in-flight receiveLwsMessageWrapper() tasks.

Because of this, a wrapper task may still run after pSignalingClient has been freed.

Relevant code paths

Async dispatch from receive path: LwsApiCalls.c (receiveLwsMessage -> THREAD_CREATE + THREAD_DETACH / threadpool push).
Async worker uses client object: LwsApiCalls.c (receiveLwsMessageWrapper).
Shutdown/free path: Signaling.c (freeSignaling).
Listener termination wait exists, but it appears to cover listener thread only, not all detached receive worker tasks.
Why this looks unsafe

### Expected Behavior

No callback worker should access pSignalingClient after shutdown/free starts.

### Current Behavior

There appears to be a race where an async receive callback can still access freed memory.

### Reproduction Steps

1. Connect as a master (or equivalent role) and keep the signaling connection active.
2. Repeatedly trigger reconnection in short intervals.
3. While reconnect is in progress, send signaling messages from a smartphone peer (especially ICE_CANDIDATE messages).

### WebRTC C SDK version being used

v1.12.0 (observed), potentially reproducible on v1.17.0

### If it was working in a previous version, which one?

_No response_

### Compiler and Version used

arm-linux-gnueabihf-gcc (GCC) 9.1.0

### Operating System and version

Linux

### Platform being used

Embedded Linux device (ARM-based custom board)

YOSI-yoshidayuji:
*Issue #, if available:*
- Fixes #2240

*What was changed?*
- Added reference counting to keep `SignalingClient` alive while listener, reconnect, and asynchronous receive workers can still access it.
- Synchronized receive callback admission with signaling shutdown.
- Added deterministic regression coverage for receive-worker shutdown races and freeing the signaling client from inside a receive callback.

*Why was it changed?*
- `receiveLwsMessage()` dispatches `receiveLwsMessageWrapper()` asynchronously using either a detached thread or the KVS thread pool.
- `freeSignaling()` previously destroyed `SignalingClient` without accounting for receive workers that had already been dispatched but had not completed.
- A delayed worker could therefore access the freed client, its synchronization primitives, diagnostics, or callback fields, causing a heap-use-after-free.

*How was it changed?*
- The signaling client starts with one owner reference.
- Listener, reconnect, and receive worker threads acquire a reference before they are created or queued and release it on completion.
- Failed thread creation or thread-pool submission rolls back the acquired reference and frees the pending message wrapper.
- `freeSignaling()` marks the client as shutting down, releases the owner reference, and defers physical destruction until the final asynchronous reference is released.
- A re-entrant receive-callback lock serializes callback admission with shutdown. This prevents a new receive callback from starting after shutdown while allowing `freeSignaling()` to be called safely from inside the callback.

- Before the fix, the deterministic regression test reproduced a `heap-use-after-free` in `receiveLwsMessageWrapper()` under AddressSanitizer.
- After the fix, the same regression scenario passes without AddressSanitizer errors or leaks.
- Verified normal builds with `ENABLE_KVS_THREADPOOL` both `OFF` and `ON`.
- Verified AddressSanitizer builds with `ENABLE_KVS_THREADPOOL` both `OFF` and `ON`.
- Verified ThreadSanitizer builds with `ENABLE_KVS_THREADPOOL` both `OFF` and `ON`, with no reported data races.
- Verified callback-triggered signaling client cleanup does not deadlock.
- Verified invalid-input cleanup and existing signaling message parsing tests.

### DIFF
--- a/src/source/Signaling/LwsApiCalls.c
+++ b/src/source/Signaling/LwsApiCalls.c
@@ -2318,7 +2318,6 @@ CleanUp:
             retStatus = pSignalingClient->signalingClientCallbacks.errorReportFn(pSignalingClient->signalingClientCallbacks.customData, retStatus,
                                                                                  pMessage, messageLen);
         }
-
     }
 
     if (!receiveWorkerScheduled) {
--- a/src/source/Signaling/LwsApiCalls.c
+++ b/src/source/Signaling/LwsApiCalls.c
@@ -361,8 +361,10 @@ INT32 lwsWssCallbackRoutine(PVOID wsi, INT32 reason, PVOID user, PVOID pDataIn, 
                 // Handle re-connection in a reconnect handler thread. Set the terminated indicator before the thread
                 // creation and the thread itself will reset it. NOTE: Need to check for a failure and reset.
                 ATOMIC_STORE_BOOL(&pSignalingClient->reconnecterTracker.terminated, FALSE);
+                acquireSignalingClient(pSignalingClient);
                 retStatus = THREAD_CREATE(&pSignalingClient->reconnecterTracker.threadId, reconnectHandler, (PVOID) pSignalingClient);
                 if (STATUS_FAILED(retStatus)) {
+                    releaseSignalingClient(pSignalingClient);
                     ATOMIC_STORE_BOOL(&pSignalingClient->reconnecterTracker.terminated, TRUE);
                     CHK(FALSE, retStatus);
                 }
@@ -407,8 +409,10 @@ INT32 lwsWssCallbackRoutine(PVOID wsi, INT32 reason, PVOID user, PVOID pDataIn, 
                 // Handle re-connection in a reconnect handler thread. Set the terminated indicator before the thread
                 // creation and the thread itself will reset it. NOTE: Need to check for a failure and reset.
                 ATOMIC_STORE_BOOL(&pSignalingClient->reconnecterTracker.terminated, FALSE);
+                acquireSignalingClient(pSignalingClient);
                 retStatus = THREAD_CREATE(&pSignalingClient->reconnecterTracker.threadId, reconnectHandler, (PVOID) pSignalingClient);
                 if (STATUS_FAILED(retStatus)) {
+                    releaseSignalingClient(pSignalingClient);
                     ATOMIC_STORE_BOOL(&pSignalingClient->reconnecterTracker.terminated, TRUE);
                     CHK(FALSE, retStatus);
                 }
@@ -1504,7 +1508,12 @@ STATUS connectSignalingChannelLws(PSignalingClient pSignalingClient, UINT64 time
 
     // The actual connection will be handled in a separate thread
     // Start the request/response thread
-    CHK_STATUS(THREAD_CREATE(&pSignalingClient->listenerTracker.threadId, lwsListenerHandler, (PVOID) pLwsCallInfo));
+    acquireSignalingClient(pSignalingClient);
+    retStatus = THREAD_CREATE(&pSignalingClient->listenerTracker.threadId, lwsListenerHandler, (PVOID) pLwsCallInfo);
+    if (STATUS_FAILED(retStatus)) {
+        releaseSignalingClient(pSignalingClient);
+    }
+    CHK_STATUS(retStatus);
     CHK_STATUS(THREAD_DETACH(pSignalingClient->listenerTracker.threadId));
 
     timeout = (pSignalingClient->clientInfo.connectTimeout != 0) ? pSignalingClient->clientInfo.connectTimeout : SIGNALING_CONNECT_TIMEOUT;
@@ -1808,6 +1817,8 @@ CleanUp:
         MUTEX_UNLOCK(pSignalingClient->listenerTracker.lock);
     }
 
+    releaseSignalingClient(pSignalingClient);
+
     LEAVES();
     return (PVOID) (ULONG_PTR) retStatus;
 }
@@ -1859,6 +1870,8 @@ CleanUp:
         // Notify the listeners to unlock
         CVAR_BROADCAST(pSignalingClient->reconnecterTracker.await);
     }
+
+    releaseSignalingClient(pSignalingClient);
 
     LEAVES();
     return (PVOID) (ULONG_PTR) retStatus;
@@ -2167,8 +2180,12 @@ STATUS receiveLwsMessage(PSignalingClient pSignalingClient, PCHAR pMessage, UINT
     STATUS retStatus = STATUS_SUCCESS;
     UINT32 i, strLen;
     PSignalingMessageWrapper pSignalingMessageWrapper = NULL;
+#ifndef ENABLE_KVS_THREADPOOL
     TID receivedTid = INVALID_TID_VALUE;
+#endif
     PSignalingMessage pOngoingMessage;
+    BOOL receiveWorkerReferenceAcquired = FALSE;
+    BOOL receiveWorkerScheduled = FALSE;
 
     CHK(pSignalingClient != NULL, STATUS_NULL_ARG);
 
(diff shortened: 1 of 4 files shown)


YPandas:
Thank you for your contribution, we are currently looking into it. Can you please change the target to develop? We will merge develop into main as part of the regular release cycle.

YOSI-yoshidayuji:
Thank you for reviewing the PR. I have changed the target branch to  develop .

src/source/Signaling/LwsApiCalls.c
```diff
@@ -2452,16 +2495,32 @@ PVOID receiveLwsMessageWrapper(PVOID args)
...
         MUTEX_UNLOCK(pSignalingClient->offerSendReceiveTimeLock);
     }
-    // Calling client receive message callback if specified
-    if (pSignalingClient->signalingClientCallbacks.messageReceivedFn != NULL) {
+    if (receiveMessagePreCallbackHookFn != NULL) {
+        CHK_STATUS(receiveMessagePreCallbackHookFn(hookCustomData));
+    }
+
+    MUTEX_LOCK(pSignalingClient->receiveCallbackLock);
+    receiveCallbackLocked = TRUE;
```
sirknightj: `receiveCallbackLock` is held across the user callback, while `freeSignaling()` also takes it

It can result in an AB‑BA deadlock in an application that acquires a lock inside `messageReceivedFn`, including the current samples (`samples/common/Common.c`: `sessionCleanupWait()` holds `sampleConfigurationObjLock` from L1587 across `freeSignalingClient()` at L1637, and `signalingMessageReceived()` takes that same lock at L1724).

| Thread | Order |
|---|---|
| Receive worker | `receiveCallbackLock` (`LwsApiCalls.c:2502`) → user `messageReceivedFn` (`:2507`) → app lock |
| App teardown | app lock → `freeSignaling()` → `receiveCallbackLock` (`Signaling.c:404`) |

src/source/Signaling/LwsApiCalls.c
```diff
@@ -2270,12 +2287,24 @@ STATUS receiveLwsMessage(PSignalingClient pSignalingClient, PCHAR pMessage, UINT
     DLOGD("Client received message of type: %s",
           getMessageTypeInString(pSignalingMessageWrapper->receivedSignalingMessage.signalingMessage.messageType));
 
+    CHK(!ATOMIC_LOAD_BOOL(&pSignalingClient->shutdown), retStatus);
+    acquireSignalingClient(pSignalingClient);
```
sirknightj, line 2291: Under `ENABLE_KVS_THREADPOOL`, it may be possible for the refCount not to reach 0, when a task is queued but not executed (only during the shutdown path).

If `deinitKvsWebRtc()` runs while a receive task is queued and has not yet run, `refCount` never reaches 0 and `destroySignalingClient()` won't run. `threadpoolFree()` does `safeBlockingQueueClear(pThreadpool->taskQueue, TRUE)`, which discards queued tasks without invoking them.

src/source/Signaling/Signaling.c
```diff
@@ -378,11 +372,48 @@ STATUS freeSignaling(PSignalingClient* ppSignalingClient)
...
+{
+    if (pSignalingClient != NULL) {
+        ATOMIC_INCREMENT(&pSignalingClient->refCount);
+    }
+}
+
+VOID releaseSignalingClient(PSignalingClient pSignalingClient)
+{
+    // ATOMIC_DECREMENT returns the value before decrementing.
+    if (pSignalingClient != NULL && ATOMIC_DECREMENT(&pSignalingClient->refCount) == 1) {
```
sirknightj: Suggest to add a way to debug this refcount if it gets stuck.

One idea: have this function return the post-decrement count and log the deferral. Compute it from a local — you must not read `pSignalingClient->refCount` after releasing, since another thread may already have freed the object:
```c
SIZE_T releaseSignalingClient(PSignalingClient pSignalingClient)
{
    SIZE_T remaining = 0;
    if (pSignalingClient != NULL) {
        remaining = ATOMIC_DECREMENT(&pSignalingClient->refCount) - 1;   // returns pre-decrement value
        if (remaining == 0) {
            destroySignalingClient(pSignalingClient);
        }
    }
    return remaining;
}
```

then in `freeSignaling()`: `if (remaining != 0) { DLOGW("Signaling client destruction deferred; %u async reference(s) outstanding", (UINT32) remaining); }`. A `DLOGV` of the count in acquire/release would also help increase visibility and make future debugging easier.

src/source/Signaling/Signaling.c
```diff
@@ -378,11 +372,48 @@ STATUS freeSignaling(PSignalingClient* ppSignalingClient)
...
+
+    if (IS_VALID_MUTEX_VALUE(pSignalingClient->receiveCallbackLock)) {
+        MUTEX_LOCK(pSignalingClient->receiveCallbackLock);
+        ATOMIC_STORE_BOOL(&pSignalingClient->shutdown, TRUE);
+        MUTEX_UNLOCK(pSignalingClient->receiveCallbackLock);
+    } else {
+        ATOMIC_STORE_BOOL(&pSignalingClient->shutdown, TRUE);
+    }
+
+    terminateOngoingOperations(pSignalingClient);
```
sirknightj: Suggest to check the return status (log via `CHK_LOG_ERR`) to make it easier to debug any stuck threads or errors.

Want to make it easy to spot any listener or reconnect thread that misses its 9 s `SIGNALING_CLIENT_SHUTDOWN_TIMEOUT`.

src/source/Signaling/LwsApiCalls.c
YOSI-yoshidayuji: I removed  receiveCallbackLock , which was the cause of the deadlock. The lifetime of the internal signaling client remains protected by reference counting. Application-owned callback data is not managed by the SDK, so the application is responsible for keeping it valid until any in-flight callback has completed.

src/source/Signaling/LwsApiCalls.c
YOSI-yoshidayuji: Ideally, the PIC thread pool would support a cleanup callback for discarded tasks. However, that would require changes to a separate project and a dependency update. To keep this fix within the SDK and avoid affecting other tasks such as DTLS and STUN, I introduced a dedicated thread pool for signaling receive tasks. During shutdown, the SDK rejects new receive tasks and waits for all queued or running receive tasks to complete before destroying the dedicated thread pool.

src/source/Signaling/Signaling.c
YOSI-yoshidayuji: Added a warning log when signaling client destruction is deferred due to outstanding asynchronous references.

src/source/Signaling/Signaling.c
YOSI-yoshidayuji: Added error logging for listener and reconnect thread termination failures.

YPandas:
Thanks so much for the contribution, we really appreciate you taking the time to dig into this.

As it happens, we've been working on fixes for this bug on our side as well. If you get a chance, would you mind trying out this branch and letting us know whether it resolves the issue you're seeing? https://github.com/awslabs/amazon-kinesis-video-streams-webrtc-sdk-c/tree/uaf-fix-2-3

We'll review your PR in parallel, so nothing is blocked on that. Thanks again!

YOSI-yoshidayuji:
> As it happens, we've been working on fixes for this bug on our side as well. If you get a chance, would you mind trying out this branch and letting us know whether it resolves the issue you're seeing? https://github.com/awslabs/amazon-kinesis-video-streams-webrtc-sdk-c/tree/uaf-fix-2-3

Thanks for sharing the branch. I tested  uaf-fix-2-3  with the deterministic regression test I originally created for this issue. Since this implementation makes  freeSignaling()  wait for the receive worker to complete, I adjusted only the test synchronization accordingly. The test passed under AddressSanitizer with the thread pool both enabled and disabled, with no use-after-free or memory leak detected. This branch appears to resolve the original issue I reported.

YPandas:
It's great to hear that it works for you. Based on your confirmation, we'll raise a PR for our two patches to merge them into mainline, along with our unit tests. Once we hear back from other customers who have reported that the reconnection path does not gracefully shut down when the application terminates, and confirm that this patch also fixes their issue, we will merge the PR into develop. We will also add credit and reference your PR accordingly.

